PR #1558, opened 2012-12-12
Simplify loading of view restrictions in the menu.

adaptivelogic:
This refactoring aims to reduce the complexity of the query run when loading view restrictions in the menu code, solving #1224. When running with CMS_PERMISSION = False, the loading should be omitted entirely, saving considerable time.
The other aim is to move most of the view-restriction related code to cms.utils.permissions, allowing for sharing of cache and ensuring that both has_view_restriction and get_visible_sites are changed in parallel. Further, parts of the code was rewritten to enhance readability.


mkoistinen:
Wow, this PR is nearly 2 years old now. Still looks useful, but I worry about its ability to be merged against any recent code-base.

@yakky, can we get an update from your side? Is this still pegged for 3.1?

yakky:
It was still really hard to merge this in 3.0, but I'll hard try to port it to 3.1.
It's not mergeable as-is but I hope to be able to port the code anyway

PR #2971, opened 2014-03-25
Implements a different live/draft switcher for placeholders outside of the CMS.

kotyy:
Placeholders outside of the CMS currently cause the live/draft switcher to be displayed, which toggles the editing mode. However, for non-CMS models, draft versions are not implemented. This can be confusing to users.

This PR adds another switcher template that uses the language "Editing Off" in place of "Draft" and "Editing Live" in place of "Live." The PageToolbar.add_draft_live() method has been modified to take the parameter "is_page" which determines the template used.

Tests have not been implemented as there are currently no tests for cms_toolbar.py.

Some thoughts: Should there be a separate method for adding the new toggle, as opposed to piggybacking off of add_draft_live? Should there be one template for the switcher that we hand text to? Can anyone offer guidance on implementing tests?

### DIFF
--- a/cms/cms_toolbar.py
+++ b/cms/cms_toolbar.py
@@ -247,7 +247,8 @@ class PageToolbar(CMSToolbar):
             self.toolbar.add_item(TemplateItem("cms/toolbar/items/live_draft.html", extra_context={'request': self.request},
                                            side=self.toolbar.RIGHT), len(self.toolbar.right_items))
         else:
-            self.toolbar.add_item(TextItem(text="Editing Live", side=self.toolbar.RIGHT))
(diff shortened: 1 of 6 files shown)


mkoistinen:
This is about 9 months old/stale now. It seems like a very useful feature. However, I wonder if this would be superseded by another implementation at this point.

@GaretJax Comment?

GaretJax:
@mkoistinen I don't know yet. I'm just working on versioning right now, but we have plans to add generic draft/published support for arbitrary models, not just placeholders. Maybe @intellectronica knows more.

yakky:
#3868 addresses this by just renaming Live/Draft to Edit/Save and close, so no expectation of published/unpublished content exists

Issue #2639, opened 2014-02-13
Better support for custom models


DylannCordel:
We, [WebU](/webu) (www.webu.coop), are using DjangoCMS and we need some custom model
support improvments. We would be happy if it could be done in DjangoCMS.
Here is the features we ([ollb](/ollb) an [I](/DylannCordel)) would like to develop :
## CMS page like integration of custom models in admin interface

Purpose :
- Add a link in change_list to load the detail view in the main iframe
- Add an icon link in change_list to edit the model without placeholder

Technical notes :
- The idea is to add a CustomModelAdmin class.
## Easy use of models in Pages

When manipulating a lots of custom models, it would be very handy to have
generic list plugin and generic detail view.
### Generic custom model list plugin

Provide a way to create generic plugin that displays a list of custom models instances.

Available dynamic options to configure the display of the list :
- max number of instances to display
- use a pager or not (is so define a min number of instances per page)
- title of the list
- define parameters for dynamic filters (cf [filters](#filters))
- enable a search form and choose fields available
- enable a sort form and choose fields available
- link to the view page

Available static options to configure the display of the list :
- define a static filter
- define the dynamic filters
- define fields available for search
- define fields available for sort
- enable or not a "create new" link

<a name="filters"></a>Dynamic and static filters

We may want to restrict the displaid list of instances (ie : news.published = True) :
- When creating the Plugin Class, to have a base restriction (static filter)
- When creating the Plugin instance, to have a "contextual" restriction 
  (dynamic filter). ie : 
  - on home page, we only want news.home = True
  - on news page, we want all news from this year
  - on archive page, we want all news from past years.
### Generic custom model detail view
- Provide a generic DetailView with specific permission verification from model, else
  generic permission fallback.
- URL configuration should be automagically discovered if not specified 
  (ie : find a slugfield, else a unique (combination of) field(s).)
## Add a generic way to integrate a custom model in cms_toolbar

Purpose :
- One click access to the custom model admin change list
- Remove page specific stuff menu in cms_toolbar on a detail custom model view  
  In a detail custom model view, the page menu missleads the user because actions
  are related to the app hook page (not the current one, whicho doens't exist in
  fact.
- Add optionaly a custom model specific menu like the page one
- Improve language menu to support hvad if custom model use it
- Improve history menu to support custom model

mkoistinen:
Hey all, this looks like it'd be a nice addition. @DylannCordel whats the status from your side? Been nearly a year since this was opened now.

[...]

PR #3034, opened 2014-04-09
[WIP] Better support of custom models

DylannCordel:
We,  [WebU](/webu) (www.webu.coop), are using DjangoCMS and we need some custom model
support improvments. We would be happy if it could be done in DjangoCMS.
Features we ([ollb](/ollb) an [I](/DylannCordel)) would like to develop are explained in issue #2639

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -18,7 +18,7 @@ django CMS
 Open source enterprise content management system based on the django framework.
 
 .. ATTENTION:: To propose features, always open pull requests on the **develop** branch.
-   It's the branch for features that will go into the next django CMS feature relese.
+   It's the branch for features that will go into the next django CMS feature release.
 
    For fixes for 3.0.x releases, please work on **support/3.0.x** which contains
    the next patch release for 3.0.x series.
--- a/cms/admin/generic.py
+++ b/cms/admin/generic.py
@@ -1,0 +1,6 @@
+# -*- coding: utf-8 -*-
+from django.contrib import admin
+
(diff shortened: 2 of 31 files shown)


FinalAngel:
Thank you for the contribution @DylannCordel Would you be able to rebase this to the current develop branch?

auvipy:
@DylannCordel whats going on your side about this PR? is everything OK with latest django cms and django framework versions? are you planning to work on it? or looking for someone else for take over?

@FinalAngel is the PR accepted by the technical board?

FinalAngel:
@evildmp @yakky @mkoistinen can you guys evaluate?

DylannCordel:
@auvipy : we (Webu) would like some feedbacks and discussions about "how / what to implement". Unfortunately, we are a small team without a lot of time to work on open source projects and must be sure it won't be thrown to the trash. We would like to work "with you", and not "alone". It is for this reason that we wrote the documentation and few code : to have some feedback about coherence with your future plans for Django-CMS.

yakky:
There a lot of different things going on here and in #2639
And as far as I can see it's a very opinionated way to build custom models (which is not wrong at all per-se, but it might not fit into the CMS core itself)
Most of the code here adds non-existing modules and packages while touch very little the existing core: this suggests me that it's more suitable for an external application that provides hooks and features for other applications to build custom models on top of it, rather than integrating this framework into the core

FinalAngel:
please re-open once the discussion has reached a conclusion in https://github.com/divio/django-cms/issues/2639

PR #3118, opened 2014-04-30
Fix #3117 - added a (basic) navigation bar in the sideframe

### LINKED ISSUE #3117: Adding navigation buttons in the side frame.
I've had some feedback about the fact that navigating in the side frame is not always intuitive.
People seems to have some trouble finding how to go "back" in the sideframe and are confused if yes or no they should hit the back button of the browser to do so.
Their first thought is that it would impact the main frame instead.

I just added a toolbar on top of the sideframe - I know much more can be done on the design side of thing but there are much more talented people than me around.

PR follows

johnraz:
see #3117


yakky:
@johnraz any progress on this?

FinalAngel:
There has been some modifications in the sideframe for 3.2 @johnraz that would make this obsolete.

PR #3516, opened 2014-10-13
Sane add plugin

ojii:
Historically, when a user wanted to add a plugin to a placeholder, upon selecting the plugin, the CMS would create a plugin instance in the database, *then* show the user the add form. In various cases where the user would not complete the process, "ghost plugins" would be created. These have always been somewhere between a nuisance and an actual problem, sometimes going as far as crashing an site.

This pull request attempts to solve this by not creating a plugin until *after* the user hits save on the add form.

This **changes the contract** for ``CMSPluginBase`` subclasses and brings them more in line with it's parent, ``ModelAdmin``. Specifically ``has_add_permission`` now needs to be properly implemented, instead of just using ``has_change_permission``, ``get_form`` actually has to respect the ``obj`` parameter (which can be ``None``) and ``add_view`` will actually be called. Further, a ``add_view_check_request`` helper function was added to ``CMSPluginBase`` which verifies a ``HttpRequest`` to ``add_view`` has the required GET parameters. This may be used by subclasses wishing to modify ``add_view``.

This pull request **breaks** at least djangocms-text-ckeditor and djangocms-link, both of which have pull requests to make them compatible (see https://github.com/divio/djangocms-text-ckeditor/pull/179 and https://github.com/divio/djangocms-link/pull/29).

This will **break** any plugin that assumes an object to be around in ``get_form`` or ``add_view`` either by assuming ``self.cms_plugin_instance`` to be available or in the case of ``get_form`` that the ``obj`` argument is not ``None``.

## Todo

- [x] Write code
- [x] Frontend integration
- [x] Make tests pass
- [x] Testing frontend integration


ojii:
@yakky @mkoistinen looks like the failures are unrelated to code (within selenium failing to connect to the remote webdriver), any clue how to fix this? Just keep hitting "restart" until it works?

yakky:
All failures are related to custom user models. What happens locally when using it?

ojii:
The problem is ``SELENIUM=1`` not the user model, see for example https://travis-ci.org/divio/django-cms/jobs/47177191 it's failing to authenticate with saucelabs. @digi604 it uses your account, any idea why this is failing?

yakky:
@ojii can't we just switch to --xvfb?

yakky:
I cannot get the selenium tests to run locally

yakky:
In other pull requests tests run fine by using the same travis configuration.
I can't get what's wrong here...

ojii:
Okay I give up. I'm happy to fix any issues with the code but simply give up getting this to work on travis. If anyone wants to take over this, that would be awesome.

ojii:
closing in favor of #3804

PR #3611, opened 2014-11-24
Exclude model base classes (MTI or otherwise) from warnings

### LINKED ISSUE #3604: `manage.py cms check` shouldn't report warning for relationship to a plugin's own "base class"
I don't know if the use of an abstract class is related here, but in this project, I define an (abstract) base plugin model class, which is then extended with a concrete class, which is itself extended for another concrete plugin model.

So, we ultimately have these models defined in modules X and Y:

- X.A (CMSPlugin), abstract
    - X.B (A), concrete
        - Y.C (B), concrete

`python manage.py cms check` reports:
> project.apps.moduleX.models.B has a foreign key from project.apps.moduleY.models.C, but no "copy_relations" method defined. [WARNING]

Also a bit strange is that this warning is emitted 3X.

I think I was the last person to touch this bit of code, and will probably fix it at some point, but I didn't want to lose sight of this issue.

mkoistinen:
Fixes #3604


yakky:
@mkoistinen can you remove the changes to develop.py and integrate the django 1.7 migrations ?

mkoistinen:
Done!

mkoistinen:
Just rebased it to current support/3.0.x

yakky:
:+1: waiting for tests to complete

develop.py
```diff
@@ -189,40 +208,62 @@ def shell():
     call_command('shell')
 
 
+class capture_all_output(object):
```
yakky: This is now useless, right?

mkoistinen: Yup, removed.


On Sun, Jan 11, 2015 at 11:20 PM, Iacopo Spalletti <notifications@github.com
> wrote:

> In develop.py
> <https://github.com/divio/django-cms/pull/3611#discussion-diff-22770187>:
>
> > @@ -189,40 +208,62 @@ def shell():
> >      call_command('shell')
> >
> >
> > +class capture_all_output(object):
>
> This is now useless, right?
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/divio/django-cms/pull/3611/files#r22770187>.
>



-- 
*MARTIN KOISTINEN*


*BG GSM +359 878 787 253UK GSM +44 7703 046 154*
*UK VoIP +44 20 3286 9932*
*US VoIP +1 (**678) **389-6676*

*100% CARBON-NEUTRAL SINCE 1-JAN-2008*

PR #3613, opened 2014-11-26
Fix makemigrations in develop.py and let it make initial migrations as required

mkoistinen:
The `call_command('schemamigration'…)` returns a `SystemExit` when it is complete. Without a try/except block around it, this will cause a loop to prematurely terminate in a seemingly normal way. This is what was happening before this fix, and this effectively caused `develop.py makemigrations` to only ever make migrations for cms, which happens to be the first appliation in the list it works with.

Also, I've added code that will first check to see if an app already has migrations and if not, it will run the equivalent of `manage.py schemamigration [app] --initial` and emit a notice of the same.

Finally, this now detects if an application in the list if not found, and emits a notice of this too.

### DIFF
--- a/develop.py
+++ b/develop.py
@@ -10,7 +10,7 @@ import os
 import sys
 import warnings
 
-from StringIO import StringIO
+from django.utils.six.moves import StringIO
 
 from django import VERSION
 from django.core.exceptions import DjangoRuntimeWarning
--- a/develop.py
+++ b/develop.py
@@ -189,7 +189,7 @@ def shell():
     call_command('shell')
 
 
-class capture_all_output(object):
+class swallow_all_output(object):
     def __init__(self):
         self.output = StringIO()
         self.original_stdout = sys.stdout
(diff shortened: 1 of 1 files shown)


develop.py
```diff
@@ -201,10 +221,28 @@ def makemigrations(migrate_plugins=True, merge=False, squash=False):
...
+                print('ATTENTION: No migrations found for {0}, creating initial migrations.'.format(application))
+                try:
+                    call_command('schemamigration', application, initial=True)
+                except SystemExit:
+                    pass
+            except ImproperlyConfigured:
+                print('WARNING: The app: {0} could not be found.'.format(application))
+            else:
+                try:
+                    with swallow_all_output():
```
yakky: @mkoistinen why do we need to swallow the output here?
We are going to miss messages and requests for input from the user

mkoistinen: I was attempting to avoid a lot of clutter. I don't think there's any input prompts, but there will be a lot of noise and developers will miss what they need to see, which is provided by the print() lines.

yakky: schemamigration requests for input in different cases (e.g. if default is not defined)

Issue #3659, opened 2014-12-10
Toolbar middleware breaks custom logout template


jsma:
If you customize the standard `registration/logged_out.html` template, this cannot inherit from a base template that has `{% cms_toolbar %}`. If you do, you'll get:

    TypeError: int() argument must be a string or a number, not 'AnonymousUser'

If you refresh the page, the error goes away. This is because the middleware sets attributes on the request object during process_request at which point the user is still logged in because the logout view hasn't been called yet. The tag picks up from there assuming all is good but of course is not.

mkoistinen:
Since I've just merged #3755, can we close this issue? Not clear from this thread.

jsma:
This can be closed, it appears to have been accidentally fixed in an unrelated change.

PR #3660, opened 2014-12-10
Added toolbar sanity check to ensure there is valid request.user before ...

jsma:
...building the toolbar

Refs #3659


yakky:
Merged in #3672 
Will be forward ported in later merge from support/3.0.x to develop
Thanks @jsma

Issue #3637, opened 2014-12-02
Publish page does not preserve plugin order


jsma:
Add three link plugins to a placeholder in the following order and with the following names:
* Link 1
* Link 2
* Link 3

Now publish your changes to create the 'live' version. Now go back to draft mode and drag Link 3 so it comes after Link 2. Both structure and content mode show the plugins now in the correct 1,2,3 order. Click publish and note that the live version still has the plugins in the original order 1,3,2.

jsma:
I should be clear that my pull request is against the develop branch. Your pull request #3709 is against the support/3.0.x branch which doesn't appear to do the same plugin position wackiness that was introduced after the django-treebeard change. I'm not clear on what the bug is you are fixing in your pull request, but agree that the code is confusing & possibly buggy.

jsma:
@yakky I finally added a proper test case after all this time! If you temporarily revert 41d988e2df09da21a4567d3d612238fc30d4cba2 you'll see the tests fail due to the scrambled ordering on the live page and that with the offending lines removed, the ordering is preserved. I've punted on the question of refactoring the plugin position changing code for now.

jsma:
Anything else needed (aside from free time!) to get this merged? Let me know.

yakky:
@jsma I should be able to review the PR in the next few days.
Thanks for your efforts

PR #3661, opened 2014-12-10
Fixed plugin ordering for live plugins.

jsma:
Refs #3637

When publishing a page, the draft plugins are copied to new 'live' versions of themselves. The code was incorrectly overwriting the position of the plugins. This change means the position remains identical to the position of the draft plugin, which is the only correct value.

This pull request is incomplete because it lacks a test case demonstrating that the issue exists as well as proving that this change fixes it. I've had challenges getting a test case to exercise the appropriate code paths since setting up the test data appears to require front-end editing and can't be triggered via the api as the current `test_plugin_order` does.


yakky:
I reworked this a bit in #3825 to add the proposed test, instead of the replacing the existing one.
thanks @jsma !

PR #3664, opened 2014-12-10
speed up travis build times by using pip-accel

piquadrat:
A few notes:

 * AWS S3 is used as the cache. When travis enables caching for
   free accounts, this dependency can be removed
 * it currently uses credentials for my personal AWS account. I'm
   fine with this as long as costs don't balloon. I'll keep an
   eye on it
 * build time of this PR isn't an indication of the speed-up, since
   the cache is empty

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,6 @@ env:
   global:
     - SAUCE_USERNAME=digi604
     - secure: NWiSBCHFB6LbTMget2qLIdqZlx0zeu3j+Y7Lsqb8kuYXyT2IUBGFVedcGWuGv/9Mzypb80EQWtVTokA3/3QIbesqr29uG95pMPHiYWLdnTO6UHcLMcNXiSzhBGdRDZ40iHSVv2dDHs4GNwGOH5+UCA0z3j7SWmChuFbNXh+Vsqw=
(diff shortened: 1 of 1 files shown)


piquadrat:
@yakky there's still the problem of exposing AWS keys, even if they are read only. My suggestion is to park this until Travis supports caching files natively for open source accounts.

yakky:
Sure. Feel free to remove the keys from the code
I put this on hold to clarify the status

yakky:
@piquadrat @divio/django-cms-core now that travis supports cache in open source projects, I'm going to resurrect this!
I already ported to current 3.2.x branch

mkoistinen:
This PR is against develop, so, I'm changing the milestone to 3.3. @yakky you mention a port to the current 3.2.x branch, but I can't find the PR. Do you have a reference?

yakky:
@mkoistinen I have lying somewhere in my fork, but I failed to make it work. pip-accel looks problematic, while plain travis cache looks doable.
I agree we should move this to 3.3

vxsx:
i believe this was implemented in https://github.com/divio/django-cms/pull/5198

yakky:
@vxsx I agree
@piquadrat thanks for raising this, now travis made much simpler to implement such a cache

PR #3666, opened 2014-12-10
Send custom events when the structureboard is shown and hidden.

qris:
This allows websites to adapt by changing page styles for example, so that
all placeholders are visible. This is useful for rotators and other features
that hide parts of the page or place them offscreen.

### DIFF
--- a/cms/static/cms/js/modules/cms.structureboard.js
+++ b/cms/static/cms/js/modules/cms.structureboard.js
@@ -295,6 +295,8 @@ $(document).ready(function () {
 			this.interval = setInterval(function () {
 				$(window).trigger('resize.sideframe');
 			}, interval);
+
+			$(window).trigger('structureboard_shown.sideframe');
 		},
 
 		_hideBoard: function () {
@@ -309,6 +311,8 @@ $(document).ready(function () {
 
 			// clear interval
 			clearInterval(this.interval);
+
+			$(window).trigger('structureboard_hidden.sideframe');
 		},
 
 		_resizeBoard: function () {


cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -295,6 +295,8 @@ $(document).ready(function () {
 			this.interval = setInterval(function () {
 				$(window).trigger('resize.sideframe');
 			}, interval);
+
+			$(window).trigger('structureboard_shown.sideframe');
```
mkoistinen: @FinalAngel Do we have an inventory anywhere of all the custom events we trigger/publish anywhere?

@qris raises a good point here.

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -309,6 +311,8 @@ $(document).ready(function () {
 
 			// clear interval
 			clearInterval(this.interval);
+
+			$(window).trigger('structureboard_hidden.sideframe');
```
mkoistinen: @qris Using this logic, shouldn't every event be thrown at $(window)? There's a reason why bubbling occurs. Having said this, the sideframe is actually an entirely separate document (frame) isn't it? In this case, it would seem to me that it should it indeed be sent to $(window).

@FinalAngel thoughts?

yakky:
@FinalAngel could you give a second look at this and check the comments?

cms/static/cms/js/modules/cms.structureboard.js
qris: @mkoistinen I think that bubbling makes sense for events about DOM elements. However I don't think it makes sense for custom events that don't relate to a particular DOM element. I think they should all be thrown at, and caught on, a specific element for consistency. And for reasons of efficiency stated above (to avoid the overhead of bubbling), I think that should be $(window).

FinalAngel:
`$(window)` is fine for this instance. You'd have to check on individual situations where the event should be bubbling too.

yakky:
@FinalAngel are the events correct as is or they must be corrected?

FinalAngel:
they are correct so far

yakky:
Thanks, so apparently everybody agrees on this PR :)

yakky:
Thanks @qris for this

Issue #3642, opened 2014-12-03
copy plugins broken?


jerch:
After updating from 3.0.3 to 3.0.7 copying plugins to clipboard doesnt work anymore if the plugin is nested within another plugin. Seems the `request.POST['source_placeholder_id']` in `PlaceholderAdminMixin.copy_plugins` contains a totally wrong id (points to some other placeholder in the template).
A quickfix for my case was to fall back to the placeholder_id of the source_plugin since the database still got the right value:
```patch
--- placeholderadmin.py	2014-12-03 16:07:10.773331000 +0100
+++ placeholderadmin_fix.py	2014-12-03 16:09:16.933332000 +0100
@@ -308,6 +308,7 @@
                 inst, cls = source_plugin.get_plugin_instance(self)
                 plugins = inst.placeholder_ref.get_plugins_list()
             else:
+                source_placeholder = get_object_or_404(Placeholder, pk=source_plugin.placeholder_id)
                 plugins = list(
                     source_placeholder.cmsplugin_set.filter(tree_id=source_plugin.tree_id, lft__gte=source_plugin.lft,
                                                             rght__lte=source_plugin.rght).order_by('tree_id', 'level',

```
I wonder why you dont do it this way in the first place instead of relying on the request value. I guess I missed there something and it has side effect? Maybe you could have a look at it.

yakky:
Fixed by #3728

PR #3671, opened 2014-12-12
Fixing #3642

heppstux:
static_placeholders will create a context-variable that will overrule following placeholders – better trust the plugin, what it's place holder is than a context variable

what do you think?

### DIFF
--- a/cms/templatetags/cms_tags.py
+++ b/cms/templatetags/cms_tags.py
@@ -1057,7 +1057,6 @@ class StaticPlaceholderNode(Tag):
         else:
             placeholder = static_placeholder.public
         placeholder.is_static = True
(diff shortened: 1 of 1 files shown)


yakky:
@heppstux sorry, I overlooked this: please open a PR on support/3.0.x

yakky:
@heppstux I think we may not need tests for this as https://github.com/divio/django-cms/blob/cf359e632b7d9348f3be85f1dfee63063319ee2d/cms/tests/frontend.py#L400 already cover the use case

heppstux:
hey @yakky,
pull request against support/3.0.x is created.

yakky:
Thanks, merged it.
Code will be forward ported after 3.0.8 release
Fixed by #3728

PR #3672, opened 2014-12-12
Added toolbar sanity check to ensure there is valid request.user before ...

jsma:
...building the toolbar

Refs #3659 and pull request #3660


mkoistinen:
Since I've just merged #3755, can we close this issue? Its not clear from the thread here.

jsma:
This PR was merged (which to my knowledge means this is already closed, but I'll delete my branch all the same) but it was later reverted because of a regression. #3755 replaces this.

PR #3682, opened 2014-12-17
Fix pagetree row overflow

yakky:
Control page row overflow by way of media query

This is actually to fix a bug introduced by #3603 which hides much of the extra information for the page
The same fix going to be merged in djangocms-admin-style to provide it quicker (see divio/djangocms-admin-style#33)


FinalAngel:
LGTM :+1:

PR #3683, opened 2014-12-18
Fixes missing add_sideframe_button method.

splaroche:
When trying to use the `CMSToolbar.add_sideframe_button()` method an exception is thrown because `add_sideframe_button()` is not implemented on `ButtonList()`.  This implements the missing method based on the pattern of the other methods in `ButtonList()`.

### DIFF
--- a/cms/toolbar/items.py
+++ b/cms/toolbar/items.py
@@ -451,6 +451,16 @@ class ButtonList(BaseItem):
         self.buttons.append(item)
         return item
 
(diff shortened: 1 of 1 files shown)


yakky:
LGTM
Will be merged in support/3.0 and then forward ported to develop

yakky:
Thanks @splaroche ! Closing this as it will be forward ported to develop from support

Issue #3686, opened 2014-12-18
PageField acts strange on form validation in latest Firefox versions


anatolyivanov:
Whenever you try to save a form that has cms.forms.fields.PageField in it, the value of the widget resets if form validation fails because of other fields in it.

The issue happens in Firefox, and doesn't in Chrome. When I dug into the code, I've noticed this snippet:
```javascript
(function($) {
    var handleSiteChange = function(site_name, selected_id) {
        $("#id_%(name)s_1 optgroup").remove();
        var myOptions = $("#id_%(name)s_2 optgroup[label='" + site_name + "']").clone();
        $("#id_%(name)s_1").append(myOptions);
        $("#id_%(name)s_1").change();
    };
    var handlePageChange = function(page_id) {
        if (page_id) {
            $("#id_%(name)s_2 option").removeAttr('selected');
            $("#id_%(name)s_2 option[value=" + page_id + "]").attr('selected','selected');
        } else {
            $("#id_%(name)s_2 option[value=]").attr('selected','selected');
        };
    };
    $("#id_%(name)s_0").change(function(){
        var site_label = $("#id_%(name)s_0").children(":selected").text();
        handleSiteChange( site_label );
    });
    $("#id_%(name)s_1").change(function(){
        var page_id = $(this).find('option:selected').val();
        handlePageChange( page_id );
    });
    $(function(){
        handleSiteChange( $("#id_%(name)s_0").children(":selected").text() );
        $("#add_id_%(name)s").hide();
    });
})(django.jQuery);
```

Apparently, the widget has one invisible `<select>` tag that affects the initial value passed to widget when validation fails. What happens (and I've checked it by making the third `<select>` visible) is that in FireFox, the value of this `<select>` doesn't get populated on page load.

I had a similar problem lately in another JS application, I've followed an advice from here:
http://stackoverflow.com/a/15656897/345811

Basically, attr("selected") doesn't work as intended in latest FF and Safari builds.

I am following this up with a MR containing a possible fix that works in Django 1.6+ (since jQuery 1.6+ is required and Django 1.5 ships jQuery 1.4.2).

yakky:
Fixed in #3687

PR #3687, opened 2014-12-18
Fix for PageField-containing form validation working incorrectly in FF/S...

anatolyivanov:
...afari

Follows up https://github.com/divio/django-cms/issues/3686


FinalAngel:
@yakky looks sane to me, I do not like the JS in python though but as its a widget go ahead ;)

PR #3702, opened 2014-12-22
Add render_uncached_placeholder templatetag

yakky:
This templatetag is a version of `render_placeholder` which does not use cache, in a similar way to `show_placeholder`, `show_uncached_placeholder`


yakky:
Reopening on develop

PR #3709, opened 2014-12-27
Fix copy plugins

yakky:
the low level method `CMSPlugin.copy_plugin` messes up the reference to the old plugin changing it to the new one
This has no consequence if the old plugin instance is not "reused", but it may affect lower level use (like in blueprint)

### DIFF
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -292,9 +292,8 @@ class CMSPlugin(with_metaclass(PluginModelBase, MPTTModel)):
             new_plugin._no_reorder = True
         new_plugin.save()
         if plugin_instance:
-            if plugin_instance.__class__ == CMSPlugin:
-                # get a new instance so references do not get mixed up
-                plugin_instance = CMSPlugin.objects.get(pk=plugin_instance.pk)
+            # get a new instance so references do not get mixed up
+            plugin_instance = plugin_instance.__class__.objects.get(pk=plugin_instance.pk)
             plugin_instance.pk = new_plugin.pk
             plugin_instance.id = new_plugin.pk
             plugin_instance.placeholder = target_placeholder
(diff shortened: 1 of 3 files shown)


cms/models/pluginmodel.py
```diff
@@ -292,9 +292,8 @@ def copy_plugin(self, target_placeholder, target_language, parent_cache, no_sign
             new_plugin._no_reorder = True
         new_plugin.save()
         if plugin_instance:
-            if plugin_instance.__class__ == CMSPlugin:
-                # get a new instance so references do not get mixed up
-                plugin_instance = CMSPlugin.objects.get(pk=plugin_instance.pk)
+            # get a new instance so references do not get mixed up
+            plugin_instance = plugin_instance.__class__.objects.get(pk=plugin_instance.pk)
             plugin_instance.pk = new_plugin.pk
             plugin_instance.id = new_plugin.pk
```
ojii: then why are they both set to the same value?

yakky: To keep the object consistent, I guess. 
Anyway, you should probably ask @ojii as he wrote those two lines in bab4029 :D

ojii: I'll shut up about this now. Seeing this just confused me as it doesn't feel right and I hoped that because you touched code near it maybe you'd have some insight as to what's going on.

yakky: My understanding is that this is required to return a functional / correct object with right values (`plugin_instance` is returned and is later used by other functions) without hitting the database again as Django ORM does not make this magic for us.
It's ugly? Yes, I think so.
We should investigate this further for 3.1? I totally agree

ojii:
code looks good to me. don't have a real project to properly test this for regressions though.

jsma:
Was asked to look at this in #3637, I'm not clear on what bug this is solving but have no doubts there is a bug ;) Is there a ticket I'm missing that explains the details of this? As noted in #3637, the issue there is that plugin ordering is thoroughly screwed up on publish and that bug was introduced in the develop branch after the django-treebeard changes were merged. It doesn't appear that this PR addresses that issue.

yakky:
yeah, after checking your PR the issue is slightly different, nevermind

cms/models/pluginmodel.py
motleytech: I love it when this happens... :-)

mkoistinen:
OK, I've tested this on an existing project with some pretty custom and deeply-nested plugin setups. Seems to work! Great job!

PR #3720, opened 2014-12-30
Added template filter "json" which is not vulnerable to XSS attacks

jrief:
Added template tag ``...|json`` which shall be used instead of ``...|js``, because the latter is vulnerable to XSS attacks when used in combination with unvalidated data.

For details see https://code.djangoproject.com/ticket/17419

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -293,3 +293,6 @@ Please see Install/2.4 release notes *before* attempting to upgrade to version 2
 - Added branch policy documentaion
 
 === 3.1 (unreleased) ===
+- Added template tag ``...|json`` which shall be used instead of ``...|js``,
+  because the latter is vulnerable to XSS attacks when used in combination
+  with unvalidated data.
--- a/cms/templatetags/cms_js_tags.py
+++ b/cms/templatetags/cms_js_tags.py
@@ -52,16 +52,15 @@ def bool(value):
     else:
         return 'false'
 
-   
+
 class JavascriptString(Tag):
     name = 'javascript_string'
-    
     options = Options(
         blocks=[
             ('end_javascript_string', 'nodelist'),
         ]
     )
-    
+
     def render_tag(self, context, **kwargs):
         rendered = self.nodelist.render(context)
         return u"'%s'" % javascript_quote(rendered.strip())
--- a/cms/templatetags/cms_js_tags.py
+++ b/cms/templatetags/cms_js_tags.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 import warnings
 from classytags.core import Tag, Options
-from cms.utils import SafeJSONEncoder
+from cms.utils.encoder import SafeJSONEncoder
 from cms.utils.compat import DJANGO_1_4
 from django import template
 from django.core.serializers.json import DjangoJSONEncoder
--- a/cms/templatetags/cms_js_tags.py
+++ b/cms/templatetags/cms_js_tags.py
@@ -1,9 +1,12 @@
 # -*- coding: utf-8 -*-
+import warnings
 from classytags.core import Tag, Options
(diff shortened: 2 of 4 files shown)


ojii:
Nice catch. 

Isn't the ``js`` filter an internal API? I'm asking because that means we can remove it without deprecation (and without mention in the changelog, because that would signify it's a supported API, which I think this isn't and shouldn't be).

ojii:
Also because import order is important, could you move it one module further into ``cms.utils``? ``cms.utils.helpers`` or maybe a new module, either is fine by me.

jrief:
@ojii Yes, from what I've seen, its used only twice by djangoCMS itself.
However, since it exposes to the public API (well sort of), and since Django itself does not offer such a filter, it can easily be used for this its own purpose. And if then it is used for not validated data, this can result into a XSS vulnerability.

Therefore, or we should get rid of that filter altogether (beware of backward incompatibilities), or replaced be something safer.

ojii:
> However, since it exposes to the public API (well sort of), and since Django itself does not offer such a filter, it can easily be used for this its own purpose. And if then it is used for not validated data, this can result into a XSS vulnerability.

There's no question this is an issue that needs to be fixed. My only point here is to not accidentally make this a documented API (as in, an API we need to support in the future).

Since it's neither a documented API and also a security issue, this should replace the API, not gracefully deprecate it.

On that note, it probably should've never ended up on here but rather on security@django-cms.org. Please use that email in the future for anything security related.

ojii:
Note: bumped to blocker, assigned 3.0.8 milestone.

ojii:
@yakky I feel like we need to backport this to 2.4.x just in case.

yakky:
@ojii agree.
Anyone's up to tackle a new 2.4 release? :D

ojii:
thanks for volunteering @yakky

yakky:
@ojii not at all (I actually never used 2.4 and I don't have any projects on it, so I'm not the best candidate!)

yakky:
Agree with @ojii: let's drop it without documentation.

jrief:
@ojii 
This is not a security issue in the sense that you hack djangoCMS sites. Otherwise of course I wouldn't have posted it here but used that security tracker's email address.

It only can result in a security problem if (ab)used by third parties apps, therefore they would have to be blamed. I just wanted to note this, because this template filter can easily be imported into your own project by persons not aware of that risk.

jrief:
@yakky 
I just had a look at the output from travis, but from what I see these errors have nothing to do with my patch. Can you please try to apply this patch in your environment.

ojii:
@jrief the errors are because you import ``from django.core.serializers.json import DjangoJSONEncoder`` and ``from django.utils.html import conditional_escape`` in ``cms.utils.__init__``. ``cms.utils.__init__`` is triggered by the test runner pre-django-setup (to set up Django). At least the ``DjangoJSONEncoder`` however requires Django to already be set up.

To solve this, move the ``SafeJSONEncoder`` into it's own module in ``cms.utils`` (maybe ``cms.utils.json_encoder`` or something).

To see if that worked, try running the tests locally. See http://docs.django-cms.org/en/support-3.0.x/contributing/testing.html#running-tests for help.

jrief:
My local test runner still fails with very obscure messages. They are not related to the encoder.

ojii:
> My local test runner still fails with very obscure messages. They are not related to the encoder.

Please paste these messages so we can try to help.

jrief:
http://dpaste.com/1JDMTDP

ojii:
These three errors are unrelated, updating selenium ``pip install -U selenium`` should solve the issue.

yakky:
Merging this through #3720 to let the tests run. Code is ok

yakky:
Thanks @jrief @ojii. Merged in #3746

PR #3724, opened 2015-01-02
delay importing models in commands to prevent circular imports

ojii:
A combination of a custom user model and https://github.com/justquick/django-activity-stream (with django-activity-stream used in the ``models.py`` which defines the custom user model) appears to cause circular imports when hitting cms commands. As a hotfix, this PR delays all model imports in cms commands until as late as possible.

This appears to be triggered by bugs in 3rd party libraries, but since it can be hotfixed in our codebase, it should until a solution can be found.


ojii:
closing for now, new information made my initial assumption invalid.

PR #3728, opened 2015-01-06
Backport: Fix for #3642

### LINKED ISSUE #3642: copy plugins broken?

heppstux:
compare PR here: #3671 , fixes #3642

### DIFF
--- a/cms/templatetags/cms_tags.py
+++ b/cms/templatetags/cms_tags.py
@@ -361,10 +361,7 @@ class RenderPlugin(InclusionTag):
         if not plugin:
             return {'content': ''}
 
-        try:
-            placeholder = context['cms_placeholder_instance']
-        except KeyError:
-            placeholder = plugin.placeholder
+        placeholder = plugin.placeholder
 
         processors = self.get_processors(context, plugin, placeholder)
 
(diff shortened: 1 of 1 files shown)


yakky:
Fix #3671
Fix #3642

yakky:
Thanks @heppstux for working on this, nice cleanup!

heppstux:
my pleasure :) thanks for django-cms!

PR #3729, opened 2015-01-06
Update templates_placeholders.rst

motleytech:
Adding minor edits for documentation.


motleytech:
I am new to sending pull request for the Django repository. So please bear with me for a few PRs.

Am I targeting the right branch (divio:support/3.0.x) or should I send the PR to a different branch?

Thanks

docs/introduction/templates_placeholders.rst
```diff
@@ -8,13 +8,13 @@ you can make your own HTML templates CMS-ready.
 Templates
 #########
 
-You can use HTML templates to customise the look of your website, define
+You can use HTML templates to customize the look of your website, define
```
evildmp: We use British English spellings, so "customise" is correct.

evildmp:
Thanks for this. Changes such as this can go into the support branch. See http://docs.django-cms.org/en/support-3.0.x/contributing/index.html#branch-policy for an explanation.

docs/introduction/templates_placeholders.rst
jrief: But in science the lingua franca is American English. I'm also very opposed to write "colour" or "cancelled" instead of "color" or "canceled", because we programmers use the American spelling everywhere.

motleytech: English has bad spellings and its getting worse as it borrows words from european languages. I just learnt the pronunciation for Fuchsia, Gnocchi and San Jose.... its just terrible.

Anyway, I will undo the s -> z change. Its wasn't wrong before my change, so no need to change it now.

Issue #3725, opened 2015-01-03
jQuery undefined in pages admin


mobula:
I'm getting the following error when browsing the pages admin: 

    Uncaught TypeError: undefined is not a function  tree_component.js:476 

I'm trying to get django-cms working along django-grappelli, which defines [django.jQuery](https://github.com/sehmaschine/django-grappelli/blob/master/grappelli/static/grappelli/js/grappelli.js#L12), which breaks [tree_component.js](https://github.com/divio/django-cms/blob/master/cms/static/cms/js/jstree/tree_component.js).

A fix I found was to simply move the *[load jstree](https://github.com/divio/django-cms/blob/master/cms/templates/admin/cms/page/tree/base.html#L32)* block after loading [cms.base.js](https://github.com/divio/django-cms/blob/master/cms/templates/admin/cms/page/tree/base.html#L37) in https://github.com/divio/django-cms/blob/master/cms/templates/admin/cms/page/tree/base.html:
 
    
    
    <script src="{% static "cms/js/modules/jquery.ui.custom.js" %}"></script>
    <script src="{% static "cms/js/modules/cms.base.js" %}"></script>
    <script src="{% static "cms/js/modules/cms.changelist.js" %}"></script>
    
    <script src="{% static "cms/js/jstree/_lib/_all.js" %}"></script>
    <script src="{% static "cms/js/jstree/tree_component.js" %}"></script>

I'm not too comfortable with all those jQuery noConflict dependencies, but that seems relatively safe, and could help with other modules that may define django.jQuery.

Side note: I think it could be nice to make jQuery and jQuery UI urls / loading configurable. Would allow to load them from CDN instead, and not have different versions loaded in the same page by different apps.

jrief:
As a developer I'm not a big fan of a CDN solution. This is because the software shall be self-contained and runnable without requiring any connection to the Internet.

On the other side, I also dislike the bad practice to add files into your own repository, which are revisioned in another repository.
Therefore my suggestion is to add a file ``bower.json`` to the repository. It contains all the non-Python dependencies in a reproducible manner.

A developer cloning ``django-cms`` from GitHub, then must checkout these external dependencies with ``bower install --require``. For normal installations, these files safely can be added to the egg or wheel package.

Side note: In https://github.com/jrief/djangocms-cascade first I started with adding external JS/CSS to my repo. Then I moved over to a CDN solution and had big trouble during development when I was without Internet. My final approach to the problem is to use the http://bower.io/ solution.

[...]

PR #3730, opened 2015-01-06
resolve divio/django-cms#3725: move load jstree *after* cms js libs

mobula:
resolve divio/django-cms#3725: move load jstree *after* cms js libs to fix jQuery undefined issue with 3rd party packages.


yakky:
@FinalAngel does this look sane?

FinalAngel:
should be fine to me as the cms.xxx.js scripts do not yet interact with the jstree @yakky :+1:

yakky:
Thanks @mobula!

Issue #3436, opened 2014-09-06
Swapped User logic incorrect in Django 1.7


kevingill1966:
in permissionmodels.py there is a piece of logic to determine if the user is swapped. There are two paths through the code, one for Django 1.6 and one for other versions of Django. The wrong branch is taken in Django 1.7. The logic should, in my opinion be, 

if DJANGO_1_6 or DJANGO_1_7:

and DJANGO_1_7 should be imported at the top.

linix-mod:
Moved mentioned migration-models 
> "PageUserGroup" and "PagePermission" 

from original origin in file 

> cms/migrations/0002_auto_20140816_1918.py 

to initial file:

> cms/migrations/0001_initial.py 

and did a sqlmigrate on both files (sqlmigrate cms 0001_initial & sqlmigrate cms 0002_auto_20140816_1918).

Works like a charm without any more modifications!

At https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#substituting-a-custom-user-model  its described that: 

> "AUTH_USER_MODEL is created in the first migration of its app (usually called 0001_initial); otherwise, you will have dependency issues"

 "PageUserGroup" and "PagePermission" access "AUTH_USER_MODEL" thats the reason why they need to be moved to 0001_initial.py file.

PR #3741, opened 2015-01-09
Mention MIGRATION_MODULES for Django 1.7 users in integration docs.

zerok:
If you don't have the correct `MIGRATION_MODULES` set before running migrate you end up with similar error messages as have been described in https://github.com/divio/django-cms/issues/3436.

### DIFF
--- a/docs/how_to/integrate.rst
+++ b/docs/how_to/integrate.rst
@@ -495,6 +495,27 @@ setting should look like::
     }
 
 
+Given django CMS's support for Django 1.6.x, Django 1.7 (or later) users have to
+specify where the migrations are situated using the ``MIGRATION_MODULES``
+setting::
+
+    MIGRATION_MODULES = {
+        'cms': 'cms.migrations_django',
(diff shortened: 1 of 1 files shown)


yakky:
LGTM
@evildmp, any comments?

yakky:
Fix #3741

docs/how_to/integrate.rst
```diff
@@ -495,6 +495,27 @@ setting should look like::
     }
 
 
+Given django CMS' support for Django 1.6.x, Django 1.7 (or later) users have to
```
evildmp: "django CMS's support"

zerok:
@evildmp ah, I always thought that if the word already ends with an s there is no additional s supposed to be after the apostrophe. I've looked it up and you are absolutely correct as the additional s would be pronounced in this case. Hurray for learning something :-D Corrected it and spashed the commit.

yakky:
thanks @zerok for the lightning fix!

PR #3749, opened 2015-01-11
Emailuserapp shouldn't be in the list of apps

mkoistinen:
If you want to run makemigrations for emailuserapp do this:

    AUTH_USER_MODEL='emailuserapp.EmailUser' python develop.py makemigrations


yakky:
How about adding it if ``AUTH_USER_MODEL`` is set?

PR #3750, opened 2015-01-11
Fix command block in plugins.html

dirtycoder:
Without the second ':' both lines appear as one, and '--init' as '-init' which can cause confusion because passing '-init' will result in an error message.

### DIFF
--- a/docs/introduction/plugins.rst
+++ b/docs/introduction/plugins.rst
@@ -183,7 +183,7 @@ something like this::
         <input type="submit" value="Vote" />
     </form>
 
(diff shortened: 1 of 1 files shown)


yakky:
@mkoistinen I guess we can merge this manually on support branch. Test failure is totally unrelated

mkoistinen:
This PR is based on develop, not support/3.0.x. I'm directly making the corresponding change in the 3.0.x branch, and I've re-tagged this for 3.1 so it is not forgotten for that release.

PR #3753, opened 2015-01-12
Revert "Added toolbar sanity check to ensure there is valid request.user...

yakky:
Reverts divio/django-cms#3672


jsma:
Was there an issue with the sanity check that needed to be reverted or was there more comprehensive solution in the develop branch that made this fix unnecessary?

yakky:
It's been an emergency revert because this caused a regression in which the toolbar didn't show at all for anonymous user, making impossible to login from the frontend
Correct fix is in #3755

jsma:
Got it, thanks!

PR #3755, opened 2015-01-12
Correct fix for #3672

### LINKED ISSUE #3659: Toolbar middleware breaks custom logout template

yakky:
Fix #3672 
Fix #3659

### DIFF
--- a/cms/test_utils/project/templates/registration/logged_out.html
+++ b/cms/test_utils/project/templates/registration/logged_out.html
@@ -1,12 +1,10 @@
 {% extends "base.html" %}
 {% load i18n %}
 
-{% block breadcrumbs %}<div class="breadcrumbs"><a href="{% url 'admin:index' %}">{% trans 'Home' %}</a></div>{% endblock %}
 
 {% block content %}
 
 <p>{% trans "Thanks for spending some quality time with the Web site today." %}</p>
 
-<p><a href="{% url 'admin:index' %}">{% trans 'Log in again' %}</a></p>
 
 {% endblock %}
--- a/cms/test_utils/project/templates/registration/logged_out.html
+++ b/cms/test_utils/project/templates/registration/logged_out.html
@@ -1,0 +1,12 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+{% block breadcrumbs %}<div class="breadcrumbs"><a href="{% url 'admin:index' %}">{% trans 'Home' %}</a></div>{% endblock %}
+
+{% block content %}
+
+<p>{% trans "Thanks for spending some quality time with the Web site today." %}</p>
+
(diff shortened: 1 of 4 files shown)


yakky:
This will require a different implementation for 3.1

yakky:
@mkoistinen would you check this? I'd like to avoid another regression in the toolbar for the same bug :D

jsma:
I just tried to reproduce #3659 on the latest in the develop branch (haven't tried this PR) but I can't reproduce the bug anymore, logging out works. I haven't followed recent developments in develop the past few weeks but perhaps something there fixed this issue already? Or were your tests here failing without your patch to toolbar.py? Either way, thanks! 

I started working on a test case for this but turned into an exercise in yak shaving (mostly due to my unfamiliarity with how the django-cms test suite is organized) and wasn't able to complete. Looks like I can delete that git stash now and I can learn from your example here instead for next time, thanks again!

yakky:
Actually no changes directly related to this in develop.
I'll rebase the test to check.
The toolbar framework will go through a bit of refactoring in develop, so thi will be fixed in quite a different way

cms/test_utils/project/templates/registration/logged_out.html
```diff
@@ -0,0 +1,10 @@
+{% extends "base.html" %}
+{% load i18n %}
+
+
+{% block content %}
+
+<p>{% trans "Thanks for spending some quality time with the Web site today." %}</p>
```
mkoistinen: Some may see this as "passive aggressive". "Web site" should probably just be "website".

Perhaps we should just say something to confirm the action: "You have successfully logged-out." perhaps?

yakky: This is just the standard django logout message 
This template it's just for testing purposes, it's never shown to the user

jsma:
Haven't investigated further yet but did notice that when I logout, the toolbar appears on the logged out template with the login form displayed in the toolbar. Navigate to another page, the toolbar is cleared. So maybe *something* changed the logic so it smartly shows the login form instead of assuming a logged in user which resulted in the exception I experienced previously. Again, this is only on the develop branch.

jsma:
As mentioned above the bug was apparently fixed through an unrelated change. It's been too long since I've looked at this code so have no comment on the small change to the "is_staff" check but the tests will be good to have to prevent regressions.

PR #3758, opened 2015-01-13
Removed slash from regexp when 'settings.APPEND_SLASH' is False

mmarzantowicz:
Removed slash which was accidentally (?) introduced during code refactoring. This is a minor change.


mkoistinen:
This has been committed to the 3.0.x branch in commit: 49eb3b0cfbf62c5b87ef1501ac63204f763dbfbf. Re-tagged with 3.1 so it won't be overlooked there.

PR #3768, opened 2015-01-15
Add render_uncached_placeholder templatetag

yakky:
This templatetag is a version of `render_placeholder` which does not use cache, in a similar way to `show_placeholder`, `show_uncached_placeholder`

### DIFF
--- a/docs/reference/templatetags.rst
+++ b/docs/reference/templatetags.rst
@@ -157,7 +157,7 @@ render_uncached_placeholder
 ===========================
 
 The same as :ttag:`render_placeholder`, but the placeholder contents will not be
-cached taken from the cache.
+cached or taken from the cache.
 
 Arguments:
 
--- a/docs/reference/templatetags.rst
+++ b/docs/reference/templatetags.rst
@@ -110,7 +110,7 @@ Example::
     {% static_placeholder "footer" site or %}There is no content.{% endstatic_placeholder %}
 
 
-.. templatetag:: show_placeholder
+.. templatetag:: render_placeholder
 
 render_placeholder
 ==================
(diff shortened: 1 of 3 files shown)


yakky:
#3702 rebased on develop

yakky:
Ping @ojii @mkoistinen @digi604 this fill in a gap in the api

docs/reference/templatetags.rst
```diff
@@ -147,7 +151,29 @@ only the english plugins:
     When used in this manner, the placeholder will not be displayed for
     editing when the CMS is in edit mode.
 
+.. templatetag:: render_uncached_placeholder
+
+render_uncached_placeholder
+===========================
+
+The same as :ttag:`render_placeholder`, but the placeholder contents will not be
+cached taken from the cache.
```
mkoistinen: cached **or** taken from the cache

mkoistinen:
@yakky, it is unclear to me what *should* happen if someone adds a plugin with `cache=True` to a placeholder that is rendered with either `show_uncached_placeholder` or `render_uncached_placeholder`. What is the correct behaviour? Also, once we decide this, can we add some tests to protect the decided upon behaviour? I can see this getting flip-flopped a lot over time.

Otherwise LGTM

yakky:
Caching is only at placeholder level, and `cache = True` only means: "Ok placeholder, I can be cached if you're up to".
So using `show_uncached_placeholder` / `render_uncached_placeholder` effectively disables **any** caching (both read and write), ignoring plugins attribute.
The current test coverage should already address this

yakky:
@mkoistinen task completed

mkoistinen:
lgtm

PR #3780, opened 2015-01-20
Change installer docs

yakky:


yakky:
@evildmp could you check this warning? This is needed because since 0.7.1 the installer check the project directory to be empty before running to avoid deleting user files in case of cleanup

docs/introduction/install_from_scratch.rst
```diff
@@ -46,6 +46,11 @@ Django project called ``mysite``::
 
     djangocms -p . mysite
 
+.. warning:: directory ``.`` is meant to be empty at this stage and djangocms-installer will check
+             for this; if you add some content to this directory before running the installer
+             because of you project layout, please use ``-s`` flag to skip the check.
```
evildmp: How about:

djangocms-installer expects directory ``.`` to be empty at this stage, and will check for this, and will warn if it's not. You can get it to skip the check and go ahead anyway using  the ``-s`` flag; **note that this may overwrite existing files**.

Issue #3786, opened 2015-01-21
Add management command to copy_site


peterfarrell:
We have the ability to `copy_lang` but no ability to "clone" an entire cms site to a new site.

PR #3785, opened 2015-01-21
Fixes Issue #3782

peterfarrell:
The circular import issue with this leaves something to be desired....

PR #3787, opened 2015-01-21
Adds support for `cms copy-site --from X --to X` Issue #3786 

peterfarrell:
Adds a management subcommand to support "cloning" an entire cms site to a new site.


peterfarrell:
FYI, you must apply PR 3785 in order for this PR to work.

https://github.com/divio/django-cms/pull/3785

yakky:
@peterfarrell thanks for this, I really like it
waiting to merge this after we fix the circular import issue

yakky:
Fix #3786

peterfarrell:
Great, glad you like it! Let me know how you fixed the circular import issue?

yakky:
Documentation (see http://django-cms.readthedocs.org/en/support-3.0.x/reference/cli.html) needed, and at least test for it  (see https://github.com/divio/django-cms/blob/develop/cms/tests/management.py#L318 for reference)

yakky:
subcommands can only use arguments, not options (see https://github.com/divio/django-cms/blob/develop/cms/management/commands/subcommands/copy_lang.py#L27 as example)

yakky:
In #3826 ``Page.copy_page`` has been reworked to allow ``None`` target, please change the command accordingly, as passing a page as first parameter will result in broken tree

yakky:
@peterfarrell #3785 has been merged. could you update this as per comment above?

peterfarrell:
@yakky - definitely, apologies -- I didn't realize that the custom sub-commands don't work like regular commands.

yakky:
No need to apologise, and thanks for spending your time contributing to django CMS

yakky:
@peterfarrell I'd like to merge this in time for 3.1: any ETA for this?

PR #3805, opened 2015-01-28
Fixed KeyError and UnboundLocalError exceptions on page revision view

747project:
- centralized check to see if an admin URL request is for history/recovery
- fixed KeyError exception ("Key u'<problematic_placeholder_name>' not found in Form"),
  raised when a page revision is viewed and template for the page has changed
  since the time the revision has been stored, specifically when a new
  placeholder has been added that did not exist in the saved revision
- fixed UnboundLocalError exception ("local variable 'placeholder' referenced
  before assignment"), which occurs when a page revision is viewed and no
  placeholders are found for some reason.

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -240,7 +240,7 @@ class PageAdmin(ModelAdmin):
         target = request.GET.get('target', None)
         position = request.GET.get('position', None)
 
-        if 'recover' in request.path:
+        if self._is_recovery(request):
             pk = obj.pk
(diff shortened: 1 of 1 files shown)


747project:
This fixes Issue #3814 and Issue #3815

czpython:
@yakky is this still relevant? given we've dropped 2.4 support.

yakky:
@czpython I think we can merge this, even if we're not going to release any further 2.4 releases

PR #3808, opened 2015-01-31
Do not enter structure mode if the mode switcher is not present

### LINKED ISSUE #3667: Stuck in draft/structure mode
When the user who has edit permission on certain page goes to draft/structure view and then navigates to some other page, one that he has no edit permission, he is stuck in draft/structure without any intuitive way back. This issue is easy to reproduce and very unfortunate for the end user.

yakky:
Fix #3667


yakky:
@mkoistinen issue should have been fixed for real, please consider this for 3.0.10

Issue #3818, opened 2015-02-04
Logging back in via toolbar on the logout page logs the user out


jsma:
The toolbar login form is set to always redirect to the current page on submission. When the user logs out, the login form in the toolbar is displayed on the logged out page prompting users to log back in. However, if the user logs in from this page, they will be redirected back to the logged out view which of course logs them back out.

mkoistinen:
This was closed in 3.0.10.

PR #3819, opened 2015-02-04
Prevents login form from redirecting to logout.

jsma:
Refs #3818


yakky:
is ``logout_url`` a standard context variable?

jsma:
I'm using the "url as" syntax, the actual url name is "logout" which is the Django default.

yakky:
Sorry, I missed the first added line

jsma:
No worries. If there is an alternative preferred approach, let me know.

PR #3835, opened 2015-02-08
hide all submit-rows in cms.modal.js

febsn:
As buttons are added to the CMS interface, all submit-rows provided by the changeform should be hidden, not only the first one (e.g. when save-on-top is active).
I guess my proposal should be reviewed by someone familiar with django-suit.


yakky:
@FinalAngel could you review this?

FinalAngel:
Does this also work without django-suit @yakky ?

yakky:
Yes, and actually save_on_top is currently broken, so pushing this to 3.0.x milestone
@mkoistinen this should be fixed in 3.0, but will require extensive *manual* testing with different admin setups: basically the more admin-in-modal window we can test, the better

PR #3844, opened 2015-02-09
Fix get_title_obj call

### LINKED ISSUE #3812: Non-existent method Page.get_title_object
If you specify a language to `ExtensionToolbar.get_title_extension_admin`, it calls `Page.get_title_object`, which doesn’t exist.  This should probably be `Page.get_title_obj` instead.

yakky:
Fix #3812


yakky:
@BertrandBordage this should fix the related issue

BertrandBordage:
Yes :) I already applied it as a workaround.

Issue #3845, opened 2015-02-09
cms.api.publish_page not setting changed_by user on page


morganwahl:
The `publish_page` function in `cms.api` takes a `User` instance in order to check permissions, but the `Page` published then has it's `changed_by` field set to `'script'`.

Looking at the `Page` save method it seems it gets the user from `cms.utils.permissions._thread_locals`. My quick-fix would be to just set that attribute in the `publish_page` function, but I can't say I understand what all the repercussions of that would be.

morganwahl:
Fix PR'd in #3848

PR #3848, opened 2015-02-10
Set Page.changed_by correctly in cms.api.publish_page()

morganwahl:
Fix for issue: https://github.com/divio/django-cms/issues/3845

Since I'm not entirely sure of the repercussions of touching `cms.utils.permissions._thread_locals`, I wrote a little contextmanager that uses `get_current_user` and `set_current_user` and then used that in `publish_page`. I figured this was least likely to have surprising effects.

I also didn't see a test for `publish_page` at all, so I added one that does a basic test of permissions checking, and a regression test for this bug.

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -450,8 +450,9 @@ def publish_page(page, user, language):
         raise PermissionDenied()
     # Set the current_user to have the page's changed_by
     # attribute set correctly.
-    # 'user' is a User object, but current_user() just wants the username.
-    with current_user(user.username):
+    # 'user' is a user object, but current_user() just wants the username. Note
+    # that not all user models have `username`.
+    with current_user(unicode(user)):
         page.publish(language)
     return page.reload()
 
--- a/cms/api.py
+++ b/cms/api.py
@@ -32,7 +32,7 @@ from cms.plugin_pool import plugin_pool
 from cms.utils import copy_plugins
 from cms.utils.conf import get_cms_setting
 from cms.utils.i18n import get_language_list
-from cms.utils.permissions import _thread_locals
+from cms.utils.permissions import _thread_locals, current_user
 from menus.menu_pool import menu_pool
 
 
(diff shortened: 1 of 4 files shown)


cms/models/pagemodel.py
```diff
@@ -179,7 +181,7 @@ def move_page(self, target, position='first-child'):
                 if target.get_root().get_next_sibling().pk == public.get_root().pk:
                     target = target.publisher_public
                 else:
-                    Logger.warning('tree may need rebuilding: run manage.py cms fix-tree')
+                    logger.warning('tree may need rebuilding: run manage.py cms fix-tree')
```
morganwahl: Unrelated bugfix. `Logger` doesn't have this method. Presumably there's no test for this path, but I have no idea where to begin writing one.

cms/api.py
```diff
@@ -448,7 +448,12 @@ def __init__(self, user):
     request = FakeRequest(user)
     if not page.has_publish_permission(request):
         raise PermissionDenied()
-    page.publish(language)
+    # Set the current_user to have the page's changed_by
+    # attribute set correctly.
+    # 'user' is a user object, but current_user() just wants the username. Note
+    # that not all user models have `username`.
+    with current_user(unicode(user)):
```
yakky: Check User.get_username() ( https://docs.djangoproject.com/en/1.7/ref/contrib/auth/#django.contrib.auth.models.User.get_username) which is guaranteed to exist on any user model

yakky:
Nice job!
Tests failures are related to the unicode function; In this case use the mentioned get_username method
Generally speaking in Django projects just use ``force_text`` from ``django.utils.encoding``.

morganwahl:
Ah; thank you for the pointers. Should've read the docs before using the model...

cms/models/pagemodel.py
yakky: This is a good starting point https://github.com/divio/django-cms/blob/develop/cms/tests/page.py#L1226-L1234

Corrupting the parent node by setting ``parent.get_public_object().get_root().pk`` to something else should be enough to trigger the warning

yakky:
Thanks @morganwahl!

PR #3853, opened 2015-02-11
Trigger urls_need_reloading signal on slug change

### LINKED ISSUE #3689: urls_need_reloading signal not firing when slugs are changed
On 3.0.6, 3.0.7 and master (dabf832a10a7f0687610fb33d3103bcb984d3973) the ``urls_need_reloading`` signal does not fire in the following cases. But it should.:

* the slug of an app-hooked page changes (and the change is published)
* the app-hooked page is moved
* the slug of a parent page changes (and the change is published)

This might be a regression from #2911 .

ojii:
Fixes #3689

### DIFF
--- a/cms/signals/apphook.py
+++ b/cms/signals/apphook.py
@@ -41,12 +41,25 @@ def apphook_post_title_checker(instance, **kwargs):
     old_title = getattr(instance, '_old_data', None)
     if not old_title:
         if instance.page.application_urls:
-            request_finished.connect(trigger_restart, dispatch_uid=DISPATCH_UID)
+            request_finished.connect(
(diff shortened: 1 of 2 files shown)


ojii:
@stefanfoulis please verify this fixes your issue.

mkoistinen:
@ojii thanks for taking a stab at this! Travis is failing because of the pyflakes linkter. We simply need to remove the unused import here: https://github.com/ojii/django-cms/blob/fix-3689/cms/tests/signals.py#L9

I've run tests locally and they all pass. I'm also chasing @stefanfoulis for a concise test that this fixes the original issue.

PR #3861, opened 2015-02-13
Fix permission detection on placeholders

yakky:
Currently placeholders in model is pushed to ``request.placeholders`` disregarding actual user permission on them; the user can't add content to them anyway as permission is enforced upon plugin insertion, but this can be confusing for users.
This PR check for user permissions on placeholders when populating ``request.placeholders`` ensuring that the users can only see the placeholders they have rights on.


yakky:
@mkoistinen please consider this for 3.0.10

PR #3868, opened 2015-02-14
Live/Draft renaming

yakky:
This changes mode switcher labels from **Live / Draft** to **Edit / Save and close**
This should be easier and more understandable for users, withouth the need to know the differences between the two modes
Documentation changes needed

Fix #2971


mkoistinen:
@Yakky, why is this marked "needs design decision"?

yakky:
A way to say that it needs not just a formal review, but an agreement on the change proposal

PR #3870, opened 2015-02-14
Implement ajax_add action

yakky:
Allows to define a toolbar action that adds a plugin
Adds the ability to define custom per-toolbar item actions in the process

### DIFF
--- a/cms/static/cms/js/modules/cms.plugins.js
+++ b/cms/static/cms/js/modules/cms.plugins.js
@@ -456,6 +456,16 @@ $(document).ready(function () {
 			}
 		},
 
+		editPluginPostAjax: function(caller, toolbar, response){
+			if (typeof toolbar == 'undefined' || typeof response == 'undefined') {
+				return function(toolbar, response) {
+					var that = caller;
+					that.editPlugin(response['url'], that.options.plugin_name, response['breadcrumb']);
+				}
+			}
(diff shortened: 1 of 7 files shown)


yakky:
@mkoistinen @ojii @FinalAngel  ready  for review

mkoistinen:
lgtm

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -456,6 +456,16 @@ $(document).ready(function () {
 			}
 		},
 
+		editPluginPostAjax: function(caller, toolbar, response){
+			if (typeof toolbar == 'undefined' || typeof response == 'undefined') {
+				return function(toolbar, response) {
+					var that = caller;
+					that.editPlugin(response['url'], that.options.plugin_name, response['breadcrumb']);
```
FinalAngel: just use response.url

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -456,6 +456,16 @@ $(document).ready(function () {
...
+				}
+			}
+			that.editPlugin(response['url'], that.options.plugin_name, response['breadcrumb']);
```
FinalAngel: same here

FinalAngel:
added some comments, even if too late :)

yakky:
@FinalAngel it's never too late :)

Issue #3873, opened 2015-02-14
Templatetag: "render_model_add" should allow you to wrap any arbitrary content


mkoistinen:
Currently, {% render_model_add %} is sorta limited to presenting an icon, which, to be fair you can replace and re-position with CSS. I think it would be tremendously more useful if it could be used to wrap any arbitrary content. This would provide the developer a lot more options in creating frontend interfaces that are more in-keeping with the style and format of the site they are building.

yakky:
Fixed by #3875

PR #3875, opened 2015-02-15
Add render_model_add_block tag

mkoistinen:
OK, this works nicely to satisfy #3873, but would be even better if ClassyTags would let me declare the nodelist block "argument" as, optional. Anyone know if this is possible?


mkoistinen:
@ojii, is it possible to have a single, but optional block/nodelist defined?

mkoistinen:
Thats odd. The GitHub integration with Travis says this build failed, but if you click through, it shows all passing tests...

yakky:
Travis, go home, you're drunk! :)

yakky:
@mkoistinen can you revert the doc change regarding ``render_model_add``? I have a fix for that
Or just leave it and I'll rebase my PR on top of this after the merge

ojii:
> @ojii, is it possible to have a single, but optional block/nodelist defined in an InclusionTag?

No, not without a heavy dose of dark magic. "endblock detection" happens at parse time, not at render time. Argument processing (including the one that would signal if it should scan for an endblock or not) happens at render time.

mkoistinen:
@yakky - done
@ojii, ok, thanks, this implementation will stand then.

Issue #3880, opened 2015-02-18
Placeholder "Empty all" remove plugins in all the languages


pevar:
The "Empty all" functionality in the "Draft | Structure" view, removes the plugins from the placeholder for all the localized versions of the page, while the "Delete Confirm Panel" message is only listing the plugins for the page in the current languages.

In depth:
Here https://github.com/divio/django-cms/blob/master/cms/admin/placeholderadmin.py#L542 you collect the plugins for the "Delete Confirm Panel", filtering by language.

Here https://github.com/divio/django-cms/blob/master/cms/admin/placeholderadmin.py#L553 you call the placeholder.clear() method without the language information.

Very dangerous for multi-languages sites, loosing a lot of contents without realizing it.

yakky:
Fixed in #3881

PR #3881, opened 2015-02-18
Empty All in placeholder remove plugins from all the localized version of the page

pevar:
Solve Issue #3880


yakky:
Nice, thanks!

yakky:
How about writing a test for this?

mkoistinen:
I'd like to get this into 3.0.11, which has a target "cutoff point" of 28-Feb.

yakky:
@pevar do you need any help to write the tests to get this merged?

pevar:
Yes please, first time for me.

yakky:
@pevar steps are:

 * create a page
 * create a few plugins (see https://github.com/divio/django-cms/blob/support/3.0.x/cms/tests/plugins.py#L324-L336)
 * create a title (this means: translate the page at first step by using create_title) as in https://github.com/divio/django-cms/blob/support/3.0.x/cms/tests/multilingual.py#L182
 * add plugins to the title (as in step 2, but setting the language according to the title you just created)
 * use login_context to login as admin (https://github.com/divio/django-cms/blob/support/3.0.x/cms/tests/admin.py#L82)
 * run the clear_placeholder view (as in https://github.com/divio/django-cms/blob/support/3.0.x/cms/tests/admin.py#L1149-L1151) (by default it runs in english)
 * after the view has run check that the placeholder still contains plugin in the other language (something like `self.assertEqual(ph_en.get_plugins().count(), #)` would do, where "#" is the number of plugins you added in step 4

Join #django-cms IRC channel for further help

pevar:
@yakky 
Thanks, which module in test folder I should put my code in?

yakky:
I think ``cms.tests.admin`` is the best one

yakky:
@pevar wow, really nice!
Te failing test it's just a pyflake check, you just need to adjust the test code style (please see the failure text on travis or run ``develop.py test cms.AboveStaticAnalysisCode`` on you local machine)

pevar:
@yakky, it's not really clear to me what travis is saying. I've now removed some unused variables to get the style right.

yakky:
@pevar exactly :)
We have this pyflake check in the testsuite that runs pyflake on the codebase and it fails if pyflake report some issues.
You can run this check locally with `develop.py test cms.AboveStaticAnalysisCode`

yakky:
@mkoistinen LGTM

yakky:
@mkoistinen this may be ported to support/3.0.x too, I think

yakky:
@pevar congrats for your first contribution to django CMS, well done!

pevar:
@yakky thanks for the help! I'm very happy to contribute.

Issue #3882, opened 2015-02-18
Tutorial documentation should include the property "cache=False"


mkoistinen:
The plugin tutorial located here: http://django-cms.readthedocs.org/en/latest/how_to/custom_plugins.html provides example code for displaying the current, logged-in user. Unfortunately, this is cached and displays incorrect output.

Fortunately, it provides a good example of the `cache` property!

yakky:
Fixed in #3883

PR #3883, opened 2015-02-19
Include `cache = False` in the 'simplest plugin' example.

mkoistinen:
... and provide description of this property. This should satisfy #3882.


yakky:
LGTM

PR #3886, opened 2015-02-20
Non serializable objects shall be handled by `default()` function

jrief:
The Django docs state: https://docs.djangoproject.com/en/1.7/topics/serialization/#serialization-formats-json

REST framework does it as well: https://github.com/tomchristie/django-rest-framework/blob/4248a8d3fc725d9ae3fe7aaaad7ee12479ab07ab/rest_framework/utils/encoders.py#L21

we could even consider to support more types inside that ``default()`` method.


yakky:
Thanks @jrief

PR #3887, opened 2015-02-20
Use "content_type" in Djangos > 1.4

mkoistinen:
Should address #3885.

I also searched the code base for "mimetype" to ensure we don't have any other lingering cases. I think we're good this time.


yakky:
LGTM

PR #3888, opened 2015-02-21
Allow using render_model_add by passing a class

### LINKED ISSUE #3874: Templatetag "render_model_add" requires a model instance
Allow to use ``render_model_add`` by passing a class as argument instead of  an instance

yakky:
Fix #3874

### DIFF
--- a/cms/tests/templatetags.py
+++ b/cms/tests/templatetags.py
@@ -400,7 +400,7 @@ class NoFixtureDatabaseTemplateTagTests(CMSTestCase):
         context = RequestContext(request, {'category': Category()})
         with self.assertNumQueries(0):
             output = template.render(context)
-        expected = 'cms_plugin cms_plugin-sampleapp-category-add-None '
+        expected = 'cms_plugin cms_plugin-sampleapp-category-add-0 '
         'cms_render_model_add'
         self.assertIn(expected, output)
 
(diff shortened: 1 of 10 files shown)


yakky:
@mkoistinen you ok with this?

cms/tests/templatetags.py
```diff
@@ -400,7 +400,7 @@ def test_render_model_add(self):
         context = RequestContext(request, {'category': Category()})
         with self.assertNumQueries(0):
             output = template.render(context)
-        expected = 'cms_plugin cms_plugin-sampleapp-category-add-None '
+        expected = 'cms_plugin cms_plugin-sampleapp-category-add-0 '
```
mkoistinen: Awesome, this bothered me a lot. Glad its fixed.

mkoistinen:
Looks like this allows a model class to be passed, which is nice. Does this also allow picking up pre-set values from a passed instance? I can't immediately tell, but the changes to the docs don't mention it.

yakky:
Nope, this is just to add support for class. #3874 has been reverted to just this use case.
The default values use case has been opened separately

mkoistinen:
I vote we apply this to 3.0.x as this is how it was supposed to work from the beginning.

yakky:
Ok, I'll close this and reopen on 3.0.x

yakky:
Supersed by #3899

Issue #3889, opened 2015-02-21
Document redirection API


yakky:
Redirection after editing object in the frontend requires action by the developer.
Currently documentation is kind of scattered around, and we need to fix this by providing a clear an unified documentation on how to properly implement this.

Key topics:
* implement ``toolbar.watch_models`` attribute to declare supported models
* implement ``object.get_public_url()``, ``object.get_draft_url()`` to take the publishing state into account
* implement ``request.toolbar.set_object()`` to properly set the current object on the toolbar
* implement the detail view such as the object can always be accessed by staff users when the toolbar is active even if the object is not published (according to the application-specific semantic of *published objects*)
* keep the unpublished objects out of the listviews when in edit mode

evildmp:
Does this now have the documentation it requires? There's such a tangled web of issues and patches here that it's hard to see.

yakky:
Current implementation (still to be merged) is https://github.com/divio/django-cms/pull/3894/files#diff-c47ae0bcfc004b5520e9654aa182096eR11

yakky:
Fixed in #3894

PR #3894, opened 2015-02-22
Fix redirection

### LINKED ISSUE #3654: Delete page shouldn't result in 404
Currently when you delete a page, the user is redirected to the url of the page that was just deleted. It should instead redirect to the parent page, if one exists or to the home page rather than giving the user an error page.

### LINKED ISSUE #3655: Save after updating slug should redirect to new slug
Currently if you change the slug for a page and save, you are redirected back to the page at the old slug, which gives a 404 error. The user should instead be redirected to the page's new slug.

### LINKED ISSUE #3766: 404 when logging out after creating a new page without publishing
I get a 404 with the following steps:
- Login to cms
- create a new page
- leave it unpublished
- logout via toolbar menu sitename > logout username

Might be related to #2770

### LINKED ISSUE #3895: Document redirection behavior
A proper documentation for redirection when in frontend mode must be written.

Currently the behavior is the following:
- after creating or editing a new page or a model with toolbar which implements ``Toolbar.watch_models`` (see #3889) the frontend is redirected to the instance absolute URL
- after deleting the object, frontend is redirected to the home page (a better behavior would be to redirect the user to the parent page)
- after logout, user is redirected to the home page (ditto)

yakky:
Fix #3766 by redirecting anonymous users to the home page
Fix #3654 by not reloading the page after setting the new URL
Fix #3655 by not reloading the page after setting the new URL
Fix #3895 by documenting redirection behavior

This depends on #3892 

Linked to #3889 to document the attribute needed to handle redirection with normal Django models

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -1360,6 +1360,8 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
                 return HttpResponse('/', mimetype='text/plain')
             else:
                 return HttpResponse('/', content_type='text/plain')
+        obj = False
+        url = False
         if request.session.get('cms_log_latest', False):
             log = LogEntry.objects.get(pk=request.session['cms_log_latest'])
             try:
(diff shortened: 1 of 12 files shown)


yakky:
@mkoistinen this must be thoroughtfully checked.

yakky:
refactoring of redirection logic to take unpublished pages into account

mkoistinen:
Pushed to 3.0.12 to get more time for testing this rather sizeable PR.

evildmp:
Looking at docs now

mkoistinen:
This really needs to have supporting tests, IMHO.

yakky:
@mkoistinen tests are already included in this PR

mkoistinen:
My bad, sorry. Assigning myself for testing.

PR #3896, opened 2015-02-22
Change marker for custom js string substitution

### LINKED ISSUE #3641: wrong mdgid for "Are you sure you want to %s this page?"
The msgid in django.po is "Are you sure you want to %%s this page?" but the real msgid should be "Are you sure you want to %s this page?" (only one "%").

The original msgid is from here : https://github.com/divio/django-cms/blob/support/3.0.x/cms/templates/admin/cms/page/tree/base.html#L57
And the msgid in django.po is here : https://github.com/divio/django-cms/blob/support/3.0.x/cms/locale/fr/LC_MESSAGES/django.po#L1256

yakky:
Fix #3641


yakky:
Ping @FinalAngel as discussed in the issue, this changes the syntax for the string substitution in the JS

FinalAngel:
Interesting character choice @yakky but is fine for me :)
can you also add some comment referring to this issue?

yakky:
@FinalAngel it's the most "special" in the non-special charset :)
Sure will add

mkoistinen:
LGTM

PR #3899, opened 2015-02-23
Allow using render_model_add by passing a class

### LINKED ISSUE #3874: Templatetag "render_model_add" requires a model instance

yakky:
Fix #3874 in 3.0
Supersedes #3888


yakky:
@mkoistinen ready!

PR #3901, opened 2015-02-24
Toolbar API update

### LINKED ISSUE #3384: Pages 404 when editing a child page whos parent has an unpublished slug change
I've built a site structured heavily around a few pages having many children.

It has recently been handed over to the client & they've decided to change a few of the parent page titles. They've also changed the slugs, and then gone to make edits to child pages. When they switch to edit mode the page 404s because the parent is at a different slug in edit mode.

The obvious solution here is to publish your changes immediately after changing a title/slug but I wondered if this had ever been raised before. I suspect the way CMS switches between edit/live is heavily based in javascript and to avoid this you'd actually have to redirect in order to pick up the changed url and avoid this.

yakky:
Extends the Toolbar API, adding a few features:

 * system plugins: plugins marked with this attribute skip the placeholder configuration checks;
 * add "addons" concept which allows to render extra content from the toolbar (like clipboard plugins, but it's rather generic);
 * add the ``get_public_url`` / ``get_draft_url`` which allows to implement proper redirection when switching draft / live mode; this requires changes to the existing models to support this semantic; extended documentation for this feature must be implemented as part of #3889 

This also adds the system flag to PlaceholderPlugin and AliasPlugin

Fix #3384

### DIFF
--- a/cms/cms_toolbar.py
+++ b/cms/cms_toolbar.py
@@ -98,17 +98,19 @@ class BasicToolbar(CMSToolbar):
     """
     Basic Toolbar for site and languages menu
     """
+    page = None
 
     def init_from_request(self):
         self.page = get_page_draft(self.request.current_page)
 
     def populate(self):
-        self.init_from_request()
+        if not self.page:
+            self.init_from_request()
 
-        self.add_admin_menu()
-        self.add_language_menu()
-        user_settings = self.request.toolbar.get_user_settings()
-        self.clipboard = user_settings.clipboard
+            self.add_admin_menu()
+            self.add_language_menu()
+            user_settings = self.request.toolbar.get_user_settings()
+            self.clipboard = user_settings.clipboard
 
     def add_admin_menu(self):
         admin_menu = self.toolbar.get_or_create_menu(ADMIN_MENU_IDENTIFIER, self.current_site.name)
--- a/cms/cms_toolbar.py
+++ b/cms/cms_toolbar.py
@@ -99,9 +99,6 @@ class BasicToolbar(CMSToolbar):
     Basic Toolbar for site and languages menu
     """
 
-    def __init__(self, request, toolbar, is_current_app, app_path):
-        super(BasicToolbar, self).__init__(request, toolbar, is_current_app, app_path)
-
     def init_from_request(self):
         self.page = get_page_draft(self.request.current_page)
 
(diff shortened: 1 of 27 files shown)


yakky:
@mkoistinen @ojii are you ok to make PlaceholderPlugin and AliasPlugin droppable in any placeholder disregarding of configuration?
Both are actually containters for plugins, somewhat artificial wrappers we need to treat multiple plugins as a whole, so it's not really sensible to be forced  to add to allowed plugins in every placeholder, if plugin restrictions are used

yakky:
Documentation added. Other changes are only part of the internal API

cms/cms_toolbar.py
```diff
@@ -189,6 +195,18 @@ def get_username(self, user=None, default=''):
...
             return default
 
+    def get_clipboard_plugins(self):
+        self.populate()
+        if not hasattr(self, "clipboard"):
+            return []
+        return self.clipboard.get_plugins()
+
+    def render_addons(self, context):
+        context.push()
```
ojii: I'd suggest using ``with context.push():``

docs/reference/plugins.rst
```diff
@@ -182,6 +182,16 @@ Can the plugin be inserted inside the text plugin?  If this is ``True`` then
...
 
+.. _system_plugin:
+
+system
+------
+
+Default: ``False``
+
+Plugins with this attribute set to True can be dropped in any placeholder, overriding
+any placeholder configuration in :setting:`CMS_PLACEHOLDER_CONF`.
```
ojii: do we want this to be a public API? The naming of it alone sounds like something internal that shouldn't be public (possibly even not using that attribute but a hardcoded list of whitelisted plugins). I'm not necessarily saying it shouldn't be a public API, just wondering.

yakky: We can easily change the name of the attribute. I think this might be useful for special custom plugins and it does not really deliver "special powers".
I'm not good at naming, so I am open to less deceiving naming

cms/cms_toolbar.py
yakky: Is it a context manager in django 1.6?

yakky:
@mkoistinen @ojii please check the ``get_draft_ulr`` / ``get_public_url`` addition
This is just a first step to a cleaner generic Django models publishing framework, but just in this phase it will allow easier redirection (albeit it still requires custom implementation)

mkoistinen:
LGTM. I agree with Jonas, though, that this 'master' flag is going to be abused by people.

yakky:
I'm all for changing the name for it (though it won't actually provide any superpower)

ojii:
how about for now we simply don't document that flag to keep it "secret" and less likely to be abused?

mkoistinen:
+1

yakky:
Happily comply.

PR #3903, opened 2015-02-25
Fix extensions copy

### LINKED ISSUE #3551: Copying page does not copy it's extensions
When copying a page, it's page and title extensions do not get copied over. Not sure if this was on purpose though..

yakky:
Add page/title extensions copy logic into ``Page.copy_page`` method
Fix #3551


yakky:
@mkoistinen I think this is a nice fit for 3.0.11

PR #3908, opened 2015-03-01
Fix copy page when permission is disabled

yakky:
Due to a bug in ``SafeJSONEncoder`` boolean ``False`` was rendered as js value ``true`` disabling copy when permissions is disabled


yakky:
Backport #3886 in the process

Issue #3909, opened 2015-03-01
Allow complete deactivation of frontend editing


pgcd:
Apparently, nobody really cares about this - at least I couldn't find anything about it -  but I think it's a pretty simple enhancement, so...

I would propose to add a switch somewhere - it might be a session cookie, or a User field, or a server-level boolean in settings.py, it doesn't really matter - to **completely disable** frontend editing without logging out.
The reason for this is that there are cases where even the menu toggle button affects the usability of the layout, and there are reasons for me to not want always to load the extra CSS & JS used for frontend editing. At the same time, I need to be logged in to do other stuff, so that's not really an option.

Ideally, this could be implemented as a simple URL toggle (like ?edit/?edit_off), that stores the override in session - if you agree with the idea, I can try and implement it myself as a pull request.

yakky:
@pgcd seems totally legit.
Please go ahead.
I think it would be nice to have this as an toggle in the toolbar too (probably in the site menu)

yakky:
Closed by #3912

PR #3912, opened 2015-03-01
Allow disabling toolbar

pgcd:
Basic implementation of a mechanism to entirely disable the toolbar and front-end editing for logged in staff, with link from admin menu.


yakky:
Fix #3909
Nice job!

pgcd:
Thanks, glad to help =)

PR #3934, opened 2015-03-07
Fix resize sideframe timeout

### LINKED ISSUE #3663: Structureboard eats 100% CPU, overheats my laptop
Opening structure view starts an interval timer that triggers 100 times a second, trying to replace all the placeholders, I think:

    var interval = 10;
    ...
    // setup an interval
    this.interval = setInterval(function () {
        $(window).trigger('resize.sideframe');
    }, interval);

<https://github.com/divio/django-cms/blame/develop/cms/static/cms/js/modules/cms.structureboard.js#L265>

Please could you reduce this to something sensible like once per second so that my poor laptop doesn't melt or shutdown from overheating?

yakky:
Fix #3663

### DIFF
--- a/cms/static/cms/js/modules/cms.sideframe.js
+++ b/cms/static/cms/js/modules/cms.sideframe.js
@@ -301,6 +301,9 @@ $(document).ready(function () {
 				// update settings
 				that.settings.sideframe.position = e.clientX;
 
+                // trigger the resize event
+                $(window).trigger('resize.sideframe');
+
 				// save position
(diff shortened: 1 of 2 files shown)


yakky:
This should be a complete fix. See comments on #3663

yakky:
@mkoistinen this is a small change in code, but merging into support will require a bit of live projects testing.

mkoistinen:
Will test, thanks for the patch!

PR #3939, opened 2015-03-12
Fix cache placeholder

jrief:
If a placeholder or one of its child plugins is marked as non-cachable, this information is not propagated down to the caching middleware and thus in https://github.com/divio/django-cms/blob/support/3.0.x/cms/views.py#L245  ``add_never_cache_headers(response)`` is not set, hence these pages are cached upstream, while they shouldn't.


yakky:
@jrief fancy a test?

mkoistinen:
@jrief What's the status on this? Can we get a test for this so we can commit it for 3.0.13?

PR #3943, opened 2015-03-12
Fix #3942

### LINKED ISSUE #3942: Nested plugins are not rendered when using fallback due to recursion in assign_plugins
Due to the recursion, `build_plugin_tree` is called twice, which on the second pass deletes all the `child_plugin_instances` that were assigned in the first pass. The result is that only the root plugins can ever be rendered.

samluescher:
`build_plugin_tree` should only be called once, in the end, after recursion.

### DIFF
--- a/cms/utils/plugins.py
+++ b/cms/utils/plugins.py
@@ -38,7 +38,7 @@ def requires_reload(action, plugins):
                for p in plugins)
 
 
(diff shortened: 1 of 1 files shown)


yakky:
@samluescher thanks for this enhancement

PR #3947, opened 2015-03-13
Fixes #3931 -- Installation documentation is confusing

### LINKED ISSUE #3931: Documentation is rather confusing when it comes to installation
The documentation structure has a menu-item: Tutorials, which has a sub item: "Installing django CMS from scratch". However, this tutorial then goes on to instruct someone to use the django-CMS installer to install. I don't think many of us would consider this "from scratch" anymore than making a cake from a box-mix, rather than from raw ingredients.

Also, I don't think many people would consider installation instructions to be a "Tutorial" so much, as rather "Installation instructions". I realise this may be mincing words a bit, but I think a tutorial here would be more like "how to make a plugin", or other things once you have installed the product.

Finally, in a complete different section of the menu, is another menu: "How-to Guides", which includes the item: "Integrate django CMS into an existing project", which is another set of installation instructions (more akin to installing "from scratch", really).

This is confusing (we see comments on this in IRC, G+ on this regularly).

I would like to propose that installation be separate from "Tutorials" and "How-To Guides". A sub-menu could then provide different installation paths including:

1. djangocms-installer;
2. manual installation;
3. integration with existing Django projects.

evildmp:
This patch retitles the two installation documents (one for automated, one for manual, installation)
and clarifies their purpose.

### DIFF
--- a/docs/how_to/index.rst
+++ b/docs/how_to/index.rst
@@ -8,7 +8,7 @@ the same territory as the :doc:`/introduction/index`, but in more detail.
 .. toctree::
     :maxdepth: 2
 
-    integrate
+    install
     custom_plugins
     menus
     apphooks
--- a/docs/how_to/install.rst
+++ b/docs/how_to/install.rst
@@ -1,6 +1,15 @@
-#############################################
-Integrate django CMS into an existing project
-#############################################
+#############################
+Installing django CMS by hand
+#############################
+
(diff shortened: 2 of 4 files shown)


yakky:
LGTM

evildmp:
@mkoistinen Your comments?

This should address the concerns raised in https://github.com/divio/django-cms/issues/3931, though it doesn't move either of the installation chapters out of their respective sections, since I think they do properly belong there.

PR #3951, opened 2015-03-16
Fixes #3784 -- adds notes on ordering in INSTALLED_APPS

### LINKED ISSUE #3784: Document the plugin order in INSTALLED_APPS
Since develop **all** plugins **must** be defined **after** ``cms``
We need to add this to the documentation (both in the upgrade section and in the normal project setup.
Especially ``djangocms_text_ckeditor`` which still is documented to be installed before ``cms``

evildmp:
`cms` should come before applications with plugins. Notes added to
several pages within the documentation.


yakky:
LGTM

PR #3953, opened 2015-03-16
polish translations corrected

pawelmarkowski:


yakky:
Please use transifex https://www.transifex.com/projects/p/django-cms/language/pl/ for this

Issue #3437, opened 2014-09-08
feature: custom html representation of text plugins in editor


ollb:
This is a proposal for an enhancement related to cmsplugin-link but I think it is wise to open a ticket here since it would make evolve text plugins support.

Actualy, when you add a child plugin in djangocms_text_ckeditor (or whatever text editor plugin), the way to remember the plugin is to insert an img element whose src attribute is given by the `icon_src` plugin method and `id` is computed by the parent plugin. It is quite annoying to be limited to an image. For example, I would like to propose an enhancement to cmsplugin-link which would show an icon and the text of the link.

So here is my proposal :

- add a method to text plugins, let say, `in_text_editor_repr`, which would return html content intended to be put in a span element. The span element would be managed by the parent plugin like the img current one.
- in order to maintain backward compat, it would be used only if both parent plugin and text plugin support it.

Does it fit well ? (I might have missed something)

FinalAngel:
Closing in favour of divio/djangocms-text-ckeditor#143

PR #3954, opened 2015-03-16
[WIP] feature: custom text enabled plugin representation (documentation)

ollb:
This is a proposal for the issue #3437 , it would be nice to have it in 3.1. Feel free to comment.
If it seems ok, I'll continue and wrote all necessary code and then I'll provide an implementation in djangocms-text-ckeditor and main text enabled plugins.

### DIFF
--- a/docs/reference/plugins.rst
+++ b/docs/reference/plugins.rst
@@ -178,9 +178,9 @@ text_enabled
 Default: ``False``
 
 Can the plugin be inserted inside the text plugin?  If this is ``True`` then
-:meth:`icon_src` must be overridden.
+:meth:`in_text_editor_repr` should be overridden.
 
-See also: `icon_src`_, `icon_alt`_
+See also: `in_text_editor_tag`_, `in_text_editor_repr`_
 
 
 Methods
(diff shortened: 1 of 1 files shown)


docs/reference/plugins.rst
```diff
@@ -301,6 +301,44 @@ for `djangocms-text-ckeditor <https://github.com/divio/djangocms-text-ckeditor/>
...
+
+When a plugin can be inserted as a child in a text plugin, this method provides
+a way to choose the tag used to encapsulate the plugin representation.
+
+:meth:`in_text_editor_tag` takes 1 arguments:
+
+* ``instance``: The instance of the plugin model
+* ``editor_name``: The name of the parent plugin
+
+This method must return a string. Default implementation returns ``'div'``.
```
evildmp: Does it really return:
    'div'

or does it return:

    div

(i.e. with or without the single quotes)?

ollb: In the documentation when we are refering to the python boolean true, we use ``True``. When refering to a to python string, we might want to use ``'whatever string'`` in order to show the value is a string. That said, I'm fine with ``whatever string``.

evildmp: OK, I think that ``whatever string`` should be preferred, because we already know it's a string.

ollb:
We would love to work on that issue but we are waiting for comments.

FinalAngel:
@evildmp can you provide additional feedback here? @ollb sorry for the long delay :( are you still able to work on it?

Issue #3957, opened 2015-03-17
Documentation for 'FrontendEditableAdminMixin' should have more explanation


mkoistinen:
The current documentation for `FrontendEditableAdminMixin` is rather declarative but doesn't tell the developer why they should add the mixin to their Admin classes. In fact, I've been using FE editing for some time without using this mixin without issue. The developer should understand the benefits (and any pitfalls) of using the mixin and the documentation should provide this.

Current docs here: https://django-cms.readthedocs.org/en/support-3.0.x/how_to/frontend_models.html

ATTN: @jrief, @evildmp

jrief:
I only converted the former ``FrontendEditableAdmin`` class (derived from ``admin.ModelAdmin``) to a mixin class, so that alternatives to the ``ModelAdmin`` can be used. I don't know myself exactly what this class is for.

yakky:
``FrontendEditableAdminMixin`` is the base class to allow editing standard django models using the django CMS frontend editor.
Order matters as it defines methods that can be defined in other mixin/objects the created class inherit from

yakky:
@evildmp do you want me to handle a first draft of this one?

mkoistinen:
Closed by #3997

PR #3958, opened 2015-03-17
explained need for FrontendEditableAdminMixin

jrief:
and fixed issue #3957


mkoistinen:
I think we can all agree this patch isn't adding anything to the existing, sparse documentation on this topic. Closing this one in hopes there are other takers.

PR #3963, opened 2015-03-18
Work in progress, release notes for 3.1

evildmp:
**Do not merge**!

### DIFF
--- a/docs/reference/cli.rst
+++ b/docs/reference/cli.rst
@@ -173,3 +173,15 @@ Example::
 
     This command publishes drafts. You should review drafts before using this
     command, because they will become public.
+
+**********************
+Maintenance and repair
+**********************
+
+.. _fix-mptt:
+
+``fix-mptt``
+============
+
+Occasionally, the MPTT tree can become corrupted (this is one of the reasons for :doc:`our move away from MPTT to MP in django CMS 3.1 </upgrade/3.1>`). Typical symptoms include problems when
+trying to copy or delete plugins or pages.
--- a/docs/reference/cli.rst
+++ b/docs/reference/cli.rst
@@ -183,5 +183,8 @@ Maintenance and repair
 ``fix-mptt``
 ============
 
-Occasionally, the MPTT tree can become corrupted (this is one of the reasons for :doc:`our move away from MPTT to MP in django CMS 3.1 </upgrade/3.1>`). Typical symptoms include problems when
+Occasionally, the MPTT tree can become corrupted (this is one of the reasons for :doc:`our move
+away from MPTT to MP in django CMS 3.1 </upgrade/3.1>`). Typical symptoms include problems when
 trying to copy or delete plugins or pages.
+
+Once a database has been migrated from MPTT to MP, there is no use for this command.
--- a/docs/upgrade/3.1.rst
+++ b/docs/upgrade/3.1.rst
@@ -4,6 +4,18 @@
 3.1 release notes
 #################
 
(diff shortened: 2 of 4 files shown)


mkoistinen:
Here's some missing templatetag additions:

Add render_model_add_block tag #3875 

Add new template tag render_plugin_block #3595

docs/reference/cli.rst
```diff
@@ -173,3 +173,15 @@ Example::
...
     This command publishes drafts. You should review drafts before using this
     command, because they will become public.
+
+**********************
+Maintenance and repair
+**********************
+
+.. _fix-mptt:
+
+``fix-mptt``
```
mkoistinen: Is this command going to remain in 3.1? I thought I had heard it wouldn't.

docs/upgrade/3.1.rst
```diff
@@ -4,6 +4,18 @@
...
+in 3.0 and establish a safe, solid base for more ambitious work in the future.
+
+In this release we have tried to maintain maximum backwards-compatibility, particularly for
+third-party applications, and endeavoured to identify and tidy loose ends in the system wherever
+possible.
+
+.. warning:: Upgrading from previous versions
+
+    3.1 introduces some changes that **require** action if you are upgrading
+    from a previous version. PLease read this document with care.
```
mkoistinen: Upper-case "L" in PLease here.

docs/reference/cli.rst
evildmp: Possibly not, but since we tell people to use it in the release docs, it had better be explained!

docs/upgrade/3.1.rst
```diff
@@ -11,57 +23,140 @@ What's new in 3.1
...
+Since django CMS 2.0 we have relied on MPTT (Modified Preorder Tree Traversal) for efficiently
+handling tree structures in the database.
+
+In 3.1, `Django MPTT <https://github.com/django-mptt/django-mptt>`_ has been replaced by
+`django-treebeard <https://github.com/tabo/django-treebeard>`_, to improve performance and
+reliability.
+
+Over the years MPTT has proved not to be fast enough for big tree operations (>1000 pages); tree
+corruption, because of transactional errors, has also been a problem.
+
```
mkoistinen: It is my understanding that MPTT (Modified Preorder Tree Traversal) is a general logical topic and that both django-mptt and django-treebeard implement concrete solutions around this. This above reads like MPTT is a product that we have abandoned in favour of Materialized Path. I don't think this is correct. I think we should replace "MPTT" with "django-mptt" in the above paragraph, and in the one below.

docs/reference/cli.rst
mkoistinen: Perhaps a pointer to the previous version's docs then? I can't make sense to document a command that isn't present.

evildmp: I can't see a way around it - there *weren't* any docs previously, which is why I added this.

yakky: ``fix-mptt`` is replaced by ``tree`` in 3.1 (I totally overlooked this change!) which has the same role

docs/upgrade/3.1.rst
yakky: @mkoistinen MPTT is a specific algorithm, different from the one implemented by treebeard, so I think current wording is ok: we're not replacing an MPTT implementation with another, but one tree-to-relational algorithm with another

yakky:
As agreed with @evildmp I'm merging this as preliminary 3.1 release notes. More work will follow before final release

PR #3964, opened 2015-03-18
Allow SITE_ID to be an integer-valued object.

jbazik:
This allows for the use of dynamic site mechanisms like django-multisite.  I think it's actually a better test, since you want to use the site id in a hash, and this tests that directly.  Fails for None and succeeds for True/False, same as the current test.


yakky:
Thanks for this nice addition
Merged in develop

Issue #3973, opened 2015-03-30
ImportError: cannot import name PageUser


jrief:
How to reproduce:

* django-1.6.x
* django-cms==3.0.x

substitute the usermodel as described [here](https://docs.djangoproject.com/en/1.6/topics/auth/customizing/#substituting-a-custom-user-model).

Run the application in uwsgi.

Then this import error will happen:

```
(...snip...)
   from cms.models.fields import PlaceholderField
  File "django-cms/cms/models/fields.py", line 2, in <module>
    from cms.forms.fields import PageSelectFormField
  File "django-cms/cms/forms/fields.py", line 8, in <module>
    from cms.forms.widgets import PageSelectWidget, PageSmartLinkWidget
  File "django-cms/cms/forms/widgets.py", line 14, in <module>
    from cms.models import Page, PageUser
ImportError: cannot import name PageUser
```

The problem here is that model ``PageUser`` wants to inherit from a customized model. This can't work and even if, one would have to recreate separate migrations.

There must be a reason, why referring to an alternative User model is not done the way, Django emphasizes it in its docs, say somethings like:

```
class PageUser(models.Model):
    user_ptr = models.OneToOneField(settings.AUTH_USER_MODEL, primary_key=True)
    created_by = models.ForeignKey(user_model_label, related_name="created_users")

    class Meta:
        verbose_name = _('User (page)')
        verbose_name_plural = _('Users (page)')
        app_label = 'cms'
```

jrief:
By reading the code of **django-cms**, it seems impossible to substitute the usermodel. The reason is, that many models (``GlobalPagePermission``, ``PagePermission``, ``PageUser`` and ``UserSettings``) in django-cms refer onto the table ``auth_user``. This is hard coded and can only be changed if one would recreate all migrations for the cms. This certainly is not a very robust solution. 

Therefore I evaluated some alternative possibilities:

#### Replace the foreign key onto ``auth.User`` by an ``IntegerField``.
This is the preferred method for polymorphic models, but could be applied here as well. The tables ``GlobalPagePermission``, ``PagePermission``, ``PageUser`` and ``UserSettings`` then  would have a simple integer key pointing onto ``auth.User.id`` or onto a primary key from any model substituting ``auth.User``.

The disadvantage is, that we loose the database constraint on foreign keys.

#### Remove the current implementation hack
[...]

PR #3975, opened 2015-04-01
Removed obsolete compatibility layer

jrief:
This compatibility layer added

* *“sort of a hack”* (citation @yakky )
* added non robust code ``importlib.import_module(app + ".models")`` – this caused issue #3973
* accessed private class members: ``User._meta.swapped``
* duplicated code, available since Django-1.4

I therefore would strongly suggest to do some code cleaning here

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -21,8 +21,9 @@ from cms.app_base import CMSApp
 from cms.apphook_pool import apphook_pool
 from cms.constants import TEMPLATE_INHERITANCE_MAGIC
 from cms.models.pagemodel import Page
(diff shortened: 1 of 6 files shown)


jrief:
This PR fixed a nasty circular dependency problem, if ``auth.User`` has been substituted.

The problem is this:

``PageUser`` normally is derived from ``auth.User``, but if a customized User model is used, it inherits from that class.

This customized User class is loaded with other models, some of which may import models from the CMS, for instance ``PlaceholderField``. The latter imports ``PageUser`` which causes a circular dependency.

By separating the classes and moving some import statements into the files, this problem can be fixed.

PR #3988, opened 2015-04-07
Fix UnboundLocalError in cms_tags

stefanw:
`content` is used in line 311 without being defined first.


yakky:
@stefanw thanks for this!

PR #3998, opened 2015-04-08
Fixes #3978 -- user documentation

### LINKED ISSUE #3978: Documentation: User tutorial - publish a page: suggestion for improvement
I just checked out your user tutorial (to see if it contains something which should be linked from our own translators' guide).
Here (http://docs.django-cms.org/en/support-3.0.x/user/tutorial/publish_page.html) I read that 
'only you can see it.' (the draft). In my experience, this is not correct, as drafts are not saved per editor.
So I think it should maybe better be something like 'only people with permission to edit the page can see it'. Else an editor may believe that he/she is *really* the only person who can see a specific draft ;)

Another little improvement suggestion: Maybe the actual number for 'Description is restricted by the number of characters the search engines actually use to show content.' could be added here: http://docs.django-cms.org/en/support-3.0.x/user/reference/page_admin.html ? (For people who copy-paste text there and then hit the limit... I already did ;) )

evildmp:
Fixes #3978

### DIFF
--- a/docs/user/reference/page_admin.rst
+++ b/docs/user/reference/page_admin.rst
@@ -210,12 +210,11 @@ element of the page (which will otherwise simply use the *Title* field). If
 provided, it will be the *Page title* that appears in the browser's title bar
 and bookmarks, and in search engine results.
 
-*Description meta tag* is expected to be used to populate a `<meta>` tag in the
-document `<head>`. This is not displayed on the page, but is used for example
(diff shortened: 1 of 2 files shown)


evildmp:
@mkoistinen This can go into 3.0.13 too.

yakky:
@mkoistinen If you are going to merge this into support I'll forward port it later

yakky:
Merged into support, will port to develop when forward porting support

PR #4006, opened 2015-04-10
Fixes #4005

### LINKED ISSUE #4005: Circular dependency problem, if auth.User has been substituted
This is the problem:

``PageUser`` normally is derived from ``auth.User``, but if a customized User model is used, it inherits from that class.

This customized User class is loaded with other models, some of which may import models from the CMS, for instance ``PlaceholderField``. The latter imports ``PageUser`` which causes a circular dependency.

By separating the classes and moving some import statements into the calling functions, this problem can be fixed.

I will fix this.

jrief:
Fixes #4005 by moving potential circular dependencies into the calling functions.

This PR adds less magicness to ``permissionmodels.py``.

Renamed ``force_unicode`` -> ``force_text``, since thats the correct name.

Obsoletes PR #3975.

### DIFF
--- a/cms/admin/useradmin.py
+++ b/cms/admin/useradmin.py
@@ -21,7 +21,7 @@ class PageUserAdmin(admin_class, GenericCmsPermissionAdmin):
     form = PageUserForm
     add_form = PageUserForm
     model = PageUser
-    
+
     def queryset(self, request):
         qs = super(PageUserAdmin, self).queryset(request)
         try:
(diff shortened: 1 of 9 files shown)


mkoistinen:
Hi Jacob, what's the status on this one?

jrief:
This patch fixes the problem, but for some combinations of Django and Python the unit tests fail. Grrr.

If 3.1 is ready, I will retry since then support, for Django-1.4 and 1.5 will be dropped.

yakky:
@jrief 3.1 is due to release any day now

mkoistinen:
What is the status of this PR?

jrief:
closes via #4005

PR #4009, opened 2015-04-11
Add position rebuild to CMSPlugin.fix_tree

### LINKED ISSUE #3735: fix-tree command does not set position
For some as yet unknown reason, I've found plugins that have their position set to None (will open a ticket once I trace that bug down). Calling the fix-tree command does nothing to repair this since for CMSPlugin's, it just defers to django-treebeard's `fix_tree()` method, but of course django-treebeard knows nothing about django-cms's custom 'position' field.

This breaks the page since `cms.utils.plugins.build_plugin_tree()` attempts to [sort plugins](https://github.com/divio/django-cms/blob/develop/cms/utils/plugins.py#L283-L299) by ('parent_id', 'position'), resulting in a `TypeError: unorderable types: int() < NoneType()`

yakky:
Add position rebuild to CMSPlugin.fix_tree to allow complete rebuild of plugin tree

Fix #3735

### DIFF
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -5,6 +5,7 @@ from operator import itemgetter
 import os
 import warnings
 
+from django.conf import settings
 from django.core.urlresolvers import NoReverseMatch
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
 from django.db import models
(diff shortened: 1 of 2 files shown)


yakky:
@jsma would you have a look at this?

PR #4017, opened 2015-04-14
don't install django 1.8

jedie:
If somebody runs "pip install --upgrade django-cms" on current 3.0.x branch, it will run into a unusable environment, because django 1.8 will be installed :(


mkoistinen:
Thanks for this!

PR #4018, opened 2015-04-14
Issues/fix copy extensions

### LINKED ISSUE #3987: Page- and title extensions not copied properly when duplicating pages
1. Duplicate a page that has a page extension using the cms tree copy function
2. Notice that the duplicated page seems to have copied the page extension as expected
3. Delete the duplicated page
4. Notice that the source page has lost its page extension, too.

I suppose it's because the extension copy mechanism was originally written to publish pages which works perfectly well. However, when I followed the duplication in the database I noticed that the duplicated page gets the 'published' entry of the extension, where as the source page keeps the 'unpublished' entry of the extension. Since the latter has a foreign key to the published extension, it is deleted as a consequence of the default 'cascade' behavior once the duplicated page is deleted.

mkoistinen:
Fixes #3987


mkoistinen:
Sorry, I opened this against the wrong branch =(

PR #4026, opened 2015-04-15
ensure that apphooked views preserve their function name and docstrings

piquadrat:
Besides being a good citizen and preserving name and docstring of decorated view names, this also allows easier introspection of views for tools like the django debug toolbar or opbeat.

### DIFF
--- a/cms/tests/apphooks.py
+++ b/cms/tests/apphooks.py
@@ -4,7 +4,7 @@ import sys
 
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Permission
-from django.core.urlresolvers import clear_url_caches, reverse
+from django.core.urlresolvers import clear_url_caches, reverse, resolve
 from django.test.utils import override_settings
 from django.utils import six
 from django.utils.timezone import now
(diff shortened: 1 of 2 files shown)


yakky:
@piquadrat nice ...

piquadrat:
@yakky cool, thanks! Any chance it could make it into the next 3.0.x release?

yakky:
@piquadrat my concern is about regressions regarding ``is_current_app`` flag for the toolbar. I don't trust 100% currents tests, so we may want to avoid regressions
I let @mkoistinen decide whether including this in 3.0

piquadrat:
@yakky @mkoistinen completely up to you of course :) we have some customers that are affected by this, but it's not critical

yakky:
@piquadrat on my side nothing against having this in 3.0, we just need to properly testing it does not interfere with the toolbar

mkoistinen:
@piquadrat Can I ask you to rebase this against current support/3.0.x? I'll then pull it into the 3.0.13 release (if you can do it quickly), then it will by default end up in 3.1 too.

mkoistinen:
OK @yakky, your instincts serve you well. Tests run clean on 3.0.x, but real world tests are failing all over the place. So, this tells me that this isn't ready for 3.0.13 and that the tests are not sufficient =)

piquadrat:
The problem seems to be what `functool.wraps` stumbles over class views (not class _based_ views, those work fine), like the ones from `django.contrib.syndication`. I'll see if I can get a workaround for those

piquadrat:
replaced by #4036

PR #4031, opened 2015-04-15
Always force language when reversing page model URLs

stefanw:
I discovered a problem when editing pages in e.g. English when the toolbar user settings language is German (ie. they are different).

Navigating to the English page and clicking the "Edit" button would navigate me to the Edit-Page (`?edit`) of the German page. Similarly publishing changes from the English page would redirect me to the German page.

The reason is the use of [`force_language` in the toolbar code](https://github.com/divio/django-cms/blob/develop/cms/toolbar/toolbar.py#L335) which activates the `user_settings.language` of the toolbar (German in my case). This results in the page being reversed as German even though `request.LANGUAGE_CODE` is English.

The `get_absolute_url` of the page model might get the `language` from the request as a parameter but since it has been temporarily deactivated and replaced by the toolbar's language in the local thread, the [call to `reverse`](https://github.com/divio/django-cms/blob/develop/cms/models/pagemodel.py#L156-L158) will reverse with the wrong language.

This PR activates the language in `get_absolute_url` so that the call to `reverse` actually returns the language given as a parameter. This fixes the observed problem.

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -152,10 +152,11 @@ class Page(six.with_metaclass(PageMetaClass, MP_Node)):
     def get_absolute_url(self, language=None, fallback=True):
         if not language:
             language = get_language()
-        if self.is_home:
-            return reverse('pages-root')
-        path = self.get_path(language, fallback) or self.get_slug(language, fallback)
-        return reverse('pages-details-by-slug', kwargs={"slug": path})
+        with i18n.force_language(language):
+            if self.is_home:
+                return reverse('pages-root')
+            path = self.get_path(language, fallback) or self.get_slug(language, fallback)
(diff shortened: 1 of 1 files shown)


yakky:
@stefanw thanks for this, it's indeed a nasty behavior, we need a test for this, though

yakky:
Test merged in #4110

yakky:
Thanks @stefanw for your contribution!

Moini:
Good to see this is already fixed! :D
I just noticed, and really started to worry about our translators accidentally translating the English pages - sometimes, the only indicator that you're on the wrong page is the url... Which version will this be in? It's essential for us...

yakky:
This will be released in 3.1.1

Moini:
Thanks!

PR #4036, opened 2015-04-16
preserve apphook view name (3.0.x branch)

piquadrat:
This replaces #4026. The original usage of `functool.wraps` turned out to have problem with "class" views that use the `__call__` pattern. Instead, in this branch we do the work manually and copy `__doc__`, `__module__` and `__name__` (for functions) or `__class__.__name__` (for classes) to the decorating function.


yakky:
LGTM

PR #4042, opened 2015-04-16
Issues/fix breadcrumbs

### LINKED ISSUE #4041: Breadcrumbs may disappear in one of published/draft mode

mkoistinen:
Fixes #4041

### DIFF
--- a/cms/admin/forms.py
+++ b/cms/admin/forms.py
@@ -245,7 +245,7 @@ class AdvancedSettingsForm(forms.ModelForm):
             self.fields['language'].initial = get_language()
         if 'navigation_extenders' in self.fields:
             self.fields['navigation_extenders'].widget = forms.Select(
-                {}, [('', "---------")] + menu_pool.get_menus_by_attribute("cms_enabled", True))
+                {}, [('', "---------")] + menu_pool.get_menus_by_attribute("cms_enabled", True, distinct=True))
         if 'application_urls' in self.fields:
             # Prepare a dict mapping the apps by class name ('PollApp') to
             # their app_name attribute ('polls'), if any.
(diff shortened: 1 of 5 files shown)


mkoistinen:
@yakky please review, thanks.

yakky:
LGTM, but I'd add a test for #4041

yakky:
Tested in live cases and it's +1 for me, but we need a test for the case that's supposed to fix

mkoistinen:
Test(s) coming...

mkoistinen:
Rebased

yakky:
@mkoistinen do you need help with tests? This is definitely and urgently needed in both 3.0 and 3.1

mkoistinen:
@yakky, actually, ~~I think~~ I need help with the code itself. As of *now* (in this branch) here's what works:

1. CMSApp (normal) apps with menus declared within them work as expected;
1. CMSConfigApp apps with menus declared within them work as expected;
1. In these cases, the attach menu setting in the advanced setting correctly only shows a single item and the correct menu items are displayed in the corresponding menus.

Here's what doesn't:
When we do not declared the menus inside the CMSApp/CMSConfigApp class, but rather allow the user to independently assign the menu in the advanced settings. In this case, things get confused.

For example, if you have CMSConfigApp app with a CMSAttachMenu registered, then you hook this app to 2 or more pages and manually assign the CMSAttachMenu to them, it is not guaranteed that the correct instance of the CMSAttachMenu gets assigned to each page.

Is this even possible?

yakky:
@mkoistinen I'll give it a run on my way to cardiff

yakky:
@mkoistinen can't reproduce the described behavior with your test project. Assigign multi menu to multi-a and multi-b pages, publishing and reloading the project provides dropdown wih "menu-a" and "menu-b" items

mkoistinen:
Merging this as a partial solution which is then augmented with #4159.

mkoistinen:
Thanks, @yakky, this definitely helps with this issue. I think there are still some edge cases that need to be addressed, but I'll open new issues specifically for these.

PR #4044, opened 2015-04-17
Fixed list numbering in permissions.rst

mikko-ahlroth-vincit:
List was previously shown with all elements concatenated into a paragraph.


yakky:
Thanks for spotting this, I'd use RST syntax ``#.`` for automatic numbering

PR #4055, opened 2015-04-19
fix issue #4046, update toolbar documentation

mrosati84:


evildmp:
Taking a look now.

docs/upgrade/3.1.rst
```diff
@@ -152,6 +152,9 @@ Improvements to the toolbar user interface
 ==========================================
 
 Some minor changes have been implemented to improve the toolbar user interface.
+The old *Draft/Live* switch has been replaced to achieve a more clear
```
evildmp: Note that buttons that perform an action such **Draft/Live** as this should be marked up with strong emphasis: http://docs.django-cms.org/en/develop/contributing/contributing.html#inline-markup (don't worry, I'll make these changes here).

evildmp:
Thanks Matteo, this code has now been merged in https://github.com/divio/django-cms/pull/4057.

yakky:
Fixed in #4057

PR #4056, opened 2015-04-19
hides the toolbar for pages out of django-cms

pa0lin082:


yakky:
This fixes an issue in which the toolbar appears even if outside the django CMS paths

yakky:
@mkoistinen @evildmp what's your opinion on this solution? The bug is legitimate

yakky:
Thanks @pa0lin082 !

PR #4064, opened 2015-04-20
Always use South 1.0.2

yakky:


yakky:
@mkoistinen I propose to force south 1.0.2 in 3.0 to allow easier plugin upgrade for the Django 1.7/South support

Issue #3747, opened 2015-01-11
Disclosure triangles in the structure-mode tree should cascade with opt/alt key


mkoistinen:
Very common UI is that when there is a collapsible tree, opt/alt-clicking the disclosure triangle will make the action cascade. So, if you click to open with the opt/alt key held, then all sub-members of the item you clicked on would open as well. Same for close.

astagi:
I started working on this feature, check https://github.com/astagi/django-cms/tree/feature/openclosecascade

mkoistinen:
Awesome!

yakky:
@mkoistinen could you check the opened PR?

FinalAngel:
confirmed fixed

PR #4065, opened 2015-04-20
Open and close collapsible trees using alt key

astagi:
Implements #3747

### DIFF
--- a/cms/static/cms/js/modules/cms.plugins.js
+++ b/cms/static/cms/js/modules/cms.plugins.js
@@ -687,9 +687,24 @@ $(document).ready(function () {
 				if(el.hasClass('cms_dragitem-expanded')) {
 					settings.states.splice($.inArray(id, settings.states), 1);
 					el.removeClass('cms_dragitem-expanded').parent().find('> .cms_draggables').hide();
+					if ($(document).data('expandmode')) {
+						var items = draggable.find('.cms_draggable').find('.cms_dragitem-collapsable');
+						if(!items.length) return false;
+						items.each(function () {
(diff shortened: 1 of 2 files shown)


yakky:
Ping @mkoistinen @FinalAngel for review

FinalAngel:
codewise all good, question is, if we want the "alt" key to trigger this

yakky:
Alt is not the best option in my opinion as it triggers window moving on Linux

astagi:
What about "shift"?

FinalAngel:
I'd rather use a combo of Shift + E (for expand) or something similar

astagi:
@FinalAngel do you mean Shift + E + click on the placeholder? In that case it won't behave as a toggle (close all if open, open all otherwise on key+click).

mkoistinen:
It is unnatural already to use SHIFT for the advanced page settings shortcut. I think its time we used ALT for Macs, and SHIFT for linux for both this and for the advanced page setting shortcut.

yakky:
@mkoistinen AFAIK in the page tree the Shift is used on any operating system (https://github.com/divio/django-cms/blob/develop/cms/static/cms/js/modules/cms.changelist.js#L564)

astagi:
any update here?

FinalAngel:
alright +1 for shift key

astagi:
Thank you @FinalAngel I'll update this PR later :)

FinalAngel:
awesome @astagi

astagi:
Updated! :tada:

mkoistinen:
+1

FinalAngel:
:+1:

yakky:
Awesome @astagi

PR #4077, opened 2015-04-29
bug : can not add a plugin to an instance I can change

DylannCordel:
Currently, we must have the "add" permission to a specific model when we want to "add" a new plugin to the placeholder of this model's existing instance. This is the same for "delete" permission. Problem is : maybe I can "change"  the instance but not add a new one. So, I'm not able to add (or delete) a plugin if I only have a "change" perm on the instance. 

This PR change the permission logic for placeholder and plugins : 

* to `change` a placeholder/plugin, we must have `change` or `add` perm on attached object
* idem for `add`
* to `delete` a placeholder/plugin, we must have `change`, `add` or `delete` perm on attached object

This PR is in three commits :

* first is full backward compatible : 
  * if there are attached objects : only test the first one and return True or False
  * if there is not any attached object, return nothing
  * if given key param is something else than `add`, `change` or `delete` : test this perm key on attached object
* second one raise an Exception if perm key is not `add`, `change` or `delete` (cf docstring which list availaible values for key param). This is because perm logic must explicitely set (e.g : a `cut` perm should have same tests than `delete`, but a `copy` perm should have same tests than `add`.)
* the last one is to do what the docstring says : 

  > "For **each** attached object, permission has to be granted either on attached model or on attached object.". 

  Test is not only performed on the first attached object if one of attached objects fail perm tests, then False is returned. None is still returned if there is not any attached objects.

### DIFF
--- a/cms/models/placeholdermodel.py
+++ b/cms/models/placeholdermodel.py
@@ -107,13 +107,24 @@ class Placeholder(models.Model):
         Generic method to check the permissions for a request for a given key,
         the key can be: 'add', 'change' or 'delete'. For each attached object
         permission has to be granted either on attached model or on attached object.
+          * 'add' and 'change' permissions on placeholder need either on add or change 
+            permission on attached object to be granted.
+          * 'delete' need either on add, change or delete
         """
         if getattr(request, 'user', None) and request.user.is_superuser:
             return True
-        if self.page:
-            return self._get_object_permission(self.page, request, key)
-        for obj in self._get_attached_objects():
-            return self._get_object_permission(obj, request, key)
+        perm_keys = {
+            'add': ('add', 'change',),
+            'change': ('add', 'change',),
+            'delete': ('add', 'change', 'delete'),
+        }
+        objects = [self.page] if self.page else self._get_attached_objects()
+        keys = perm_keys[key] if key in perm_keys else (key,)
+        for obj in objects:
+            for key in keys:
+                if self._get_object_permission(obj, request, key):
+                    return True
+            return False
 
     def _get_object_permission(self, obj, request, key):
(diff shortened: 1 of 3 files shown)


yakky:
Nice change @DylannCordel!
I'd like a test to assert the new behaviour and the relevant documentation

DylannCordel:
@yakky : I change the existing `PlaceholderPluginPermissionTests` to test new behaviour and all (usefull) perms combinations. Old test was using real request : this is not any more possible because it takes too much time and is not the job of those tests to check that admin works well (it is already tested elsewhere). So I directly call admin's interface's methods which check permissions with the same request. Unfortunately, to achieve this, I need to [remove the user._perm_cache](https://github.com/webu/django-cms/commit/1e7a41af81b9a416346f91c669b462db6d5c151d#diff-49912f4c94ec08af931c0d53c3b9612aR1311) each time his perms are changed to force the perm backend to retrieve __all__ user permissions. If someone has a better idea to avoid to remove a protected attribute, it would be great.

yakky:
@DylannCordel great job. I'm totally fine with tinkering with ``user._perm_cache``. There is no public API to do this. There's a lot of hacks like this in our tests to really test low-level behavior. It will cost some test refactoring somewhere in the future if the privati API will change, but I think it's a good tradeoff

DylannCordel:
@yakky Thx for your feed back. If you are okay with the `_user_perm_cache` trick, than  I am done with this PR :)

yakky:
@mkoistinen @evildmp it'd lean to merge this into 3.1 as it's legitimate fix, but it substantially changes the way permissions are applied and I think we should give users proper information regarding this

yakky:
@DylannCordel great job here!

PR #4087, opened 2015-05-03
Pin html5lib version

yakky:
Due to html5lib/html5lib-python#189 html5lib newer than 0.999 causes regressions


yakky:
@mkoistinen I think we need to merge and release a version that merges this due to the above upstream bug.

gsnedders:
FWIW, there's now a release with the regressions fixed. (As far as I and the tests can tell, at least.)

yakky:
@gsnedders awesome, thanks! We will change the dependency soon

PR #4096, opened 2015-05-05
Allow the migrate_use_structure migration to work with django 1.7

doctormo:
The ContentType can be gotten much more simply in django 1.7 and because of the ContentType error masks the rest of the code it was impossible to see the next error from the Permission get.

With this code the migration runs without the ContentType warning which happened every test run for our website.


yakky:
@doctormo thanks for this improvement

PR #4107, opened 2015-05-12
added fix for show_menu_below_id and soft_root

douwevandermeij:
When using show_menu_below_id you want to get past soft_root, as mentioned in the documentation.

http://docs.django-cms.org/en/latest/reference/navigation.html#show-menu-below-id
"Note that soft roots will not affect the menu when using show_menu_below_id."

### DIFF
--- a/cms/tests/menu.py
+++ b/cms/tests/menu.py
@@ -990,6 +990,107 @@ class ShowMenuBelowIdTests(BaseMenuTest):
         node = nodes[0]
         self.assertEqual(node.id, b.publisher_public.id)
 
+    def test_menu_beyond_soft_root(self):
+        """
+        Test for issue 4107
+
+        Build the following tree:
+
+            A
+            |-B (soft_root)
+              |-C
+        """
+        stdkwargs = {
+            'template': 'nav_playground.html',
+            'language': 'en',
+            'published': True,
+            'in_navigation': True,
+        }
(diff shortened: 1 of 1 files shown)


yakky:
@douwevandermeij I'd like a regression test for this

douwevandermeij:
I added a test, but it seems to work both on develop as on master... Any ideas?

douwevandermeij:
I can't reproduce the error (so without my fix) on 3.1.0 but only on 3.0.9.
It fixes the 3.0.9 issue, but doesn't help or harm in 3.1.0.
I think this commit did fix the problem: https://github.com/divio/django-cms/commit/afd69f8c8e8a2bc0d6d3578c3dbca97144e79227
I'm not sure which fix is better, but problem is resolved anyhow.

yakky:
@douwevandermeij are versions of 3.0.x later than 3.0.9 affected by this?

douwevandermeij:
Yes, 3.0.13 is still affected.

yakky:
@douwevandermeij Leave just the test in this PR and we'll merge it
@mkoistinen I'd merge this and #3509 into 3.0.x branch and then forward port it. I can take care of doing this.

yakky:
Merged in #4175 
Thanks @douwevandermeij for your contribution!

PR #4114, opened 2015-05-14
[WiP] Let's try to really fix languages

yakky:


yakky:
@mkoistinen @evildmp do this behavior look sane to you? See https://github.com/yakky/django-cms/blob/feature/fix_languages_update/cms/tests/page.py#L83-L91

yakky:
Wrong way to solve this. Closing this as the current implementation is better than thi fix

PR #4121, opened 2015-05-20
Fix order of which application urls are injected into urlpatterns

vstoykov:
This will fix problem when there are nested apphooks and urls of top level apphook are more greedy.

This fix is for Django CMS 3.0.x. I know that 3.1 but fix for it will be little different because of django-treebeard and we are not currently upgraded to 3.1.

I'm planning to add tests for that but first I need your opinion about the solutions of the problem.

### DIFF
--- a/cms/appresolver.py
+++ b/cms/appresolver.py
@@ -1,6 +1,12 @@
 # -*- coding: utf-8 -*-
 from __future__ import with_statement
 
+try:
+    from collections import OrderedDict
+except ImportError:
+    # Python < 2.7
+    from django.utils.datastructures import SortedDict as OrderedDict
+
 from django.conf import settings
 from django.conf.urls import patterns
 from django.contrib.sites.models import Site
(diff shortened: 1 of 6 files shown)


yakky:
Nice one!
Tests definitely needed, but a big fat +1 on  my side

yakky:
BTW would you add a TODO comment to remember to port this to 3.1 when forward porting from 3.0.x branch?

vstoykov:
I updated my pull request with tests.

yakky:
Thanks @vstoykov !

yakky:
:+1:

yakky:
@mkoistinen is this something for 3.0? It's actualy a bugfix, but with potential backward incompatibilities (albeit incompatibilities based on a flawed implementation)

yakky:
Fix by #4169

mkoistinen:
LGTM, merging.

PR #4127, opened 2015-05-28
Update year.

sam-m888:


sam-m888:
Really, Travis-ci fails because I updated the year?

yakky:
Thanks!

Issue #4134, opened 2015-05-29
Add setting to not show the CMS Toolbar for anonymous users


stefanw:
[This line](https://github.com/divio/django-cms/blob/develop/cms/middleware/toolbar.py#L69) lets anonymous users see the CMS toolbar to login when `?edit` is added.

I wish there was a setting to disable showing the toolbar for anonymous users. My concern is that a link containing `?edit` (e.g. [a link like this](http://divio.ch/en/?edit)) will inadvertently spread around and visitors will be confused by the toolbar showing up. Developers may also wish to additionally protect their login form, reduce attack surface against XSS, hide the identity of their CMS etc.

If you feel like this is a reasonable feature request and worth a setting, I'm happy to provide a PR.

yakky:
I think a setting to disable the toolbar login is acceptable, but it should be off by default, as otherwise it's not possible to login from the frontend, which is a feature we want to be enabled by default

PR #4139, opened 2015-06-01
Add setting to disable toolbar for anonymous users

stefanw:
This adds the setting `CMS_TOOLBAR_ANONYMOUS_ON` defaulting to `True` which maintains the current behaviour. If set to False, anonymous users (like non-staff logged in users) will not see the CMS toolbar when `?edit` is appended to a URL. 

This PR contains a test and documentation for the new setting. Relates to #4134.


yakky:
Thanks @stefanw

PR #4142, opened 2015-06-02
Fix X-Frame-Options on top-level pages

### LINKED ISSUE #4140: Top-level inherited X-Frame-Options are not applied
If your page has X-Frame-Options set to INHERIT and if it's a top-level page then `get_xframe_options` for this page will return `None` (https://github.com/divio/django-cms/blob/f6277b127c463e3188b8b19c823ff4f8aac8f82b/cms/models/pagemodel.py#L1261).

The problem is, the delegation of the X-Frame-Options handler to Django is only done if the `xframe_options` is set to `INHERIT` (https://github.com/divio/django-cms/blob/f6277b127c463e3188b8b19c823ff4f8aac8f82b/cms/views.py#L205). Since it doesn't check for a `None` value, it then sets `response.xframe_options_exempt = True` and no X-Frame option will ever be set on this page.

This actually prevents DjangoCMS users from getting ponycheck stickers at the DjangoCon so I think this issue is VERY important.

I guess the right way to do it would be to let Django's middleware handle the case also if `xframe_options` is `None.`.

sephii:
This should fix #4140

### DIFF
--- a/cms/tests/page.py
+++ b/cms/tests/page.py
@@ -13,7 +13,7 @@ from django.core.urlresolvers import reverse
 from django.db import transaction
 from django.db.models import signals
 from django.http import HttpRequest, HttpResponse, HttpResponseNotFound
-from django.test.utils import modify_settings, override_settings
+from django.test.utils import override_settings
 from django.utils.encoding import force_text
 from django.utils.timezone import now as tz_now, make_aware, get_current_timezone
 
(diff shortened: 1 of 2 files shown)


yakky:
@mkoistinen I think this can be merged in 3.0 and ported to 3.1. I think it's important to support this on legacy version too

yakky:
@sephii thanks for this contribution!

yakky:
Fixed by #4168

Issue #3981, opened 2015-04-03
PageField lists pages available in current user language instead of page languages


sephii:
If you use a `PageField` in one of your plugins, then the only pages the user can link to are the pages that are in the same language as the preferred language of the user. The correct behaviour should be to put the pages that are in the same language as the page that is currently edited instead.

Here are the references in the code:
https://github.com/divio/django-cms/blob/38955c35a6e137343d5d1ecf5e355c38dce3f093/cms/forms/fields.py#L42
https://github.com/divio/django-cms/blob/38955c35a6e137343d5d1ecf5e355c38dce3f093/cms/forms/utils.py#L76

Claimed by @sephii

sephii:
I started working on this here: https://github.com/sephii/django-cms/commit/1d2c875db1ce0c3ae9003b0aa6331e95215d82fe

I don't know what would be the right approach to get the language from the current page in the `PageField` though. Any suggestion?

Aiky30:
Hello @sephii,

Thank you very much for your contribution to the django CMS project. The django CMS project has recently joined forces under the banner of the newly founded django CMS Association to revive the development of django CMS. For this reason we have analyzed all pull requests and issues in detail and decided which ones fit the product vision we want to go with django CMS. 

Due to the age of this issue we are unable to safely accept the contribution. If you feel this issue is still required please inform a member of our technical committee: [nicolai@django-cms.org](mailto:nicolai@django-cms.org ).

Nevertheless we thank you for your participation and would like to emphasize that we do not want to discourage you in your contribution to the django CMS project. On the contrary, we encourage everyone who is interested in the success of django CMS to join us!

If you want to learn more about the work of the technical committee, our product vision and how you can become a member of the dCA, you can find all information here:

- [The Technical Committee of the django CMS Association](https://github.com/django-cms/django-cms-mgmt/blob/master/tech-committee/about.md)
- [Technical Roadmap and Product Vision for django CMS](https://github.com/django-cms/django-cms-mgmt/blob/master/django-cms/roadmap.md)
- [How to become a member of the django CMS Association](https://github.com/django-cms/django-cms-mgmt/blob/master/association/about.md)

We are looking forward to every companion who would like to shape the future of django CMS together with us! 

If you have any questions, feel free to reach out to Nicolai, our Community Manager: nicolai@django-cms.org 


Sign up for the Association now!

PR #4153, opened 2015-06-05
Make PageField only list pages in the same language as the edited page

sephii:
This is a workaround for #3981. It introduces a new `CMSPluginBase` derived class named `CMSPagePluginBase`, that, if inherited, will make `PageField` fields only list pages in the same language as the edited page. I think it would be better if it could be included in `CMSPluginBase` since `PageField` is still a core feature.

If we leave it as `CMSPagePluginBase`, then this should be documented somewhere.

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -361,6 +361,15 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
         )
 
 
+class CMSPagePluginBase(CMSPluginBase):
+    def formfield_for_dbfield(self, db_field, **kwargs):
+        from cms.models import fields
+        if isinstance(db_field, fields.PageField):
+            kwargs['language'] = self.cms_plugin_instance.language
+
+        return super(CMSPagePluginBase, self).formfield_for_dbfield(db_field, **kwargs)
+
+
 class PluginMenuItem(object):
     def __init__(self, name, url, data, question=None, action='ajax'):
         """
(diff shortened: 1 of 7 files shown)


sephii:
Please note that the tests are currently failing, mainly because of the check on the number or executed SQL queries. I'm not sure how to fix this.

yakky:
@sephii for number of queries just update the allowed range. It's normal that query number is larger whenver you add plugins to the test suite (like in this case) or if the plugins does more queries

yakky:
I'll review and check whether the new mixin class is really needed.

cms/plugin_base.py
```diff
@@ -361,6 +361,15 @@ def get_plugin_media(self, request, context, plugin):
         )
 
 
+class CMSPagePluginBase(CMSPluginBase):
+    def formfield_for_dbfield(self, db_field, **kwargs):
+        from cms.models import fields
+        if isinstance(db_field, fields.PageField):
+            kwargs['language'] = self.cms_plugin_instance.language
```
czpython: cms_plugin_instance is no longer used when adding a plugin on 3.3

czpython:
@sephii 
Hello,
Thanks for the contribution.
I've reviewed the PR and seems to me that adding the logic to the plugin class couples plugins with the page field.

While agree that usually it's desired to see the pages available for the plugin the field is used on, I think it's the responsibility of the plugin author to set this up.
We can improve the plugin and page field internals to make it easier for a plugin author to filter pages based on the plugin's language. /cc @divio/django-cms-core

PR #4160, opened 2015-06-06
Fix language fallback

### LINKED ISSUE #4083: Language fallbacks broken for nested plugins
This one has stretched my tiny little mind almost to breaking point.

I have a site with 2 languages, English and French. When someone is browsing the French site and requests a page for which there isn't a French translation, I'd like to return the English page (without redirecting).

This works well enough, except when it comes to nested plugins. For example, if you have a text plugin directly in a placeholder, it works - however if you have a text plugin inside a style plugin, or a multi-column plugin, the child plugin (the text) is never rendered in language fallback mode.

Here is why I think that is:

The function `cms.utils.plugins.assign_plugins` iterates over a placeholder for the request language and builds up a list of plugins. If it can't find any plugins for a given language, it'll recursively look for plugins in the fallback language (if any).

It then assigns the `placeholder._plugins_cache` attribute with whatever plugins it finds. however, due to the recursion, the fallback language `_plugins_cache` list is always clobbered by the request language.

I think I've solved this by making `assign_plugins` check if the `_plugins_cache` attribute exists first, (i.e., if it was populated by the fallback language call), and if so, don't clobber it:

```
    for placeholder in placeholders:
        if not hasattr(placeholder, '_plugins_cache'):
            setattr(placeholder, '_plugins_cache', list(groups.get(placeholder.pk, [])))
```

Although I'm not sure if this is the correct approach or if I'm breaking something else by doing this.

yakky:
Nested plugins did not respect the language fallback

@mkoistinen this has been tested in live projects
Test should be comprehensive enough to demonstrate the issue

Fix #4083

### DIFF
--- a/cms/tests/multilingual.py
+++ b/cms/tests/multilingual.py
@@ -38,7 +38,6 @@ class MultilingualTestCase(SettingsOverrideTestCase):
                           ('nav_playground.html', 'nav_playground.html')],
     }
 
-
     def test_create_page(self):
         """
         Test that a page can be created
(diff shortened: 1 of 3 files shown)


mkoistinen:
LGTM

PR #4163, opened 2015-06-07
Fix #4135 - Cast watch_models to list

### LINKED ISSUE #4135: Ensure that ToolbarPool.get_watch_models is a list
Cast ``toolbar.watch_models`` to a list at https://github.com/divio/django-cms/blob/develop/cms/toolbar_pool.py#L58

Without this change, you'll get a ``*** TypeError: can only concatenate list (not "tuple") to list`` when ``watch_models`` is a tuple.

yakky:


yakky:
Please @mkoistinen @stefanfoulis review this

PR #4169, opened 2015-06-07
Backport #4121

yakky:
This will fix problem when there are nested apphooks and urls of top level apphook are more greedy

Fix #4121


vstoykov:
In this merge I see one strange file ` cms/test_utils/project/sampleapp/views.py.rej ` Is this intentional or a mistake?

yakky:
@vstoykov Sorry, just a mistake when backporting your patch. File removed

PR #4170, opened 2015-06-07
Fix #3945

### LINKED ISSUE #3945: Scanning for placeholders fails on include tags with a variable as an argument
I believe this is the underlying problem for issue #3852. I am opening a new one as the other thread is rather noisy. Here's a concise bit of evidence that people can hopefully work with:

```python
from django import setup
from django.template import Template

setup()

t = Template('{% include name %}')

from cms.utils.plugins import _scan_placeholders
_scan_placeholders(t.nodelist)
```

Running this with Django v1.7.6 and Django CMS v3.0.12, the last line will throw an AttributeError.

I suppose these nodes will have to be ignored as there's no way to resolve the variable without a context.

yakky:
Fix #3945


yakky:
Sorry @mkoistinen I missed this

mkoistinen:
LGTM

PR #4183, opened 2015-06-10
Fix icon positioning on sideframe

FinalAngel:
It should be:
image

rather than:
image

The coding guidelines from the documentation`http://docs.django-cms.org/en/develop/contributing/contributing.html#frontend` apply


Brankub:
LGTM...

PR #4185, opened 2015-06-10
Simple structure board rendering

FinalAngel:
As the structureboard can cause several issues in regards with the positioning, we tend to move towards a simple version where all placeholders are on the same level. The following enhancements are still in consideration:

- [x] Visually separate placeholders from static_placeholders
- [x] Let the user overwrite the padding (via `cms_structure-content-simple`)
- [x] Add python option to enable / disable the simplistic view
- [x] Finalise the code implementation
- [x] Add documentation

You can currently enable/disable the view by editing ``cms/templates/cms/toolbar/toolbar_javascript.html`` new setting ``'simpleStructureBoard': false``

**New Proposal:**
image

**Current Integration:**
image


yakky:
:+1:

yakky:
Welcome 3.2!

PR #4189, opened 2015-06-11
Added page attribute to page navigation node

donce:
Maybe it's worth keeping 'page' in NavigationNode.attrs? It used to be there and our code depends on it for creating custom menus using navigation modifiers.

I see that it was removed in https://github.com/divio/django-cms/pull/1558, but maybe keeping a reference to a page shouldn't effect performance that much.

### DIFF
--- a/cms/menu.py
+++ b/cms/menu.py
@@ -105,7 +105,8 @@ def page_to_node(page, home, cut):
     """
     # Theses are simple to port over, since they are not calculated.
     # Other attributes will be added conditionnally later.
-    attr = {'soft_root': page.soft_root,
+    attr = {'page': page,
+            'soft_root': page.soft_root,
             'auth_required': page.login_required,
             'reverse_id': page.reverse_id, }
 


yakky:
@mkoistinen It sounds reasonable to me, but it may have an impact on the performance (thinking about large menus) and the cache occupation

mkoistinen:
I don't recall there being any complaints about memory usage (perhaps I've missed them), as a general rule, we shouldn't put things in memory without a purpose, but this sounds like it has one =)

I do know that there are some projects with many, many pages, perhaps an owner of one of these could weight in?

yakky:
@donce closing this as this have negative performance impact, as suspected. See #4276

vstoykov:
@donce we added a page to a menu node lazily only when it is needed by this menu modifier and it works quite ok.

    from django.utils.functional import SimpleLazyObject

    from cms.models import Page
    from menus.base import Modifier


    class PageDescription(Modifier):
        """Add page meta description to the menu items"""
        post_cut = True

        def modify(self, request, nodes, namespace, root_id, post_cut, breadcrumb):
            if not breadcrumb and post_cut:
                pages_ids = []

                def cache_page_ids(nodes):
                    for node in nodes:
                        pages_ids.append(node.id)
                        cache_page_ids(node.children)

                cache_page_ids(nodes)
                pages_cache = SimpleLazyObject(lambda: Page.objects.in_bulk(pages_ids))

                for node in nodes:
                    self.set_page_metadata(node, pages_cache)

            return nodes

        def set_page_metadata(self, node, pages_cache):
            node.page = SimpleLazyObject(lambda: pages_cache[node.id])
            node.description = SimpleLazyObject(lambda: node.page.get_meta_description())
            for child in node.children:
                self.set_page_metadata(child, pages_cache)

    menu_pool.register_modifier(PageDescription)

Probably a better approach need to be implemented in the core but for now this works ok.

yakky:
@vstoykov this is really nice! How about contributing a PR? :smile:

vstoykov:
Because looks too hacky to be  in core (my opinion).

I also thought for another approach. In `cms.menu.page_to_node` when node is created we can  use custom NavigationNode class (for example PageNavigationNode) inherited from NavigationNode which will have method or `cached_property` for getting the page instance.

Pro: Benefit of that is that there will be no need of another modifier which will iterate over nodes on every rendering (even for menus which will not use that feature)

Cons: The downside of this approach compared to the modifier is that it will generate query for every page instead of getting them at once with `in_bulk` and then get the page from the dict.

If we can combine both methods custom NavigationNode and getting page instances in bulk for current menu only this can be better.

donce:
Thanks @vstoykov , I'll probably use modifier with bulk selection for now.

PR #4190, opened 2015-06-12
Fix #4029 - first point of the issue + test

### LINKED ISSUE #4029: cms modal window issues
I encounter some issues with the cms modal window on the 3.0.x version:

- If the content of a numeric input is not valid and you still try to save the model.
  The browser (recent browser with validation type) displays a "Please enter a number" error and then gets stuck on a the loading frame. I guess the loading function is attached to the onclick event of the "submit" button, instead of the submit event of the form. That would be a reason why you get the loading frame but nothing happens after that.

- The second one is a bit more tricky :

  - First open a ckeditor plugin in a popup window then close it with the closing or cancel button.
  Log => POST /admin/cms/page/delete-plugin/10648/ HTTP/1.1" 302

  - Then open it again: 

    - If you close it using the button cancel it'll work. 
    Log => POST /admin/cms/page/delete-plugin/10649/ HTTP/1.1" 302

    - But if you use the button close you'll get a 404, because it keeps the old id instead of the new one. 
    Log => POST /admin/cms/page/delete-plugin/10648/ HTTP/1.1" 404 

I hope the explanation is clear , I can get you some screenshots if you want.

basilelegal:
This PR fix the first point of the issue #4029.

In the case of a decimal field in the form, when submitting the cms_modal iframe would "hide" before the browser's form validation. That would lead to the iframe stuck in hide, no Django's validation performed and nothing the user could do.

It adds the corresponding as well.

### DIFF
--- a/cms/static/cms/js/modules/cms.modal.js
+++ b/cms/static/cms/js/modules/cms.modal.js
@@ -465,6 +465,15 @@ $(document).ready(function () {
 				row = iframe.contents().find('.save-box:eq(0)');
 			}
 			row.hide(); // hide submit-row
+			form = iframe.contents().find('form');
+			//avoids conflict between the browser's form validation and Django's validation
+			form.submit(function(e){
+				if (that.hideFrame) { //submit button was clicked
+					that.modal.find('.cms_modal-frame iframe').hide();
+					//page has been saved, run checkup
+					that.saved = true;
+				}
+			});
 			var buttons = row.find('input, a, button');
 			var render = $('<span />'); // seriously jquery...
 
@@ -507,10 +516,13 @@ $(document).ready(function () {
  							that.options.newPlugin = null;
  							// reset onClose when delete is triggered
 							if(item.hasClass('deletelink')) that.options.onClose = null;
-							// hide iframe
-							that.modal.find('.cms_modal-frame iframe').hide();
-							// page has been saved or deleted, run checkup
-							that.saved = true;
+							if (! item.hasClass('default')) {// hide iframe when using buttons other than submit
+								that.modal.find('.cms_modal-frame iframe').hide();
+								// page has been saved or deleted, run checkup
+								that.saved = true;
+							} else { // submit button uses the form's submit event
+								that.hideFrame = true;
+							}
 						}
 					});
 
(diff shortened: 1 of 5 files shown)


yakky:
Thanks @basilelegal 
Nice test indeed!
Could you add south and django 1.7 migrations? ``develop.py makemigrations`` will take care of this

@FinalAngel could you review the frontend part?

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -507,10 +516,13 @@ $(document).ready(function () {
  							that.options.newPlugin = null;
  							// reset onClose when delete is triggered
 							if(item.hasClass('deletelink')) that.options.onClose = null;
-							// hide iframe
-							that.modal.find('.cms_modal-frame iframe').hide();
-							// page has been saved or deleted, run checkup
-							that.saved = true;
+							if (! item.hasClass('default')) {// hide iframe when using buttons other than submit
```
FinalAngel: watch the spacings, no space after `! ` and space before the comment `{ //´

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -465,6 +465,15 @@ $(document).ready(function () {
 				row = iframe.contents().find('.save-box:eq(0)');
 			}
 			row.hide(); // hide submit-row
+			form = iframe.contents().find('form');
+			//avoids conflict between the browser's form validation and Django's validation
+			form.submit(function(e){
```
FinalAngel: if you don't use `e` than don't add it and mind spacing: ´function () {´

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -465,6 +465,15 @@ $(document).ready(function () {
 				row = iframe.contents().find('.save-box:eq(0)');
 			}
 			row.hide(); // hide submit-row
+			form = iframe.contents().find('form');
```
vxsx: missing declaration, this would be implied global

basilelegal:
Ok tank you for the feedback, I'll do the changes and the migations as soon as I can.

basilelegal:
Okay I had the time to finally fix my error.
@yakky Just a quick question should I push the latest correction with another commit in this PR, or is it better if I open a new PR with the whole correction in one clean commit ?

yakky:
@basilelegal You can continue here, thanks

yakky:
@FinalAngel could you review the frontend?

FinalAngel:
@yakky FE LGTM

basilelegal:
@yakky Just realised I introduced a malfunction due to javascript event priority. I'll fix it ASAP

yakky:
Thanks @basilelegal merged via 27045ad2ab1b37e2b4b26cb073e3a4d11f522dd2

PR #4200, opened 2015-06-17
make shift tab work correctly in submenu

vxsx:
So currently when you click on the submenu in the placeholder or a plugin in structure mode and then start pressing tab the focus would go over menu items. However when you press shift+tab the expected behaviour would be to go to the previous item, which is not the case at the moment.

This PR fixes that.


FinalAngel:
LGTM

PR #4216, opened 2015-06-27
Fix cache invalidation

### LINKED ISSUE #4202: Publishing does not correctly invalidate cache
On Django-Cms 3.1 with Django 1.7 : when publishing a page that contains a placeholder, the live page does not reflect the changes immediately. At least not until cache expires (according to CMS_CACHE_DURATIONS). So it seems that cache invalidation does not work correctly on publish.
This can be bypassed by setting CMS_PLACEHOLDER_CACHE to False.

This bug does not appear to be existing in Django-Cms 3.0

yakky:
And refactor the cache in the process

Refactoring is not needed to fix the issue but it makes way easier.
The scope of the fix is to use version for **every** cached item, not only pages. When a page is published the version is updated for placeholders, pages and menus not only the pages as it is now

Fix #4202

### DIFF
--- a/cms/cache/__init__.py
+++ b/cms/cache/__init__.py
@@ -1,0 +1,87 @@
+# -*- coding: utf-8 -*-
+import re
+from cms.utils import get_cms_setting
+
+CMS_PAGE_CACHE_VERSION_KEY = get_cms_setting("CACHE_PREFIX") + 'CMS_PAGE_CACHE_VERSION'
+
+
+def _get_cache_version():
+    """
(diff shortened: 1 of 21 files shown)


yakky:
@mkoistinen are we brave enough to merge this now? :smile:

mkoistinen:
LGTM, thank you for leaving my comments intact =) I don't think we comment enough in this code.

yakky:
@mkoistinen agree I'm planning more refactoring and documentation for 3.2

PR #4085, opened 2015-05-01
Django 1.8 support

yakky:
Fix #4023
Fix #3993
Fix #4025

PR #4227, opened 2015-07-01
Fix supported django version

yakky:


yakky:
Thanks @blueyed (https://github.com/divio/django-cms/pull/4085#issuecomment-117771567)

PR #4229, opened 2015-07-03
Fix redirect when deleting a page

### LINKED ISSUE #3654: Delete page shouldn't result in 404

yakky:
Fix #3654 

This works by preserving ``this.options.onClose`` when deleting a page.

This does not fix the redirect when deleting an object in a apphooked application from the frontend

### DIFF
--- a/cms/static/cms/js/modules/cms.base.js
+++ b/cms/static/cms/js/modules/cms.base.js
@@ -53,7 +53,7 @@ $(document).ready(function () {
 						if(response === '' && !url) {
 							// cancel if response is empty
 							return false;
-						} else if(parent.location.pathname !== response) {
+						} else if(response !== '' && parent.location.pathname !== response) {
 							// api call to the backend to check if the current path is still the same
 							that.reloadBrowser(response);
 						} else if(url === 'REFRESH_PAGE') {
--- a/cms/static/cms/js/modules/cms.modal.js
+++ b/cms/static/cms/js/modules/cms.modal.js
@@ -501,8 +501,6 @@ $(document).ready(function () {
 						// trigger only when blue action buttons are triggered
 						if(item.hasClass('default') || item.hasClass('deletelink')) {
  							that.options.newPlugin = null;
- 							// reset onClose when delete is triggered
-							//if(item.hasClass('deletelink')) that.options.onClose = null;
 							// hide iframe
 							that.modal.find('.cms_modal-frame iframe').hide();
 							// page has been saved or deleted, run checkup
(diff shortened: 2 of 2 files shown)


yakky:
@FinalAngel Could you have a look at this? I've done some manual tests and apparently nothing is broken, but I'm still away from being fully confident

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -584,7 +584,12 @@ $(document).ready(function () {
 				// when the window has been changed pressing the blue or red button, we need to run a reload check
 				// also check that no delete-confirmation is required
 				if(that.saved && !contents.find('.delete-confirmation').length) {
-					that.reloadBrowser(window.location.href, false, true);
+					if(that.options.onClose){
+						that.reloadBrowser(that.options.onClose);
+					}
```
FinalAngel: add the else on the same line

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -502,7 +502,7 @@ $(document).ready(function () {
 						if(item.hasClass('default') || item.hasClass('deletelink')) {
  							that.options.newPlugin = null;
  							// reset onClose when delete is triggered
-							if(item.hasClass('deletelink')) that.options.onClose = null;
+							//if(item.hasClass('deletelink')) that.options.onClose = null;
```
FinalAngel: space after `//`

yakky:
@mkoistinen This is also present in 3.0.x 
Should I merge this in 3.0.x branch and then forwardport to 3.1?

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -584,7 +582,11 @@ $(document).ready(function () {
 				// when the window has been changed pressing the blue or red button, we need to run a reload check
 				// also check that no delete-confirmation is required
 				if(that.saved && !contents.find('.delete-confirmation').length) {
-					that.reloadBrowser(window.location.href, false, true);
+					if(that.options.onClose){
```
intellectronica: ```if (that.options.onClose) {``` (though i see that this format is consistent with what's currently in the file so maybe it's better to change en masse in a later branch).

cms/static/cms/js/modules/cms.base.js
```diff
@@ -53,7 +53,7 @@ $(document).ready(function () {
 						if(response === '' && !url) {
 							// cancel if response is empty
 							return false;
-						} else if(parent.location.pathname !== response) {
+						} else if(response !== '' && parent.location.pathname !== response) {
```
FinalAngel: space after `if`

vxsx:
I think js whitespace should be adressed later on, since we are already working on that https://github.com/divio/django-cms/compare/develop...FinalAngel:feature/3.2-javascript

FinalAngel:
LGTM

yakky:
@vxsx @FinalAngel @intellectronica thanks
keep in mind that this is targeted to 3.1, so the style has to be consistent with the same codebase
In must be fixed when ported to develop

@mkoistinen do you want this in 3.0 too?

yakky:
Merged in 3.0.x branch via b7866d7b0cd903282136275be8fd487f645f6983

PR #4231, opened 2015-07-03
[WiP] Use python-urlify instead of slugify

yakky:
To be compatible with urlify.js


yakky:
Github gone mad
Closed in favor of #4272

PR #4235, opened 2015-07-03
[WiP] Reversion transaction

### LINKED ISSUE #3811: Broken undo causes corruption in page. Kills page view and page publish operations.
An undo error 'Could not revert revision, due to database integrity errors.' caused Page field 'publisher_public' to become unhinged. It's a double link between draft and live version, but if one of those links is broken (not both!) then the page is in an unstable state.

We had publisher_public==None in the draft(pk=2) and publisher_public==2 set in the live version(pk=942). This meant viewing the page threw error A and attempting to publish the page threw error B.

A:
```
File "/var/www/www.inkscape.org/pythonenv/lib/python2.7/site-packages/cms/models/pagemodel.py", line 954, in has_view_permission
    return self.publisher_draft.has_view_permission(request, user)
DoesNotExist: Page has no publisher_draft.
```

B:
```
File "/var/www/www.inkscape.org/pythonenv/lib/python2.7/site-packages/cms/models/pagemodel.py", line 574, in publish
    public_page.save()
IntegrityError: (1062, "Duplicate entry '2' for key 'publisher_public_id'")
```

The error had to be fixed manually from a django shell:
```
Page.objects.filter(pk=2).update(publisher_public=Page.objects.filter(publisher_public=2)[0])
```
But idealy the software would be able to detect these issues and repair itself, especially when hitting the publish button.

yakky:
Wrapping ``Page._apply_revision`` in transaction should avoid damages if reversion fails

Fix #3811

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -790,7 +790,9 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
             return HttpResponseForbidden(force_text(_("You do not have permission to change this page")))
         try:
             reverted, clean = page.undo()
-            if not clean:
+            if clean is None:
+                messages.error(request, _("Error when applying the undo. Page has not been modified."))
+            elif not clean:
                 messages.error(request, _("Page reverted but slug stays the same because of url collisions."))
         except IndexError as e:
             return HttpResponseBadRequest(e.message)
@@ -809,7 +811,9 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
             return HttpResponseForbidden(force_text(_("You do not have permission to change this page")))
         try:
             reverted, clean = page.redo()
-            if not clean:
+            if clean is None:
+                messages.error(request, _("Error when applying the undo. Page has not been modified."))
+            elif not clean:
                 messages.error(request, _("Page reverted but slug stays the same because of url collisions."))
         except IndexError as e:
             return HttpResponseBadRequest(e.message)
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -7,7 +7,7 @@ from django.contrib.auth import get_permission_codename
 from django.contrib.sites.models import Site
 from django.core.exceptions import ValidationError
 from django.core.urlresolvers import reverse
-from django.db import models
+from django.db import models, transaction
 from django.shortcuts import get_object_or_404
 from django.utils import six
 from django.utils.encoding import force_text, python_2_unicode_compatible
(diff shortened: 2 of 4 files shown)


yakky:
@mkoistinen @ojii I'm a bit halted here. Can't understand why reverting in the transaction (https://github.com/divio/django-cms/pull/4235/files#diff-960fbbc1702d77aa625a17bcb0c37191R1341) does not work on MySQL (and mysql does not make any easier by not having a query error log ...)

ojii:
how does it fail? Does it raise an exception (inside the try/except)? Have you tried enabling db debugging in django and looking at connection.queries, maybe they have a hint.

General (not related to your problem) observations:
Instead of blank try/except and returning ``True/False/None``, why not let it raise the exception and check for ``True/False/Exception`` in the calling function? I find using ``bool/None`` as return values a bit risky (due to the falsyness of ``None``).
And a small nitpick, instead of ``assertTrue(x != y)``, use ``assertNotEqual(x, y)``. 

PS: Installing mysql :(

yakky:
@ojii transaction is set to ``needs_rollback=True`` during ``target_revision.revert(True)`` (here https://github.com/django/django/blob/stable/1.8.x/django/db/transaction.py#L249) with no error reported by the mysql driver apparently because savepoints id has ``None`` value (https://github.com/django/django/blob/stable/1.8.x/django/db/transaction.py#L200) :/
I tried debugging this with no success, any idea?
Letting the Exception bubble might be an option, at this stage I just tried to see if using a transaction would fix the issue (which it does, except the weird mysql behavior ...)

ojii:
I'll try to dig in as soon as I get ``--parallel`` to work with mysql in better-tests ;-)

yakky:
For the scope of this PR you can just run ``cms.tests.reversion_tests.ReversionTestCase.test_undo_redo``

yakky:
@mkoistinen the only solution is to ignore transactions in case of Mysql
Problem is this set of queries:

                 DELETE FROM `cms_page_placeholders` WHERE `cms_page_placeholders`.`page_id` = 1
                 SELECT `cms_page_placeholders`.`placeholder_id` FROM `cms_page_placeholders` WHERE (`cms_page_placeholders`.`placeholder_id` IN (2, 3) AND `cms_page_placeholders`.`page_id` = 1)
                 INSERT INTO `cms_page_placeholders` (`page_id`, `placeholder_id`) VALUES (1, 2), (1, 3)

Probably due to isolation level, the last one fails due to duplicate values, which should not occurr because the records is dropped in query n. 1 ... I give up on understanding the mysql behavior

yakky:
Closing this in favor of new reversion support in 3.2.1

PR #4240, opened 2015-07-05
Extended "Other Plugins" to be consistent with tutorial

creakings:
Added djangocms-flash to "Other Plugins" list to be consistent with the rest of the tutorial e.g. INSTALLED_APPS settings in settings.py.


yakky:
Thanks!

PR #4241, opened 2015-07-05
django.core.context_processors.request also needed for Django 1.8

douwevandermeij:
I got errors with CMS 3.1.2 on Django 1.8.2 that "django.template.context_processors.request" was not in the TEMPLATE_CONTEXT_PROCESSORS due to this constraint.

### DIFF
--- a/cms/utils/setup.py
+++ b/cms/utils/setup.py
@@ -1,7 +1,7 @@
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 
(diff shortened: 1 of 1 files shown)


yakky:
@douwevandermeij that checks must stay with DJANGO_17 which means "Django 1.7 and below"
The template context processor has changed in 1.8

douwevandermeij:
@yakky you're right indeed, I was referring to the CMS docs but Django docs tell the existing way: https://docs.djangoproject.com/en/1.8/ref/templates/upgrading/

PR #4242, opened 2015-07-05
Added missing plugins, clarification of 'SITE_ID' setting, filling the 'static' directory

creakings:
I am currently working through this tutorial and some parts are not fully clear to me. So i try to add a few more comments and clarifications.

- some recommended other plugins were still missing but referenced later in the tutorial
- the static files have to be collected before serving the app or copied to the '/static/' directory (a little subtile for a django novice)
- the SITE_ID has to be set to an integer value, presumably 1 (in my case starting with a completely empty project/app)


docs/how_to/install.rst
```diff
@@ -465,6 +470,17 @@ setting::
 
     Please make sure both the ``static`` and ``media`` subfolders exist in your
     project and are writable.
+
```
yakky: During the tutorial this is not required as runserver takes care of this. Also: the correct way to do this is ``manage.py collectstatic`` command

yakky:
@evildmp would you have a look at this?

docs/how_to/install.rst
```diff
@@ -68,6 +68,11 @@ Other Plugins
 * djangocms-oembed
 * djangocms-table
 * djangocms-flash
+* djangocms-googlemap
+* djangocms-inherit
+* djangocms-picture
+* djangocms-teaser
+* djangocms-video
```
evildmp: This is just a suggested list, with some examples. I don't see the need to add more than a minimum. The user will decide what they need.

evildmp:
@creakings sorry to be so slow about this.

The only change here that I think we really need is the clarification of ``SITE_ID``.

If you're able to rebase on the latest *release* version, I'll be happy to merge the ``SITE_ID`` change.

FinalAngel:
@creakings any chance from your end to update the PR according to the feedback?

PR #4256, opened 2015-07-11
Relax html5lib versions

yakky:


yakky:
@mkoistinen merging into 3.0 and then forward porting this

PR #4261, opened 2015-07-12
[WiP] Fix support for unpublished language pages

### LINKED ISSUE #4214: unpublish page translation on multilingual site does not actually unpublish the language version
Let's say I have a multilanguage site. My page name is
"About" for en language
"O nas" for sk language

For the first time, after the sk "O nas" has been created, but it's not published yet, I see the correct "About" fallback on site and grey indicator in the administration. After publishing sk version "O nas", everything is ok, the indicator changes to green, language version can be accessed by visitors. I decide to unpublish this sk translation for some reason, I click on "UNPUBLISH", but the page remains published in navigation menu and is publicly accessible, it only changes the indicator from green to grey. 

I'm not sure if this is an actual issue (even I expected behavior similar to hide_untranslated), but it at least causes the grey indicator to behave in two different ways. 

I've tried it from cms 3.0.12/django 1.6 to 3.1/django 1.7

thanks, 
j.

yakky:
Fix #4214

### DIFF
--- a/cms/menu.py
+++ b/cms/menu.py
@@ -253,6 +253,8 @@ class CMSMenu(Menu):
 
         if hide_untranslated(lang, site.pk):
             filters['title_set__language'] = lang
+            if not use_draft(request):
+                filters['title_set__published'] = True
 
         if not use_draft(request):
             page_queryset = page_queryset.published()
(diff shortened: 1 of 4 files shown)


yakky:
@mkoistinen this is a pretty bad bug and I think it worth to be fixed in 3.0
In 3.1 we can use a ``languages_published`` field to save the published languages and save some queries. Adding a migration in 3.0 will make upgrades much more complicated due to mismatch migration number

mkoistinen:
@yakky I don't understand your comment about migrations. Are you recommending we commit this or not?

yakky:
@mkoistinen Merge as is in 3.0, and add a field/migration in 3.1

PR #4270, opened 2015-07-14
Moved image to an image file

ojii:


yakky:
Much needed, Jonas. Thanks!

PR #4272, opened 2015-07-15
Use urlify compatible with Djang urlify.js instead of slugify

yakky:


yakky:
Ping @mkoistinen, @mikek for review

mkoistinen:
:100:

Bashar:
How about adding arabic to UBERMAP since it's from the unicode list

As I'm on vacation and on mobile now it'll be little hard to commit it :) 

ا أ آ = a
ب = b
پ = p
ت = t
پ = th
ج = j
چ = ch
ح = ha
خ = kh
د = d
ذ = th
ر = r
إ = e
س = s
ش = sh
ص = sa
ض = dh
ط = ta
ظ = tha
ع = aa
غ = gh
ف = f
ق = q
ك = k
ل = l
م = m 
ن = n
ه = h
ة = tt
و = o
 ؤ = oo
ي ى = y
١ = 1
٢ = 2
٣ = 3
٤ = 4
٥ = 5 
٦ = 6
٧ = 7
٨ = 8
٩ = 9
٠ = 0

I added couple of urdu/farsi letters which might be used but the rest not available on my mobile keymap 

I'll review the Arabic later from pc if you can add them, alternatively if you can't copy paste them I will add them from pc once I'm back :) 

We'll done 👍

Bashar:
پ = th
Is duplicate and can be ignored

yakky:
@Bashar thanks for the suggestion. Apparently, some translitteration (such as ص = sa) does not match django's urlify.js (https://github.com/django/django/blob/stable/1.8.x/django/contrib/admin/static/admin/js/urlify.js#L63)
I'd keep the Django ones, if they are correct, and add the missing ones

Bashar:
@yakky Ah didn't know one exists in django, i didn't know you can map two arabic letters to one english letter that's why i tried to make little alternation based on the sound of the actual arabic letter

Do you think I should submit a PR for django's urlify.js with the other letters and optional urdu/farsi ones?

Bashar:
and on the other hand if a letter in arabic doesn't exist what would urlify.js do to replace it? or it will show a warning and won't allow the user to submit/save it?

yakky:
@Bashar if anything needs to be fixed in urlify.js, yes, just send it over to Django itself. Here we can just do whatever sounds better to you, as I cannot definitely speaks about arabic transliteration :)
If a letter is not found I think it's just removed from the slug

yakky:
@mkoistinen yay or nay?

Bashar:
@yakky I've created a PR at django's https://github.com/django/django/pull/5678 once accepted you can update here maybe :)

Thanks

yakky:
@Bashar awesome!

yakky:
I propose to retire this.
Django 1.9 accepts now unicode in slugs and both python and urlifyjs have been updated. I propose to stick with what django does
@mkoistinen @evildmp @czpython what do you think?

yakky:
Closing this as it's hardly relevant anymore: upstream ``slugify``/``urlify`` it's getting much better and Django it's a better place to fix inconsistencies in this regard

PR #4273, opened 2015-07-15
fix get User model in 0010 migration

igorkud90:


yakky:
@igorkud90 Could you elaborate this?

igorkud90:
@yakky I have a custom User model. When i apply migrations on empty database, i have a error: ProgrammingError: column User.same_field does not exist. This field added in my 0004 migration in my app, but django-cms migrations perfomed before this 0004 migration. We can't import the User model directly as it may be a newer version than this migration expects. We can use the historical version.

yakky:
@igorkud90 I'll merge this into support branch with a further fix for south migration.
Thanks for spotting it

yakky:
Linked PR merged

PR #4274, opened 2015-07-15
JS Bundling

vxsx:
This PR adds the ability to create bundles out of js files so they do not trigger unnecessary requests.

Work was done in:
- https://github.com/FinalAngel/django-cms/pull/55
- https://github.com/FinalAngel/django-cms/pull/56

Additionally there is a change in the CMS.TreeManager because `tree_component.js` relied on it's name to get additional assets

/cc @FinalAngel @yakky


FinalAngel:
LGTM :+1: @yakky

PR #4277, opened 2015-07-16
Remove page from menu items for performance reason.

### LINKED ISSUE #4276: Performance regression in menus in 3.0.x
I found that in 3.0.11 there is a performance regression in menus. One of our site (with more than thousand pages) is now 4 times slower which leads to seconds to load the HTML when there is a cache miss for the whole page or the page can't be cached because some plugin has `cache=False`.

I used `git bisect` to investigate the issue and found that the problematic commit was https://github.com/divio/django-cms/commit/cf82e7d0d7b7bb59bbeb479fad63db95e54284b4

In that change I saw that the the page instance is also attached as attr to menu node. When I remove that line the performance is back to normal.

This change is not forward ported to 3.1.x and for that reason versions >=3.1 are not affected.

I think that the reason of slowdown is because all pages (more than thousand) need to be pickled to put in cache and then restored from a cache every time when menu need to be rendered.

I can't see where this page instance is used in the menu.
@mkoistinen Is there some special need of page instance to be put in `node.attr` only for versions >=3.0.11<3.1.0?

yakky:
Fix #4276 

Fix #4189

### DIFF
--- a/cms/menu.py
+++ b/cms/menu.py
@@ -174,7 +174,6 @@ def page_to_node(page, home, cut):
     # Theses are simple to port over, since they are not calculated.
     # Other attributes will be added conditionnally later.
     attr = {
-        'page': page,
         'soft_root': page.soft_root,
         'auth_required': page.login_required,
         'reverse_id': page.reverse_id,


yakky:
@mkoistinen are you ok with this?

PR #4284, opened 2015-07-18
Fix reachability of pages with expired ancestors

### LINKED ISSUE #4098: Use Publishing Dates to control publishing of sub-tree
A use case could be to create an entire sub-tree of pages, but have them held back from being viewable for visitors until a specific point in time.

If I understand the current workflow correctly, this would involve setting a publishing date for *each* of the sub pages. What I propose is to have pages in a sub-tree inherit their publishing dates from ancestors.

This would line up the "publishing dates" feature with the behavior of the "publish" buttons, which affects the sub-tree for a page.

Have I misunderstood something, is there currently a way to solve this?

yakky:
Fix #4098

### DIFF
--- a/cms/tests/page.py
+++ b/cms/tests/page.py
@@ -697,6 +697,30 @@ class PagesTestCase(CMSTestCase):
         page = get_page_from_request(request)
         self.assertEqual(page, None)
 
+    def test_ancestor_expired(self):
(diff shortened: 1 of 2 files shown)


yakky:
@mkoistinen please review this

PR #4285, opened 2015-07-18
Faster login in frontend tests

ojii:
Drastically speeds up frontend tests (some basic timing I made locally showed that tests that use ``_fastlogin`` instead of ``_login`` are about 2x as fast (``PlaceholderBasicTests`` ran in ~9s instead of ~20s).

The way this works is by using the Django login (``django.contrib.auth.authenticate``/``django.contrib.auth.login``) and setting the session cookie in selenium.


ojii:
@yakky (or anyone else) review?

yakky:
@ojii LGTM, ready for merge?

ojii:
@yakky ready from my side

mkoistinen:
:+1:

yakky:
@ojii more goodies in!

mkoistinen:
I vote we revert this. It causes tests to fail, for example in cms.tests.test_frontend.PlaceholderBasicTests on Mac OS.

PR #4295, opened 2015-07-20
make sure version returned is an int

josjevv:
while deploying to heroku together with memcachedcloud, we ran into a TypeError comparing a byte string to an int.

cms/cache/permissions.py in clear_permission_cache
    if version > 1:

This does not occur when running memcached locally. 

python 3.4.3
django 1.7.9
django-cms 3.1.2


yakky:
Nice, thanks

PR #4301, opened 2015-07-22
Only attempt to delete cache keys if there are some to be deleted

tomwardill:
If no cached keys are found for a particular menu, do not attempt to remove them as this will cause 

`redis.exceptions.ResponseError: wrong number of arguments for 'del' command` 

when used with Redis as the cache backend.

Referenced here: https://github.com/andymccurdy/redis-py/issues/513


yakky:
@tomwardill thanks, will apply this through #4308 on 3.0 branch to include the fix in currently supported versions

yakky:
Fixed by #4308, will be later forward ported

boydjohnson:
This is a blocker for 3.1.x. We are getting the same error as above. Please forward port it to 3.1.x.

PR #4306, opened 2015-07-23
Additional style changes

vxsx:
New look for buttons.
Potentially a breaking change if people relied on behavior that cms-btn-active is a black button instead of a pressed button. But it's way cleaner now.

Also affects buttons in sideframe.

screen shot 2015-07-23 at 15 04 59


vxsx:
not sure why commits from Daniele are in my branch :(

cms/static/cms/sass/mixins/_buttons.scss
```diff
@@ -0,0 +1,64 @@
+// taken from bootstrap with adaptations
+@function important($important) {
+    @if($important == true) {
+        @return !important;
+    } @else {
+        @return;
+    }
+}
+@mixin button-variant($color, $background, $border, $important: false) {
```
FinalAngel: awesome

FinalAngel:
LGTM

PR #4316, opened 2015-07-26
Subclassing Admin classes made possible

LeGast00n:
Here you go !!

### DIFF
--- a/cms/admin/permissionadmin.py
+++ b/cms/admin/permissionadmin.py
@@ -51,8 +51,9 @@ class PagePermissionInlineAdmin(TabularInline):
         # can see only permissions for users which are under him in tree
 
         # here an exception can be thrown
+        model = self.model
         try:
-            qs = PagePermission.objects.subordinate_to_user(request.user)
+            qs = model.objects.subordinate_to_user(request.user)
             return qs.filter(can_view=False)
         except NoPermissionsException:
             return self.objects.get_empty_query_set()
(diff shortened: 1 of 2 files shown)


cms/admin/permissionadmin.py
```diff
@@ -51,8 +51,9 @@ def get_queryset(self, request):
         # can see only permissions for users which are under him in tree
 
         # here an exception can be thrown
+        model = self.model
```
yakky: Why using a different code style instead of doin ``qs = self.model...`` like for PageAdmin?

yakky:
@LeGast00n nice!

cms/admin/permissionadmin.py
LeGast00n: In fact I initially started with the qs = self.model…. just to see that another portion of code had model = self.model.
Finally I did not change it all as I thought it did not make real sense :)

yakky: Could you point me to the ``model = self.model`` syntax?

LeGast00n: Altogether:

permissionadmin.py:54:        model = self.model

permissionadmin.py:117:        model = self.model

settingsadmin.py:42:        model = self.model

yakky: Ok for settingsadmin.py:42
I can't find permissionadmin.py:54, permissionadmin.py:117: aren't those in this PR?
Use the ``qs = self.model...`` syntax everywhere in this PR, we will change settingsadmin later

LeGast00n: Yes I think the last 2 ones are in the PR.

I’ve done and committed the adjustments you asked for now !

cheers

PR #4321, opened 2015-07-28
fix sideframe opening and closing immediately on touch devices

vxsx:
basically first touchend opened the thing and click closed it


FinalAngel:
:+1:

PR #4326, opened 2015-07-29
More welcoming welcome page

ojii:
Currently, when starting a new django CMS project, the first thing you get greeted with is our welcome page, which asks you to add a page or enter edit mode. The first problem with this is that new users might just be confused by the idea to enter an edit mode in the first place, but the bigger issue is what happens if a user hits "add a page". In that case, the user is sent to the admin to add a page, and then is in the admin page list, likely completely lost and not sure where to go next.

This PR proposes we change this welcome page to ask you to log in, then choose the title of your home page, then sends you to the home page in edit mode.

The form asks you for a single input: the title of your home page. This is to keep friction as low as possible and make the whole process as fast/easy as possible. For the template it uses the first template in the list, the language is the current request language and since it's the home page, slug doesn't matter.

TODO:
- [x] Implement logic
- [ ] Actually style/design the pages

### DIFF
--- a/cms/page_rendering.py
+++ b/cms/page_rendering.py
@@ -1,13 +1,22 @@
 # -*- coding: utf-8 -*-
+from cms import constants
+from cms.api import create_page
+from cms.utils.permissions import has_page_add_permission
+from django import forms
 from django.conf import settings
+from django.contrib.admin.forms import AdminAuthenticationForm
+from django.contrib.auth.views import login
 from django.core.urlresolvers import resolve, Resolver404
-from django.http import Http404
+from django.http import Http404, HttpResponseRedirect
+from django.shortcuts import render
 from django.template import RequestContext
 from django.template.response import TemplateResponse
+from django.utils.translation import ugettext_lazy as _, \
+    get_language_from_request
 
 from cms.cache.page import set_page_cache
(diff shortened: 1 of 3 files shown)


evildmp:
The idea sounds great, big thumbs-up from me. What about streamlining it further still, so that there is always a first page ready in the system, with a ``menu_title`` *Home* and a ``title`` *My new django CMS website* (I'd prefer something less cutesy than that, but it's late and I'm tired)?

Then the user can retitle the page, but doesn't even need to if just exploring.

ojii:
The idea of the welcome page is that it's shown when there's no pages yet. So the question would be when/how this "default" page gets created. It can't really be created at setup because that makes life harder for people who want to pre-populate their database. So it would have to be created by the welcome page (after login). I somewhat agree that auto-creating it would be streamlining it, but I kind of like the idea of less magic and empowering the user somewhat by letting them create their very first cms page themselves. A compromise might be to pre-fill the field on the create page step.

mkoistinen:
+1

mkoistinen:
@ojii I believe the spirit of this has been addressed in 3.2. Closing this. If you feel this is in error, please feel free to re-open.

Issue #4323, opened 2015-07-28
Django CMSPlugin.position null raises unorderable types: NoneType() < int()


pawelmarkowski:
Traceback (python 33 dj 183 cms 3.1.2) latest releases of plugins from pip:

```
...
File "/srv/venv/py33dj183/lib/python3.3/site-packages/cms/templatetags/cms_tags.py" in render_tag
  302.             content = get_placeholder_content(context, request, page, name, inherit, nodelist)
File "/srv/venv/py33dj183/lib/python3.3/site-packages/cms/templatetags/cms_tags.py" in get_placeholder_content
  221.         placeholder = _get_placeholder(current_page, page, context, name)
File "/srv/venv/py33dj183/lib/python3.3/site-packages/cms/templatetags/cms_tags.py" in _get_placeholder
  200.         assign_plugins(context['request'], fetch_placeholders, page.get_template(),  get_language())
File "/srv/venv/py33dj183/lib/python3.3/site-packages/cms/utils/plugins.py" in assign_plugins
  80.         plugin_groups[group] = build_plugin_tree(plugin_groups[group])
File "/srv/venv/py33dj183/lib/python3.3/site-packages/cms/utils/plugins.py" in build_plugin_tree
  138.                       key=attrgetter('parent_id', 'position'))
 
Exception Type: TypeError at /
Exception Value: unorderable types: NoneType() < int()
```

1. Restore the database
2. Run the command `manage.py cms fix-tree`
3. Make alias of plugin instance and put into the placeholder
4. Click "publish"
and the problem occures each time in my case ... after step 3 and 4. Script `manage.py cms fix-tree` is solving that problem temporary
In webbrowser console I can see: 
a) Uncaught TypeError: Cannot read property 'placeholder_id' of undefined cms.structuredboard.js:330 which is id = item.data('settings').placeholder_id;
b) GET http://xxxx.xxx/?edit_off 500 (INTERNAL SERVER ERROR)
c) cms.toolbar.js:234 Uncaught Error: [object Object]

regards

ojii:
For the record, IRC logs: https://botbot.me/freenode/django-cms/2015-07-28/?msg=45802968&page=3

pawelmarkowski:
A problem is not in the copy of a plugin instance but in his alias. Click right context menu in a structure view and choose option "Alias". Put alias into a placeholder and try to click publish... This steps should reproduce the error.

ojii:
[...]

PR #4328, opened 2015-07-29
CMSPlugin null position fix

pawelmarkowski:
refs #4323


pawelmarkowski:
position after copy have to been calculated ...

PR #4329, opened 2015-07-30
CMSPlugin.position not null 

pawelmarkowski:
Django CMSPlugin.position null raises unorderable types: NoneType() < int()

### DIFF
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -75,7 +75,7 @@ class CMSPlugin(six.with_metaclass(PluginModelBase, MP_Node)):
     '''
     placeholder = models.ForeignKey(Placeholder, editable=False, null=True)
     parent = models.ForeignKey('self', blank=True, null=True, editable=False)
-    position = models.PositiveSmallIntegerField(_("position"), blank=True, null=True, editable=False)
+    position = models.PositiveSmallIntegerField(_("position"), default = 0, editable=False)
     language = models.CharField(_("language"), max_length=15, blank=False, db_index=True, editable=False)
     plugin_type = models.CharField(_("plugin_name"), max_length=50, db_index=True, editable=False)
     creation_date = models.DateTimeField(_("creation date"), editable=False, default=timezone.now)
@@ -319,6 +319,7 @@ class CMSPlugin(six.with_metaclass(PluginModelBase, MP_Node)):
             plugin_instance.language = target_language
             plugin_instance.parent = new_plugin.parent
             plugin_instance.depth = new_plugin.depth
+            plugin_instance.position = new_plugin.position
             plugin_instance.path = new_plugin.path
             plugin_instance.numchild = new_plugin.numchild
             plugin_instance._no_reorder = True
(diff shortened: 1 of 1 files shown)


pawelmarkowski:
CMSPlugin.position not null + default 0 passes tests and solving my issue #4323

pawelmarkowski:
@yakky @ojii   I do not understand save and copy_plugin methods. **copy_plugin** method rewrites _plugin_instance.depth = new_plugin.depth_ and save method will have executed. In save method is condition: _if not self.depth_ so position will be never calculated... Am I right? But if yes ... this should be detected by tests ...
```
def save(self, no_signals=False, *args, **kwargs):
        if not self.depth:
            if self.parent_id or self.parent:
                self.parent.add_child(instance=self)
            else:
                if not self.position and not self.position == 0:
                    self.position = CMSPlugin.objects.filter(parent__isnull=True,
                                                             language=self.language,
                                                             placeholder_id=self.placeholder_id).count()
                self.add_root(instance=self)
            return
        super(CMSPlugin, self).save()
[...]

    def copy_plugin(self, target_placeholder, target_language, parent_cache, no_signals=False):
[...]
        new_plugin.save()
        if plugin_instance:
            # get a new instance so references do not get mixed up
            plugin_instance = plugin_instance.__class__.objects.get(pk=plugin_instance.pk)
            plugin_instance.pk = new_plugin.pk
            plugin_instance.id = new_plugin.pk
            plugin_instance.placeholder = target_placeholder
            plugin_instance.cmsplugin_ptr = new_plugin
            plugin_instance.language = target_language
            plugin_instance.parent = new_plugin.parent
            ->plugin_instance.depth = new_plugin.depth<-
            plugin_instance.path = new_plugin.path
            plugin_instance.numchild = new_plugin.numchild
            plugin_instance._no_reorder = True
            plugin_instance.save()
            old_instance = plugin_instance.__class__.objects.get(pk=self.pk)
            plugin_instance.copy_relations(old_instance)
        if no_signals:

            signals.pre_save.connect(pre_save_plugins, sender=CMSPlugin, dispatch_uid='cms_pre_save_plugin')

        return new_plugin

yakky:
@pawelmarkowski in ``copy_plugin`` you're creating an exact copy of said plugin, so depth and position is going to be the same. Mind that ``plugin instance`` retains all the original values of the original plugin, while it gets the new primary key and other information from the "void" plugin (``new_plugin``).

pawelmarkowski:
@yakky if I good remember version with additional line after ```plugin_instance.depth = new_plugin.depth``` with  ```plugin_instance.position = new_plugin.position``` has failed the tests, so that is why I ask ... (https://github.com/pawelmarkowski/django-cms/commit/43a14876bf6f9190b60ae402b77fcd56d13df437)

yakky:
@pawelmarkowski srry I overlooked your reply. I'll have a look at this in the next few days

FinalAngel:
@yakky any update on this? :)

yakky:
@FinalAngel unfortunately not. Being it a fix, we can ship in a patch release later. We basically need to come up with a testing strategy for this

PR #4336, opened 2015-08-04
Pagetree touch support

vxsx:
this PR adds touch support to page tree.
it is not perfect, sometimes can be fairly slow

had to do changes to jstree, some are very specific to cms implementation which is very bad, but the plan is to upgrade that part of the cms in 3.x so yeah

unsolved problems so far:

- click on the row to go to the page sometimes doesn't always work
- ~~when dragging - iframe size changes~~  not possible to fix because iOS for example doesn't respect the iframe boundaries at all and always makes the iframe size fit the content
- ~~publishing tooltip is not a nice interaction yet~~ fixed

Before:

screen shot 2015-08-04 at 13 45 26

After:

screen shot 2015-08-04 at 13 42 16

/cc @FinalAngel


cms/static/cms/js/jstree/tree_component.js
```diff
@@ -869,17 +872,26 @@ function tree_component () {
...
                             event.stopPropagation();
                             return false;
-                        });
-                    jQuery(document)
-                        .bind("mousedown",    tree_component.mousedown)
-                        .bind("mouseup",    tree_component.mouseup)
-                        .bind("mousemove",    tree_component.mousemove);
-                } 
+                        })
+                        .listen("pointerup", "li", function (e) {
```
FinalAngel: single quotes / double quotes (several places)

cms/static/cms/js/modules/cms.changelist.js
```diff
@@ -211,7 +212,8 @@
 
             setupUIHacks: function () {
                 // enables tab click on title entry to open in new window
-                $('.tree').delegate('.col1 .title', 'click', function (e) {
+                $('.tree').on('click', '.col1 .title', function (e) {
+                    console.log('click from changelist');
```
FinalAngel: console

FinalAngel:
LGTM

PR #4348, opened 2015-08-10
Move Aldryn Apphooks Reload into core

mkoistinen:
Apphooks reload in core CMS with tests and documentation.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -437,3 +437,8 @@ Please see Install/2.4 release notes *before* attempting to upgrade to version 2
 
 - Fix placeholder cache invalidation under some circumstances
 - Update translations
+
+
+=== 3.1.3 (2015-xx-xx) ===
+
+- Pulls Aldryn Apphook Reload into core.
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -439,6 +439,6 @@ Please see Install/2.4 release notes *before* attempting to upgrade to version 2
 - Update translations
 
 
-=== 3.1.3 (2015-xx-xx) ===
+=== 3.2.0 (2015-xx-xx) ===
 
 - Pulls Aldryn Apphook Reload into core.
--- a/cms/signals/__init__.py
+++ b/cms/signals/__init__.py
@@ -8,6 +8,7 @@ from cms.signals.reversion_signals import post_revision
 from cms.signals.title import pre_save_title, post_save_title, pre_delete_title, post_delete_title
 from cms.utils.compat.dj import is_installed
 from cms.utils.conf import get_cms_setting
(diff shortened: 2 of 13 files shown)


cms/signals/__init__.py
```diff
@@ -26,10 +27,21 @@
 post_unpublish = Signal(providing_args=["instance", "language"])
 
 # fired if a public page with an apphook is added or changed
-urls_need_reloading = Signal(providing_args=[])
+cms_urls_need_reloading = Signal(providing_args=[])
```
yakky: I think we should not change the signal name without a deprecation period, and in a patch release

cms/signals/apphook.py
```diff
@@ -1,14 +1,25 @@
...
 from django.core.urlresolvers import clear_url_caches
 from django.core.signals import request_finished
 from cms.models import Title
+from cms.utils.apphook_reload import mark_urlconf_as_changed
 
 
 DISPATCH_UID = 'cms-restart'
 
 
+def trigger_server_restart(**kwargs):
```
yakky: This looks the same as https://github.com/divio/django-cms/pull/4348/files#diff-643105bbf8c6920853343a09e34d38acR38

cms/utils/apphook_reload.py
```diff
@@ -0,0 +1,104 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import unicode_literals, absolute_import
```
yakky: We should conform to the rest of the codebase and not use this, ATM (to be clear, I think we should update all the codebase to use this and print_function, but we should be consistent)

CHANGELOG.txt
```diff
@@ -437,3 +437,8 @@ Please see Install/2.4 release notes *before* attempting to upgrade to version 2
 
 - Fix placeholder cache invalidation under some circumstances
 - Update translations
+
+
+=== 3.1.3 (2015-xx-xx) ===
```
yakky: This should be 3.2, right?

cms/signals/__init__.py
```diff
@@ -26,10 +27,21 @@
...
-    urls_need_reloading.connect(debug_server_restart)
+    cms_urls_need_reloading.connect(debug_server_restart)
+
+
+def trigger_server_restart(**kwargs):
+    mark_urlconf_as_changed()
+
+cms_urls_need_reloading.connect(
+    trigger_server_restart,
+    dispatch_uid='aldryn-apphook-reload-handle-urls-need-reloading'
```
yakky: I'll change this to ``cms-reload-handle-urls-need-reloading``

cms/signals/apphook.py
mkoistinen: True. Sorted.

yakky:
@mkoistinen :+1:

PR #4359, opened 2015-08-14
Fix for issue #4254

### LINKED ISSUE #4254: Login form on CMS Toolbar should use `USERNAME_FIELD`
Django allows custom User models that user alternative fields instead of "username". Most common example of this is probably email-only User model, i.e. User model that does not have "username" field and instead relies on "email" field for identification.

To this end, Dajngo provides `USERNAME_FIELD`: https://docs.djangoproject.com/en/1.7/topics/auth/customizing/#django.contrib.auth.models.CustomUser.USERNAME_FIELD

Login form on the CMS Toolbar should follow this, and not hardcode "Username" for the form label. 
NB - the login form itself works fine, only the label is wrong.

`templates/cms/toolbar/items/login.html` seems like the place to start (or finish) this.

jproffitt:
Fix #4254

### DIFF
--- a/cms/toolbar/toolbar.py
+++ b/cms/toolbar/toolbar.py
@@ -8,7 +8,6 @@ from cms.utils.compat.dj import installed_apps
 from cms.utils.conf import get_cms_setting
 from cms.utils.i18n import force_language
 
-from django import forms
 from django.conf import settings
 from django.contrib.auth import login, logout, REDIRECT_FIELD_NAME
 from django.contrib.auth.forms import AuthenticationForm
--- a/cms/toolbar/toolbar.py
+++ b/cms/toolbar/toolbar.py
@@ -23,7 +23,6 @@ except ImportError:
 
 
 class CMSToolbarLoginForm(AuthenticationForm):
-    username = forms.CharField(max_length=100)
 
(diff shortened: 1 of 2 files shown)


cms/toolbar/toolbar.py
```diff
@@ -24,7 +23,7 @@
 
 
 class CMSToolbarLoginForm(AuthenticationForm):
-    username = forms.CharField(label=_("Username"), max_length=100)
+    username = forms.CharField(max_length=100)
```
yakky: I assume we can completely remove this, it's already defined in ``AuthenticationForm``, right?

jproffitt: Well yes, but the `username` in `AuthenticationForm` has a `max_length` of 254. I guess that should probably match whatever django sets. I wasn't sure if there was a reason it was set to 100 for the cms.

yakky: Legacy code from the days django had very short usernames. django CMS does not define a user, so we can now get rid of it and just use whatever Django provides

jproffitt: Ok, I have added a new commit to remove it.

yakky:
@jproffitt awesome, thanks

PR #4363, opened 2015-08-15
Fix Page.get_cached_ancestors order

### LINKED ISSUE #4305: Placeholder inherits NOT from nearest ancestor because of wrong order in list
**Environment**
Django==1.6.11
django-cms==3.1.1

**Issue Description**
On a page which is nested 4 levels deep in the page tree we discovered that the placeholder in the page is not inheriting from its nearest ancestor. Instead of inheriting from the nearest ancestor it inherits from the page which is the nearest to the root and contains such a plaeholder. 

**Possible Section of Code which causes this**
https://github.com/divio/django-cms/blob/develop/cms/templatetags/cms_tags.py#L219-L220

```
pages = chain([current_page], current_page.get_cached_ancestors())
````
pages is an array which contains the pages in the following order for traversal:
[current_page, root, page_level_1, page_level_2]
so what happens is that it starts with the current  page, then jumps to the root, then to the page nested one level deep, and  so on... As far as i understand it should instead go in this order: current_page, page_level_2, page_level_1 and so on.

I hope i could explain it good enough :) Thanks for all your hard work!

### PR DESCRIPTION
**Pull request by yakky**
treebeard has the opposite default ordering of mptt

Fix #4305

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -790,8 +790,10 @@ class Page(six.with_metaclass(PageMetaClass, MP_Node)):
         return sorted([language for language in self.get_languages() if self.is_published(language)])
 
     def get_cached_ancestors(self):
+        # Unlike MPTT, Treebeard returns this in parent->child order, so you will have to reverse
+        # this list to have the same behavior as before
         if not hasattr(self, "ancestors_ascending"):
-            self.ancestors_ascending = list(reversed(self.get_ancestors()))
+            self.ancestors_ascending = list(self.get_ancestors())
         return self.ancestors_ascending
 
     def get_cached_descendants(self):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cms/models/pagemodel.py [hunk 1]
```diff
@@ -791,7 +791,7 @@ def get_published_languages(self):
 
     def get_cached_ancestors(self):
         if not hasattr(self, "ancestors_ascending"):
-            self.ancestors_ascending = list(self.get_ancestors())
+            self.ancestors_ascending = list(reversed(self.get_ancestors()))
```
ojii: won't this make ``get_cached_ancestors`` inconsistent with ``get_ancestors``? Looking at their names I would assume they both return the same thing.

yakky: Good point. I'll fix the consumer of this method, then

yakky:
@ojii PR updated

yakky:
@ojii does it look saner?

ojii:
out of town today, will have a look tomorrow
2015/08/18 午後3:17 "Iacopo Spalletti" <notifications@github.com>:

ojii:
actually the diff is easy enough to check on mobile, merging.

yakky:
@ojii ty

PR #4365, opened 2015-08-18
Re-add support for AUTH_USER_MODEL environment variable

yakky:
I think support for custom user model has been removed from new manage.py

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -7,6 +7,15 @@ from cms.utils.compat import DJANGO_1_6, DJANGO_1_7
 
 gettext = lambda s: s
 
+__doc__ = """django CMS development helper script.
+
+Executes the standard ``manage.py`` command in the django CMS test environment, over the standard
+Django options it accepts further configuration:
+
+* To use a different database, set the DATABASE_URL environment variable to a
+  dj-database-url compatible value
+* The AUTH_USER_MODEL environment variable can be used to change the user model to a custom one'
+"""
 
 if __name__ == '__main__':
     os.environ.setdefault(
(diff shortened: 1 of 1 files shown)


ojii:
I think ideally this would be done with the built-in configuration system for django-app-manage, as for example is done for ``USE_TZ``: https://github.com/divio/django-cms/blob/develop/manage.py#L189-L193

See https://github.com/ojii/django-app-manage/blob/master/app_manage/config.py#L75

This way it will automatically support either env vars or a CLI argument.

Probably could look something like:

```
Config(
    env='AUTH_USER_MODEL',
    arg='--auth-user-model',
    default='django.contrib.auth.models.User'
)

**Comment by yakky** [comment]
@ojii should I send a PR to django-app-manage?

**Comment by yakky** [comment]
@ojii how can we add the different applications in this case?
Do we need a second setting?

**Comment by ojii** [comment]
will think of an elegant way to solve this

File: manage.py [hunk 1]
```diff
@@ -169,6 +178,11 @@
             'mti_pluginapp': 'cms.test_utils.project.mti_pluginapp.migrations',
         }
 
+    if os.environ.get('AUTH_USER_MODEL', False):
+        custom_user_app = 'cms.test_utils.project.' + os.environ.get('AUTH_USER_MODEL', False).split('.')[0]
+        INSTALLED_APPS.insert(INSTALLED_APPS.index('cms'), custom_user_app)
```
ojii: is this order important? (insert over append)

yakky: Yes, it is: http://django-cms.readthedocs.org/en/develop/reference/configuration.html#custom-user-requirements

ojii:
See #4370 for my implementation of this.

yakky:
Fixed by #4370

PR #4372, opened 2015-08-20
Toolbar touch

vxsx:
Previously in order to get to submenus before you had to drag your finger around, now simple taps work as well


Brankub:
:100:

PR #4383, opened 2015-08-26
show_placeholder do not ignore lang anymore

iddqd1:
There was the bug in show_placholder.  lang parameter was ignored

### DIFF
--- a/cms/templatetags/cms_tags.py
+++ b/cms/templatetags/cms_tags.py
@@ -581,7 +581,7 @@ def _show_placeholder_for_page(context, placeholder_name, page_lookup, lang=None
             raise
         return {'content': ''}
     watcher = Watcher(context)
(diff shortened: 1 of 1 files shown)


yakky:
This definitely needs a regression test. Looking at the code this actually behaved this way for a very long time

evildmp:
@iddqd1 Hi, can you add a regression test to this patch? If you need some help writing one, please say.

yakky:
Moving to 3.2 so that we can actually document the new behavior

evildmp:
This should be targeted at develop, no?

yakky:
@evildmp yes. Will merge post rc1. I prefer to let the new things settle down before merging bugfixes

yakky:
To merged via #4584

PR #4384, opened 2015-08-26
[WIP] Fix env test

mkoistinen:
Setting the ENV var SELENIUM to anything, including `0` previously was causing Selenium tests to be skipped. This is rather unintuitive. Now, use `SELENIUM=0` to force skip. Otherwise, it runs tests.

Heads-up.
-------------

The current .travis.yml uses ENV var 'SELENIUM' to turn on and off selenium tests. Previously, the mere usage of this ENV var was disabling all selenium tests full stop. This PR corrects the issue so now, selenium is actually being run on Travis for some of the tests (where SELENIUM=1). This has revealed that Selenium tests were actually failing for who knows how long.


mkoistinen:
@ojii OK, I'm stopping this for the night, but it looks like no matter what I do, I cannot login and then maintain the session to the next page. I've tried both the new fast login, but also the older login method here. Any ideas? (Everything works well locally, of course).

czpython:
@mkoistinen I believe this is no longer needed. Can we close?

vxsx:
Selenium tests were complete removed, no longer relevant

PR #4387, opened 2015-08-27
Fix nested plugins in text plugin

vxsx:
It is not allowed to drag anything in or out of a text plugin (or any disabled_child_plugins plugin for that matter).

So the styles and js were updated to not allow this.

Before:
screen shot 2015-08-27 at 11 57 47

After:
screen shot 2015-08-27 at 11 57 04


intellectronica:
:+1:

PR #4388, opened 2015-08-27
Structureboard speedup

vxsx:
mostly js speed optimizations, caching selectors, removing unnecessary timeouts
and a bug fix for oldschool structuremode

this PR is build from the top of https://github.com/divio/django-cms/pull/4372 so first three commits are actually from there


intellectronica:
LG™

PR #4389, opened 2015-08-27
Expand collapse speedup

vxsx:
speed of expanding/collapsing a tree of about 130 plugins went down from ~7s to ~30ms
further optimizations will follow

before:
before

after:
after


Brankub:
:trophy:

Brankub:
LGTM

intellectronica:
:+1:

PR #4391, opened 2015-08-28
Update django.po

phuong:
Translating to vietnamese


yakky:
@phuong please requeste vietnamese translation via transifex https://www.transifex.com/digi604/django-cms/dashboard/

phuong:
Ah, got it, thanks.

PR #4404, opened 2015-08-30
add page slug to navigation node

AlexProfi:
Please add it to cms. It is need in navigations with dropdowns for example topmenu here
http://136.243.204.118:8000/
data-dropdown options

### DIFF
--- a/cms/cms_menus.py
+++ b/cms/cms_menus.py
@@ -160,6 +160,8 @@ def page_to_node(page, home, cut):
     # Do we have a redirectURL?
     attr['redirect_url'] = page.get_redirect()  # save redirect URL if any
 
+    attr['slug'] = page.get_slug() #save page slug
+
     # Now finally, build the NavigationNode object and return it.
     ret_node = NavigationNode(
         page.get_menu_title(),


evildmp:
Would it be more useful to add the page id to a ``NavigationNode``? Then the menu template could have access, if needed, to all of a page's attributes.

@yakky @mkoistinen comments?

AlexProfi:
Yes it's good idea

yakky:
The only issue is potential performance effect in case of very big menus, we should be able to check this when adding attributes to a ``NavigationNode``

yakky:
@evildmp @mkoistinen @czpython I am +1 on adding slug to the menu (the memory impact is negligible even with very large menus)
As for the complete page, we already tried that and we had to remove it due to huge memory impact

mkoistinen:
I'm not opposed to the slug, but its not unique (can be the same at different levels), why not a page ID? Isn't that better?

yakky:
@mkoistinen it's not unique but it can be helpful to provide some context. page.pk it's **too** unique maybe?

mkoistinen:
@evildmp and I discussed the idea of adding a page_id attribute with the page.pk. This should only be present if the node does in fact represent a Page instance. This would solve the immediate issue, but perhaps isn't as extensible as it could be. If we later have a lot of types of things that can be menu nodes, then the nodes would get pretty cluttered with attributes of different sorts.

So, we thought it best that we should add two attributes: content_type_id and object_id. In the common case, the content_type_id would reference the Page model and the object_id, of course, would reference the instance's pk.

I think this solves the issue in the general case and is in-fact, rather Djangonic.

yakky:
I don't think primary key is of much use, actually, it's really too specific. slug and/or reverse id are good candidates as additional keys to attrs dictionary, like other.
Adding a 'type' attribute to the navigation node can be helpful, but id'r rather user something like ``app_name__model_name`` as menus are mostly used in templates, where you can't easily work with queryset to detect the required content type.
As NavigationNode is completely disconnected from the concept and implementation of models we should keep implementation details like ids/pks/foreign keys out of it.

stefanfoulis:
Is the slug (url part at the current level) really useful for this? It's only unique at the current url level. Does it still work for the ``data-dropdown`` case when the same slug appears twice on the site? Wouldn't it fall apart when using navigation nodes that are not based on cms pages (do they all have slugs?).
So the unique thing (most of the time, but there can be exceptions) would be the whole path, but that has annoying ``/``s in it.

czpython:
I think we should rethink the way menu nodes interact with pages, the end goal would be to make it possible to access any attribute of the page/title while keeping the memory footprint low.
I have nothing against adding the slug of the page but don't see it being very useful for the reasons outlined by @stefanfoulis and also can become an issue as we start adding more and more attributes to the object.

ojii:
> I think we should rethink the way menu nodes interact with pages, the end goal would be to make it possible to access any attribute of the page/title while keeping the memory footprint low.

Please keep in mind that node != page. Not every node has a page associated with it, which is why there's no page attribute on the node. The idea is/was that the menu code is mostly independent from the page code and that nodes should be mostly standalone.

yakky:
How about having a very simple way for developers to add desired page attributes for nodes (like a modifier that read some setting or an easy-to-subclass modifier)
NavigationNode should be kept totally agnostic on the item it represents

czpython:
Thanks for the contribution @AlexProfi.
I'm closing this as it requires a design decision by the core team.

PR #4409, opened 2015-09-01
Plugins cleanup

vxsx:
there could also be some savings about traversing the dropdown but this will be addressed later
there also doesn't seem to be much need of passing subnav around however there are no significant wins eliminating that


cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -7,21 +7,29 @@
 
     // TODO move out to separate module CMS-276
     var KEYS = {
```
FinalAngel: didn't we want to do that somewhere globally?

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -627,53 +649,27 @@
...
-                if ($(window).height() + $(window).scrollTop() - nav.offset().top - dropdown.height() <= 10 &&
-                    nav.offset().top - dropdown.height() >= 0) {
-                    dropdown.css('top', 'auto');
-                    dropdown.css('bottom', offset);
-                    // if parent is within a plugin, add additional offset
-                    if (dropdown.closest('.cms-draggable').length) {
-                        dropdown.css('bottom', offset - 1);
-                    }
+                if (this.ui.window.height() + this.ui.window.scrollTop() -
+                    dropdown.offset().top - dropdown.height() <= 10 && nav.offset().top - dropdown.height() >= 0) {
```
FinalAngel: too long?

cms/templates/cms/toolbar/placeholder.html
```diff
@@ -6,21 +6,16 @@
...
-        'plugin_language': '{{ language }}',
-        'plugin_restriction': [{% for module in allowed_plugins %}"{{ module}}"{% if not forloop.last %}, {% endif %}{% endfor %}],
-        'urls': { {% language request.toolbar.toolbar_language %}
-            'add_plugin': '{{ placeholder.get_add_url }}',
-            'copy_plugin': '{{ placeholder.get_copy_url }}'
-        } {% endlanguage %}
-    });
-});
-})(CMS.$);
+CMS._plugins.push(['cms-placeholder-{{ placeholder.pk|unlocalize }}', {
```
FinalAngel: nice work

cms/static/cms/js/modules/cms.plugins.js
vxsx: yep, going to move them all away together in a separate PR

cms/static/cms/js/modules/cms.plugins.js
vxsx: not really

cms/templates/cms/toolbar/placeholder.html
vxsx: actually let's get rid of unnecessary quotes as well

FinalAngel:
LGTM

Issue #4243, opened 2015-07-06
RemovedInDjango19Warnings


blueyed:
With Django 1.8 there are several `RemovedInDjango19Warning` warnings triggered by django-cms:

> RemovedInDjango19Warning: The django.contrib.admin.util module has been renamed. Use django.contrib.admin.utils instead.

via:

    cms/admin/placeholderadmin.py
    8:from django.contrib.admin.util import get_deleted_objects

    cms/admin/pageadmin.py
    15:from django.contrib.admin.util import get_deleted_objects

    cms/tests/permmod.py
    6:from django.contrib.admin.util import unquote

And `django.utils.importlib will be removed in Django 1.9.` via:

      …/django-cms/cms/utils/django_load.py(15)<module>()
         13 import traceback # changed
         14 
    ---> 15 from django.utils.importlib import import_module
         16 from django.utils.six.moves import filter, map
         17

yakky:
@blueyed which django CMS version are you using? With django CMS 3.1.2 there should be no warning regarding 1.9

blueyed:
@yakky
I was using the develop branch, but it's the same with 3.1.2.

`cms/utils/django_load.py` is using `from django.utils.importlib import import_module`, for example.

yakky:
@blueyed #4255 should remove any deprecation warning in the django CMS codebase.
Thanks for sharing this

yakky:
Fixed in #4255

jproffitt:
I'm still seeing a couple of `RemovedInDjango19Warning`'s after upgrading to CMS 3.1.3:

    cms/models/permissionmodels.py:7: RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.

    cms/utils/compat/forms.py:11: RemovedInDjango19Warning: django.db.models.get_app is deprecated.

PR #4413, opened 2015-09-02
Fixing more RemovedInDjango19Warnings fixes issue #4243

jproffitt:


bittner:
#4438 is partially a duplicate of this PR.

(Not sure about the `apps.get_app_config` change, really. Didn't see with Django 1.8 yet. When does it occur?)

vstoykov:
Now I also see from where it comes https://github.com/django/django/blob/1.8.4/django/db/models/__init__.py#L48-L64

This is triggered only when `settings.AUTH_USER_MODEL != 'auth.User'` https://github.com/jproffitt/django-cms/blob/develop/cms/utils/compat/forms.py#L12.

I think that this PR is better than mine (#4438) and also first in the queue ;). I don't know how I missed it.

Probably the only problem with this PR is that it introduces two commits where only one is needed.

jproffitt:
`models.get_model()` is deprecated, and will be removed in 1.9. Sorry about the two commits. I can try to redo this if needed.

cms/models/permissionmodels.py
```diff
@@ -4,7 +4,11 @@
 from django.contrib.auth.models import Group, UserManager
 from django.contrib.sites.models import Site
 from django.core.exceptions import ImproperlyConfigured
-from django.utils import importlib
+try:
```
bittner: This import should be moved to line 20 (after `if DJANGO_1_6:`).

And maybe, when we're at it, we should use `from importlib import import_module` instead, and `... = import_module(...)` on line 23 to make things shorter to read.

vstoykov: If the import is moved after lif DJANGO_1_6 there is no need to use `importlib` because we will brake python2.6. For dajngo 1.6 there is no problem to use `from django.utils import importlib`

bittner: :+1: Makes the code stay slick.

yakky:
I'm merging this through #4446

PR #4419, opened 2015-09-04
Fix Structure mode migration

### LINKED ISSUE #4411: [3.1.3] AttributeError: 'User' object has no attribute 'has_perm'
Hi,

Using orm['auth.Model'] in `0075_use_structure.py` migration can not have `has_perm` attribute:

>    File "/home/dcordel/workspace/webu/projects/ffcriavs/python-env/local/lib/python2.7/site-packages/cms/south_migrations/0075_use_structure.py", line 39, in forwards
>      if user.has_perm("cms.change_page"):
>  AttributeError: 'User' object has no attribute 'has_perm'

Using "real" `User` model (like in 3.1.2) is not a good solution. What do you recognize for this ? I could write a PR but I'm not sure about which solution is the best : using real User model ? Permission queryset on user and user's groups ? other ?

yakky:
Fix #4411 (hopefully)

### DIFF
--- a/cms/migrations/0010_migrate_use_structure.py
+++ b/cms/migrations/0010_migrate_use_structure.py
@@ -16,10 +16,10 @@ def forwards(apps, schema_editor):
         page_ctype = ContentType.objects.get_for_model(page_model)
         permission, _ = Permission.objects.get_or_create(
             codename='use_structure', content_type=ph_ctype, name=u"Can use Structure mode")
-        page_permission = Permission.objects.get_or_create(codename='change_page', content_type=page_ctype)
+        page_permission, __ = Permission.objects.get_or_create(codename='change_page', content_type=page_ctype)
         for user in user_model.objects.filter(is_superuser=False, is_staff=True):
-            if user.has_perm("cms.change_page"):
-                user.user_permissions.add(permission)
+            if user.user_permissions.filter(codename='change_page', content_type_id=page_ctype.pk).exists():
+                user.user_permissions.add(permission.pk)
         for group in Group.objects.all():
             if page_permission in group.permissions.all():
                 group.permissions.add(permission)
@@ -34,8 +34,7 @@ def backwards(apps, schema_editor):
     permission, _ = Permission.objects.get_or_create(
         codename='use_structure', content_type=ph_ctype, name=u"Can use Structure mode")
     for user in user_model.objects.filter(is_superuser=False, is_staff=True):
-        if user.has_perm("cms.use_structure"):
-            user.user_permissions.remove(permission)
+        user.user_permissions.remove(permission)
     for group in Group.objects.all():
         if permission in group.permissions.all():
             group.permissions.remove(permission)
--- a/cms/migrations/0010_migrate_use_structure.py
+++ b/cms/migrations/0010_migrate_use_structure.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-from __future__ import unicode_literals
 import warnings
 from django.contrib.auth.models import Permission, Group
 from django.contrib.contenttypes.models import ContentType
--- a/cms/migrations/0010_migrate_use_structure.py
+++ b/cms/migrations/0010_migrate_use_structure.py
@@ -23,7 +23,7 @@ def forwards(apps, schema_editor):
         for group in Group.objects.all():
             if page_permission in group.permissions.all():
                 group.permissions.add(permission.pk)
-    except ContentType.DoesNotExist:
+    except Exception:
         print(u'Users not migrated to use_structure permission, please add the permission manually')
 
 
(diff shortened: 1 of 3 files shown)


cms/south_migrations/0075_use_structure.py
```diff
@@ -32,12 +32,12 @@ def forwards(self, orm):
         try:
             ph_ctype = ContentType.objects.get(app_label=ph_model._meta.app_label, model=ph_model._meta.model_name)
             page_ctype = ContentType.objects.get(app_label=page_model._meta.app_label, model=page_model._meta.model_name)
-            permission, _ = Permission.objects.get_or_create(
+            permission, __ = Permission.objects.get_or_create(
                 codename='use_structure', content_type=ph_ctype, name=u"Can use Structure mode")
-            page_permission = Permission.objects.get(codename='change_page', content_type=page_ctype)
+            page_permission, __ = Permission.objects.get(codename='change_page', content_type=page_ctype)
```
DylannCordel: should be `get_or_create` or remove the `__` var

yakky: Done, thanks

DylannCordel:
Hi @yakky 

If we use a custom permission backend, it could not work because permissions won't be tested "as usual". We should add a Warning for that cases. Or we could use the `django.contrib.auth.models._user_has_perm` function (but it uses real Models and not the old ones).

Instead this special case and my previous comment on `get_or_create`, LGTM according to the ["custom user requirements"](https://django-cms.readthedocs.org/en/develop/reference/configuration.html?highlight=abstractuser#custom-user-requirements) section of django CMS doc..

vstoykov:
I tried this pull request without any custom backends or custom user model and forwards migrations from empty DB are ok. The only problem are backward migrations of 0010 and 0005 but I think that this is not so big problem.

Error for backward migrating 0010 is:

    ValueError: Cannot query "cms | placeholder | Can use Structure mode": Must be "Permission" instance.

And error for backward migrating 0005 is:

    django.db.migrations.migration.IrreversibleError: Operation <RunPython <function move_to_mp at 0x7f0bb558a7b8>> in cms.0005_auto_20140924_1039 is not reversible

vstoykov:
Probably is a good idea for new installations to use squashed migration, because these datamigrations are unnecessary for new installations and will make maitanance for future release harder. https://docs.djangoproject.com/en/1.8/topics/migrations/#squashing-migrations

yakky:
@DylannCordel @vstoykov do you agree to just catch all the exception in forward migrations and emit the warning?
@vstoykov  I updated the PR fixing the first error in backward migrating 0010.
The second one is a standard error when a bacward migration is not possible, identical to the similar message that exists in south.
Squashing migrations can be a nice option for 3.2, thanks for the suggestion

yakky:
The 'real one' is #4472 which is against the correct branch. I'll leave this open to keep the discussion intact

DylannCordel:
@yakky good enough for a migration. LGTM.

vstoykov:
My opinion for catching all errors is 50/50. For empty database there are no sense to break. Also there are no sense of warning at all.
Probably the situation where error can be raised is when custom user model is used which can be checked easily and warning to be printed in this situation.
One more thing. If forward migration raises an error for some reason I thing that there is a big chance in backward migration also error to be raised.

yakky:
@vstoykov the meaning of migrations it's just an heads up for the developer. Everything will still work despite any migration failing, but a check of the state of "use structure" permission is required.

vstoykov:
ok then there are no real reason for migrations to break if this particular migration is failing for some reason and warning can be printed

yakky:
@vstoykov @DylannCordel So, the migration now catches any exception and returns a warning allowing migration in any case

yakky:
@DylannCordel @vstoykov merging above PR, I'll be glad if you could give it a last round

vstoykov:
For me It's ok. If @DylannCordel is ok you can merge it.

cms/migrations/0010_migrate_use_structure.py
```diff
@@ -14,31 +15,33 @@ def forwards(apps, schema_editor):
     try:
         ph_ctype = ContentType.objects.get_for_model(ph_model)
         page_ctype = ContentType.objects.get_for_model(page_model)
-        permission, _ = Permission.objects.get_or_create(
+        permission, __ = Permission.objects.get_or_create(
             codename='use_structure', content_type=ph_ctype, name=u"Can use Structure mode")
```
DylannCordel: As `__future__.unicode_literals` is imported, shouldn't `u'strings'` be replaced by `'strings'` ?
Else, LGTM to be merged.

yakky: Removed, it slipped in, but it's not consisten with the rest of the codebase

PR #4427, opened 2015-09-08
django-cms 2.3.4 fixed googlemaps plugin

justine123:
In db.add_column you can't have default='...%', beacuse there are problems with migration. You should have default='...px' instead.


yakky:
2.3 is no longer supported. I can merge this (after the conflicts have been solved) but there is no plan for further 2.3 release

PR #4381, opened 2015-08-25
Sane add plugin

mkoistinen:
For posterity, this is the latest of a series of PRs for this feature: #3804 and #3516. All credit for this work goes to @ojii! Thank you!

PR #4429, opened 2015-09-08
Fixes issue with new plugin cancellation

FinalAngel:
- Create new ``Tex`` Plugin
- Go to ``History > View History``
- Click on ``cancel``
- Click on the (x) icon

**The Problem**

- History opens, it shouldn't until we cancel or save the plugin
- When pressing cancel, it creates an empty plugin
- When pressing the (x) icon, message appears (correct behaviour but wrong context)

**This PR fixes**

The window will stay frozen now when a plugin is created and caches the state. No new content can be loaded until the user either presses cancel or saves the plugin. This does not solve the reload issue where an empty plugin is still created. A better solution will be provided through https://github.com/divio/django-cms/pull/4381


cms/static/cms/js/modules/cms.modal.js
```diff
@@ -45,7 +45,7 @@
                 this.ui.modal.data('ready', true);
             },
 
-            _setupUI: function setupUI() {
+            _setupUI: function () {
```
vxsx: tsk tsk

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -666,6 +672,32 @@
...
+            },
+
+            /**
+             * _deletePlugin removes a plugin once created when clicking
+             * on delete or the close item. If we don't do this, an empty
+             * placeholder is generated
+             * https://github.com/divio/django-cms/pull/4381 will eventually
+             * provide a better solution
+             *
+             * @param hideAfter Boolean
```
vxsx: not to be picky, but can we avoid the so called "[boolean trap](http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html)"? 
would be so much easier with `_deletePlugin({ hideAfter: true })`

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -621,7 +627,7 @@
                         // case when there is no prefix
                         if (name === undefined && that.ui.titlePrefix.text() === '') {
                             var bc = iframe.contents().find('.breadcrumbs').contents();
-                            that.ui.titlePrefix.text(bc.eq(bc.length - 1).text().replace(' › ', ''));
+                            that.ui.titlePrefix.text(bc.eq(bc.length - 1).text().replace('›', '').trim());
```
vxsx: like

cms/static/cms/js/modules/cms.modal.js
FinalAngel: I think it might be a bit overkill to pass an object just yet, as there is currently only one exception. If we add more than it would make sense?

cms/static/cms/js/modules/cms.modal.js
FinalAngel: :D

cms/static/cms/js/modules/cms.modal.js
vxsx: if we want more that would be then an api change which would require you to go over every call and change it (which in case of a public api is then a matter of a major version (sic!)), which is more hassle than adding another param to the list and then it becomes a mess like `_deletePlugin(true, null, false, null, true)`. sure though it sounds extremely unlikely for this case, but for it's more a matter of principle. don't think we care that much for creating object vs variable memory/speed tradeoff

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -666,6 +672,33 @@
...
+                var data = CMS._newPlugin;
+                var post = '{ "csrfmiddlewaretoken": "' + this.config.csrf + '" }';
+                var text = this.config.lang.confirmEmpty.replace(
+                    '{1}', CMS._newPlugin.breadcrumb[0].title
+                );
+
+                // trigger an ajax request
+                return CMS.API.Toolbar.openAjax(data['delete'], post, text, function () {
+                    CMS._newPlugin = false;
+                    if (opts.hideAfter) {
```
vxsx: opts && opts.hideAfter

cms/static/cms/js/modules/cms.modal.js
FinalAngel: will be re-added on Friday on a "cleanup" PR

vxsx:
LGTM

PR #4433, opened 2015-09-09
django-cms 2.3.4 fixed googlemaps plugin

justine123:
This is request for the correct version of django-cms. 

In db.add_column you can't have default='...%', beacuse there are problems with migration. You should have default='...px' instead.


FinalAngel:
LGTM

evildmp:
@justine123 thanks for this. We can merge it into 2.3.4, but it won't ever make it into a new 2.3 release. 

Just to be clear though, are you saying that ``default='100%'`` causes a problem with migrations on a ``CharField``? Or that it's not correct for rendering a Google Map? (I can't see why it would cause an issue with a migration.)

FinalAngel:
LGTM, google maps needs a fixed height

yakky:
@FinalAngel merging it myself

PR #4437, opened 2015-09-10
Separate Basic/Advanced settings

FinalAngel:
This pull request creates a system where buttons can be grouped and than are placed to the left of the modals button section. This example can be triggered through ``Page > Page Settings``.

image


vxsx:
LGTM

PR #4438, opened 2015-09-10
Fix RemovedInDjango19Warning: django.utils.importlib

vstoykov:
Import `django.utils.importlib` only where it is needed (for Django 1.6 only). This will remove the warning:

    RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.

### DIFF
--- a/cms/utils/compat/forms.py
+++ b/cms/utils/compat/forms.py
@@ -1,6 +1,11 @@
 # -*- coding: utf-8 -*-
+try:
+    import importlib
+except ImportError:
+    # Python 2.6
+    from django.utils import importlib
+
 from django.conf import settings
-from django.utils import importlib
 from django.db import models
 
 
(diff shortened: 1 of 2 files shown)


cms/utils/compat/forms.py
```diff
@@ -1,6 +1,11 @@
 # -*- coding: utf-8 -*-
```
bittner: This line is probably unnecessary in this very source file, and could be removed.

vstoykov: If you talking about `# -*- coding: utf-8 -*-` it is there already. It's not introduced with my PR.

With git blame you can see that the commit introducing it is this https://github.com/divio/django-cms/commit/3ee64f9f470101ac7b336feb19ccef5f31a10c96

bittner: It should go away, because there's no need for it. No UTF-8 strings in the source code. (I know it's not part of the PR, but I like to clean up too when I touch a source file. It's just a comment, anyway.)

bittner:
PR #4413 also mentions [models.get_app](https://github.com/divio/django-cms/pull/4413/files#diff-0a872c8bde70e290b239f717c455b955L11) as spitting out a deprecation warning (I didn't see this ever myself, though). If you can reproduce it you may add it to your changes too. Makes your PR more complete.

yakky:
I'm merging this through #4446

PR #4441, opened 2015-09-10
Cleanup Sideframe Code

FinalAngel:
- Move all elements to ``this.ui`` (search for .find() and replace)
- add YUIdocs to general outline
- set the name for anonymous functions: ``_events: function () {`` becomes ``_events: function _events() {``
- set event handler variables like ``this.pointerup = 'pointerup.cms.sideframe pointercancel.cms';``
- replace ``bind`` with ``on`` and ``unbind`` with ``off``
- ensure for correct variable name positioning
- remove as many timeouts as possible


cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -1,14 +1,29 @@
-//##############################################################################
-// #SIDEFRAME#
-/* global CMS */
-
+/*
```
FinalAngel: I'm not sure about the general header here, maybe some feedback @evildmp @mkoistinen @yakky

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -99,14 +138,38 @@
...
+
+                // show iframe
+                this._show(width, animate);
+            },
+
+            /**
+             * handles content replacement mechanisms
+             *
+             * @method _content
+             * @module open
```
vxsx: that's not what `@module` is used for

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -143,51 +207,19 @@
...
 
-            // private methods
-            _show: function (width, animate) {
-                this.sideframe.show();
+            /**
+             * animation helper for opening the sideframe
+             *
+             * @method _show
+             * @module open
+             * @private
```
vxsx: missing params decl

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -23,73 +38,97 @@
...
-            // public methods
-            open: function (url, animate) {
-                // prepare iframe
-                var that = this;
+            /**
+             * opens a given url within a sideframe
+             *
+             * @method open
+             * @param url {String} URL string
+             * @param animate {Number} Animation speed in ms
```
vxsx: animate is optional, so it should be `@param [animate] {Number} ...`

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -196,58 +228,96 @@
...
-                        }, 0);
+                        });
                     }
                 }
 
-                // lock toolbar, set timeout to make sure CMS.API is ready
-                setTimeout(function () {
+                // trigger API handlers
+                if (CMS.API && CMS.API.Toolbar) {
+                    // FIXME: initialization needs to be done after our libs are loaded
```
vxsx: how does it work now?

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -312,7 +388,7 @@
 
                 // merge manually because jquery...
                 $.each(arr, function (index, item) {
-                    var i = $.inArray(item.param, keys);
+                    i = $.inArray(item.param, keys);
```
vxsx: why did you move declaration out of here?

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -196,58 +228,96 @@
...
-            _startResize: function () {
+            /**
+             * initiates the start resize event from _events
+             *
+             * @method _startResize
+             * @private
+             */
+            _startResize: function _startResize() {
                 var that = this;
                 var outerOffset = 30;
```
vxsx: would be cool to also get rid of magic numbers like this

cms/static/cms/js/modules/cms.sideframe.js
vxsx: i remember leaving some of the timeouts because there are some race conditions between api/structureboard (esp old one) and sideframe

cms/static/cms/js/modules/cms.sideframe.js
FinalAngel: how can I refer than to the parent "open"?

cms/static/cms/js/modules/cms.sideframe.js
FinalAngel: yep, the goal is to eliminate these race conditions using proper initialisation. The UI is not affected currently like that.

cms/static/cms/js/modules/cms.sideframe.js
FinalAngel: to have it in sync with all the other vars in that module

cms/static/cms/js/modules/cms.sideframe.js
FinalAngel: yes but I'd rather not touch that part of the code TBH

cms/static/cms/js/modules/cms.sideframe.js
vxsx: you simply don't. no need to tightly couple one method to another

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -196,58 +231,96 @@
...
+             * animation helper for closing the iframe
+             *
+             * @method _hide
+             * @module close
+             * @param opts
+             * @param opts.duration {Number} animation duration
+             * @private
+             */
+            _hide: function _hide(opts) {
+                var duration = opts.duration || this.options.sideframeDuration;
```
vxsx: i think we can be a bit more defensive and check for opts existence here

FinalAngel: fixed

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -1,14 +1,29 @@
...
-         * Sideframe
-         * Controls a cms specific sideframe
+        /**
+         * The sideframe is triggered via API calls from the backend either
+         * through the toolbar navigation or from plugins. The APIs only allow to
+         * open a url within the sideframe
+         *
+         * @class Sideframe
+         * @namespace CMS
+         * @requires CMS.API.Helpers
```
vxsx: i think `@uses` is more appropriate
http://yui.github.io/yuidoc/syntax/index.html#uses

vxsx:
LGTM, great job

Issue #3404, opened 2014-08-22
Url parameters are converted to entities in the modal iframe 


yakky:
If url variable here https://github.com/divio/django-cms/blob/develop/cms/static/cms/js/modules/cms.modal.js#L499 containes url parameters that _look like_ html entities they are converted to the corresponding entity.
This breaks some toolbar items, for example **Save as Page Type** does not allow saving pages as the original url parameters `?language=en&copy_target=5` are converted to `?language=en©_target_5` (**&copy** is converted to **©**)
This happens on latest firefox / chrome.
Quick fix is converting `&` to `&amp;`, but I'm up to wiser suggestions.

PR #4442, opened 2015-09-11
Remove custom modal and sideframe css

FinalAngel:
We now add the generic ``.cms-admin`` class to the ``<body>`` whenever we load admin resources into an iframe. If its loaded within the sideframe ``.cms-admin-sideframe`` is additional appended and for the modal ``.cms-admin-modal``. This speeds up loading and simplifies the use of other 3rd party themes.

**The old class for modal ``.cms-modal-window`` will be deprecated.**


lory87:
:+1:

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -547,20 +547,12 @@
             },
 
             /**
-             * _prepareUrl adds `?modal=1` get param to the url, which is then got by the backend
-             * and additional "modal" stylesheet is then inserted into a template that is loaded
-             * inside of an iframe
+             * start of en entity by jQuery. See #3404
```
vxsx: `Encodes the ampersands in urls, see #3404`

vxsx:
LGTM

PR #4444, opened 2015-09-11
Aligned background color

FinalAngel:
Ensure the loader also displays the appropriate color in conjunction with the modal window and sideframe.


vxsx:
LGTM

PR #4447, opened 2015-09-13
remove dimmer from structureboard

vxsx:
we don't really need extra element for it


cms/static/cms/sass/components/_structureboard.scss
```diff
@@ -8,19 +8,8 @@
     z-index: z(structure, base);
     width: 100%;
     height: 100%;
+    background: $structure-bgcolor;
 
-    .cms-structure-dimmer {
```
FinalAngel: I also implemented a dimmer for the sideframe, eventually we might want to create a component out of it?

FinalAngel:
LGTM

cms/static/cms/sass/components/_structureboard.scss
vxsx: i'm removing it, what are you talking about? :)

FinalAngel: yeah I know, just my thoughts on making it general if we might need it somewhere else (currently only used within the sideframe after this removal)

vxsx:
NB: merging this, let's see if the build will break when this is merged into `preview`

PR #4450, opened 2015-09-14
docs: i18n topic, Fixed typo on title.

rscnt:
Just a typo.


yakky:
@evildmp please review

evildmp:
Thanks @rscnt, but we use British English spellings, so ``internationalisation`` is correct.

Please see http://docs.django-cms.org/en/develop/contributing/documentation.html#contributing-documentation.

PR #4463, opened 2015-09-18
Cleanup toolbar and improve speed

FinalAngel:
This went already bigger than expected as the toolbar is intertwined into a lot of other components features, sorry about that :/

- Move all elements to ``this.ui`` (search for .find() and replace)
- add YUIdocs to general outline
- set the name for anonymous functions: ``_events: function () {`` becomes ``_events: function _events() {``
- set event handler variables like ``this.pointerup = 'pointerup.cms.sideframe pointercancel.cms';``
- replace ``bind`` with ``on`` and ``unbind`` with ``off``
- ensure for correct variable name positioning
- remove as many timeouts as possible


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -1,14 +1,29 @@
...
     'use strict';
-    // CMS.$ will be passed for $
+
+    // shorthand for jQuery(document).ready();
     $(document).ready(function () {
-        /*!
-         * Toolbar
-         * Handles all features related to the toolbar
+        /**
+         * The toolbar is the generic element which holds verious components
```
FinalAngel: I'll fix typo here

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -1,14 +1,29 @@
...
+var CMS = window.CMS || {};
+
+// #############################################################################
+// Toolbar
 (function ($) {
     'use strict';
-    // CMS.$ will be passed for $
+
+    // shorthand for jQuery(document).ready();
     $(document).ready(function () {
```
vxsx: but it aint no shorthand

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -17,106 +32,88 @@
...
+             * Stores all jQuery references within `this.ui`.
+             *
+             * @method _setupUI
+             * @private
+             */
+            _setupUI: function _setupUI() {
+                var container = $('#cms-toolbar');
+                this.ui = {
+                    container: container,
+                    body: $('html'),
```
vxsx: can we change the name to something else (similar to what's in modal)? body is misleading here

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -171,11 +168,11 @@
 
                         isTouchingTopLevelMenu = false;
                         // create the document event
-                        $(document).bind(that.click, reset);
+                        $(document).on(that.click, reset);
```
vxsx: should cache document instance somewhere outside or in `this.ui`

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -171,11 +168,11 @@
...
                     });
 
                     // attach hover
-                    lists.find('li').on('pointerover pointerout', function () {
+                    lists.find('li').on(that.poimterOverOut, function () {
```
vxsx: typo

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -218,39 +215,39 @@
...
-                        btn.on('click', function (e) {
+                        btn.on(that.click, function (e) {
                             e.preventDefault();
                             that._delegate($(this).find('a'));
                         });
                     }
 
                     // in case of the publish button
-                    btn.find('.cms-publish-page').bind(that.click, function (e) {
+                    btn.find('.cms-publish-page').on(that.click, function (e) {
```
vxsx: cache them as well?

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -270,34 +267,221 @@
...
-                if (show) {
-                    this.settings.toolbar = 'collapsed';
+            /**
+             * We check for various states on load if elements in the toolbar
+             * should appear or trigger other components. This preceeds a timeout
+             * which is not optimal and should be addressed separately.
+             *
+             * @method _initialStates
+             * @private
+             * @deprecated this method will be deprecated in > 3.2
```
vxsx: this method is deprecated _now_, it will be _removed_ in > 3.2

cms/static/cms/js/modules/cms.toolbar.js
FinalAngel: same as in modal and sideframe

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -312,153 +496,179 @@
...
-                    },
-                    'error': function (jqXHR) {
-                        CMS.API.locked = false;
-                        that.showError(jqXHR.response + ' | ' + jqXHR.status + ' ' + jqXHR.statusText);
+                this.showLoader();
+
+                return $.ajax({
+                    type: 'POST',
+                    url: url,
+                    data: (post) ? JSON.parse(post) : {}
```
vxsx: why do we have this? i though jquery does the parsing for you. so it could just be `data: post`

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -312,153 +496,179 @@
...
+                    default:
+                        window.location.href = el.attr('href');
                 }
             },
 
-            _setSwitcher: function (el) {
+            /**
+             * Sets the functionality for the switcher button.
+             *
+             * @method _hide
```
vxsx: wrong method

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -502,91 +715,76 @@
...
                 var interval = 20;
-                var blocker = this.screenBlock;
+                var blocker = this.ui.screenBlock;
                 var sideframe = $('.cms-sideframe');
 
                 // automatically resize screenblock window according to given attributes
-                $(window).on('resize.cms.screenblock', function () {
-                    var width = $(this).width() - sideframe.width();
-
+                $(window).on(this.resize, function () {
```
vxsx: we should make it position fixed fullscreen instead of this

vxsx: i think i actually did that in one of my cleanups o____O

cms/static/cms/js/modules/cms.toolbar.js
FinalAngel: Yep I agree

cms/static/cms/js/modules/cms.toolbar.js
FinalAngel: what you mean?

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -17,106 +32,89 @@
...
-                }
+                // FIXME the general initialization is handled within the toolbar
+                // rather than a separate cms.setup or similar. Yet other components
+                // are loaded after the toolbar so it can create a clash where
+                // CMS.API is not ready. This is a workaround until a proper fix
+                // will be released in 3.x
+                var that = this;
+                setTimeout(function () {
+                    that._initialStates();
+                }, 200);
```
vxsx: btw just wanted to check - is `setTimeout(fn, 0)` sufficient here?

cms/static/cms/js/modules/cms.toolbar.js
vxsx: well since screen block is a position fixed element with top 0, left 0 and width/height 100% it doesn't make any sense to listen to window resize, especially since we changed the way sideframe works.

or at the very least cache `$(window)` reference

vxsx:
overall awesome :+1:

PR #4464, opened 2015-09-19
Fix apphooks clear

yakky:
Maybe :)

### DIFF
--- a/cms/test_utils/testcases.py
+++ b/cms/test_utils/testcases.py
@@ -90,7 +90,6 @@ class BaseCMSTestCase(object):
         self.create_fixtures()
         activate("en")
 
-
     def create_fixtures(self):
         pass
 
@@ -396,6 +395,13 @@ class BaseCMSTestCase(object):
             from django.template import Template
             template_obj = Template(template)
             return template_obj.render(RequestContext(request, context))
+
+    def apphook_clear(self):
+        from cms.apphook_pool import apphook_pool
+        for name, label in list(apphook_pool.get_apphooks()):
+            if apphook_pool.apps[name].__class__.__module__ in sys.modules:
+                del sys.modules[apphook_pool.apps[name].__class__.__module__]
+        apphook_pool.clear()
 
 
 class CMSTestCase(BaseCMSTestCase, testcases.TestCase):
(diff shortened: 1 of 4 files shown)


yakky:
@FinalAngel @vxsx  please check if this really fix issues with preview branch

vxsx:
yes please!

vxsx:
looks like it solved the problem

yakky:
@mkoistinen Maybe this part https://github.com/divio/django-cms/pull/4464/files#diff-2a8a0d85dd243bf778b381658688d623R404 can be integrated in apphook_pool.clear() (even if I think only tests will really clear the the apphooks pool)

FinalAngel:
@yakky can we merge to preview (as we would like to create a PR against develop)

yakky:
@FinalAngel waiting for @mkoistinen feedback regarding https://github.com/divio/django-cms/pull/4464#issuecomment-141775102 then I'll merge

mkoistinen:
@yakky I'm afraid I don't understand the comment. =/

yakky:
@mkoistinen I added ``apphook_clear`` method to base test class that deletes ``cms_apps`` modules from ``sys.modules`` before clearing the apphooks list, so that a new discover will still work
This may be a behavior we may want to integrate into ``apphook_pool.clear`` method (https://github.com/divio/django-cms/blob/develop/cms/apphook_pool.py#L19)

mkoistinen:
@yakky, trying it now...

mkoistinen:
OK, I've tried this and it doesn't appear to do any harm. However, it does feel like the "thermonuclear" option. It seems to me we should be fixing register() to work with already-registered apphooks. Having a look at the code, I don't see why it doesn't. What am I missing? Regardless, here's the PR: https://github.com/divio/django-cms/pull/4479

yakky:
@mkoistinen ``discover`` relies on modules loading, which apparently does not work in the same way after jonas changes. The behavior should really only matters in tests, so I'm still undecided whether to incorporate the nuclear option in the ``apphooks_pool``

mkoistinen:
Understood. I'm OK to commit this PR, and disregard #4479. Up to you. I think it is safest to pull this one.

yakky:
@mkoistinen ok, let's play safe for now

PR #4465, opened 2015-09-19
Fix CMS_TOOLBAR_SIMPLE_STRUCTURE_MODE usage

yakky:
The current usage is not consistent with the rest of the code


yakky:
Please @FinalAngel review

FinalAngel:
@yakky LGTM - just to double check, this means the new structure board is the default for 3.2 or? (which we actually want)

Issue #4143, opened 2015-06-02
copy_relations() does not work for relations between cmsplugins


philippze:
In the docs, there's a recipe for [handling relations](http://docs.django-cms.org/en/develop/how_to/custom_plugins.html#handling-relations) in custom plugins.

However, this will not work when we deal with relations from one cmsplugin to another cmsplugin.

I've written four unittests [here](https://github.com/philippze/django-cms/commit/658e67f41f9029443ed3fe36d268ef09ef5e2644). Two pass and show that the docs is correct in usual cases. Two fail and show that it's not so simple for relations between cmsplugins.

Maybe one could mention this in the docs.

I'm now writing a little module that solves the problem for ManyToManyFields. Existing solutions welcome!

yakky:
Thanks for the awesome tests and a very well described issue.
The problem with ``copy_relations`` between plugins is that while the ``copy_relations`` in https://github.com/philippze/django-cms/commit/658e67f41f9029443ed3fe36d268ef09ef5e2644#diff-f813663252f41cc6f43485a6b06e7caeR58 will work when copying the ``PluginModelWithM2MToModel``, there is no (general) way for the target plugin to change the reference from the "draft" to the "live" version of itself in the referencing plugins.
I think we can just document this.

evildmp:
@yakky @philippze Should we not also incorporate at least the passing tests and keep this open until a more general solution is available?

yakky:
@philippze I totally second @evildmp proposal. Could you open a PR?

sergei-maertens:
Another use this case that fits this issue is re-using plugin content.

We have a specific use case where the end-user adds a 'Story' plugin to a certain page. It has some images, a bit of text, and nested plugins for the body of the content.

On the homepage, they need to be able to refer to this very same story, and if the initial story page (and plugin) gets updated, the homepage version has to be synced as well.

Currently this is solved by letting the `HomepageStoryPlugin` link (via FK) to the draft version of the `StoryPlugin`, but that obviously has its drawbacks.

mkoistinen:
@philippze someone else had a similar need and I put together a sample plugin for them. Perhaps this will help you too? https://gist.github.com/mkoistinen/403a81c05b4b8f783c74 and a video demo of this plugin in action: http://cl.ly/332q3R42202X/draft-public-plugin-demo.m4v

philippze:
[...]

PR #4466, opened 2015-09-19
Tests for copy_relations, see #4143

philippze:
Here's the pull request for the passing tests on copy_relations as suggested in Issue #4143 - for review...

I've set up three environments with Django 1.6, 1.7, and 1.8. In each one the two new tests pass.

Somehow, Django's built in makemigrations does not create the correct files, while the South schemamigration command works. But ... the tests work without the migration files, too...

### DIFF
--- a/cms/test_utils/project/pluginapp/plugins/manytomany_rel/cms_plugins.py
+++ b/cms/test_utils/project/pluginapp/plugins/manytomany_rel/cms_plugins.py
@@ -7,10 +7,6 @@ from cms.test_utils.project.pluginapp.plugins.manytomany_rel.models import (
     ArticlePluginModel, Article,
     PluginModelWithFKFromModel,
     PluginModelWithM2MToModel,
-    FKPluginModel,
-    M2MTargetPluginModel,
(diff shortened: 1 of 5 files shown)


mkoistinen:
Any idea why tests are failing?

philippze:
@mkoistinen Do you mean why one commit is called "Tests for copy_relations. Two fail."?  
copy_relations does not work for relations between plugins. This was dicussed in issue #4143 .

If you ask why the Travis build fails ... I could have a closer look at it if you want. But when I checked out my branch, the build for the development branch failed as well.

yakky:
@philippze I think here https://github.com/divio/django-cms/pull/4466/files#diff-1a778a76252565e5890a771389dff1f9R12 you should use ``get_user_model`` function instead of importing ``User`` model. Also: check if migrations involves ``User`` and eventually change like https://github.com/divio/django-cms/blob/develop/cms/south_migrations/0041_auto__add_usersettings.py

yakky:
@philippze thanks for this contribution!

PR #4458, opened 2015-09-17
Bugfix spaceless

vxsx:
https://github.com/divio/django-cms/issues/4337

additionally changed html comment to be django comment since it's not necessary to render it

PR #4467, opened 2015-09-21
bugfix spaceless  3.0.x

vxsx:
https://github.com/divio/django-cms/pull/4458

### DIFF
--- a/cms/tests/toolbar.py
+++ b/cms/tests/toolbar.py
@@ -1036,7 +1036,7 @@ class EditModelTemplateTagTest(ToolbarTestBase):
                 'placeholderapp', 'example1', ex1.pk))
         self.assertContains(response, '<h1>%s - %s</h1>' % (ex1.char_1, ex1.char_2))
         self.assertContains(response, '<span class="date">%s</span>' % (ex1.date_field.strftime("%Y")))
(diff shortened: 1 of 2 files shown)


yakky:
@vxsx thanks, will forward port later

cms/tests/toolbar.py
```diff
@@ -1467,7 +1467,7 @@ def test_edit_page(self):
                 page.pk, page.get_title(language)))
         self.assertContains(
             response,
-            '<div class="cms_plugin cms_plugin-cms-page-changelist-%s cms_render_model cms_render_model_block"><h3>Menu</h3></div>' % page.pk)
+            '<div class="cms_plugin cms_plugin-cms-page-changelist-%s cms_render_model cms_render_model_block">\n        <h3>Menu</h3>' % page.pk)
         self.assertContains(
```
foobacca: On this line, is the deletion of the closing `</div>` at the end of the line deliberate?

FinalAngel: probably as there will be a menu rendered afterwards, @yakky can you confirm?
@foobacca did you encounter any issues?

vxsx: the change was deliberate not to clutter the  assert string with `\n    \n  {not sure about amount of space symbols}      \n</div>`  
it would probably make sense to not compare html as strings, but rather as a tree of dom nodes in future

PR #4470, opened 2015-09-21
adjust styles

vxsx:
screen shot 2015-09-21 at 10 43 45

now hover / focus are different, and made plugin items overflow just in case


FinalAngel:
LGTM

PR #4471, opened 2015-09-21
Set the new structure board as default

FinalAngel:
The new look & feel works best with the new structure board alignment. We recommend using this version but allow legacy users to switch back to the old behaviour.

This PR adapts the docs from https://github.com/divio/django-cms/pull/4465


vxsx:
:+1:

docs/reference/configuration.rst
```diff
@@ -955,16 +955,16 @@ CMS_TOOLBAR_SIMPLE_STRUCTURE_MODE
...
-    ``False``
+    ``True``
 
 Enable or disable the absolute positioning within the structure board.
 
-By default the absolute positioned board is activated. Setting this attribute
-to ``False`` will prevent the absolute positioning and order the placeholders
-beneath each other. Static placeholders have a visual gap in the simple mode.
+By default the absolute positioned board is deactivated. Setting this attribute
+to ``False`` will allow the absolute positioning used in versions prior to 3.2.
```
mkoistinen: This isn't very clear at all. Consider:

"By default, starting in django CMS 3.2, the structure board will now ignore absolutely positioned elements. To use a structure board which respects absolutely positioned elements, set this flag to False. NOTE: This option will be removed in version 3.3."

FinalAngel: @mkoistinen I'll adapt your suggestions within my next PR!

PR #4472, opened 2015-09-21
Fix use structure migration

yakky:
Fix #4419


fabiocaccamo:
Could you release 3.1.4 ?

PR #4481, opened 2015-09-22
fix clipboard tests

vxsx:
on local this one still fails for me test_cms_modal_html5_validation_error but i have no idea how to fix it

### DIFF
--- a/cms/tests/test_frontend.py
+++ b/cms/tests/test_frontend.py
@@ -342,7 +342,7 @@ class PlaceholderBasicTests(FastLogin, CMSLiveTests):
         build_button = self.driver.find_element_by_css_selector('.cms-toolbar-item-cms-mode-switcher a[href="?%s"]' % get_cms_setting('CMS_TOOLBAR_URL__BUILD'))
         build_button.click()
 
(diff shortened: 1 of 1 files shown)


cms/tests/test_frontend.py
```diff
@@ -384,17 +384,15 @@ def test_copy_to_from_clipboard(self):
...
         WebDriverWait(self.driver, 10).until(lambda driver: clipboard.is_displayed())
 
-        hov = ActionChains(self.driver).move_to_element(clipboard)
-        hov.perform()
+        clipboard.find_element_by_css_selector('.cms-clipboard-numbers a:first-child').click()
 
         # necessary sleeps for making a "real" drag and drop, that works with the clipboard
-
-        time.sleep(0.1)
+        time.sleep(0.3)
```
vxsx: here it waits for modal to open
i'm pretty sure there are better ways, but my selenium fu is lacking

mkoistinen:
Looks ok to me, you've already pointed out the area that needs a better solution, but I also don't know what it is in the context of drag-and-drop.

PR #4487, opened 2015-09-23
Fix html5 validation test

vxsx:
> This is an attempt to reverse https://github.com/divio/django-cms/pull/4190/
I don't know what to do exactly with migrations introduced there so please advise
@mkoistinen @yakky 

> So the whole point of #4190 was that if there is an input field with html5 type like "number" then there is some browser validation. However in django 1.7 there was this commit https://github.com/django/django/commit/307eef20e35e78b1e812dc347c6c959e380267cf
which introduced `novalidate` attribute on all the forms in the admin, which tells browsers to not validate these kind of inputs. 

> Related issue here → https://code.djangoproject.com/ticket/23075

> It is only possible to bump into this if you have modified `templates/admin/change_form.html` to not have the `novalidate` attribute. An option would be to create the test with custom change_form template but sadly I have absolutely no idea how to do this.

> Here I still reintroduce the changes made for modal, just in case.~~

disregard all that.

We are going to skip this test in Django>=1.7 as it's not really valid there.
Reintroduced modal changes that were lost somewhere on the way to 3.2, adapted the test case


FinalAngel:
I'm in favour of removing this test as django adds ``novalidate`` which is the intended behaviour. Maintaining this test would break django's default behaviour and we would have to include some monkey patching to the cms.

So for me :+1: on this PR

vxsx:
also this test case was only relevant for firefox, as for example chrome and safari do not allow typing anything but numbers into the number field

yakky:
If it's a real thing, I'd just do a git revert on support/3.0.x branch. Or is it an issue only in 3.2 ?

yakky:
@vxsx I'll take care of this and merging in the various branchs

yakky:
Merged via #4503

PR #4488, opened 2015-09-23
Use new frame for admin

mkoistinen:
Creates a new CMSFrameItem toolbar item. This opens urls in a new, named frame "cms_frame", which is generally going to be a new browser tab, depending on the user's browser and personal browser settings.

This has now been applied to the menu items: "Administration …" and "Users …".


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -649,6 +649,12 @@ var CMS = window.CMS || {};
                             message: el.data('text')
                         });
                         break;
+                    case 'cms_frame':
+                        var cms_frame = new CMS.CMSFrame();
+                        cms_frame.open({
```
FinalAngel: wouldn't it be just enough to use ``window.open(el.attr('href'), 'cms_frame')`` here to avoid an extra file (would be to much overkill IMO) like:

case 'cms_frame':
    frame = window.open(url, 'cms_frame');
    frame.focus();

vxsx:
uhm, what's wrong with `<a href="url" target="_blank">text</a>` ?

cms/templates/cms/toolbar/items/item_cms_frame.html
```diff
@@ -0,0 +1,5 @@
+<li class="{% if active %}cms-toolbar-item-navigation-active{% endif %}{% if disabled %} cms-toolbar-item-navigation-disabled{% endif %} {{ extra_classes|join:' ' }}">
+    <a href="{{ url }}" data-rel="cms_frame"{% if on_close %} data-on-close="{{ on_close }}"{% endif %}>
```
vxsx: likely we are not able to handle `on_close`

mkoistinen: Oh, yes, I forgot to remove this, thanks.

mkoistinen:
@FinalAngel @vxsx, Thanks for the review. I'm of the opinion to retain the encapsulation we have here. I don't think the extra file is much a big deal, and in the end, it'll be a few dozen symbols in the minified file. It provides a framework for extensibility this way, which is what I think the CMS is for. Besides, removing CMS.CMSFrame would then require a "magic string" constant of "cms_frame" to be stored and managed anyway.

Having said all of this, I think what I should change, though, is to make the frame name a parameter so that this can be used for any number of named tabs by cms developers going forward. Agree?

FinalAngel:
LGTM

vxsx:
but this cms_frame string is generally useless, since the title will be
overridden by the title of the page being opened

On Wed, Sep 23, 2015 at 10:28 PM, Martin Koistinen <notifications@github.com
> wrote:

> @FinalAngel <https://github.com/FinalAngel> @vxsx
> <https://github.com/vxsx>, Thanks for the review. I'm of the opinion to
> retain the encapsulation we have here. I don't think the extra file is much
> a big deal, and in the end, it'll be a few dozen symbols in the minified
> file. It provides a framework for extensibility this way, which is what I
> think the CMS is for. Besides, removing CMS.CMSFrame would then require a
> "magic string" constant of "cms_frame" to be stored and managed anyway.
>
> Having said all of this, I think what I should change, though, is to make
> the frame name a parameter so that this can be used for any number of named
> tabs by cms developers going forward. Agree?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/divio/django-cms/pull/4488#issuecomment-142719115>.
>



-- 
Vadim Sikora

Frontend Teamlead
Divio AG | Riedtlistrasse 23, CH-8006 Zürich
Tel.: +41 44 480 12 70 | Mobile: +41 79 104-43-50
http://divio.com | http://aldryn.com

Divio is a core contributor to the open source content management system
django CMS: http://www.django-cms.org

PR #4495, opened 2015-09-24
Closing a message unlocks toolbar

vxsx:
This PR fixes a bug where messages were autohidden when they shouldn't have been,
also removes the locking/unlocking the toolbar while showing messages, as it's not really required


FinalAngel:
LGTM

PR #4496, opened 2015-09-24
Url helper

vxsx:


cms/static/cms/js/modules/cms.base.js
```diff
@@ -261,8 +261,71 @@ var CMS = {
...
+            },
+
+            /**
+             * Modifies the url with new params and sanitises the ampersand within the url for #3404.
+             *
+             * @method makeURL
+             * @param url {String} original url
+             * @param [params] {String[]} array of `param=value` strings to update the url
+             */
+            makeURL: function makeURL(url, params) {
```
FinalAngel: why not setURL or cleanURL?

FinalAngel:
LGTM

cms/static/cms/js/modules/cms.base.js
vxsx: ~~because naming is hard~~
cleanURL doesn't fit, because you can add parameters
setURL i just personally don't like :D

PR #4498, opened 2015-09-24
Prepare draft release notes for 3.0.16

yakky:


yakky:
Merging now as it's just a doc change and we don't need the full matrix to validate it

Issue #4344, opened 2015-08-07
`on_close` parameter on CMSToolbar.add_modal_item doesn't do anything


mkoistinen:
If a URL is supplied, it should redirect to given URL on close.

yakky:
@mkoistinen have you tested on latest support? This should have been fixed by #4229

mkoistinen:
Can this be forward-ported to 3.1.x?

mkoistinen:
I've tested this with latest develop and the issue persists.

yakky:
@mkoistinen do we need this in 3.0.15 or we can postpone it?

mkoistinen:
I marked it blocker, but it doesn't have to be.

vxsx:
I think we fixed this in 3.2 preview.

vxsx:
Should've been fixed as part of https://github.com/divio/django-cms/pull/4473 and subsequent change in https://github.com/divio/django-cms/pull/4478

vxsx:
I was wrong

yakky:
@vxsx could you backport the fix to 3.0/3.1 ?

yakky:
@vxsx @FinalAngel any update on this?

vxsx:
@yakky sadly no, had been quite busy with 3.2, but this one is still on my radar

yakky:
@vxsx thanks

FinalAngel:
@qnub we fixed the on_close behaviour for modal and sideframe?

czpython:
@vxsx any updates on this? is it still relevant?

vxsx:
it is still relevant

vxsx:
Sorry, highly unlikely that it's going to happen for 3.0/3.1.

PR #4499, opened 2015-09-24
Bugfix on close not respected

vxsx:
based off https://github.com/divio/django-cms/pull/4487

should fix the https://github.com/divio/django-cms/issues/4344 for 3.2


FinalAngel:
LGTM

PR #4500, opened 2015-09-24
Rendering Issue

FinalAngel:
GitHub apparently doesn't take width and height attributes into account when rendering on their page


vxsx:
:+1:

PR #4454, opened 2015-09-15
Extend modal to inject html markup

FinalAngel:
**This PR extends the modals functionality to inject html markup.**

image

The API to open a modal has been adapted:

```javascript
// old
var modal = new CMS.Modal();
modal.open('http://www.google.ch', 'Modal name', ['breadcrumb items']);
// new
var modal = new CMS.Modal();
modal.open({
    url: 'http://www.google.ch',
    title: 'Modal name',
    breadcrumb: ['breadcrumb items']
});
```

the new features can be invoked using:

```javascript
var modal = new CMS.Modal();
modal.open({
    html: '<p>Hello World</p>',
    title: 'Modal name'
});
```

Please not that url has precedence over html. If both are provided url wins. Also note that the html version doesn't load a breadcrumb as of this version. All params can be inspected through the `comments <https://github.com/divio/django-cms/pull/4454/files#diff-1995b31ad4aa5ec90e5a1a3aeba9f113R144>`_.

The example above (screenshot) was generated using:

```javascript
var modal = new CMS.Modal();
modal.open({
    html: $('.js-modal-test'),
    title: 'Modal',
    subtitle: 'Adding HTML Markup',
    width: 940,
    height: 325
});
```

```html
<div class="js-modal-test" style="padding: 18px;">
    <ul class="test-css-blah">
        <li><a href="#1">Create new page</a></li>
        <li><a href="#2">Choose page from Template</a></li>
        <li><a href="#3">Create new template</a></li>
        <li><a href="#4">Cancel</a></li>
    </ul>
    <style>
        .cms-reset .test-css-blah { list-style-type: none; padding: 0; margin: 0; }
        .cms-reset .test-css-blah li { list-style-type: none; padding: 0; margin: 0; text-align: center; }
        .cms-reset .test-css-blah a { float: left; width: 46%; padding: 25px ; margin: 2%; border: 1px solid #ccc; border-radius: 3px; }
        .cms-reset .test-css-blah a:hover { background: #ddd; }
    </style>
</div>
```

**Cleanup Changes**

- Move all elements to ``this.ui`` (search for .find() and replace)
- add YUIdocs to general outline
- set the name for anonymous functions: ``_events: function () {`` becomes ``_events: function _events() {``
- set event handler variables like ``this.pointerup = 'pointerup.cms.sideframe pointercancel.cms';``
- replace ``bind`` with ``on`` and ``unbind`` with ``off``
- ensure for correct variable name positioning
- remove as many timeouts as possible

PR #4501, opened 2015-09-24
Style subtitle differently if no title provided

vxsx:
toolbar items should now use the title from the name attribute for the main title that was implemented in https://github.com/divio/django-cms/pull/4454

screen shot 2015-09-24 at 16 11 23


FinalAngel:
LGTM

vxsx:
ended up revering the thing with templates because there're several issues with changing content inside of the iframe

FinalAngel:
LGTM

PR #4504, opened 2015-09-25
Build mode url handling

vxsx:
[In the docs](http://docs.django-cms.org/en/latest/reference/api_references.html?highlight=build_mode#module-cms.toolbar.toolbar) we mention the `build_mode` property on the toolbar, but it is never used to determine in which mode we actually initialize it. This PR attempts to fix that.

New behaviour is - always respect GET param (which helps with sharing urls) and if the browser is fresh enough to support history api (http://caniuse.com/#feat=history support is good) the url will always be updated whenever you change mode.

I'm pretty sure that the behaviour of python code is incorrect, so please advise /cc @yakky @mkoistinen


cms/toolbar/toolbar.py
```diff
@@ -97,7 +98,7 @@ def init_toolbar(self, request):
         self.request = request
         self.is_staff = self.request.user.is_staff
         self.edit_mode = self.is_staff and self.request.session.get('cms_edit', False)
-        self.build_mode = self.is_staff and self.request.session.get('cms_build', False)
+        self.build_mode = self.is_staff and request.GET.get(get_cms_setting('CMS_TOOLBAR_URL__BUILD')) != None
```
vxsx: so here I don't really know what to do (or rather how to do it correctly). if i keep the old code - it works for the first time you visit the the page with `?build=1` but then it _never_ resets back to false.

yakky: What's the intended behavior?
If ``?build=1`` is in the URL -> switch to build mode and set it in the session
If ``?build=0`` is in the URL -> switch to build mode and set it in the session
If no ``?build`` is in the URL -> remember the session

FinalAngel:
FE LGTM

cms/toolbar/toolbar.py
vxsx: i was thinking intended behaviour is if `\?build=?.*` in the url - store `build_mode = True` into session. if no build is in the url - reset value to `False`

vxsx: build mode = structure mode

mkoistinen: @vxsx that sounds reasonable. What are the use-cases for this? Is this a documented feature?

vxsx: the usecase is sharing a link to another cms user (will go into correct mode). it is documented here http://docs.django-cms.org/en/latest/reference/api_references.html?highlight=build_mode#module-cms.toolbar.toolbar but it's not respected in any way at the moment

vxsx: however keep in mind that my adjustment doesn't store anything in the session because i don't know how to do that, it only achieves expected frontend behaviour

mkoistinen: My feeling is that we're not doing this right. Since changing between structure and preview mode does not require a page load, I don't think it should have a GET parameter to do, but rather a # parameter.

PR #4508, opened 2015-09-25
Content creation wizard

mkoistinen:


cms/wizards/wizard_pool.py
```diff
@@ -0,0 +1,111 @@
...
+    def register(self, entry):
+        """
+        Registers the provided «entry».
+
+        Raises AlreadyRegisteredException if the entry is already registered.
+
+        NOTE: This method triggers pool discovery.
+        """
+        assert isinstance(entry, Wizard), u"entry must be an instance of Wizard"
+        if self.is_registered(entry):
```
czpython: I don't think we should call is_registered() here because it could result in some sort of infinite loop.
when registering an item we shouldn't trigger the discovery

cms/wizards/wizard_pool.py
```diff
@@ -0,0 +1,111 @@
...
+            model = entry.get_model()
+            raise AlreadyRegisteredException(
+                _(u"A wizard has already been registered for model: %s") %
+                model.__name__)
+        else:
+            self._entries[entry.id] = entry
+
+    def unregister(self, entry):
```
czpython: same as above, please don't trigger discovery

cms/wizards/wizard_base.py
```diff
@@ -0,0 +1,90 @@
...
+    override as force_language,
+    force_text,
+    ugettext as _
+)
+
+
+class WizardBase(object):
+    template_name = None
+
+    def __init__(self, title, weight, form, model=None, template_name=None,
```
czpython: is ``weight`` really appropriate? why not just call it ``position``?

cms/wizards/wizard_base.py
```diff
@@ -0,0 +1,90 @@
...
+        # NOTE: If class attributes or properties are changed, consider updating
+        # cms.templatetags.cms_wizard_tags.WizardProperty too.
+        self.title = title
+        self.weight = weight
+        self.form = form
+        self.model = model
+        if description is not None:
+            self.description = description
+        else:
+            self.description = _("Create a new %s instance.") % model.__name__
```
czpython: use ``model._meta.verbose_name`` here please

cms/wizards/urls.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+
+from django.conf.urls import patterns, url
+
+from .views import WizardCreateView
+
+
+urlpatterns = patterns('',  # NOQA
+    url(r'^create/$',
+        WizardCreateView.as_view(), name='wizard_create'),
```
czpython: I highly recommend either using namespaced urls when including this or changing the url name to ``cms_wizard_create`` this is because ``wizard_create`` is too generic and can easily conflict with other apps

cms/cms_wizards.py
```diff
@@ -0,0 +1,23 @@
...
+    CreateCMSPageForm,
+)
+
+
+cms_page_wizard = Wizard(
+    title=_(u"New page"),
+    weight=100,
+    form=CreateCMSPageForm,
+    model=Page,
+    description="Start with a new blank page."
```
czpython: should this not be ``_("Start with a new blank page.")``?

cms/forms/wizards.py
```diff
@@ -0,0 +1,78 @@
...
+            if isinstance(title_obj, EmptyTitle):
+                create_title(language, title, page)
+
+    @staticmethod
+    def get_placeholder_slot(page):
+        """
+        Returns the slot name of the first editable, non-static placeholder
+        or None.
+        """
+        for ph in page.placeholders.all():
```
czpython: use ``page.get_placeholders()``

cms/forms/wizards.py
```diff
@@ -0,0 +1,78 @@
...
+        page = create_page(
+            title=title,
+            template=TEMPLATE_INHERITANCE_MAGIC,
+            language=self.language_code,
+            created_by=smart_text(self.user),
+            parent=self.page,
+            in_navigation=True,
+            published=False
+        )
+        self.create_page_titles(page, title, [self.language_code])
```
czpython: why try and create the page title if it already was created for self.language_code above when creating the page?

cms/forms/wizards.py
```diff
@@ -0,0 +1,78 @@
...
+
+        content = self.cleaned_data['content']
+        if content:
+            slot = self.get_placeholder_slot(page)
+            if slot:
+                placeholder = page.placeholders.get(slot=slot)
```
czpython: instead of doing double calls, why not rename get_placeholder_slot to get_first_placeholder and have it return a placeholder object?

cms/forms/wizards.py
```diff
@@ -0,0 +1,78 @@
...
+                add_plugin(
+                    placeholder=placeholder,
+                    plugin_type='TextPlugin',
```
czpython: this assumes TextPlugin is registered which might not always be the case.
We should check this when checking if content

cms/wizards/wizard_base.py
mkoistinen: Because it isn't a position. Its a sorting weight. I considered `sort_metric`, but didn't feel like it was worth changing from the source code I started with.

cms/wizards/wizard_base.py
mkoistinen: :+1:

cms/wizards/forms.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.get('wizard_user', None)
+        super(WizardStep1Form, self).__init__(*args, **kwargs)
+        # set the entries here to get an up to date list of entries.
+        self.fields['entry'].choices = entry_choices(user=self.user)
+
+    def clean(self):
+        page = self.cleaned_data.get('page')
+        space = self.cleaned_data.get('space')
```
czpython: there's no concept of spaces here afaik

cms/cms_wizards.py
mkoistinen: +1

cms/wizards/forms.py
```diff
@@ -0,0 +1,88 @@
...
+        invalid_request_message = _(u"We're unable to process your request.")
+
+        if page and space:
+            if page.application_urls:
+                # user is creating an object for an apphooked page.
+                raise forms.ValidationError(invalid_request_message)
+
+            else:
+                # user is creating an object for a page outside of space.
+                raise forms.ValidationError(invalid_request_message)
```
czpython: can be removed

cms/forms/wizards.py
mkoistinen: :+1:

cms/forms/wizards.py
mkoistinen: :+1:

cms/wizards/urls.py
mkoistinen: :+1:

cms/wizards/wizard_pool.py
mkoistinen: I'm not sure about any infinite loop possibilities, but I agree it might create unexpected results. To counter any of this, I've publicly exposed the property `discovered`. It can be used to first check if the pool is populated before using any of these other methods. If the pool hasn't been populated, then it is already nonsensical to use any of those methods. The outcome of each is already known.

However to put any further concerns to rest, I'll create a kwarg on is_registered() that will not trigger discovery.

cms/forms/wizards.py
```diff
@@ -0,0 +1,78 @@
...
+    def save(self, **kwargs):
+        from cms.api import create_page, add_plugin
+
+        title = self.cleaned_data['title']
+        page = create_page(
+            title=title,
+            template=TEMPLATE_INHERITANCE_MAGIC,
+            language=self.language_code,
+            created_by=smart_text(self.user),
+            parent=self.page,
```
czpython: there's a known issue with accepting the parent page id like this, basically anyone could modify this id to point to a page they don't have access to or just any page and be able to add a subpage to it

cms/wizards/wizard_pool.py
mkoistinen: See previous comment

cms/forms/wizards.py
mkoistinen: True, probably wouldn't hurt to be more defensive here. :+1:

cms/wizards/forms.py
mkoistinen: Oops! Thanks.

cms/wizards/forms.py
mkoistinen: Yep, thanks.

cms/forms/wizards.py
mkoistinen: What do you propose?

cms/forms/wizards.py
mkoistinen: :+1:

cms/static/cms/sass/components/_modal.scss
```diff
@@ -131,8 +131,10 @@
         text-overflow: ellipsis;
         margin-right: $modal-header-button-area-size * 3 + $padding-normal;
         .cms-modal-title-suffix {
+            display: block;
```
vxsx: this is not needed

cms/static/cms/sass/components/_modal.scss
```diff
@@ -131,8 +131,10 @@
...
             font-weight: normal;
             padding-left: $padding-normal;
+            height: $modal-header-height;
```
vxsx: and this

cms/cms_wizards.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def user_has_add_permission(self, user):
+        opts = Page._meta
+        site = Site.objects.get_current()
+        global_add_perm = GlobalPagePermission.objects.user_has_add_permission(
+            user, site).exists()
+        perm_str = opts.app_label + '.' + get_permission_codename('add', opts)
+        if user.is_superuser or (user.has_perm(perm_str) and global_add_perm):
+            return True
+        print("No permission: {0}, {1}".format(user.has_perm(perm_str), global_add_perm))
```
czpython: print?

czpython:
Ok to merge. We can look into the apphooks concern at a later point.

PR #4513, opened 2015-09-27
fix default params

vxsx:
minor fix to the way default values are determined (since 0 is a valid value, although it is falsy)


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -566,7 +566,7 @@ var CMS = window.CMS || {};
                 var that = this;
                 // url, post, text, callback, onSuccess
                 var url = opts.url;
-                var post = opts.post || {};
+                var post = opts.post || '{}';
```
FinalAngel: why a string?

vxsx: because later in this file we try to `JSON.parse` it

FinalAngel:
LGTM

PR #4514, opened 2015-09-27
Add plugin UX

vxsx:
This PR adds the ability to pick first plugin in the list by pressing enter in the quicksearch field.

Also adds generic debounce helper.


FinalAngel:
LGTM

PR #4515, opened 2015-09-27
Proposal: sharper icon font using fixed 16px font-size

SachaMPS:
Hi guys,
this is just a proposal top optimize the icons on non retina displays. How this approach works is crucial to consider if you want to merge that changes in.

- All icons are 16px tall in css
- Different icon sizes are made via the svg file NOT via font-size.
- The svg files are normalized by hand to contain only icons which use a 512px grid (Exception: logo.svg)
- This way the text rendering does not produce blurry lines as long as you aligned the svg paths properly
- The $...icon-size definitions in the scss are obsolete. Instead 16px is hardcoded in _iconfont.scss

The benefit are sharper icons in some cases. Just an example:
image

The obvious downside of this approach is that you must align the icons properly and adjust the to your needs. The second downside is that I recommend to use different icon files for different font sizes if an icon is required in different sizes. 

I can explain how to do that if you want. Actually it's quite simple... Use a 512px Artboard in Illustrator. Subdivide the artboard by 32px with 16px subdivisions. If you now want an 12px tall icon you just have to calculate: 32px * 12px = 384px and apply that dimensions to the icon. Adjust some points while "snap to grid" is activated and voila :)


yakky:
@FinalAngel @vxsx please review

FinalAngel:
Thank you very much for the PR. Sharpness on retina is quite important, I don't mind having a default of 16px for an icon, I guess that can be applied to almost all places. However it might sometimes be required to scale them down (change the font-site) as adding the same icon with different sizes breaks DRY.

We would have to see how the modal gets affected. Can this be applied to the rest of the structure mode @vxsx ?

SachaMPS:
@FinalAngel The sharpness actually visually comes more into effect on NON-Retina Displays due to less pixels per inch. It is still possible to scale them down or up via css font-size but subpixel antialiasing might kick in again. It's just a way of a i would say a clean start. For the modal icons i adjusted the icons to be 12px tall  if a font-size of 16px is used.

vxsx:
thanks for the PR! i'm ok with merging (and having consistent icon size for that matter), but scissors icon seems cut off on the sides

SachaMPS:
@vxsx can you gimme a hint where the scissor icon occurs in the ui? Will fix that.

vxsx:
also if possible could you create a PR against develop? preview is obsolete and will be removed shortly

SachaMPS:
@vxsx alright, no problem will make a new PR

vxsx:
@SachaMPS it's `cut` in the plugin menu. but it's visible on the diff 
screen shot 2015-09-28 at 10 56 16

vxsx:
thanks!

SachaMPS:
@vxsx Ragarding the scissor: If you mean "cut off" by comparing the diff images. That doesnt matter. Actually its for purpose. The 512px Artboard is only a reference. Icons CAN overlap the artboard. In case of the scissor i adjusted it to match the intended height. It can happen that icons overlap the artboard if they are more wide than high. Anyway will check if the icon appears properly in the UI.

vxsx:
Implemented in #4521

PR #4520, opened 2015-09-28
Clipboard changes

vxsx:
This PR moves clipboard trigger and clearing to the toolbar menu.


cms/cms_toolbars.py
```diff
@@ -141,6 +142,16 @@ def add_admin_menu(self):
             self._admin_menu.add_modal_item(_('User settings'), url=admin_reverse('cms_usersettings_change'))
             self._admin_menu.add_break(USER_SETTINGS_BREAK)
 
+            # clipboard
```
FinalAngel: @yakky @mkoistinen do we need to document this for users who have customized the toolbar?

FinalAngel:
FE LGTM

cms/cms_toolbars.py
mkoistinen: Documentation is always great. Have you changed anything that goes here: https://django-cms.readthedocs.org/en/latest/reference/api_references.html#cms-toolbar?

vxsx: that doesn't really concern people who customise the toolbar as they can only _add_ to it, not remove things. i didn't change anything api-wise, only used what existed

cms/tests/test_toolbar.py
```diff
@@ -428,7 +428,7 @@ def test_no_change_button(self):
         # Logo + page-menu + admin-menu + logout
         self.assertEqual(len(items), 3, items)
         admin_items = toolbar.get_or_create_menu(ADMIN_MENU_IDENTIFIER, 'Test').get_items()
-        self.assertEqual(len(admin_items), 9, admin_items)
+        self.assertEqual(len(admin_items), 12, admin_items)
```
vxsx: clipboard, clear clipboard and a break

cms/toolbar/items.py
```diff
@@ -286,7 +286,7 @@ class FrameItem(BaseItem):
     def __init__(self, name, url, active=False, disabled=False,
                  extra_classes=None, on_close=None, side=LEFT):
         super(FrameItem, self).__init__(side)
-        self.name = "%s ..." % force_text(name)
+        self.name = "%s..." % force_text(name)
```
yakky: Any specific reason to do this? This may breaks tests for the toolbar in third party applications

cms/cms_toolbars.py
yakky: Actually people **can** remove items from the toolbar

cms/toolbar/items.py
vxsx: well, there should be no space between a word and the ellipsis in the ui, same as in all the applications or operating systems

Issue #4443, opened 2015-09-11
Multiple PageFields are not possible in a single form


creimers:
```
Django==1.8.4
django-cms=3.1.3
```
I have a plugin with a PageField and an Inline, which itself also has a PageField.

Minimal example:

``models.py``
```
class Teaser(CMSPlugin):
    internal_link = PageField(
        blank=True,
        null=True,
        verbose_name='interner Link'
    )
    ... (copy_relations, etc.)


class TeaserButton(models.Model):
    teaser = models.ForeignKeys(Teaser, related_name='buttons')
    internal_link = PageField(
        blank=True,
        null=True,
        verbose_name='interner Link'
    )
```

``cms_plugins.py``
```
class ButtonInline(admin.StackedInline):
    model = TeaserButton
    ...


class TeaserPlugin(CMSPluginBase):
    ...
    inlines = [ButtonInline, ]
```

When adding an instance of the plugin, I observer a strange behavior. The PageField's select works in the plugin's form:
<img width="1048" alt="bildschirmfoto 2015-09-11 um 14 26 13" src="https://cloud.githubusercontent.com/assets/6090492/9814839/7773acc4-5891-11e5-954b-4c001df4ca3c.png">

In the inline's form, however, the PageField's select is *empty*:
<img width="1050" alt="bildschirmfoto 2015-09-11 um 14 26 21" src="https://cloud.githubusercontent.com/assets/6090492/9814849/935fd39a-5891-11e5-85fe-090b2f6c0817.png">

Additionally, I get the following browser console error:

<img width="639" alt="bildschirmfoto 2015-09-11 um 14 26 33" src="https://cloud.githubusercontent.com/assets/6090492/9814854/a5fd974e-5891-11e5-92c3-e8d65cd3ce1a.png">

Once the plugin instance is saved with a not-empty PageField and then edited again, the first inline's PageField select is populated properly. Subsequent inline instances' PageField selects are empty again.

<img width="984" alt="bildschirmfoto 2015-09-11 um 14 32 05" src="https://cloud.githubusercontent.com/assets/6090492/9814920/4edcd348-5892-11e5-93e8-6b6b540d6b77.png">

# **TL;DR**
Apparently, there can only be one PageField per form.

evildmp:
Shouldn't the title of this issue be: "Multiple PageFields are not possible in a single form"?

creimers:
It should.

mkoistinen:
[...]

PR #4522, opened 2015-09-29
Fix PageSelectWidget JS syntax

qnub:
Original JS raise exception: `Syntax error, unrecognized expression #id_page_2 option[value]`

### DIFF
--- a/cms/forms/widgets.py
+++ b/cms/forms/widgets.py
@@ -102,10 +102,10 @@ class PageSelectWidget(MultiWidget):
     var handlePageChange = function(page_id) {
         if (page_id) {
             $("#id_%(name)s_2 option").attr('selected', false);
-            $("#id_%(name)s_2 option[value=" + page_id + "]").attr('selected', true);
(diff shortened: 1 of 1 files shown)


cms/forms/widgets.py
```diff
@@ -102,10 +102,10 @@ def render(self, name, value, attrs=None):
     var handlePageChange = function(page_id) {
         if (page_id) {
             $("#id_%(name)s_2 option").attr('selected', false);
-            $("#id_%(name)s_2 option[value=" + page_id + "]").attr('selected', true);
+            $("#id_%(name)s_2 option[value='" + page_id + "']").attr('selected', true);
         } else {
             if ($("#id_%(name)s_2").length) {
-                $("#id_%(name)s_2 option[value=]").attr('selected', true);
+                $("#id_%(name)s_2 option[value]").attr('selected', true);
```
vxsx: shouldn't it be `option[value='']`, since we are trying to pick the empty value not every value?

vxsx:
related to https://github.com/divio/django-cms/issues/4443

PR #4523, opened 2015-09-29
interrupting dragging through ESC issue

dhorelik:
Interrupting dragging through ESC button was not working as expected, the latest interaction got cancelled


vxsx:
LGTM

PR #4525, opened 2015-09-30
tooltip text on touch devices

dhorelik:
* tooltip text updated
* tooltip behavior on touch slightly updated to avoid unnecessary events binding
* minor typo


cms/static/cms/js/modules/cms.base.js
```diff
@@ -39,6 +39,8 @@ var CMS = {
     $(document).ready(function () {
         CMS.API.Helpers = {
 
+            isTouch: window.ontouchstart !== undefined,
```
vxsx: having ontouchstart in window sadly doesn't really mean anything. 
we have the surface tablet that handles both mouse events and touch events, but the user might be hovering with mouse and we would still show the `tap here`.

can we instead detect if the sequence of events that started the showTooltip started with touchstart or mouseover?

cms/static/cms/js/modules/cms.base.js
```diff
@@ -118,8 +120,8 @@ var CMS = {
             },
 
             // handles the tooltip for the plugins
-            showTooltip: function (name, id) {
-                var tooltip = $('.cms-tooltip');
+            showTooltip: function (e, name, id) {
+                var tooltip = CMS.API.Helpers.pickTooltip();
```
vxsx: it would actually be amazing if we could move tooltip init / related methods out of helpers (leaving old api of showTooltip / hideTooltip)  into it's own cms.tooltip.js

vxsx: what do you think @FinalAngel ?

cms/static/cms/js/modules/cms.base.js
dhorelik: well, the events are normalized with pep.js, so showTooltip comes from `pointerover`. There is a `originalEvent.pointerType` property, but it gets `mouse` on ipad, so it's not robust at all.

I believe we'll need a solid solution to define whether we are dealing with a touch device anyway, so I'll look into better wayto implement touch detection. Modernizr does it somehow

cms/static/cms/js/modules/cms.base.js
dhorelik: yep, that makes sense, will do

cms/static/cms/js/modules/cms.base.js
vxsx: modernizr officially says that it it not possible to detect touch device, just the availablity of touch events

it comes from pointerover of course, but we can at least detect if there was a touch recently / nearby, and trigger some flag to change toolbar to the touch mode, and then after showing it - reset the flag. and keep doing it over and over again

(that's roughly what i was thinking)

cms/static/cms/js/modules/cms.base.js
FinalAngel: yeah we also plan to do that for messages, I thing those components should be handled separately

cms/static/cms/js/modules/cms.base.js
dhorelik: checking this every time is an overhead, I think. My thought is that if a user taps on the page at least once, it makes sense to suggest that he is going to keep tapping through the ui. So I've introduced the `isTouch` flag, which defaults to `false` and updates on first body tap

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -225,7 +225,7 @@ var CMS = window.CMS || {};
                 this.ui.modal.removeClass('cms-modal-has-breadcrumb');
 
                 // hide tooltip
-                this.hideTooltip();
+                CMS.API.Tooltip.displayToggle(false);
```
vxsx: i think i like `hide` more than `displayToggle(false)`, explicit looks much better

cms/static/cms/js/modules/cms.tooltip.js
```diff
@@ -0,0 +1,108 @@
...
+            },
+
+            hide: function () {
+                var tooltip = this.domElem;
+
+                // change css
+                tooltip.css('visibility', 'hidden').hide();
+
+                // unbind events
+                $('body').unbind('mousemove.cms');
```
vxsx: we should cache the body

cms/static/cms/js/modules/cms.tooltip.js
```diff
@@ -0,0 +1,108 @@
...
+            position: function (e, tooltip) {
+                // so lets figure out where we are
+                var offset = 20;
+                var relX = e.pageX - $(tooltip).offsetParent().offset().left;
+                var relY = e.pageY - $(tooltip).offsetParent().offset().top;
+                var bound = $(tooltip).offsetParent().width();
+                var pos = relX + tooltip.outerWidth(true) + offset;
+
+                tooltip.css({
+                    'left': (pos >= bound) ? relX - tooltip.outerWidth(true) - offset : relX + offset,
```
vxsx: no need for quoted keys

cms/static/cms/js/modules/cms.tooltip.js
```diff
@@ -0,0 +1,108 @@
...
+                var that = this;
+
+                $('body').one('touchstart.cms', function () {
+                    that.isTouch = true;
+                    that.domElem = that.pick();
+                });
+            },
+
+            displayToggle: function (isShown, e, name, id) {
+                isShown ? CMS.API.Tooltip.show(e, name, id) : CMS.API.Tooltip.hide();
```
vxsx: `this.show` ?

cms/static/cms/js/modules/cms.tooltip.js
```diff
@@ -0,0 +1,108 @@
...
+                tooltip.css('visibility', 'visible')
+                    .data('plugin_id', id || null)
+                    .show()
+                    .find('span').html(name);
+
+                if (this.isTouch) {
+                    this.position(e, tooltip);
+
+                    // attach tooltip event for touch devices
+                    tooltip.bind('touchstart.cms', function () {
```
vxsx: `bind` -> `on`

vxsx: also i think we can bind it once, no need to bind/unbind

cms/static/cms/js/modules/cms.tooltip.js
```diff
@@ -0,0 +1,108 @@
...
+        CMS.Tooltip = new CMS.Class({
+
+            initialize: function () {
+                this.isTouch = false;
+                this.domElem = this.pick();
+
+                this.checkTouch();
+            },
+
+            checkTouch: function () {
```
vxsx: can you add jsdoc for all methods, like in https://github.com/divio/django-cms/blob/preview/cms/static/cms/js/modules/cms.modal.js ?

vxsx:
looks awesome

cms/static/cms/js/modules/cms.base.js
vxsx: :+1:

cms/static/cms/js/modules/cms.tooltip.js
FinalAngel: checkout this.ui in toolbar, modal or sideframe

dhorelik: `this.ui` seems like an overhead, tooltip just needs body, cached it via `this.body`

FinalAngel:
LGTM

vxsx:
can be merged after resolving the conflicts

PR #4526, opened 2015-10-01
Separate openMessage into cms.message.js

FinalAngel:
This PR creates a new component out of the ``openMessage`` and ``closeMessage`` API calls within ``CMS.API.Toolbar``.

The new format is:

```javascript
var message = new CMS.Message();
message.open({
    message: 'hello world'
});
message.close();
```


cms/static/cms/js/modules/cms.message.js
```diff
@@ -0,0 +1,166 @@
...
+
+// #############################################################################
+// MODAL
+(function ($) {
+    'use strict';
+
+    // shorthand for jQuery(document).ready();
+    $(function () {
+        /**
+         * Displays a message underneath the toolbar
```
vxsx: dot

cms/static/cms/js/modules/cms.message.js
```diff
@@ -0,0 +1,166 @@
...
+             *
+             * @method open
+             * @param opts
+             * @param opts.message {String|HTMLNode} message to be displayed
+             * @param [opts.dir='center'] {String} direction to be displayed `center` `left` or `right`
+             * @param [opts.delay=this.options.messageDelay] {Number} delay until message is closed
+             * @param [opts.error] {Boolean} if true sets the style to `.cms-messages-error`
+             */
+            open: function open(opts) {
+                if (!(opts && opts.message)) {
```
FinalAngel: this part is copied from https://github.com/divio/django-cms/pull/4526/files#diff-f33b2ef979aa78d0eeb29b02871daa5eL457 and changes there are no part within this PR

cms/static/cms/js/modules/cms.message.js
```diff
@@ -0,0 +1,166 @@
...
+                });
+
+                // set top to 0 if toolbar is collapsed
+                if (this.settings.toolbar === 'collapsed') {
+                    top = 0;
+                }
+
+                // do we need to add debug styles?
+                if (this.config.debug) {
+                    top = top + 5;
```
vxsx: cache the debug height somewhere, it can be changed in css

cms/static/cms/js/modules/cms.message.js
```diff
@@ -0,0 +1,166 @@
...
+                this.ui = {
+                    container: container,
+                    body: $('html'),
+                    toolbar: container.find('.cms-toolbar'),
+                    messages: container.find('.cms-messages')
+                };
+            },
+
+            /**
+             * Opens a message window underneath the toolbar.
```
vxsx: we should tell somewhere here that if delay is 0 it's actually not gonna hide until you explicitly close it

cms/static/cms/js/modules/cms.modal.js
```diff
@@ -50,6 +50,9 @@ var CMS = window.CMS || {};
                 // event emitter
                 this._setupEventEmitter();
 
+                // reference messaging system
+                this.messages = new CMS.Message();
```
vxsx: we could also add it to the `CMS.API` and use the single instance

cms/static/cms/js/modules/cms.message.js
FinalAngel: check the comment above, will not change the functionality in this PR

cms/static/cms/js/modules/cms.modal.js
FinalAngel: yeah thought about it too, though handling in separately gives more freedom and is detached from CMS.API

vxsx: doing it separately means that you're creating an instance for _each_ plugin which in turn means higher memory footprint etc.

vxsx:
:+1:

PR #4532, opened 2015-10-02
no click on dblclick

dhorelik:
double click on link in edit mode didn't open the edit form, since the click event took over and users got redirected


cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -134,6 +137,22 @@
                     );
                 });
 
+                // prevents single click from messing up the edit call
```
FinalAngel: we actually removed this once as it creates several complications, for example:

- What if javascript is attached to the link
- What happens when you do CTRL + click

I'm not sure introducing this behaviour will solve our problems

dhorelik: hmm, but now it's basically impossible to edit link or button. We have to decide upon the behavior then

FinalAngel: yeah you would have to switch to the structure mode. I'd really love alternatives to this, yet I have no idea how to achieve this. /cc @vxsx @yakky

yakky: @dhorelik @FinalAngel either way, we're screwed... I'm ignorant about frontend, but how about an overlay on top of the link which "steals" the events in content mode/draft?

vxsx: it's very tricky, but similar to what i was thinking. what we could also do is figure out if the `cms-plugin` contains a link/button and show some kind of "edit" icon right after it so the link is clickable and editable. in order for it not to change layout we'd have to position it absolutely somehow. but i think it doesn't play that well with touch and a lot of complicated trickery involved. 

or we keep the behaviour Dmytro did, just add workarounds for ctrlKey and shiftKey to pass through
and if there's js involved - we use http://api.jquery.com/event.stopimmediatepropagation/ which stops other event handlers from executing (although we still need usual stopPropagation in case event is delegated)

FinalAngel: alright, than I'd suggest @dhorelik implements:
- js propagnation
- ctrlKey / shiftKey workarounds

/cc @vxsx

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -183,6 +186,27 @@ var CMS = window.CMS || {};
...
+                this.ui.container.find('a').on('click', function (e) {
+                    if (e.shiftKey || e.ctrlKey || e.metaKey) {
+                        return;
+                    }
+
+                    e.preventDefault();
+                    e.stopImmediatePropagation();
```
vxsx: why stopImmediate? we don't mind that they having custom js executed there? e.g. accordions

dhorelik: I've done it intentionally to avoid custom js, but yes, it makes sense to let it run, though. Removed.

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -183,6 +186,26 @@ var CMS = window.CMS || {};
                     );
                 });
 
+                // prevents single click from messing up the edit call
+                this.ui.container.find('a').on('click', function (e) {
+                    if (e.shiftKey || e.ctrlKey || e.metaKey) {
+                        return;
+                    }
+
+                    e.preventDefault();
```
vxsx: we can also check if the `e.isDefaultPrevented()` and immediately bail out right here by returning
this way if the user has a custom js on a link with the url and he does the preventing already - we don't go to the link

dhorelik: @vxsx added the check along with the minor comment, custom js handlers are going to work now.

another thing - I played around with the default addons (ex.accordions), and their handlers often aren't attached to `a`, but rely upon event propagation. So I suggest getting rid of `e.stopPropagation()` here.

FinalAngel:
LGTM

vxsx:
:+1:

PR #4534, opened 2015-10-02
change width of the "mobile" dropdown

vxsx:
previously it was set to fit the content of the body, but that means
that if someone's content is overflowing on mobile the dropdown would
also go out of the screen, which is not ideal

also fixes "invisible button" problem in that same menu


FinalAngel:
LGTM

PR #4540, opened 2015-10-04
change the way modal is resized

vxsx:
if you get to the minimum width or height previously it stopped
resizing completely in both directions. this commit makes resizing sides
independent so it seems more natural

Before: http://d.pr/v/183RS
After: http://d.pr/v/1g7mt

/cc @mkoistinen @FinalAngel


FinalAngel:
LGTM :)

PR #4547, opened 2015-10-06
Remove last declared ID

FinalAngel:
This removes ``#cms-toolbar`` and replaces it with ``.cms`` to reserve the "cms" namespace. Some struggles:

CSS specificity is a charm (irony off) we need to use ``div.cms`` as wrapper to take precedence over ``.cms-reset a``.. selectors. Other than that, we removed the css specificity for modules where its not required anymore.

The css files are thereafter more readable as we save one tab space :)

You can review this best using: https://github.com/divio/django-cms/pull/4547/files?w=1


cms/templates/cms/toolbar/toolbar.html
```diff
@@ -1,7 +1,11 @@
 {% load i18n l10n %}
 
 {% language request.toolbar.toolbar_language %}
-<div id="cms-toolbar" class="cms-reset{% if not user.is_authenticated %} cms-toolbar-auth{% endif %}{% if debug %} cms-toolbar-debug{% endif %}" dir="ltr" data-touch-action="none">
+<div id="cms-top"
```
vxsx: why keep an id?

vxsx:
looks ok

cms/templates/cms/toolbar/toolbar.html
FinalAngel: for the tests

PR #4548, opened 2015-10-06
Numerous small fixes

FinalAngel:
- Expand / Collapse added a "#" to the url
- Navigation can be again closed once opened when clicking on the root entry
- Aligned installation screen background with djangocms-admin-style
- Set correct paddings / margin between each navigation entry (between breaklines...)


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -149,6 +150,12 @@ var CMS = window.CMS || {};
                         e.stopPropagation();
                         var el = $(this);
 
+                        // close navigation once it's pressed again
+                        if (el.parent().hasClass(root) && open) {
+                            that.ui.body.trigger(that.click);
```
vxsx: is it the same as calling `reset()` ?

cms/static/cms/sass/components/_toolbar.scss
```diff
@@ -179,7 +179,7 @@
         @include hide-content();
         height: 1px;
         margin: 0 0 4px;
-        padding: 0 0 3px;
+        padding: 0 0 4px;
```
vxsx: :D

cms/static/cms/js/modules/cms.toolbar.js
FinalAngel: apparently not, calling reset has some weird effects. There might be other events also intertwined

vxsx:
:+1:

PR #4549, opened 2015-10-06
Adds "move_a_copy" as an optional boolean flag on move_plugin

mkoistinen:
This PR updates existing code to work with the new changes to the clipboard.


cms/admin/placeholderadmin.py
```diff
@@ -342,6 +351,7 @@ def copy_plugins(self, request):
             )
         self.post_copy_plugins(request, source_placeholder, target_placeholder, plugins)
         json_response = {'plugin_list': reduced_list, 'reload': reload_required}
+        print(CMSPlugin.find_problems())
```
ojii: this should probably be removed ;-)

cms/admin/placeholderadmin.py
```diff
@@ -439,39 +457,63 @@ def move_plugin(self, request):
...
+            return HttpResponseBadRequest('order parameter did not have all '
+                                          'plugins of the same level in it' + str(order))
 
         self.post_move_plugin(request, source_placeholder, placeholder, plugin)
-        json_response = {'reload': requires_reload(PLUGIN_MOVE_ACTION, [plugin])}
-        return HttpResponse(json.dumps(json_response), content_type='application/json')
+        json_response = {
+            'reload': move_a_copy or requires_reload(
+                PLUGIN_MOVE_ACTION, [plugin])}
+        print(CMSPlugin.find_problems())
```
ojii: ^^

cms/admin/placeholderadmin.py
```diff
@@ -419,59 +447,128 @@ def edit_plugin(self, request, plugin_id):
...
+                    to_language=language)
+            else:
+                source_plugins = [plugin] + list(plugin.get_descendants())
+                new_plugins = copy_plugins.copy_plugins_to(
+                    source_plugins, placeholder, language, parent_id)
+
+            new_plugins = [nu for nu, _old in new_plugins]
+            top_parent = new_plugins[0].parent_id
+            top_plugins = [p for p in new_plugins if p.parent_id == top_parent]
+            top_plugins_pks = [str(p.pk) for p in top_plugins]
```
cyberstar: maybe it would be better to combine ``top_plugins`` and ``top_plugins_pks`` (and also maybe ``new_plugins ``) into one for loop instead of 2 (3) list comprehensions?

Though minor importance.

cms/admin/placeholderadmin.py
```diff
@@ -419,59 +447,128 @@ def edit_plugin(self, request, plugin_id):
...
-            child.placeholder = placeholder
-            child.language = language
-            child.save()
+            # Regular move
+            if parent_id:
+                if plugin.parent_id != parent_id:
+                    parent = CMSPlugin.objects.get(pk=parent_id)
+                    if parent.placeholder_id != placeholder.pk:
+                        return HttpResponseBadRequest(force_text(
+                            'parent must be in the same placeholder'))
```
cyberstar: not translated?

cms/admin/placeholderadmin.py
```diff
@@ -419,59 +447,128 @@ def edit_plugin(self, request, plugin_id):
...
+                    if parent.language != language:
+                        return HttpResponseBadRequest(force_text(
+                            'parent must be in the same language as '
```
cyberstar: not translated?

cms/admin/placeholderadmin.py
```diff
@@ -419,59 +447,128 @@ def edit_plugin(self, request, plugin_id):
...
+            # Don't neglect the children
+            for child in [plugin] + list(plugin.get_descendants()):
+                child.placeholder = placeholder
+                child.language = language
+                child.save()
+
         plugins = reorder_plugins(placeholder, parent_id, language, order)
         if not plugins:
-            return HttpResponseBadRequest('order parameter did not have all plugins of the same level in it')
+            return HttpResponseBadRequest('order parameter did not have all '
```
cyberstar: All ``HttpResponseBadRequest`` should be translatable or only part of them?

cms/utils/copy_plugins.py
```diff
@@ -7,20 +7,33 @@ def copy_plugins_to(old_plugins, to_placeholder,
     """
     Copies a list of plugins to a placeholder to a language.
     """
+    # import pdb; pdb.set_trace()
```
cyberstar: maybe this could be deleted?

cms/utils/copy_plugins.py
```diff
@@ -7,20 +7,33 @@ def copy_plugins_to(old_plugins, to_placeholder,
...
     old_parent_cache = {}
-    # For subplugin copy, first plugin's parent must be nulled before copying.
+    # For subplugin copy, top-level plugin's parent must be nulled
+    # before copying.
     if old_plugins:
-        old_plugins[0].parent = old_plugins[0].parent_id = None
-    new_plugins = [old.copy_plugin(to_placeholder, to_language or old.language,
-                                   old_parent_cache, no_signals)
-                   for old in old_plugins]
+        old_parent = old_plugins[0].parent
```
cyberstar: is ``id(old_parent) != old_plugins[0].parent``? Otherwise you would override old_parent in the first forloop iteration.
If that is the case maybe consider using ``for old_plugin in old_plugins[1:]`` instead ``for old_plugin in old_plugins``

cms/utils/copy_plugins.py
```diff
@@ -7,20 +7,33 @@ def copy_plugins_to(old_plugins, to_placeholder,
...
+
     if new_plugins and parent_plugin_id:
         from cms.models import CMSPlugin
-        new_plugins[0].parent_id = parent_plugin_id
-        new_plugins[0].save()
-        new_plugins[0] = new_plugins[0].move(CMSPlugin.objects.get(pk=parent_plugin_id), pos='last-child')
+        parent_plugin = CMSPlugin.objects.get(pk=parent_plugin_id)
+        top_plugins = [p for p in new_plugins if p.parent_id is None]
+        for idx, plugin in enumerate(new_plugins):
+            if plugin in top_plugins:
```
cyberstar: Why not check ``if plugin.parent_id is None`` here? That might save some time for not top plugins by not walking through each element in top_plugins list...
Also in this case you don't need L#30 (``top_plugins = [p for p in new_plugins if p.parent_id is None]``)

cms/admin/placeholderadmin.py
mkoistinen: Fair point.

cms/admin/placeholderadmin.py
mkoistinen: I'll add.

cms/admin/placeholderadmin.py
mkoistinen: Thanks, will add.

cms/admin/placeholderadmin.py
mkoistinen: I didn't change these, but will add.

cms/utils/copy_plugins.py
mkoistinen: =) Could be =)

cms/utils/copy_plugins.py
mkoistinen: No, the point of this is to find all the "old plugins" that are at the top level and set their parent to `None`. We need to consider the first item so we can set its parent to `None` too. Sure, I could move it out of the loop and consider it separately, but then it wouldn't be very DRY.

cms/utils/copy_plugins.py
mkoistinen: :+1:

cms/admin/placeholderadmin.py
cyberstar: But it is also 3 walks on the same list, instead of 1.
PS. well ok 2 walks on same and one for only top (p.parent_id == top_parent)

cyberstar:
LGTM :+1:

cms/admin/placeholderadmin.py
mkoistinen: OK, fair point.

Issue #4423, opened 2015-09-04
Improvement: Use custom plugin queryset


o-zander:
Hi, I use custom plugins which uses django-filer fields. These fields are foreign keys. I noticed multiple queries to the filer table from one plugin type, which could be avoided by using ``select_related``. But I can't find a feature to set ``select_related`` on the queryset because the queryset is build by the function ``downcast_plugins``. It would be easy to use a custom queryset since in this line: https://github.com/divio/django-cms/blob/develop/cms/utils/plugins.py#L156 the plugin admin is used to obtain the model.

My proposal is to use ``cls.get_queryset()`` or use a custom ``cls.get_plugin_queryset()`` function on the admin class which can be overwritten.

evildmp:
Hi @o-zander, thanks for this suggestion, which looks like a useful improvement. Would you like to contribute the patch?

o-zander:
Yes, I can do that. Maybe I can get it done this week. Any advice on forking which branch to ease the process of merging the changes?

yakky:
@o-zander use develop. As it changes the models API is targeted to 3.2

FinalAngel:
@mkoistinen I assume that this hasn't been solved yet (as of the reverts)?

czpython:
Fixed by https://github.com/divio/django-cms/pull/4556

PR #4554, opened 2015-10-07
#4423: added hook for using custom query sets for plugins

o-zander:
Improvement for #4423

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -156,6 +156,9 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
         elif getattr(self, 'render_template', False):
             return getattr(self, 'render_template', False)
 
+    def get_render_queryset(self):
+        return self.model._default_manager.all()
+
     def render(self, context, instance, placeholder):
         context['instance'] = instance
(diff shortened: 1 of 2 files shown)


yakky:
@o-zander nice!
@mkoistinen are you with merging this for 3.2? 
@evildmp do we need any docs for this?

mkoistinen:
This is nice. Would like to see this get into 3.2.

mkoistinen:
@o-zander, Tests aren't passing. Some of this may be my fault (working on them now), but there are some tests failing because of this API change.

mkoistinen:
Actually, I don't think this does require any docs changes... committing...

mkoistinen:
@o-zander I've committed this in, but we still need to fix the tests.

mkoistinen:
Actually, reverting until we have tests that pass.

o-zander:
I nearly changed anything except a little wrapper func so I'm very suprised this breaks something. Maybe it helps to use objects instead of _default_manager. Then there isn't really a change at all. But I will have a look at it. Could be a little bit difficult for me to distinguish your test failures from the one caused by the api change ;)

o-zander:
Opened a new merge request: #4556

PR #4555, opened 2015-10-07
Revert "#4423: added hook for using custom query sets for plugins"

mkoistinen:
Reverts divio/django-cms#4554


mkoistinen:
@o-zander sorry about this hassle, I was overly eager to get this in, and since it was failing tests all over the place, I've reverted this PR. Can you resubmit with passing tests? If you can do this very soon, we can get this into a 3.2 RC1 release tomorrow.

PR #4556, opened 2015-10-07
#4423: added hook for using custom query sets for plugins [fixed]

o-zander:
This time it should work :) I tested it with django 1.8 but I don't think this is in any way related to a specific django version. Sometime it should be explained in the docs so plugin developers can use this api to improve their queries.


mkoistinen:
Let's rebase this once #4550 gets committed.

mkoistinen:
Can you rebase and retest now?

mkoistinen:
Thanks @o-zander!

PR #4557, opened 2015-10-07
bind the keyboard traversing to CMS.Plugin instance

vxsx:
fixes a bug introduced in https://github.com/divio/django-cms/pull/4544
where keyboard events were not applied when "Add plugin" modal was
opened


mkoistinen:
LGTM, please rebase and re-test

PR #4558, opened 2015-10-07
Fix "add plugin" url

vxsx:
before:
```
add_plugin: "/en/admin/cms/page/add-plugin/}"
```
after:
```
add_plugin: "/en/admin/cms/page/add-plugin/"
```

wasn't really affecting functionality


mkoistinen:
LGTM, please rebased though, so tests will pass.

PR #4559, opened 2015-10-07
Parent restriction

vxsx:
This PR adds the handling of parent classes restrictions in structureboard.
It was handled by the backend for _adding_ plugins, but after adding a plugin to the appropriate parent you could still drag it to a different parent which shouldn't have been allowed. This PR fixes that.

There is one case though, where it's still not working correctly - and that case is clipboard. When a clipboard plugin is registered in the `CMS._plugins` pool in js - it doesn't have parent classes and child classes.

@mkoistinen could you please take a look?


cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -285,8 +286,11 @@ var CMS = window.CMS || {};
 
                 var bounds = this.options.plugin_restriction;
                 var type = clipboardPlugin.data('settings').plugin_type;
+                var parent_bounds = clipboardPlugin.data('settings').plugin_parent_restriction;
+                var currentPluginType = this.options.plugin_type;
 
-                if (bounds.length && $.inArray(type, bounds) === -1) {
+                if ((bounds.length && $.inArray(type, bounds) === -1) ||
+                    (parent_bounds.length && $.inArray(currentPluginType, parent_bounds) === -1)) {
```
vxsx: this is what is supposed to fix clipboard, but parent bounds is always an empty array

FinalAngel:
LGTM

mkoistinen:
LGTM

PR #4560, opened 2015-10-08
Minor Frontend Fixes

FinalAngel:
- Set correct z-index for messages so it appears underneath the toolbar
- Cutting a plugin causes an error as Message.open provides incorrect params
- Restore default ``.cms {}`` css as this causes the tooltip to be underneath certain content elements with ``position: relative``.


cms/static/cms/sass/components/_toolbar.scss
```diff
@@ -2,13 +2,6 @@
 // TOOLBAR
 
 .cms-toolbar {
-    position: absolute;
-    top: $toolbar-debug-height;
-    left: 0;
-    z-index: z(base);
-    width: 100%;
-    margin-top: -$toolbar-height - 10px;
```
vxsx: i think margin top was actually needed here

FinalAngel: that was actually not in original code

FinalAngel: restored

vxsx:
:+1:

mkoistinen:
:+1:

PR #4564, opened 2015-10-08
Modal fixes

FinalAngel:
- Use proper title assignment, first prefix than suffix
- Use the new wizard for installation screen


cms/static/cms/js/modules/cms.modal.js
```diff
@@ -916,7 +916,11 @@ var CMS = window.CMS || {};
                             that.ui.titlePrefix.text(bc.eq(bc.length - 1).text().replace('›', '').trim());
                         }
 
-                        titleSuffix.text(innerTitle.text());
+                        if (titlePrefix.text() === '') {
```
vxsx: `titlePrefix.text().trim()` maybe?

vxsx:
:+1:  (even without trimming the string)

mkoistinen:
:+1:

Issue #4313, opened 2015-07-24
Disable toolbar option


leesolway:
Disabling the toolbar using the button within the toolbar itself (?toolbar_off) generates lots of the following errors within the browser console.

"Uncaught ReferenceError: CMS is not defined" 

Tested with:
Django 1.7
django-cms 3.1.2

FinalAngel:
Fixed in 3.2

PR #4565, opened 2015-10-08
Don't render editmode templates when there's no toolbar

mkoistinen:
Also, some defensive refactoring...

### DIFF
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -127,9 +127,11 @@ def render_placeholder(placeholder, context_to_copy, name_fallback="Placeholder"
 
     # Prepend frontedit toolbar output if applicable
     toolbar = getattr(request, 'toolbar', None)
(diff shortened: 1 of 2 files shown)


vxsx:
LGTM

vxsx:
#4313 
/cc @ojii

cms/plugin_rendering.py
```diff
@@ -127,9 +127,11 @@ def render_placeholder(placeholder, context_to_copy, name_fallback="Placeholder"
 
     # Prepend frontedit toolbar output if applicable
     toolbar = getattr(request, 'toolbar', None)
-    if getattr(toolbar, 'edit_mode', False) and getattr(placeholder, 'is_editable', True) and editable:
+    if (getattr(toolbar, 'edit_mode', False) and
+            getattr(toolbar, "show_toolbar", False) and
+            getattr(placeholder, 'is_editable', True) and editable):
```
ojii: I feel like this check should be a property/method on the toolbar object, so if this changes again in the future we don't have to change it in more than one place.

PR #4572, opened 2015-10-11
only autofocus field when not touching

vxsx:
when opening add plugin modal it autofocuses the search field which is very useful on desktop but painful on touch devices because it brings up a keyboard 

this attempts to fix it for most touch devices


FinalAngel:
LGTM

Issue #4569, opened 2015-10-09
Various UI/UX issues


mkoistinen:
Here be a list of small-ish UI/UX-related issues that should be repaired before 3.2.0 final.

- [x] Page tree becomes "broken" when one clicks the menu toggle. Here's a screen-shot taken after clicking to enable the "Sample" page in the menu. Note the "columns" with the published/unpublished/changed status indicators:
![image](https://cloud.githubusercontent.com/assets/615759/10394752/269068ba-6e67-11e5-9203-028bb8c8db0b.png)
- [x] add translate3d to toolbar to split up rendering (marginal rendering speedup on scroll #4573)
- [x] this is incorrect https://github.com/divio/django-cms/blob/8e9e540d19a61cef8fdfe28badf4fc084580cae4/cms/static/cms/sass/components/_modal.scss#L136-L151

PR #4573, opened 2015-10-12
Bugfixes

vxsx:
- less scrolling repaints #4569
- fix double scroll
- fix toolbar width jumping when switching content mode / structure mode
- removed unused clipboard code
- use box-sizing border-box by default
- change the gray-lighter shade
- fix buttons on ios without user styles


FinalAngel:
LGTM

Issue #4516, opened 2015-09-28
Page admin should have View on site re-enabled


evildmp:
Here's a typical use-case:

* I visit the CMS > Pages admin to have a look at my pages
* I decide I need a new one, so add it
* I hit Save and continue until I am happy with the Description etc
* now I want to add some content - and should be able to hit **View on site**, but it's not there. Instead I have to go back to the admin page list and (confusingly, though that's a different problem) open the page frontend, not the page admin, by selecting it from the list

evildmp:
Related issue: https://github.com/divio/django-cms/issues/4517

PR #4577, opened 2015-10-13
Fix for issue #4516

FinalAngel:
image

### DIFF
--- a/cms/templates/admin/cms/page/change_form.html
+++ b/cms/templates/admin/cms/page/change_form.html
@@ -17,6 +17,17 @@
 <script src="{% static "cms/js/dist/bundle.admin.changeform.min.js" %}" type="text/javascript"></script>
 {% endblock %}
 
+{% block content_title %}
+    {% if title %}<h1>{{ title }}</h1>{% endif %}
+    {% if not popup %}
+        <ul class="object-tools">
+            <li>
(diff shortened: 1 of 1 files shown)


cms/templates/admin/cms/page/change_form.html
```diff
@@ -17,6 +17,17 @@
 <script src="{% static "cms/js/dist/bundle.admin.changeform.min.js" %}" type="text/javascript"></script>
 {% endblock %}
 
+{% block content_title %}
+    {% if title %}<h1>{{ title }}</h1>{% endif %}
+    {% if not popup %}
+        <ul class="object-tools">
+            <li>
+                <a href="../{{ page.id }}/{{ language }}/preview/" target="_blank">{% trans "View on site" %}</a>
```
vxsx: is it ok to manually build the url? we don't have any url reverse helpers?

FinalAngel: I checked the history and cms, there seems to be no url for this case (it was once there like this) @yakky maybe you know?

yakky: @FinalAngel @vxsx indeed. I'll open an issue for 3.3: all the admin views are anonymous :(

vxsx:
alright for now :+1:

PR #4592, opened 2015-10-15
Fix formtools issue in django 1.8

FinalAngel:
https://docs.djangoproject.com/en/1.8/ref/contrib/formtools/

Formtools is deprecated in django 1.8 and moved into its own package. This needs to be defined within ``INSTALL_REQUIREMENTS``


mkoistinen:
lgtm

PR #4595, opened 2015-10-17
Clipboard fixes

vxsx:
cherry-picked few improvements from #4549 
and fixed a problem with menu not being reset correctly when opening / clearing the clipboard


FinalAngel:
LGTM

PR #4601, opened 2015-10-20
add placeholders_list to request.toolbar

vxsx:
to have correct order of placeholders in the template

/cc @mkoistinen

### DIFF
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -164,10 +164,15 @@ def render_placeholder(placeholder, context_to_copy, name_fallback="Placeholder"
     toolbar_content = ''
 
     if edit and editable:
+        # TODO remove ``placeholders`` in 3.3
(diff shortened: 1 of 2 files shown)


yakky:
Wondering if instead we should add a DeprecationWarning and wait for 1 more release before removing toolbar.placeholders

vxsx:
+1 for DeprecationWarning, but i'm not really sure how to do it correctly :)

mkoistinen:
@yakky since this is an internal structure, I don't think we require a warning, but we'll definitely have a note in the release notes. As you can see, in this PR, we plan to remove the old in 3.3.0.

FinalAngel:
@mkoistinen I'd suggest creating a ticket for 3.3 then to remove that line, otherwise comments get lost. I'm also not for a deprecation warning.

yakky:
:+1: for comment in release notes and a ticket to track the removal
Ok to not have a warning

PR #4602, opened 2015-10-20
Wizard refactoring

mkoistinen:
This PR:
- Removes "one wizard per Content Type" restriction
- Greatly improves permission checks for "Add page" wizard
- Adds a new "Add sub page" wizard (also using new permissions checks)
- Adds page type support for both wizards
- Refactor for clarity, extensibility


cms/wizards/wizard_base.py
```diff
@@ -48,8 +48,7 @@ class Wizard(WizardBase):
 
     @property
     def id(self):
-        content_type = ContentType.objects.get_for_model(self.get_model())
-        return content_type.pk
+        return self.__class__.__name__
```
czpython: agreed to do appname.classname

cms/wizards/views.py
```diff
@@ -85,8 +87,10 @@ def get_context_data(self, **kwargs):
     def get_form_kwargs(self, step=None):
         kwargs = super(WizardCreateView, self).get_form_kwargs()
         kwargs['wizard_user'] = self.request.user
-
-        if self.is_second_step(step):
+        if self.is_first_step(step):
+            kwargs['wizard_page'] = Page.objects.filter(
```
czpython: move the code from get_form_initial to this block.
```python
page = Page.objects.filter(pk=page_pk).first()
kwargs['initial']['page'] = page
kwargs['wizard_page'] = page
```

cms/cms_wizards.py
```diff
@@ -1,20 +1,85 @@
...
+        elif position in ("left", "right") and target.parent_id:
+            return permissions.has_generic_permission(
+                target.parent_id, user, "add", target.site_id)
+    else:
+        if site is None:
+            site = Site.objects.get_current()
+        global_add_perm = GlobalPagePermission.objects.user_has_add_permission(
+            user, site).exists()
+        perm_str = opts.app_label + '.' + get_permission_codename('add', opts)
+        if user.has_perm(perm_str) and global_add_perm:
```
czpython: ``return user.has_perm(perm_str) and global_add_perm``

cms/cms_wizards.py
```diff
@@ -1,20 +1,85 @@
...
+        if position in ("first-child", "last-child"):
+            return permissions.has_generic_permission(
+                target.pk, user, "add", target.site_id)
+        elif position in ("left", "right") and target.parent_id:
+            return permissions.has_generic_permission(
+                target.parent_id, user, "add", target.site_id)
+    else:
+        if site is None:
+            site = Site.objects.get_current()
+        global_add_perm = GlobalPagePermission.objects.user_has_add_permission(
```
czpython: this is done regardless of target, so is better to move it out of condition and define it above ``if target``

cms/static/cms/js/modules/cms.page_type_select.js
```diff
@@ -0,0 +1,26 @@
+//##############################################################################
```
FinalAngel: I would say we name it ``cms.wizzards.js`` and create and encapsulate all js related functions within the wizard (there more in the html files)

cms/static/cms/js/modules/cms.page_type_select.js
```diff
@@ -0,0 +1,26 @@
...
+        var pageTypeField = $('.form-row.page_type select'),
+            contentGroup = $('.form-row.content'),
+            contentField = contentGroup.find('textarea');
+
+        // Show it if we change to an app_hook that requires a namespace
+        pageTypeField.on('change', function () {
+            if ($(this).val()){
+                contentGroup.hide();
+                contentField.prop('disabled', true);
+            }
```
FinalAngel: same line

cms/static/cms/js/modules/cms.page_type_select.js
```diff
@@ -0,0 +1,26 @@
+//##############################################################################
+// PAGE WIZARD
+
+/* global CMS */
+(function ($) {
+    'use strict';
+    // CMS.$ will be passed for $
+    $(document).ready(function () {
```
FinalAngel: we use the shorthand now:

```
// shorthand for jQuery(document).ready();
$(function () {
```

at best copy style/structure from: https://github.com/divio/django-cms/blob/develop/cms/static/cms/js/modules/cms.messages.js#L1

cms/templates/cms/wizards/base.html
```diff
@@ -15,6 +15,7 @@
 {% block extrahead %}
     {{ block.super }}
     <script src="{% static "cms/js/dist/bundle.admin.base.min.js" %}" type="text/javascript"></script>
+    {{ form.media }}
     <script>
     $(function () {
```
FinalAngel: we might want to move this also into ``cms.wizards.js`` and include it globally.

cms/static/cms/js/modules/cms.page_type_select.js
```diff
@@ -0,0 +1,26 @@
...
+
+        // Show it if we change to an app_hook that requires a namespace
+        pageTypeField.on('change', function () {
+            if ($(this).val()){
+                contentGroup.hide();
+                contentField.prop('disabled', true);
+            }
+            else {
+                contentGroup.show();
+                contentField.removeProp('disabled');
```
vxsx: `.prop('disabled', false)` maybe, just for consistency?

cms/static/cms/js/modules/cms.page_type_select.js
```diff
@@ -0,0 +1,26 @@
...
+    // CMS.$ will be passed for $
+    $(document).ready(function () {
+
+        var pageTypeField = $('.form-row.page_type select'),
+            contentGroup = $('.form-row.content'),
+            contentField = contentGroup.find('textarea');
+
+        // Show it if we change to an app_hook that requires a namespace
+        pageTypeField.on('change', function () {
+            if ($(this).val()){
```
vxsx: space before the curly bracket

cyberstar:
LGTM.

cms/static/cms/js/modules/cms.page_type_select.js
mkoistinen: This JS file is part of a specific "widget" for a specific field for a specific cms_wizard. It doesn't make sense to combine it with anything else. This should *not* be included on all cms_wizard forms.

cms/templates/cms/wizards/base.html
mkoistinen: This JS file is part of a specific "widget" for a specific field for a specific cms_wizard. It doesn't make sense to combine it with anything else. This should not be included on all cms_wizard forms.

cms/forms/wizards.py
```diff
@@ -3,18 +3,81 @@
...
+
+
+class PageTypeSelect(forms.widgets.Select):
+    """
+    Special widget for the page_type choice-field. This simply adds some JS for
+    hiding/showing the content field based on the selection of this select.
+    """
+    class Media:
+        js = (
+            'cms/js/modules/cms.base.js',
```
vxsx: `cms.base.js` is already part of `bundle.admin.base.min.js`

cms/static/cms/js/modules/cms.page_type_select.js
FinalAngel: It would make sense to centralize all wizard related js in one file. Once it gets bigger > than separate

mkoistinen: I would agree with you if this was something that all wizards used, but it is not. This is JS for a wizard-specific field and if included for all wizards could affect 3rd party wizards. I have, however, pulled out the inline JS and CSS that you contributed from the base template into their own files.

PR #4604, opened 2015-10-21
add min-height to html when in structure mode

vxsx:
on clean slate projects on iOS9 `overflow: hidden` on html causes really
weird zooming issues (when there's not enough content)

strangely enough it worked correctly on iOS8


FinalAngel:
LGTM

PR #4607, opened 2015-10-23
Fix publish tooltip cutoff on touch devices

vxsx:
Before:
screen shot 2015-10-23 at 11 57 43

after:
screen shot 2015-10-23 at 11 57 09


FinalAngel:
LGTM

PR #4608, opened 2015-10-23
Also don't show page_type widget if the ...

mkoistinen:
... page type root exists, but has no children


cyberstar:
LGTM :+1:

FinalAngel:
LGTM

PR #4613, opened 2015-10-23
Visually differentiate currently viewed page

FinalAngel:
image

This also fixes the icon issues introduced within the latest djangocms-admin-style commits where icons are hidden.

PR #4614, opened 2015-10-23
Improved publishing UI

FinalAngel:
Fixed the issue when pressing **Publish changes** staying in edit mode. After pressing the button you will be redirected to the live version now. The general improvements to the workflow now is:

**Being in live mode**:
image

**Switch to edit without any changes**:
image

**Made changes through dragging**:
image

Now you can hit ``Save as Draft`` to go back to the live mode without publishing (same old behaviour) or you can ``Publish changes`` and the loop starts from the beginning. ``Save as draft`` stays until you publish your changes. ``View published`` is only showed when there are **no changes** to publish.

This also sadly includes https://github.com/divio/django-cms/pull/4613 because of my messup, will merge anyways and fix tests.


FinalAngel:
LGTM /cc @evildmp

evildmp:
It's beautiful!

PR #4615, opened 2015-10-23
Fixes for the creation wizard and bonus

FinalAngel:
This centralises the wizards javascript into one separate file which is more favourable than splitting up.

Also fixes an issue where wizard does not show up on welcome screen.

Also sneaks in a lighter gray variant for disabled button state:
image


mkoistinen:
LGTM for both the lighter button and the JS moves

mkoistinen:
LGTM!

cms/static/cms/sass/mixins/_buttons.scss
```diff
@@ -68,4 +68,7 @@
             }
         }
     }
+    &.cms-btn-disabled {
+        color: $gray-lighter;
```
vxsx: that's going to work only for default buttons, not for action / danger ones.
maybe it's better to fix line 63 in this file to be more transparent?

cms/static/cms/sass/mixins/_buttons.scss
```diff
@@ -68,4 +68,7 @@
             }
         }
     }
+    &.cms-btn-disabled {
+        color: lighten($color, 50%);
```
vxsx: @FinalAngel can you move it up to line 63? this one doesn't cover all cases and doesn't have the important function

vxsx:
:+1:

PR #4619, opened 2015-10-26
Extracts widgets javascript into "js/widgets" folder

FinalAngel:
- ``cms/forms/widgets.py`` PageSelectWidget
- ``cms/forms/widgets.py`` PageSmartLinkWidget
- ``cms/forms/widgets.py`` AppHookSelect & ApplicationConfigSelect

We have to avoid injecting javascript through python directly. This PR extracts those snippets and moves them into the ``/static/cms/js/widgets/`` folder.

The only JS injected by python are initial states for ``_PageSelectWidget`` and ``_PageSmartLinkWidget ``.

I also took up the liberty to clean up the code, ensure its consistent and included the JS linter.

Thanks to @mkoistinen for the help


mkoistinen:
LGTM

cms/forms/widgets.py
```diff
@@ -92,44 +98,32 @@ def render(self, name, value, attrs=None):
...
 
+    class Media:
+        css = {
+            'all': (
+                'cms/js/select2/select2.css',
+                'cms/js/select2/select2-bootstrap.css',
+            )
+        }
+        js = (
+            'cms/js/modules/cms.base.js',
```
vxsx: is base not already there as part of the bundle?

cms/forms/widgets.py
```diff
@@ -92,44 +98,32 @@ def render(self, name, value, attrs=None):
...
-        var page_id = $(this).find('option:selected').val();
-        handlePageChange( page_id );
-    });
-    $(function(){
-        handleSiteChange( $("#id_%(name)s_0").children(":selected").text() );
-        $("#add_id_%(name)s").hide();
-    });
-})(django.jQuery);
-</script>''' % {'name': name})
+            window._PageSelectWidget = {
```
vxsx: can the page contain two widgets of the same type? if yes, then data will be overridden by the last one. maybe it's better to provide a factory for the widgets then and initialize them directly here?

cms/static/cms/js/widgets/forms.pageselectwidget.js
```diff
@@ -0,0 +1,56 @@
...
+                group1.trigger('change');
+            }, 0);
+        }).trigger('change');
+
+        // sets the correct value
+        group1.on('change', function () {
+            tmp = $(this).find('option:selected').val();
+
+            if (tmp) {
+                group2.find('option').attr('selected', false);
```
vxsx: strictly speaking it should be `removeAttr('selected')` or `prop('selected', false)` but i don't know if it makes sense to change things now

cms/forms/widgets.py
vxsx: (even if it's not the case now, i think it would be better to make things future proof)

yakky:
LGTM

cms/forms/widgets.py
```diff
@@ -148,44 +145,15 @@ def render(self, name=None, value=None, attrs=None):
...
-            multiple: false,
-            initSelection : function (element, callback) {
-                var initialValue = element.val()
-                callback({id:initialValue, text: initialValue});
-            }
-        });
-    })
-})(CMS.$);
-</script>''' % {
+            $(function () {
```
vxsx: `CMS.$` ?

vxsx:
:+1:

PR #4622, opened 2015-10-27
Change menu behaviour for the admin

FinalAngel:
We have been struggling with the new feature to open the "Administration…" or "Users…" menu items in a separate tab. The goal of this was to reduce unnecessary the number of page loads. And the original idea was to offer a button/link in each tab to switch to the other fairly seamlessly, and without reloading the content again. This can be important when users are editing remote sites (Aldryn users) or users who are using a tablet, or worse, tablet users on Aldryn.

Unfortunately, browser manufacturers are removing any API to do this button/link convenience and we're left in a position where we either leave a link there which causes duplicate copies of your project in multiple tabs and the confusion that creates or, we remove it and some users may struggle getting back to their project.

For this case a better solution is, to return to opening the Administration views in the sideframe again but to add a "power-user" feature that if they hold the ``ctrl/cmd`` key down when invoking the "Administration…" or "Users…" menu-items, then the view will open in a separate tab (without any "return" button).

This pull request restores and adds this feature.


yakky:
:100: on keeping the admin in the sidebar!

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -130,18 +131,44 @@ var CMS = window.CMS || {};
...
+                        if (e.type === 'keyup') {
+                            cmdPressed = false;
+                        }
+
+                        if (el.attr('href') !== '' &&
+                            el.attr('href') !== '#' &&
+                            !el.parent().hasClass(disabled) &&
+                            !el.parent().hasClass(disabled)) {
+
+                            if (cmdPressed) {
```
vxsx: `e.shiftKey` and `e.ctrlKey` are also a thing

FinalAngel: added

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -130,18 +131,46 @@ var CMS = window.CMS || {};
...
+                        // we need to restore the default behaviour once a user
+                        // presses ctrl/cmd and clicks on the entry. In this
+                        // case a new tab should open. First we determine if
+                        // ctrl/cmd is pressed:
+                        if (
+                            e.keyCode === CMS.KEYS.CMD_LEFT ||
+                            e.keyCode === CMS.KEYS.CMD_RIGHT ||
+                            e.keyCode === CMS.KEYS.CMD_FIREFOX ||
+                            e.keyCode === CMS.KEYS.SHIFT ||
+                            e.keyCode === CMS.CTRL
```
vxsx: `CMS.KEYS.CTRL`

vxsx: why not `e.shiftKey` or `e.ctrlKey` ?

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -130,18 +131,46 @@ var CMS = window.CMS || {};
                     var children = 'cms-toolbar-item-navigation-children';
                     var isTouchingTopLevelMenu = false;
                     var open = false;
+                    var cmdPressed = false;
+                    var el;
```
vxsx: any reason for creating the element variable outside of the event scope?

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -130,18 +131,46 @@ var CMS = window.CMS || {};
...
+                            cmdPressed = true;
+                        }
+                        if (e.type === 'keyup') {
+                            cmdPressed = false;
+                        }
+
+                        if (el.attr('href') !== '' &&
+                            el.attr('href') !== '#' &&
+                            !el.parent().hasClass(disabled) &&
+                            !el.parent().hasClass(disabled)) {
```
vxsx: this is repeated twice

vxsx: i think before that there was two different checks

cms/static/cms/js/modules/cms.toolbar.js
FinalAngel: to be consistent with all the other key mappings

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -130,18 +131,44 @@ var CMS = window.CMS || {};
                     var children = 'cms-toolbar-item-navigation-children';
                     var isTouchingTopLevelMenu = false;
                     var open = false;
+                    var cmdPressed = false;
 
                     // remove events from first level
-                    navigation.find('a').on(that.click, function (e) {
-                        e.preventDefault();
```
vxsx: is it ok not to `preventDeafult` anymore?

vxsx:
overall LGTM provided we're going to thoroughly test it

cms/static/cms/js/modules/cms.toolbar.js
FinalAngel: yes, in this case it's already disabled by cms.navigation

vxsx: cms.navigation doesn't prevent any handlers :D

FinalAngel: than its prevented somewhere else :dancer:

PR #4623, opened 2015-10-27
Various frontend related fixes

FinalAngel:
- When closing modal don't refresh
- Sideframe can be closed by pressing "ESC"
- Fix wizard back button
- Fix wizard on installation screen
- Fix add page in page permissions
- Fix wizard style on smaller devices
- Fix wizard alignment with create button


cms/static/cms/js/modules/cms.modal.js
```diff
@@ -373,6 +373,7 @@ var CMS = window.CMS || {};
                 // add esc close event
                 this.ui.body.off('keydown.cms.close').on('keydown.cms.close', function (e) {
                     if (e.keyCode === CMS.KEYS.ESC) {
+                        that.options.onClose = null;
```
vxsx: if you press esc while creating new plugin it will not ask anymore?

mikek:
Back-end LGTM

cms/static/cms/js/modules/cms.modal.js
FinalAngel: what should it ask for? The only check happens, if you create the first plugin and thats still the case. Every close event (pressing on cancel, esc or x) cancels the reload or "onClose"

vxsx:
FE LGTM

PR #4624, opened 2015-10-28
Docs reference get_fullname instead of get_full_name

vkurup:
The AbstractBaseUser method is `get_full_name`, but the docs mention `get_fullname`.


FinalAngel:
@vkurup thank you for the contribution :+1:

vkurup:
Thanks for the quick response!

PR #4630, opened 2015-10-29
Only use "View published" button

FinalAngel:
last call on:

image


evildmp:
Big thumbs up from me: simpler and easier for users to understand.

mkoistinen:
LGTM

PR #4637, opened 2015-10-31
Update toolbar.rst

jxiewei:
Fixed spelling


yakky:
PR on master are not accepted. Mergin in support/3.1.x via #4638, already fixed in develop
Thanks for the fix

PR #4638, opened 2015-10-31
Update toolbar.rst

yakky:
Fix #4637


yakky:
Merging, and checking failure later as it's not related

PR #4641, opened 2015-11-01
Fix dates test

yakky:


cms/tests/admin.py
```diff
@@ -417,11 +431,18 @@ def test_change_dates(self):
                 })
                 self.assertEqual(response.status_code, 302)
                 draft = Page.objects.get(pk=draft.pk)
-                self.assertNotEqual(draft.publication_date.timetuple(), original_date.timetuple())
-                self.assertEqual(timezone.localtime(draft.publication_date).timetuple(), new_date.timetuple())
-                self.assertEqual(timezone.localtime(draft.publication_end_date).timetuple(), new_end_date.timetuple())
-                if original_end_date:
-                    self.assertNotEqual(draft.publication_end_date.timetuple(), original_end_date.timetuple())
+                if DJANGO_1_4:
```
FinalAngel: do we still need to test for django 1.4?

yakky: This was for 3.0 branch :)

PR #4642, opened 2015-11-01
Allow using migrations in tests

yakky:
This re-enables migrations by default during tests and makes manage.py executable
Migrations are enabled for selected test cases in travis


yakky:
@FinalAngel @mkoistinen please review

FinalAngel:
LGTM

mkoistinen:
Will slow things down even further. =( Perhaps we could enable this for only a subset of the test configs?

yakky:
@mkoistinen Sure. I'll add a modified travis configuration

yakky:
@mkoistinen revisited PR with migrations not enabled by default and only used in some tests

mkoistinen:
Looks like Py35 vs. Dj16 and Py26 vs. Dj18 need some attention =(

yakky:
@mkoistinen excluded offending combinations from matrix (unsupported matches of python/django versions)

mkoistinen:
LGTM!

PR #4645, opened 2015-11-01
Add dependency_links entry for django-formtools...

mkoistinen:
... pointing to a fork which allows installation on Django 1.6+

This is to support environments running Django 1.6.x that also are strict about dependency versions and conflict resolution.

Well commented, this can all be removed when we drop Django 1.6 support.


yakky:
@mkoistinen dependency_links are not processed by pip anymore by default and needs ``--process-dependency-links`` to be processed. The only option is to release the fork on PyPi in some way

mkoistinen:
Yes, you should note that this is already in the comments in setup.py.

yakky:
Sorry I have not noticed that

mkoistinen:
"Rebased"

PR #4650, opened 2015-11-02
Fixes an issue with the cms login rendering

FinalAngel:
before:
image

after:
image


mkoistinen:
LGTM

vxsx:
button is still off

FinalAngel:
@vxsx yep this we cannot fix unless we use vendor specific hacks. Firefox's the only one that behaves bad on buttons

Issue #4341, opened 2015-08-07
Add information to the documentation about release schedule/policy


evildmp:
Should also link to Roadmap documents on the wiki https://github.com/divio/django-cms/wiki/Roadmap, which themselves need a bit of tidying up.

See https://groups.google.com/forum/#!topic/django-cms-developers/tXfDNF5XMtk

evildmp:
See also @timgraham's proposal https://docs.google.com/document/d/1L23KhInNHc_rK5LJIJB9ZPhslGeQNaw-X9btAYD8PRY/edit

FinalAngel:
@evildmp I'd even recommend removing the Wiki and focus on the docs - but thats just my opinion :)

Issue #3625, opened 2014-11-28
"Page Type" isn't documented


mkoistinen:
Or if it is, I was unable to find it, which is another issue. =/

This should be in both the developer- and user-documentation.

PR #4651, opened 2015-11-02
Documentation adaptions

FinalAngel:
- Make installation and running of docs simpler
- Adds release policies for #4341
- Adds documentation for page types #3625


evildmp:
We have quite a mix of ``build`` and ``_build`` now. Any reason not to stick with ``build``?

FinalAngel:
@evildmp the default from sphinx is also ``_build``: https://github.com/sphinx-doc/sphinx/blob/470bac3d1c17f2728573f96c2864ab07918692d2/sphinx/quickstart.py#L1167

evildmp:
Personally I prefer ``build``, since that is what we have been using. If we do change it to ``_build``, there will be many instances where it needs to be updated, such as in ``https://github.com/divio/django-cms/blob/develop/docs/Makefile``.

docs/how_to/page_types.rst
```diff
@@ -0,0 +1,63 @@
...
+pages from predefined presets. This presets contain example plugins and content
+that will be carried over the the designated page while leaving the preset
+untouched.
+
+
+*******************
+Creating Page Types
+*******************
+
+First you need to create a new page through **Add page** from the **Page**
```
evildmp: Only buttons should be **bold**; menu items should be *italic* (fyi, fixing here anyway).

FinalAngel: thanks for clarifying @evildmp

evildmp:
Let's merge. All looks good to me.

PR #4652, opened 2015-11-02
align button correctly in firefox

vxsx:
before
screen shot 2015-11-02 at 22 57 34

after

screen shot 2015-11-02 at 22 57 15


cms/static/cms/sass/components/_toolbar.scss
```diff
@@ -408,11 +408,15 @@
         }
     }
     input[type="submit"] {
+        display: block;
```
vxsx: don't ask

FinalAngel:
LWTMWTF Firefox

PR #4654, opened 2015-11-03
adds pagetree fixes for other themes

lory87:
|BEFORE|AFTER|
|---|---|
|screen shot 2015-11-02 at 16 15 38|screen shot 2015-11-03 at 09 46 26|


cms/static/cms/sass/cms.pagetree.scss
```diff
@@ -423,6 +425,10 @@ $row-height: 41px;
         display: none;
     }
 }
+#changelist-search {
```
darioalbanesi: why ID?

ruslanmalogulko:
:+1:

cms/static/cms/sass/cms.pagetree.scss
lory87: because there is no class and I have to work with the given html.

darioalbanesi:
lgtm

vxsx:
:+1: (still wait for travis)

FinalAngel:
@vxsx didn't we wanted to use border box everwhere? :D

vxsx:
@FinalAngel we did, but not in the admin (and it's a bit late :D)

PR #4663, opened 2015-11-05
Ensure pasted plugins preserve their original order.

mkoistinen:
Fixes an issue where "pasting" plugins on the structureboard would put them in an incorrect order.


czpython:
:+1:

PR #4664, opened 2015-11-05
hopefully make TEMPLATE_INHERITANCE_LABEL translatable

SachaMPS:
The string 'Inherit the template of the nearest ancestor" doesn't appear in translations. And seems to be the same if you switch the language in the backend.

I suppose the reason is thats its a passed  as a constant and not as a string to _(...).

Maybe I just didnt get the point of a constant in this case... but nervertheless it would remain untranslatable...

Saludos,
Sacha

### DIFF
--- a/cms/constants.py
+++ b/cms/constants.py
@@ -2,7 +2,6 @@
 
 
 TEMPLATE_INHERITANCE_MAGIC = 'INHERIT'
-TEMPLATE_INHERITANCE_LABEL = 'Inherit the template of the nearest ancestor'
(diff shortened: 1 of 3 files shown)


mkoistinen:
Thanks, @SachaMPS! @yakky Can you review please? Is it safe to remove TEMPLATE_INHERITANCE_LABEL from constants and instead make the string translatable where it was used (1 "real" place, 1 "test"). Seems to be fine with (and indeed better) like this.

FinalAngel:
:+1:

PR #4671, opened 2015-11-08
Configurable options for page wizard

### LINKED ISSUE #4669: CMS Page Wizard assumes that TextPlugin exists
When filling in the "Contents" field in the CMS Page wizard, a TextPlugin is automatically created, but it may not be installed. I think we should find a more general solution

yakky:
Fixes to annoyances with current wizard setup:
* Standard ``TextPlugin`` is not guaranteed to exists
* The ``TEMPLATE_INHERITANCE_MAGIC`` template may not be the most appropriate one

Fix #4669


FinalAngel:
LGTM @mkoistinen ?

mkoistinen:
LGTM

PR #4675, opened 2015-11-09
Optimizes touch support

FinalAngel:
- Add default django search template and remove ``document.getElementById("searchbar").focus();``
- Removed edit this page note
- Enable scrollable area on action items: (green = scrollable, red = draggable)

image


vxsx:
:+1:

PR #4677, opened 2015-11-09
Tooltip on touch

vxsx:
tooltip wasn't consistently showing up in the content mode + additional fixes to make behaviour smoother


cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -277,11 +279,13 @@ var CMS = window.CMS || {};
...
-                this.ui.container.on(this.pointerOverAndOut, function (e) {
-                    e.stopPropagation();
+                this.ui.container.on(this.pointerOverAndOut + ' ' + this.touchStart, function (e) {
+                    if (e.type !== 'touchstart') {
+                        e.stopPropagation();
+                    }
                     var name = that.options.plugin_name;
                     var id = that.options.plugin_id;
-                    CMS.API.Tooltip.displayToggle(e.type === 'pointerover', e, name, id);
+                    CMS.API.Tooltip.displayToggle(e.type === 'pointerover' || e.type === 'touchstart', e, name, id);
```
FinalAngel: we should also pass an object there at some point

FinalAngel:
looks like magic to mee, you probably need to run me through

cms/static/cms/js/modules/cms.plugins.js
vxsx: yeah eventually

FinalAngel:
:+1:

PR #4682, opened 2015-11-10
Default to "true" when preventing dialogs

FinalAngel:
image

You cannot close the modal anymore when choosing "Prevent this page form creating additional dialogs." on new plugins. This PR fixes that behaviour and defaults to **true**.


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -521,7 +521,16 @@ var CMS = window.CMS || {};
...
                 var onSuccess = opts.onSuccess;
-                var question = (text) ? confirm(text) : true;
+                // a user could click on "Prevent this page form creating additional
+                // dialogs." which would prevent the modal to be closed
+                // We fallback to "true" in this case to avoid this
+                var secureConfirm = function (message) {
+                    var start = Number(new Date());
+                    var result = confirm(message);
+                    var end = Number(new Date());
+                    return (end < (start + 10) || result === true);
```
vxsx: must be `return (end < (start + 10) || result)`

vxsx: if confirm dialog is prevented - then it is false, but first condition is true - so it defaults to true.
if confirm dialog is _not_ prevented then it is false (and we need to just return it)

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -521,7 +521,16 @@ var CMS = window.CMS || {};
                 var text = opts.text || '';
                 var callback = opts.callback;
                 var onSuccess = opts.onSuccess;
-                var question = (text) ? confirm(text) : true;
+                // a user could click on "Prevent this page form creating additional
```
vxsx: typo form → from

vxsx:
:+1:

PR #4684, opened 2015-11-10
Allows locked entries to be collapsable

FinalAngel:
|BEFORE|AFTER|
|---|---|
|image|image|


cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -1129,7 +1129,9 @@ var CMS = window.CMS || {};
                 // collapsable function and save states
                 if (el.hasClass('cms-dragitem-expanded')) {
                     settings.states.splice($.inArray(id, settings.states), 1);
-                    el.removeClass('cms-dragitem-expanded').parent().find('> .cms-draggables').addClass('cms-hidden');
+                    el.removeClass('cms-dragitem-expanded').parent()
+                        .find('.cms-collapsable-container').eq(0).addClass('cms-hidden');
```
vxsx: ` .find('.cms-collapsable-container').eq(0)` → `.find('> .cms-collapsable-container')`

technically they are not the same

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -1145,7 +1147,9 @@ var CMS = window.CMS || {};
 
                 } else {
                     settings.states.push(id);
-                    el.addClass('cms-dragitem-expanded').parent().find('> .cms-draggables').removeClass('cms-hidden');
+                    el.addClass('cms-dragitem-expanded').parent()
+                        .find('.cms-collapsable-container').eq(0).removeClass('cms-hidden');
```
vxsx: same

vxsx:
don't really like this change, but codewise is ok
should still switch to a faster jquery lookup method

PR #4685, opened 2015-11-10
If success_url is None, redirect to origin page

mkoistinen:


yakky:
LGTM. Somewhat duplicated logic of PageAdmin.resolve (https://github.com/divio/django-cms/blob/develop/cms/admin/pageadmin.py#L1285)  but the serve a bit of different scopes and we can refactor them later

PR #4687, opened 2015-11-11
fixes jquery conflict with other version

lory87:

- fixes jquery conflict if other jquery version is loaded for example with another theme.
- wrapped tree_component into function to assign $ to CMS.$ and loads tree_component function globally.
- Because there were jQuery and $ mixed in tree_component I replaced them with $ to make it consistent


cms/templates/admin/cms/page/change_form.html
```diff
@@ -12,9 +12,13 @@
 
 {% block extrahead %}
 {{ block.super }}
+<!-- load nonconflict pre and post files on line 17 and 21 to fix conflict if there is another jQuery loaded -->
```
vxsx: don't use html comments, use django comments instead.
also, line numbers tend to change so no need to specify them

vxsx:
shame that github is not showing the changes in tree_component

FinalAngel:
We would have to test this extensively but makes sense to implement :+1:

vxsx:
:+1:

PR #4688, opened 2015-11-11
fixes pagination arrow in sideframe in IE

lory87:

- adds arrow right icon
- uses arrow right icon instead of arrow icon with transform: rotate(180deg); in pagination

|BEFORE|AFTER|
|---|---|
|ie10_pagination_buttons|screen shot 2015-11-11 at 14 09 08|


vxsx:
LGTM

PR #4690, opened 2015-11-11
Change migration dependency

### LINKED ISSUE #4683: Avoid use of _latest_ in migrations?
If you are migrating a site from Django 1.6 to Django 1.8 then you will want to fake the initial migrations to bridge the gap between south and django migrations. For example, for `cms` this meant faking up to `0011_auto_20150419_1006`, but not faking `0012_auto_20150607_2207`. Great so far.

In Django 1.6 the models for content types were basically at `0001_initial` and the new Django version added `0002_remove_content_type_name`. So all our own migrations only have a dependency on `0001_initial` to make sure we don't fake `0002_remove_content_type_name` (by dependency).

Unfortunately `cms.migrations.0010_migrate_use_structure` has a dependency on the `_latest_` content types migration. This means it will cause Django to fake the content types migration `0002_remove_content_type_name`. So by faking `0011_auto_20150419_1006` i implicitly fake content types `0002_remove_content_type_name`, and end up with a broken migration to Django 1.8.

I think this migration should be updated to depend on `0001_initial` rather than `0002_remove_content_type_name`, and that `_latest_` should be avoided in future.

yakky:
Fix #4683

### DIFF
--- a/cms/migrations/0010_migrate_use_structure.py
+++ b/cms/migrations/0010_migrate_use_structure.py
@@ -47,7 +47,7 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('cms', '0009_merge'),
-        ('contenttypes', '__latest__'),
+        ('contenttypes', '__first__'),
     ]
 
     operations = [


FinalAngel:
LGTM

yakky:
Closed in favor of #4725

PR #4691, opened 2015-11-11
Fix handling of --migrate option

yakky:


yakky:
@mkoistinen @FinalAngel plz review (it's not elegant, and probably we should work a bit more on app-manage integration, but we can postpone that)

FinalAngel:
@yakky just that I understand it correctly, if you would do:

``python manage.py --migrate --no-migrations`` which results in ````python manage.py --no-migrations``?

If this is correct, could we also display a warning, saying "You cannot provide --migrate when using --no-migrations"?

yakky:
Yes ``--no-migrations`` "wins". How about something like **Both --migrate and --no-migrations have been provided. --no-migrations is in effect.**

FinalAngel:
:+1: @yakky

PR #4694, opened 2015-11-12
fixes pagetree menu link and close tooltip

lory87:

- updates jQuery ui and adds effect
- closes tooltip on clicking on content

|BEFORE|AFTER|
|---|---|
|screen shot 2015-11-12 at 16 20 18|screen shot 2015-11-12 at 16 20 30|


FinalAngel:
awesome :+1:

FinalAngel:
@lory87 are you certain that styling is an issue, I see the following two errors that might also trigger that behaviour:

**on load**

``GET http://localhost:8000/static/admin/img/search.svg 404 (NOT FOUND)``

**when clicking on the menu switch button**

````
XHR finished loading: POST "http://localhost:8000/en/admin/cms/page/5/change-navigation/?language=en
Uncaught TypeError: this.effect is not a function
```

lory87:
@FinalAngel checking. This errors have nothing to do with my change. This is another issue.

FinalAngel:
@lory87 just saying that the two errors might actually cause the language icons to collapse, as some required js might not be triggered.

cms/static/cms/js/modules/cms.changelist.js
```diff
@@ -211,6 +211,13 @@
                         }
                     });
                 });
+                $('html').on('touchend', function (e) {
+                    if (!$(event.target).hasClass('unpuplushed') || !$(event.target).hasClass('published') || !$(event.target).hasClass('dirty')) {
```
FinalAngel: too long

FinalAngel:
LGTM

PR #4695, opened 2015-11-12
Bugfix/private mode fixes

vxsx:
Fixes inconsitencies in handling settings when localStorage is not available.
Also fixes some of the issues for modal / toolbar when older touch devices are not registering click events


FinalAngel:
LGTM

PR #4697, opened 2015-11-12
Remove reload magic in sideframe

FinalAngel:
This disturbs "Save and continue editing" and other mechanics natively inside an iframe.


vxsx:
not sure what was that thing doing so :+1:

PR #4701, opened 2015-11-13
Fixes for the sideframe

FinalAngel:
- Disable outer scrolling when focusing the sideframe
- Set percentage with when first loading the iframe so it works on device rotation
- Reset 80% rule when on mobile (so window appears larger)
- Separates general styles into ``_general.scss``


cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -251,6 +257,15 @@ var CMS = window.CMS || {};
                     that._addToHistory(this.contentWindow.location.href);
                 });
 
+                // prevent scrolling in underlying page
+                iframe.on('mouseenter mouseleave', function (e) {
```
FinalAngel: I'm not sure if we should use this on ``_open`` and ``_close`` might be risky, what do you think @vxsx ?

vxsx: would also prefer this on `_open` and `_close`

vxsx:
:+1:

PR #4703, opened 2015-11-13
Reload on user settings

FinalAngel:
When ``reload_window`` param is given, appropriately reload the sideframe url and parent window to apply new language settings.


vxsx:
:+1:

PR #4704, opened 2015-11-13
Custom icon for maximized stated

FinalAngel:
We show now the following icon on full screen:

image

to switch back to normal window size.


vxsx:
:+1:

PR #4705, opened 2015-11-13
Fixes sideframe action positioning

FinalAngel:
It seems like a very tiny change, basically:
- make clickable area larger
- add background so events don't fall through

Before
image
image

After
image
image


vxsx:
:+1:

PR #4707, opened 2015-11-13
Fixes message display

FinalAngel:
**Before: **
image

**After:**
image


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -454,7 +454,7 @@ var CMS = window.CMS || {};
             _show: function _show(opts) {
                 var speed = opts && opts.duration !== undefined ? opts.duration : this.options.toolbarDuration;
                 var debugHeight = $('.cms-debug-bar').height() || 0;
-                var toolbarHeight = $('.cms-toolbar').height() + 10;
```
vxsx: it was needed to speed up the performance of the rendering while dragging in structureboard, because the box-shadow that leaked from the toolbar was causing a big chunk of frames recomposing :(

vxsx:
LGTM, but fix the merge conflict

PR #4708, opened 2015-11-13
Adds admin-related JS to the base wizard class.

mkoistinen:


FinalAngel:
I actually think the issue here lies in the template rendering of the wizard. Normally django includes the django jQuery and there is probably somewhere an option to enable this. We should _not_ load our own custom jQuery there. It should be django's default before, something like:

```
django jQuery
plugin "Media"
```

and django jQuery is missing completely.

PR #4709, opened 2015-11-14
Adds admin jQuery

mkoistinen:
Some wizards use form widgets which use JS-based behavior. These will require access to jQuery from admin.


FinalAngel:
LGTM though missing a tab: https://aldryn-boilerplate-bootstrap3.readthedocs.org/en/latest/guidelines/markup.html#indentation

PR #4710, opened 2015-11-14
Make implementation reflect documentation

mkoistinen:
BACKGROUND: Since its original implementation, the corresponding documentation (http://docs.django-cms.org/en/latest/reference/plugins.html#disable-child-plugins) specifies the property: `disable_child_plugins` (plural), while the implementation was `disable_child_plugin` (singular).

Due to this, plugins have been implementing the plural of the property. Most notably djangocms-text-ckeditor, and the CMS structure-board also evaluates the plural.

Since the property's default value is the same as evaluating a non-existent property in the templates, this should be a benign change.

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -118,7 +118,8 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
     require_parent = False
     parent_classes = None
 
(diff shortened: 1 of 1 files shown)


yakky:
If I understand correctly original implementation never worked, right?
LGTM
Probably worth a mention in release notes just to clarify

mkoistinen:
Correct and agree about release notes.

PR #4715, opened 2015-11-16
Pagetree lang tooltips and prevented clicks

vxsx:
- remove preventDefault from all touchends  
that prevented every other interaction with the page on touch devices
- change how langTooltip target is determined  
previously `e.target` was used, which meant that `touchUsedNode` could actually be a span inside, which would break the hide/show tooltip functionality


lory87:
:) :+1:

FinalAngel:
LGTM

PR #4716, opened 2015-11-16
Fresh install django-cms with CMS_RAW_ID_USERS set to True throws exception

cstrap:
Environment:
* Django 1.7+
* django-cms 3.x+

How to reproduce the issue:
* install django-cms via django CMS installer and choosing
  * django and django-cms version according to environment above
  * keep the default values and the installer does all "boring things"
* Update `settings.py` with `CMS_PERMISSION = CMS_RAW_ID_USERS = True`
* delete `project.db`
* re-run `./manage.py migrate`

The result should be:
```
...
django.db.utils.OperationalError: no such table: auth_user
```

So, I provide a solution on this PR.

Maybe I'm wrong, but I didn't found other ways in order to have a fresh install that starts with `CMS_PERMISSION` and `CMS_RAW_ID_USERS` set to `True`, except to comment out the `CMS_RAW_ID_USERS = True` and uncomment the row after migrations.

Any suggestions are appreciated.
Thanks!

### DIFF
--- a/cms/admin/permissionadmin.py
+++ b/cms/admin/permissionadmin.py
@@ -4,6 +4,7 @@ from django.contrib import admin
 from django.contrib.admin import site
 from django.contrib.auth import get_user_model, get_permission_codename
 from django.contrib.auth.admin import UserAdmin
+from django.db import OperationalError
 from django.utils.translation import ugettext as _
 
 from cms.admin.forms import GlobalPagePermissionAdminForm, PagePermissionInlineAdminForm, ViewRestrictionInlineAdminForm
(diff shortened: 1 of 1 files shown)


mkoistinen:
LGTM

yakky:
@cstrap looks good. I wonder why that function is triggered during a simple migration, though.
I'd apply this  to 3.1 and forwardport to develop/3.2

yakky:
Fixed in #4723

cstrap:
You're welcome! Thanks for merging!

Btw, I spent some time in debugging and I found that the function is triggered by the checks provided by admin apps on the inlines models.
In fact, if I run `./manage.py check` the resulting exception is the same.

PR #4719, opened 2015-11-16
cms events

vxsx:
it is useful for plugin consumers to listen to specific events for their own purposes, say, if modal has been maximized, toolbar was opened or what not. this PR implements the framework for dispatching and listening to these events

also adds two events for modal for this PR: https://github.com/divio/djangocms-text-ckeditor/pull/282


FinalAngel:
LGTM

PR #4720, opened 2015-11-16
Fixes allow_children and disable_child_plugins context

FinalAngel:
Disabling the "+" add and pasting functionality should be standard when allow_children is false:

image

Additionally we also need to lock items that are ``disable_child_plugins``


mkoistinen:
Nice!

vxsx:
:+1:

Issue #4712, opened 2015-11-14
"Revert to live" not working if page title and page slug have been changed


otg-sandrey:
How to reproduce:
- create a new page
- publish it
- open the page again
- change the page title and the page slug
- save the changes
- go to "History" -> "Revert to live"

You'll get HTTP 404.

The problem can be reproduced using the live demo at www.django-cms.org

mkoistinen:
Closed by #4727 (et al.)

PR #4727, opened 2015-11-17
Fix revert to live 30

### LINKED ISSUE #4711: "Revert to live" not working for brand new pages
If you create a new page and then click "History" -> "Revert to live" you'll get HTTP 500.

The problem can be reproduced using the online demo at http://www.django-cms.org

mkoistinen:
Fixes #4711 and #4712.

This should be forward ported to 3.1.x and 3.2.x.

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -1206,10 +1206,15 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
             # if request comes from tree..
             return admin_utils.render_admin_menu_item(request, page)
 
-        referer = request.META.get('HTTP_REFERER', '')
-        path = '../../'
-        if admin_reverse('index') not in referer:
-            path = '%s?%s' % (referer.split('?')[0], get_cms_setting('CMS_TOOLBAR_URL__EDIT_OFF'))
+        try:
+            path = page.get_absolute_url(language=language)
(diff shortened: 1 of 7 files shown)


cms/admin/pageadmin.py
```diff
@@ -1206,10 +1206,15 @@ def revert_page(self, request, page_id, language):
...
 
-        referer = request.META.get('HTTP_REFERER', '')
-        path = '../../'
-        if admin_reverse('index') not in referer:
-            path = '%s?%s' % (referer.split('?')[0], get_cms_setting('CMS_TOOLBAR_URL__EDIT_OFF'))
+        try:
+            path = page.get_absolute_url(language=language)
+            path = '%s?%s' % (
+                path, get_cms_setting('CMS_TOOLBAR_URL__EDIT_OFF'))
+        except:
```
czpython: I don't really like this. Did the ```get_absolute_url``` fail at any point?

mkoistinen: Yes, in my testing, it failed during the page-creation process.

yakky: I cannot reproduce the ``get_absolute_url`` failure. How can I trigger that?

yakky:
@mkoistinen this should run fine. But let me do other tests and let me merge this when I feel confident we're not breaking things

PR #4728, opened 2015-11-17
Touch tooltip

vxsx:
if you tapped on a plugin it would stop propagating to the body to "pick" touch version of the plugin. This PR does a workaround + makes sure that we pick the tooltip only once, so there's not much of a performance overhead


cms/static/cms/js/modules/cms.tooltip.js
```diff
@@ -41,36 +41,43 @@ var CMS = window.CMS || {};
...
-                            generic.eq(0).trigger('dblclick.cms');
-                        }
-                    });
+            /**
+             * Force tooltip to be "touch".
+             *
+             * @method _forceTouch
+             * @private
+             */
+            _forceTouch: function () {
```
FinalAngel: missing _forceTouch in function

cms/static/cms/js/modules/cms.tooltip.js
```diff
@@ -41,36 +41,43 @@ var CMS = window.CMS || {};
...
             },
 
             /**
              * Checks for touch event and switches to touch tooltip if detected.
              *
-             * @method checkTouch
+             * @method _checkTouch
              * @private
              */
             _checkTouch: function () {
```
FinalAngel: missing _checkTouch in function

cms/static/cms/js/modules/cms.base.js
```diff
@@ -520,6 +520,30 @@ var CMS = {
...
+             */
+            once: function once(fn) {
+                var result;
+                var didRunOnce = false;
+
+                return function () {
+                    if (!didRunOnce) {
+                        didRunOnce = true;
+                        result = fn.apply(this, arguments);
+                    }
```
FinalAngel: why not just an else?

cms/static/cms/js/modules/cms.tooltip.js
vxsx: we're actually missing it everywhere in this file

FinalAngel:
LGTM

PR #4730, opened 2015-11-17
close info tooltip on touch devices

lory87:

- replace showing and hiding info tooltip with js function


vxsx:
codewise LGTM

PR #4733, opened 2015-11-18
truncates text in clipboard

lory87:

|BEFORE|AFTER|
|---|---|
|cms_ipad_clipboard|screen shot 2015-11-18 at 08 44 38|


cms/static/cms/sass/components/_clipboard.scss
```diff
@@ -20,6 +20,7 @@
         cursor: move;
     }
     .cms-draggable .cms-dragitem {
+        overflow: hidden;
```
vxsx: is there a need for overflow here?

lory87: yes to remove horizontal scrollbar

lory87: @vxsx was wrong with this solution it's not necessary. removed

vxsx:
:+1:

PR #4741, opened 2015-11-19
Target on anchor does not work

FinalAngel:
We prevent links to go directly to the desired page to enable "double click to edit". The current implementation always redirects on the current page and is not taking ``target="_blank"`` or other attributes into account.

This PR solves the issue so ``<a href="http://www.google.ch" target="_blank">Google</a>`` will work.


cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -47,7 +48,9 @@ var CMS = window.CMS || {};
             if (++clickCounter === 1) {
                 timer = setTimeout(function () {
                     clickCounter = 0;
-                    window.location.href = $(e.currentTarget).attr('href');
+                    anchor = $(e.currentTarget);
```
vxsx: `var` should be here. otherwise we would hold the reference to the anchor forever

vxsx:
not sure if there is no side effects when doing `window.open(url, '_self')`, but LGTM

cms/static/cms/js/modules/cms.plugins.js
FinalAngel: I'm not sure, we do the same with timer so it might be better to also cache anchor, what you think?

vxsx:
:+1:

FinalAngel:
Me neither but thats the only way to trigger the target context. Eventually its handled as a pop up and will be blocked?

vxsx:
unlikely, but if you do window.open inside of that new window it'll probably stick to that same window then

vxsx:
:+1:

PR #4744, opened 2015-11-19
Modal form submit

vxsx:
we want to keep the browser behaviour, so when user submits the form inside of an iframe by any means it triggers the click on the button, and if the button is the default one we trigger the default behaviour with the hiding frame, reloading, checking `that.saved` etc.


FinalAngel:
LGTM

PR #4746, opened 2015-11-19
add `?_popup=1` to add apphook url

vxsx:
to fix the popup not dismissing correctly in django 1.8


mkoistinen:
LGTM =)

FinalAngel:
Sweet :+1:

mkoistinen:
@vxsx Shouldn't we have logic here to use '?' or '&' depending on the url itself?

vxsx:
I don't think so, as it never has it in the apphook urls

FinalAngel:
agreed, even if, normally you have the popup param first (or last) doing evaluation here is overkill

mkoistinen:
Well, the url comes from a method which can be overridden by the developer in their own apphook. Someone may wish to include their own GET params at some point. At that point, this may break.

PR #4757, opened 2015-11-23
Trick Django to get us a login URL even in version 1.6.

intellectronica:
Django 1.6 does not provide a reversible /login URL, so
we need to rely on it providing a login handler when attemtpting
to get a protected URL without being logged in.

Credit @czpython for the fix.

### DIFF
--- a/cms/templates/cms/welcome.html
+++ b/cms/templates/cms/welcome.html
@@ -88,9 +88,11 @@
                 btn.trigger('click');
             });
         {% else %}
-            {# Django <= 1.6 doesn't provide a login URL handler. Rely on #}
-            {# it redirecting to the login page when attempting to access a #}
-            {# protected handler while not logged in. #}
+            {% comment %}
+            Django <= 1.6 doesn't provide a login URL handler. Rely on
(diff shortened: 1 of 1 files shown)


cms/templates/cms/welcome.html
```diff
@@ -88,7 +88,12 @@ <h2 class="cms-hidden">{% trans "Installation Notes" %}</h2>
                 btn.trigger('click');
             });
         {% else %}
-            window.location.href = '{% url "admin:login" %}?next=/{{ LANGUAGE_CODE }}/?{{ cms_edit_on }}';
+            {# Django <= 1.6 doesn't provide a login URL handler. Rely on #}
```
czpython: please use ```{% comment %}{% endcomment %}``` for block comments.

vxsx:
:+1:

FinalAngel:
LGTM

czpython:
Not sure why the ``?`` character is used twice on the url.. but that's been there before this pr so ok for me. :+1:

FinalAngel:
@czpython the second question mark is used for the redirect to turn edit on, all good :)

PR #4759, opened 2015-11-23
remove search_form template, adapt changelog and docs

vxsx:
since the template is used from inside of admin templatetag it doesn't
really work

https://github.com/django/django/blob/53ccffdb8c8e47a4d4304df453d8c79a9be295ab/django/contrib/admin/templatetags/admin_list.py#L407-L416


docs/topics/touch.rst
```diff
@@ -110,3 +110,10 @@ CKEditor issues
   portions of the toolbar when the CKEditor is opened - even though sections may appear to missing
   or moved, they can still be activated by touching the part of the screen where they should have
   been found.
+
+
+Django Admin issues
+-------------------
+
+* In the page tree first touch on the page would open the keyboard because
```
FinalAngel: We might want to describe that in a bit more detail

vxsx: yeah, but i'm not the best person for this :(

FinalAngel:
LGTM

mkoistinen:
I made minor tweaks for clarity in the known issues list. Otherwise LGTM.

PR #4779, opened 2015-11-27
added new screenshots for people app

darioalbanesi:


FinalAngel:
LGTM @evildmp what do you think?

docs/user/tutorial/40-people.rst
```diff
@@ -49,7 +49,11 @@ To create a new Person:
         Description
             Freda oversees all technical activities in our workshop.
 
-    .. todo:: screenshot of above
+.. image:: /user/tutorial/images/create_new_person.png
```
evildmp: Is it possible to make the image less high, so we can make it wider (and the text easier to see) without taking up too much space?

docs/user/tutorial/40-people.rst
```diff
@@ -80,11 +84,40 @@ automatically.
...
-
 .. todo:: write up steps properly
 
-* create new page called People
-* add Apphook
-* publish page
-* view page with automatic list of people in it
+1. Create new page called People
+
+.. image:: /user/tutorial/images/create_people_page.png
```
evildmp: Same note as above: *if* it is possible to make these images wider and less tall, we should.

docs/user/tutorial/40-people.rst
```diff
@@ -80,11 +84,40 @@ automatically.
...
+   :alt: Create new person
```
evildmp: Don't forget the ``alt`` text - ``Create a new 'People' page``

docs/user/tutorial/40-people.rst
```diff
@@ -80,11 +84,40 @@ automatically.
...
+   :align: center
+
+2. Select advanced settings
+
+.. image:: /user/tutorial/images/select_advanced_settings.png
+   :alt: Create new person
+   :width: 150
+   :align: center
+
+3. Add Apphook
```
evildmp: Can you confirm that the settings here are all default settings? Are there any that need to be specified manually by the user?

evildmp:
@darioalbanesi Thanks, I left a few notes, but this fundamentally looks good to merge once you have checked those. 

* [ ] Let me know about the images
* [ ] Add the alt text as appropriate
* [ ] confirm whether settings in apphook are automatic or need to be manually supplied

docs/user/tutorial/40-people.rst
darioalbanesi: sure it is, but that would mean that we have to cut the images a bit more. Which can have the effect that the user doesn't understand where that step happens because of the lack of information on the screenshot. And please consider that scrolling is a natural concept of getting information while clicking to enlarge an image that is too small can also result in bad ergonomics.

docs/user/tutorial/40-people.rst
darioalbanesi: same as above

docs/user/tutorial/40-people.rst
darioalbanesi: as far as I know they are automatically set. To be 100% sure I would like to ask @ruslanmalogulko because he setup the page. @ruslanmalogulko can you confirm that the app-hook settings are automatically set? thank you

ruslanmalogulko: as far as i've seen, if there is only one app-hook added to the news, it's set up automatically. if several, additional selectbox appear to choose proper app-hook for current news blog.

docs/user/tutorial/40-people.rst
evildmp: I meant not by cropping important information, but by resizing the CKEditor and dialog box etc before making the screenshot - possible? Sometimes these elements have minimum heights so it's not always possible.

darioalbanesi: min width is 800px

docs/user/tutorial/40-people.rst
darioalbanesi: thank you

docs/user/tutorial/40-people.rst
evildmp: Height, not width!

PR #4792, opened 2015-11-30
added screenshots

genti94:
added images to the django admin page user tutorial


docs/user/tutorial/60-admin.rst
```diff
@@ -18,23 +18,25 @@ Using the Filer in the Django Admin
...
    :alt: The zoom button
    :align: right
-   :width: 10%
+   :width: 50%
 
 This opens the django CMS *overlay*, which provides a quick view onto some extra control
 functionality.
 
-.. todo:: screenshot of the overlay
+.. image:: images/zoom_panel.png
```
evildmp: We seem to have zoom_panel.png in here twice.

evildmp: There is a lot of white space in the image of the overlay - would it be possible to make the whole browser window narrower, and re-take the image at the minimum width possible so its contents are clearer?

docs/user/tutorial/60-admin.rst
```diff
@@ -45,10 +47,29 @@ Add an image to the Filer
...
+
 * open folder
+
+.. image:: images/open_folder.png
+   :alt: open the new folder
+   :width: 80%
+
 * hit upload button
+
+.. image:: images/upload_filer.png
```
evildmp: Can we also have an image of the upload button itself, so it can be added inline in the text please?

docs/user/tutorial/60-admin.rst
```diff
@@ -45,10 +47,29 @@ Add an image to the Filer
 .. todo:: write up steps properly
 
 * hit New folder in Filer, create folder called "People"
```
evildmp: Can we also have an image of the New Folder button, to be used inline?

docs/user/tutorial/60-admin.rst
```diff
@@ -45,10 +47,29 @@ Add an image to the Filer
...
+
 * hit upload button
+
+.. image:: images/upload_filer.png
+   :alt: hit the upload button
+   :width: 80%
+
 * hit Paste button
 
+.. image:: images/image_paste.png
```
evildmp: Can we also have an image of the Paste button, to be used inline?

docs/user/tutorial/60-admin.rst
```diff
@@ -66,9 +87,20 @@ Moving items
 .. todo:: write up steps properly
 
 * move file to clipboard
+
+.. image:: images/move_image.png
```
evildmp: Image of the Cut icon please for inline use.

docs/user/tutorial/60-admin.rst
```diff
@@ -81,10 +113,24 @@ Editing items
...
+.. image:: images/image_description.png
+   :alt: Shows image information
+   :width: 80%
+
 * Choose the image's *focus point* - the point around which any cropping will occur - by dragging
   the red circle in the image pane
 
+.. image:: images/image_focus.png
+   :alt: red dot on image
+   :width: 80%
```
evildmp: The ``image`` directives like this one that you have added document are part of the list items, therefore they should be indented by the same amount as the list content.

In the case of this one, the indentation of the following paragraph means it raises an error. It should look like:

```
*   Choose the image's *focus point* - the point around which any cropping will occur - by dragging
    the red circle in the image pane

    .. image:: images/image_focus.png
       :alt: red dot on image
       :width: 80%

    This helps ensure for example that however a portrait is cropped, the subject's head will not be
    chopped off. If your image doesn't have a particular focus point, leave the circle in the middle
    of the image.
```

docs/user/tutorial/60-admin.rst
```diff
@@ -103,7 +149,17 @@ Using images from the Filer
...
 
 * close the admin pane
+
+.. image:: images/close_admin.png
+   :alt: close icon of the admin page
+   :width: 80%
+
 * go to People page
+
+.. image:: images/people_page.png
```
evildmp: This is the wrong image - we need the People page in the site, not the admin.

evildmp:
Hi @genti94 thanks, have added some notes.

One thing we must decide is whether/how to highlight items (i.e. where you have used the arrows). We should have a quick chat with @darioalbanesi. 

My preference is that we should use them only where strictly necessary, and where possible, carefully take images that mean they are not necessary.

darioalbanesi:
@evildmp 
Spoke to @genti94 and we agreed on taking of the red arrows and stick to the style we used before. If something has to be highlighted we should use the hover effect or a zoomed version of the button/switch etc.. If this is not possible, we can still use the mouse-pointer to highlight elements.

@genti94 please exchange the images as discussed and ping Daniele to review.

thanks!

docs/user/tutorial/60-admin.rst
```diff
@@ -44,12 +40,33 @@ Add an image to the Filer
...
+
+  .. image:: images/new_folder_open.png
+   :alt: The new folder button
+   :width: 80%
+
 * open folder
+
+.. image:: images/open_folder.png
+   :alt: open the new folder
+   :width: 40%
```
evildmp: Please always use exact sizes for buttons, not % (just for future reference).

docs/user/tutorial/60-admin.rst
```diff
@@ -44,12 +40,33 @@ Add an image to the Filer
 
 .. todo:: write up steps properly
 
+.. image:: images/new_folder_button.png
+   :alt: The new folder button
+   :width: 20%
```
evildmp: Please use actual pixel sizes for buttons, not % (just for future reference, I will change this).

docs/user/tutorial/60-admin.rst
```diff
@@ -44,12 +40,33 @@ Add an image to the Filer
 
 .. todo:: write up steps properly
 
+.. image:: images/new_folder_button.png
```
evildmp: Please use Screenshot Node (in Firefox) or similar in Chrome so you don't get whitespace around buttons. Can you retake this image please?

evildmp: Also if you can, please take screenshots on a retina display for better resolution.

docs/user/tutorial/60-admin.rst
```diff
@@ -44,12 +40,33 @@ Add an image to the Filer
...
 * open folder
+
+.. image:: images/open_folder.png
+   :alt: open the new folder
+   :width: 40%
+
 * hit upload button
-* hit Paste button
 
+.. image:: images/upload_button.png
```
evildmp: New image with no whitespace please.

evildmp:
@genti94 Hi, thanks for this, I have merged, but have left some comments on the PR, and new todos on the page itself. Please take a look!

Issue #4767, opened 2015-11-25
Docs » Tutorials » Plugins Typo


bartmika:
**How to replicate:**
(1) Go to: http://docs.django-cms.org/en/latest/introduction/plugins.html
(2) Scroll down to this code:
  ```
  from cms.plugin_base import CMSPluginBase
  from cms.plugin_pool import plugin_pool
  from polls_plugin.models import PollPluginModel
  from django.utils.translation import ugettext as _


  class PollPluginPublisher(CMSPluginBase):
      model = PollPlugin  # model where plugin data are saved
      module = _("Polls")
      name = _("Poll Plugin")  # name of the plugin in the interface
      render_template = "djangocms_polls/poll_plugin.html"

  def render(self, context, instance, placeholder):
    context.update({'instance': instance})
    return context

plugin_pool.register_plugin(PollPluginPublisher)  # register the plugin
  ```

(3) The typo exists on this line:
  ```
  model = PollPlugin  # model where plugin data are saved
  ```

**How to Fix:**
Please change the line to:
  ```
  model = PollPluginModel  # model where plugin data are saved
  ```

FinalAngel:
Fixed in https://github.com/divio/django-cms/pull/4794

FinalAngel:
closed as already merged in https://github.com/divio/django-cms/pull/4778/files

PR #4794, opened 2015-11-30
Fixes issue #4767

FinalAngel:
typo on ``model = PollPlugin``


evildmp:
@FinalAngel thanks but isn't this already merged? https://github.com/divio/django-cms/pull/4778/files

PR #4796, opened 2015-11-30
adds template tag to render dictionary of i18n urls

### LINKED ISSUE #4797: Side Frame contents don't change
Steps to reproduce

1. Login to /?edit
2. Click example.com > "User settings"
3. Change user settings to another one
4. Click "Save"

The steps above will show that the content is not changing in the side frame after changing the language.
This is because the javascript code from #4703 does not know what the url for the selected language is.

czpython:
The purpose of this template tag is to expose a dictionary of language codes that maps to the current url per language. This allows the frontend code to easily redirect to the corresponding url without having to have logic on the template or javascript.

This addresses an issue where changing the user language in settings does not update the sideframe.

Needs frontend work from @vxsx 

Fixes #4797

### DIFF
--- a/cms/templatetags/cms_admin.py
+++ b/cms/templatetags/cms_admin.py
@@ -6,6 +6,7 @@ from cms.constants import PUBLISHER_STATE_PENDING
 from cms.utils import get_cms_setting
 from cms.utils.admin import get_admin_menu_item_context
 from cms.utils.permissions import get_any_page_view_permissions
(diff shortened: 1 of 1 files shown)


mkoistinen:
:+1:

vxsx:
turns out the problem was a bit different #4799

czpython:
Closing since the issue was more in the javascript side.

PR #4799, opened 2015-12-01
Fix usersettings reload

vxsx:
https://github.com/divio/django-cms/pull/4796 our original thought was that the language part of the url
wasn't correctly replaced, but it turns out that was not the case

when the page is loaded in sideframe we want to override the url of the
sideframe and reload the parent so the change in language is reflected
in the UI. however there's also a load handler for the sideframe to save
the url in the settings so that after you reload - correct frame is
restored. problem is, this load event happens after we try to modify the
sideframe url manually, so we want ours to trigger last.

i'm not very happy with the solution, but we definitely need both behaviours

also, we do not actually update the url in settings correctly, but that's going to be a different pull request

refs #4797

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,9 @@
+=== 3.2.1 (Unreleased) ===
+
+- Fixed an issue with refreshing the UI when switching CMS language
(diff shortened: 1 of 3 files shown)


FinalAngel:
LGTM > changelog

PR #4800, opened 2015-12-01
change the way iframe url is retrieved for saving

vxsx:
iframe src never changes, so we need to check window location instead

videos!
before: http://d.pr/v/fPq4
after: http://d.pr/v/1d9g2


FinalAngel:
LGTM

vxsx:
changelog is in #4799

PR #4803, opened 2015-12-01
Fixes/reload plugin urls

czpython:
* Makes plugin action urls come from a consistent location
* Adds plugin action urls to json response on plugin move


czpython:
@FinalAngel Please have a look.

mkoistinen:
LGTM. However, we need a test for this. Can be in a future PR though, let's just not forget.

yakky:
Is this for 3.2?

mkoistinen:
Yes, this is a bug-fix.

czpython:
The real fix will be done by frontend.

yakky:
This requires editing the documentation here https://github.com/divio/django-cms/blob/develop/docs/reference/plugins.rst#add_url

czpython:
@yakky how so? no change has been made to those properties. Do you mean to add the get_ methods as public api?

FinalAngel:
@czpython we probably want to add the frontend fix in here. Let's see if I can do that today quickly. Please keep open

yakky:
@czpython exactly. We can even deprecate the attributes (which were actually born as properties) and make the new get_ methods official

FinalAngel:
@czpython I added the JS logic for this. There are some inconsistency problems though. The current urls scheme requires urls to be:

```json
{
    add_plugin: "/en/admin/cms/page/add-plugin/", 
    edit_plugin: "/en/admin/cms/page/edit-plugin/76/", 
    move_plugin: "/en/admin/cms/page/move-plugin/", 
    copy_plugin: "/en/admin/cms/page/copy-plugins/", 
   delete_plugin: "/en/admin/cms/page/delete-plugin/76/"}
```

but your are suffixed with ``_url``. This causes a mashup which doesn't work.

I added console.log so you can debug it as well.

FinalAngel:
please see the PR here @czpython https://github.com/czpython/django-cms/pull/1

czpython:
@yakky if it's ok with you I'd like to tackle the deprecation of the properties in a separate pr along with the documentation changes.

The plan would be that each get_action_url() method checks if a value comes back from the property, if so then each method is responsible for raising the deprecation warning for that property.
Ok for you?

yakky:
@czpython :+1:

PR #4805, opened 2015-12-02
only open welcome modal if sideframe is not open

vxsx:
if you close the modal and open a sideframe and then for whatever reason you refresh the page - the modal is opened on top of the sideframe which is not something that should happen


FinalAngel:
nice LGTM

PR #4806, opened 2015-12-02
added new images and step by step description

darioalbanesi:
fullscreen3


evildmp:
@darioalbanesi thanks will merge after writing up my part.

docs/user/tutorial/70-page-management.rst
```diff
@@ -4,7 +4,10 @@ Managing pages
 
 Our menu of pages is growing.
 
-.. todo:: screenshots of menu
+|page_tree|
```
evildmp: By the way you don't need to do this - it's only necessary to create a ``|substitution|`` when you want an inline image.

For an image as a block, just do:

```
.. image:: /path/image.url
    etc etc
```

docs/user/tutorial/70-page-management.rst
```diff
@@ -22,9 +25,40 @@ it's::
...
+
+    .. image:: /user/tutorial/images/name_of_parent_page.png
+       :alt: Enter title of parent page
+       :width: 100%
+
+#. Hit **save** to create the page. |save_button|
+
+    .. |save_button| image:: /user/tutorial/images/save_button.png
+       :alt: 'Hit save button'
+       :width: 60px
```
evildmp: You don't need to specify the width of inline images such as buttons or icons - unless you have taken them on a Retina screen, in which case make them 50% of the actual size (in pixels).

docs/user/tutorial/70-page-management.rst
darioalbanesi: the thing is, if I want to align the image to the left, it floats and the text that is below, will move up and stick to the right edge of the image.
Using the above method will keep the space on the right side empty.

evildmp:
Thanks @darioalbanesi for the additional changes; merged.

PR #4810, opened 2015-12-03
use `addPluginHelpTitle` from options instead of dom node

vxsx:
sometimes it might be that the plugin doesn't have a dom representation
and that line would throw an error then


FinalAngel:
LGTM

PR #4817, opened 2015-12-06
Placeholder name in tooltip

vxsx:
before:
screen shot 2015-12-06 at 13 04 33

after: 
screen shot 2015-12-06 at 13 04 03


FinalAngel:
very nice :+1:

PR #4823, opened 2015-12-07
Add slug to page wizards

mkoistinen:
Non-latin character-set languages won't produce non-null slugs in the CMS because of this, the page and sub-page wizards produce a 500 error when used with these languages.

This patch provides an optional slug-field that, if left empty will be filled with the automatically-created slug as usual, but in the event the CMS cannot produce a non-null slug, allows the user to provide one.


cms/forms/wizards.py
```diff
@@ -78,10 +79,15 @@ class Media:
...
 
 class BaseCMSPageForm(forms.Form):
-    title = forms.CharField(label=_(u'Title'), max_length=255,
-                            help_text=_(u"Provide a title for the new page."))
-    page_type = forms.ChoiceField(label=_(u'Page type'), required=False,
-                                  widget=PageTypeSelect())
+    title = forms.CharField(
+        label=_(u'Title'), max_length=255,
+        help_text=_(u"Provide a title for the new page."))
+    slug = forms.CharField(
```
evildmp: Would it be possible to have such fields in an "Advanced options" fieldset? Slugs are definitely non-obvious to 99% of new users.

PR #4824, opened 2015-12-07
Add slug field to page wizards

mkoistinen:
Non-latin character-set languages won't produce non-null slugs in the CMS because of this, the page and sub-page wizards produce a 500 error when used with these languages.

This patch provides an optional slug-field that, if left empty will be filled with the automatically-created slug as usual, but in the event the CMS cannot produce a non-null slug, allows the user to provide one.

### DIFF
--- a/cms/api.py
+++ b/cms/api.py
@@ -41,8 +41,7 @@ from menus.menu_pool import menu_pool
 # Helpers/Internals
 #===============================================================================
 
-
-def _generate_valid_slug(source, parent, language):
+def generate_valid_slug(source, parent, language):
     """
     Generate a valid slug for a page from source for the given language.
     Parent is passed so we can make sure the slug is unique for this level in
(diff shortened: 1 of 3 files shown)


yakky:
LGTM
Next step is fixing the slug generation :)

yakky:
sorry, closed by mistake

mkoistinen:
Agreed.

yakky:
@mkoistinen it will require moving from plain ``slugify`` to some more general tool, I think. It's more a 3.3/3.4 thing I guess

mkoistinen:
Agreed. I'm addressing the more immediate issue here, but I think it would be great to get a transliterating slugify in sooner than later.

yakky:
@mkoistinen go ahead for the quickfix, we will find a better solution in the next versions

PR #4831, opened 2015-12-08
Use rich text editors for page wizards

mkoistinen:


cms/forms/wizards.py
```diff
@@ -17,6 +17,8 @@
 
 from cms.utils.conf import get_cms_setting
 
+from djangocms_text_ckeditor.widgets import TextEditorWidget
```
czpython: here need to do try/except

czpython:
:+1:

Issue #4798, opened 2015-12-01
Column does not exist in tests with multiple databases


igorkud90:
I use multiple databases in my project: master and log_db. When i updated to django-cms==3.1.4 and started the tests, i've got error in 0005_auto_20140924_1039.py migration, when creating test log database:
```
Creating test database for alias 'default'...
Creating test database for alias 'log_db'...
Traceback (most recent call last):
  File "/home/user/Downloads/pycharm-5.0/helpers/pycharm/django_test_manage.py", line 129, in <module>
    utility.execute()
  File "/home/user/Downloads/pycharm-5.0/helpers/pycharm/django_test_manage.py", line 104, in execute
    PycharmTestCommand().run_from_argv(self.argv)
  File "/home/user/project-env/project/local/lib/python2.7/site-packages/django/core/management/commands/test.py", line 50, in run_from_argv
    super(Command, self).run_from_argv(argv)
  File "/home/user/project-env/project/local/lib/python2.7/site-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "/home/user/project-env/project/local/lib/python2.7/site-packages/django/core/management/commands/test.py", line 71, in execute
    super(Command, self).execute(*args, **options)
  File "/home/user/project-env/project/local/lib/python2.7/site-packages/django/core/management/base.py", line 338, in execute
    output = self.handle(*args, **options)
  File "/home/user/Downloads/pycharm-5.0/helpers/pycharm/django_test_manage.py", line 91, in handle
    failures = TestRunner(test_labels, verbosity=verbosity, interactive=interactive, failfast=failfast, keepdb='--keepdb' in sys.argv)
  File "/home/user/Downloads/pycharm-5.0/helpers/pycharm/django_test_runner.py", line 256, in run_tests
    extra_tests=extra_tests, **options)
  File "/home/user/Downloads/pycharm-5.0/helpers/pycharm/django_test_runner.py", line 156, in run_tests
    return super(DjangoTeamcityTestRunner, self).run_tests(test_labels, extra_tests, **kwargs)
  File "/home/user/project-env/project/local/lib/python2.7/site-packages/django/test/runner.py", line 147, in run_tests
    old_config = self.setup_databases()
  File "/home/user/project-env/project/local/lib/python2.7/site-packages/django/test/runner.py", line 109, in setup_databases
    return setup_databases(self.verbosity, self.interactive, **kwargs)
[...]

PR #4833, opened 2015-12-09
fix migration for multiple databases

igorkud90:
fix for https://github.com/divio/django-cms/issues/4798

### DIFF
--- a/cms/migrations/0005_auto_20140924_1039.py
+++ b/cms/migrations/0005_auto_20140924_1039.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 
 from django.db import migrations
 from django.db.models import F
+from django.conf import settings
 
 from treebeard.numconv import NumConv
 
@@ -126,6 +127,9 @@ class MP_AddChildHandler(MP_AddHandler):
 
 
 def move_to_mp(apps, schema_editor):
+    if not schema_editor.connection.alias == getattr(settings, 'CMS_DATABASE_NAME', 'default'):
+        return
+
     Page = apps.get_model("cms", "Page")
     CMSPlugin = apps.get_model("cms", "CMSPlugin")
     pages = Page.objects.all().order_by('tree_id', 'level', 'lft')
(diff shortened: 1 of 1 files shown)


yakky:
I'm -1 on this, as this should be handled by the database router in the project

igorkud90:
yakky, how should i fix my router? Code of my router in https://github.com/divio/django-cms/issues/4798.

clifff:
@yakky: The Django documentation specifies [this is the responsibility of the `RunPython` code](https://docs.djangoproject.com/en/1.9/howto/writing-migrations/#data-migrations-and-multiple-databases), not the router. Accordingly, the `RunPython` operation doesn't check the `allow_migrate` method, so the Router cannot influence this.

I'm running into this problem in my application too, would love to see this merged. :+1:

stefanw:
I ran into this and it turned out my database router still had the old pre-1.8 signature which then doesn't handle `RunPython` operations.

Compare the API for [`allow_migrate` in Django 1.7](https://docs.djangoproject.com/en/1.7/topics/db/multi-db/#allow_migrate) and [`allow_migrate` in Django 1.8](https://docs.djangoproject.com/en/1.8/topics/db/multi-db/#allow_migrate)

This is the [`RunPython` code in Django 1.7](https://github.com/django/django/blob/1.7/django/db/migrations/operations/special.py#L112-L117) and [in 1.8](https://github.com/django/django/blob/1.8/django/db/migrations/operations/special.py#L175-L181).

If you use the old signature, your database router will [not be called for `RunPython` operations](https://github.com/django/django/blob/1.8/django/db/utils.py#L337-L345).

In conclusion, starting in 1.8 the database router is able to decide on `RunPython` operations and the decision to run on the current db connection does not have to be in the migration.

clifff:
@stefanfoulis Thanks for the heads up! I was using the old signature style and this fixed my problem.

yakky:
@stefanw thanks for the information. Closing this PR, then. It looks a much better approach

PR #4802, opened 2015-12-01
Obliviate djangocms-flash

mkoistinen:
We wish to no longer propagate the djangocms-flash plugin. This PR removes it from code, management and documentation. (backward-facing documentation, still retains references).

PR #4834, opened 2015-12-09
Remove remaining flash traces

FinalAngel:
Goes along with https://github.com/divio/django-cms/pull/4802


vxsx:
:+1: (though missing changelog) 
not sure if can be 3.2.x?

FinalAngel:
#4802 also went into develop

PR #4838, opened 2015-12-09
050 - added fullscreen image

darioalbanesi:


docs/user/tutorial/50-news.rst
```diff
@@ -129,5 +129,5 @@ home page too. So let's do that.
 The news articles will now be listed in your home page too, and this list will be always be kept
 up-to-date with a selection of latest news items.
 
-    .. image:: /user/tutorial/images/home-page-news-articles.png
-        :alt: 'page settings button'
+    .. image:: /user/tutorial/images/news_entire_page.png
+        :alt: 'Entire news page'
```
evildmp: ``the News page`` is enough.

I only use ``'`` when for example I say: ``the 'Self-destruct' button`` (where that is the exact name of something, like a button or field). Also it has to be single quotes otherwise it interferes with the RTD HTML.

PR #4849, opened 2015-12-13
Feature Suggestion

armysick:
Feature_explanation.txt describes it.


yakky:
I don't think a PR is the best way to propose a feature. Please post this in django-cms-developer group https://groups.google.com/forum/#!forum/django-cms-developers

PR #4851, opened 2015-12-13
Fix redirect when disabling toolbar

### LINKED ISSUE #4850: Confusing redirect-page behavior when logged in with CMS toolbar hidden
Seen with Django CMS 3.1.3:

* Fill in Redirect on Advanced Settings for a page
* Save/publish
* View page with CMS toolbar enabled
* See "This page has no preview!/It is being redirected to: ``<target_URL>``"
* Disable toolbar
* See CMS page template rendered with empty content.

It should either display the "This page has no preview!" message or it should redirect.

yakky:
Fix #4850


FinalAngel:
LGTM

PR #4854, opened 2015-12-15
Don't assume toolbar is present

### LINKED ISSUE #4537: CMS_TOOLBAR_HIDE causes 'WSGIRequest' object has no attribute 'toolbar'
When `CMS_TOOLBAR_HIDE = True` then saving plugin causes `'WSGIRequest' object has no attribute 'toolbar'` `AttributeError` in `pageadmin.py`.

In `pageadmin.py`this:

```py
if not request.toolbar
```

should probably be replaced with:

```py
if not getattr(request, 'toolbar', False)
```

mkoistinen:
Fixes #4537 for me.

### DIFF
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -922,7 +922,8 @@ CMS_TOOLBAR_HIDE
 default
     ``False``
 
-If True, the toolbar is hidden in the pages out django CMS.
+If True, the toolbar is hidden, even on CMS pages.
+
 To determine the internal url of django cms, you need to assign CMS_APP_NAME to use when you include ``'cms.urls'``
 
 Example::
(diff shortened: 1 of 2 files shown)


docs/reference/configuration.rst
```diff
@@ -922,7 +922,8 @@ CMS_TOOLBAR_HIDE
 default
     ``False``
 
-If True, the toolbar is hidden in the pages out django CMS.
+If True, the toolbar is hidden, even on CMS pages.
```
evildmp: Use double backticks around literals - ``True``

evildmp:
Shouldn't this include a regression test?

mkoistinen:
Well, we have tests for the toolbar working in general for both states of the setting `CMS_TOOLBAR_HIDE`, and we have tests for creating and saving plugins. All of which pass, so, I think we're ok.

yakky:
:+1:

PR #4546, opened 2015-10-05
Fix pagewidget

yakky:
Merge #4522 in support 3.0 with fixes
Fix #4522
Fix #4334
Fix #4443

PR #4858, opened 2015-12-17
Transfer changes of PR #4546 into support/3.1.x

leture:
In the referenced issues it was said that the change will be ported forward to 3.1.x but the bug is still in the in between published release 3.1.4.


yakky:
Sorry @leture my mistake.
I merged in 3.0 and porting to 3.1 through #4888

yakky:
Fixed in #4888

leture:
Thank you @yakky

PR #4865, opened 2015-12-22
Do not call transaction.atomic when used as a decorator

stephenpaulger:
Without this change I kept getting errors saying 

    TypeError: 'GeneratorContextManager' object is not callable

https://docs.djangoproject.com/en/1.8/topics/db/transactions/#controlling-transactions-explicitly

This is in line with the Django documentation. The test suite outcome is the same with and without this change.


yakky:
Merging this via #4918

Thanks @stephenpaulger for proposing this

PR #4866, opened 2015-12-23
Feature/deprecate plugin url properties

czpython:
* Deprecates the url properties on plugins
* Updates plugin documentation

This is a continuation of #4803


mkoistinen:
LGTM

PR #4868, opened 2015-12-26
Add aldryn_common to installed apps to avoid error when adding people…

kburts:
…: https://github.com/aldryn/aldryn-common/issues/14


FinalAngel:
This makes absolutely sense. Thank you @kburts for spotting this. Good to go @mkoistinen or @yakky

yakky:
thanks @kburts

PR #4869, opened 2015-12-26
Fixed broken imports

kburts:
cms_menus.py should be in polls_plugin/ not in env/src/... right? Or at least that's how I got it to work and editing the parent package as little as possible seems like a good thing.


mkoistinen:
Yes, thank you!

Issue #4862, opened 2015-12-21
Initiation of foundation() breaks "Add plugin" button


Ichag:
when including foundation 6 in your project the initiation of foundations 6 javascript will break the functionality of adding new content plugins to a page. the dialog to select a content element won't show up.

```javascript
<script src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/modernizr/2.8.3/modernizr.min.js"></script>
<script src="https://cdn.jsdelivr.net/foundation/6.0.5/foundation.min.js"></script>
<script>
    $(document).foundation();
</script>
```
to reproduce this issue: kickstart a new cms 3.2 version and include the foundation 6 javascript. 
after this you can't add a new content plugin in structure editing mode.

leesolway:
I can confirm this as i am having the same issue. 

Foundation Version:     "foundation-sites": "~6.0.6",

creimers:
I am experiencing a similar problem. Updating from djangoCMS 3.1 to 3.2, `$(document).foundation()` crashes Firefox and IE (both latest) in edit mode (script not responding). Chrome and Safari are fine. I'm using foundation 5.5.3.

vxsx:
@Ichag @leesolway would you mind trying to reproduce on `release/3.2.x` branch?

leesolway:
Tested with the 3.2.x release branch -- seems to be working

PR #4870, opened 2015-12-28
Foundation clash over data-tooltip

vxsx:
apparently [foundation](http://foundation.zurb.com/) initialises tooltips when it sees the `data-tooltip` attribute. This PR attempts to fix it by using more specific attribute for CMS css tooltips

additionally fixes future compatibility with libsass

should fix https://github.com/divio/django-cms/issues/4862


Brankub:
:+1:

Flight:
LGTM

PR #4871, opened 2015-12-28
More resilient apphooks

mkoistinen:
This PR allows the CMS to actually start even when there's an apphook that is configured on a page, but is no longer installed. A warning is emitted, but otherwise, the CMS will start and even the apphooked page can be viewed.

### DIFF
--- a/cms/appresolver.py
+++ b/cms/appresolver.py
@@ -17,7 +17,6 @@ except ImportError:
     from django.utils.importlib import import_module
 
 from django.conf import settings
-from django.conf.urls import patterns
 from django.contrib.sites.models import Site
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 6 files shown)


stefanfoulis:
:+1:

cms/appresolver.py
```diff
@@ -223,18 +237,22 @@ def get_app_patterns():
         if title.page_id not in hooked_applications:
             hooked_applications[title.page_id] = {}
         app = apphook_pool.get_apphook(title.page.application_urls)
+        if not app:
```
czpython: is this intentional? doing this would mean that one missing apphook interrupts the rest no?

mkoistinen: Ooh, good catch.

czpython:
:+1:

yakky:
@mkoistinen what's the status here?

mkoistinen:
Rebased

Issue #4808, opened 2015-12-02
Browser caches old cms.base.css when upgrading to 3.2


jsma:
Recently upgraded a site to 3.2 but staff were complaining that the admin toolbar is broken. This was due to their browser using the old `cms.bas.css` from cache since the url didn't change. Don't have time at the moment to look into this myself but is it possible to have the new build tools generate this file with the version in the file name e.g. `cms.base.3.2.css` so it forces the browser to download the latest?

FinalAngel:
@jsma thanks for the report, we are thinking of something like ``cms.base.css?randomtoken``. Just changing the filename causes reference problem inside different places.

vxsx:
well for css we actually could use different file names but since we cannot do that for js files (because they are used in external addons) i think we'll do that consistently and use a query param (even though it's not the best solution)

mkoistinen:
Please rename the file, adding a param is not the way to go here.

FinalAngel:
@mkoistinen we can't rename the file as it is referenced across multiple places.

mkoistinen:
Fixed by #4874.

PR #4872, opened 2015-12-28
add query string to static assets with cms version

vxsx:
in order to bust cache every time new release is made

PR #4874, opened 2015-12-28
add query string to static assets with cms version

vxsx:
in order to bust cache every time new release is made

reopen from #4872 
should fix https://github.com/divio/django-cms/issues/4808


mkoistinen:
We need to come up with a better solution for managing our static assets then query-params, but this LGTM otherwise.

PR #4875, opened 2015-12-28
Adds setting CMS_INTERNAL_IPS

mkoistinen:
Implements a means of restricting the toolbar to specific IP address or ranges (if iptools is used).

Includes tests and docs.


cms/utils/setup.py
```diff
@@ -38,6 +42,10 @@ def validate_settings():
                 'django.template.context_processors.request' not in context_processors):
             raise ImproperlyConfigured("django CMS requires django.template.context_processors.request in "
                                        "'django.template.backends.django.DjangoTemplates' context processors.")
+    # Warn on empty CMS_INTERNAL_IPS
+    if not get_cms_setting('INTERNAL_IPS'):
+        warn('The setting `CMS_INTERNAL_IPS` was not defined or is empty, this '
```
czpython: is this really necessary? I think it can get very annoying as usually this level of configuration is not needed. Also systems where ips are dynamically assigned can't really use this feature so I'd say we don't need to warn about it.
Documentation should be enough.

cms/middleware/toolbar.py
```diff
@@ -104,6 +137,9 @@ def process_request(self, request):
         else:
             request.session['cms_build'] = False
             request.session['cms_edit'] = False
+            request.session['cms_toolbar_disabled'] = True
+            # menu_pool.clear()
```
czpython: cleanup?

cms/utils/setup.py
mkoistinen: I think this is a reasonable startup check/warning. Anyone else care to weigh-in?

mkoistinen: I'll add a note in documentation about how to suppress this if someone really doesn't want to be reminded...

czpython:
:+1:

yakky:
In -0 on introducing this within 3.2.1, especially with a warning emitted by default

cms/utils/setup.py
yakky: I'm against emitting this warning: the overall majority of the CMS installations won't need this feature and I don't think we should force developers to add a setting to silence the warning

czpython: This is my feeling as well.

mkoistinen: Will remove the warning.

mkoistinen:
Warning removed.

mkoistinen:
Rebased

mkoistinen:
Rebased to develop (3.3)

czpython:
@mkoistinen looks like the rebase messed up the branch.

mkoistinen:
It *looks* like it, yes, but only because the PR is still wanting to merge to 3.2.x.

czpython:
@mkoistinen afaik you can't.. need to close the pr and open it again targeting the develop branch

mkoistinen:
I have never seen so many +1's in my life: https://github.com/isaacs/github/issues/18

mkoistinen:
Closing to move to new target branch ('develop')

PR #4876, opened 2015-12-29
declare forceTouchOnce before using it

vxsx:
the code didn't work previously and error was swallowed by $.proxy

in essence - now touch on the body actually forces tooltip to become a "touch" tooltip


Flight:
LGTM

Issue #4856, opened 2015-12-15
Treebeard error when copying/pasting plugin


czpython:
All plugins have an internal treebeard field called ``numchild`` whose value should be how many direct children does the plugin have.

Given the structure below:

![cms_plugin_copy_error](https://cloud.githubusercontent.com/assets/994951/11824945/7edebf36-a34a-11e5-8a71-d5a533eee0bb.png)

If I copy the nested Row 1 Column plugin and paste it under another plugin like the style plugin, the new plugin will have its ``numchild`` attribute set to 0, this means that this new plugin will raise an exception every time I try to add a new plugin to it.

Here's a screencast of the issue:

https://drive.google.com/file/d/0B9SwgyV0g1hSS3RLa0ptUEg5U3c/view

Progress:

- [x] Merge fix on CMS 3.2.x (https://github.com/divio/django-cms/pull/4878)
- [x] Add upgrade notes for CMS 3.2.1 (https://github.com/divio/django-cms/pull/4880)
- [x] Merge fix on CMS 3.1.x (https://github.com/divio/django-cms/pull/4881)
- [x] Add upgrade notes for CMS 3.1.5

jrief:
Today I ran into the same problem in CMS-3.2, when using the function ``cms.api.add_plugin``.

Something I don't understand, if someone adds a plugin as child to an existing plugin, then ``target is not None`` and line 345 ``plugin_base = plugin_base.move(target, pos=position)`` triggers.

Wouldn't this be the right place to recalculate the ``numchild`` for the target plugin?

czpython:
Hello,

The numchild value is calculated by treebeard automatically.
Can you post a gist of the code you're using to reproduce the problem?

jrief:
Sure, the working code is here: https://github.com/jrief/djangocms-cascade/blob/master/cmsplugin_cascade/clipboard/admin.py#L109

I'll try to extract the fundamental parts:

```
clipboard = Placeholder.objects.get(slot='clipboard')
language = 'en'
plugin = plugin_pool.get_plugin('BootstrapRowPlugin')
data = {'glossary': {}, 'numchild': 7}
root_instance = add_plugin(clipboard, plugin, language, **data)
```

here ``root_instance.numchild`` is set to 0 despite of its initial value of 7. I could live with that, if adding children would correct the problem.

Now, if I add another instance to the current one, say

```
child_instance = add_plugin(clipboard, plugin, language, target=root_instance, **data)
```

[...]

PR #4878, opened 2015-12-29
rewrites/fixes test for issue #4856

czpython:
The test will fail.
This is good because pr https://github.com/czpython/django-cms/pull/2 fixes it.


mkoistinen:
LGTM

PR #4883, opened 2016-01-02
Fix support for reversion >=1.9

### LINKED ISSUE #4784: Fix compatibility with reversion 1.9
This is an issue in 3.1 and 3.2

Edit: reversion 1.10 has been released with modified API. I suggest to support only 1.8.5 (for Django 1.6) and 1.10 (for Django 1.7+ as stated in the documentation)

yakky:
Slowly approaching to full compatibility.
Reversion logic has changed a lot: 1.8 & 1.9+ compatibiliy layer added, to be dropped in 3.3 when we drop Django 1.6/1.7 and can thus focus on reversion 1.10+ 

Fix #4784 
This is a blocker for #4812 (which is complete as far as Django 1.9 is involved)

### DIFF
--- a/cms/utils/reversion_hacks.py
+++ b/cms/utils/reversion_hacks.py
@@ -1,6 +1,26 @@
 # -*- coding: utf-8 -*-
-from reversion import revisions as reversion
-from reversion.revisions import RegistrationError, VersionAdapter
+
+try:
+    from reversion import revisions as reversion
+    from reversion.admin import VersionAdmin as ModelAdmin, RollBackRevisionView  # NOQA
+    from reversion.models import Revision, Version  # NOQA
+    from reversion.revisions import create_revision, RegistrationError, VersionAdapter  # NOQA
+    from reversion.signals import post_revision_commit  # NOQA
+
+    revision_manager = reversion.default_revision_manager
+    revision_context = reversion.revision_context_manager
+except ImportError:
+    import reversion
+    from reversion import create_revision  # NOQA
+    from reversion.admin import VersionAdmin as ModelAdmin  # NOQA
+    from reversion.models import Revision, Version, post_revision_commit  # NOQA
+    from reversion.revisions import RegistrationError, VersionAdapter  # NOQA
+
+    revision_manager = reversion.revision
+    revision_context = reversion.revision_context_manager
+
+    class RollBackRevisionView(Exception):
+        pass
 
 
 def register_draft_only(model_class, fields, follow, format):
@@ -8,7 +28,7 @@ def register_draft_only(model_class, fields, follow, format):
     version of the reversion register function that only registers drafts and
     ignores public models
     """
-    revision_manager = reversion.revision
+
     if revision_manager.is_registered(model_class):
         raise RegistrationError(
             "%r has already been registered with Reversion." % model_class)
--- a/cms/utils/reversion_hacks.py
+++ b/cms/utils/reversion_hacks.py
@@ -2,19 +2,19 @@
 
 try:
     from reversion import revisions as reversion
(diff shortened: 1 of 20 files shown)


yakky:
@FinalAngel @mkoistinen this is complete, but it requires extensive testing as many moving parts have changed and 1.10 has a very different admin integration logic

cms/utils/reversion_hacks.py
```diff
@@ -1,6 +1,26 @@
 # -*- coding: utf-8 -*-
-import reversion
-from reversion.revisions import RegistrationError, VersionAdapter
+
```
czpython: It looks like you're importing the same stuff even on ImportError, it's best to explicitly target the imports that will fail and add some comment as to why they would fail ie. "Reversion >= 1.x compat".

Also the revision manager assignments can easily be outside of the ImportError try/except and placed on their own try/except that catches the AttributeError, more specifically only the revision_manager seems to change location

yakky: Well, I could easily change this in a simple version checking.
Having a simple try / except will make easier to remove the legacy imports in 3.3 (as we can use reversion 1.10 for all the supported Django versions) and a single section makes the code more readable. I'll happily add a comment to clarify the reason of the code style

test_requirements/django-1.8.txt
```diff
@@ -1,9 +1,6 @@
 -r requirements_base.txt
 Django>=1.8,<1.9
-django-reversion>=1.8,<1.9
-django-classy-tags>=0.6.1
+django-reversion==1.10.0
```
mkoistinen: Shouldn't this be `...>=1.10,<1.11`?

test_requirements/django-1.7.txt
```diff
@@ -1,9 +1,6 @@
 -r requirements_base.txt
 django>=1.7,<1.8
-django-reversion>=1.8,<1.9
-django-classy-tags>=0.5
+django-reversion==1.10.0
```
mkoistinen: Shouldn't this be `...>=1.10,<1.11`?

test_requirements/django-1.9.txt
```diff
@@ -1,10 +1,6 @@
 -r requirements_base.txt
 https://github.com/django/django/tarball/stable/1.9.x/django.tar.gz#egg=django
-django-reversion>=1.8,<1.9
-django-classy-tags>=0.6.1
-djangocms-text-ckeditor>=2.8.1
-django-hvad==1.5
+django-reversion==1.10.0
```
mkoistinen: Shouldn't this be `...>=1.10,<1.11`?

mkoistinen:
Good work, @yakky I have personally tested this a lot (in a test project, not a live project) and seems to work well. I've made a couple of comments about the requirements, but otherwise LGTM.

mkoistinen:
OK, I've now tested this in existing projects. Restoring revisions created recently (since the new django-revision was installed?), seem to work OK. Restoring older revisions fails with a Database Integrity issue as show here: https://gist.github.com/mkoistinen/140ff90d2ea077629588. It is possible (but rather unlikely) that this could be related to some previous corruption in the DB.

Also, after restoring even a "working" revision, when I attempt to publish them, the AJAX request returns this: https://gist.github.com/mkoistinen/5aae3398498316659f56. I've been able to reproduce this errors on-demand using a new project (using djangocms-installer).

yakky:
@mkoistinen I'll have a look at this shortly

yakky:
@mkoistinen I've been able to reproduce both issues and they are hopefully fixed

mkoistinen:
Awesome! Will re-test.

PR #4884, opened 2016-01-04
bring back debug version

vxsx:
1. fix incorrect check for event type, making the message with cms version show up again
2. remove the check for debug that resulted in incorrect positioning of the messages


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -721,7 +720,7 @@ var CMS = window.CMS || {};
                 debug.on(this.mouseEnter + ' ' + this.mouseLeave, function (e) {
                     clearTimeout(timer);
 
-                    if (e.type === that.mouseEnter) {
+                    if (e.type === 'mouseenter') {
```
FinalAngel: that looks like a weird fix, does it also work when toolbar is opened / closed?

vxsx: yes, it works
it's not weird :)
when event happens it has the type, and the type is always the _original_ event name, which in this case is `mouseenter`. otherwise we were always comparing `'mouseenter' === 'mouseenter.cms.toolbar'` which in turn never showed the message

FinalAngel: alright :)

FinalAngel:
LGTM

PR #4885, opened 2016-01-04
Sideframe hide duration

vxsx:
condition was not accepting falsy value such as 0,
now it does, but we changed the closing duration of the sideframe to be half of the opening duration


FinalAngel:
LGTM

PR #4889, opened 2016-01-05
Replace removed javascript_quote with escapejs

evenicoulddoit:
`javascript_quote` is removed in Django19


yakky:
@evenicoulddoit thanks for contributing this. This fix is included in #4812 and we'll track it there

PR #4891, opened 2016-01-06
fixes incorrect language on placeholder text for redirect field

czpython:
``widget_attrs`` is called when the field is instantiated which means it picks up the language of the project (``LANGUAGE_CODE``) instead of the thread language.


mkoistinen:
LGTM

PR #4895, opened 2016-01-07
replace data-icon with data-cms-icon

vxsx:
sometimes styles from the website use the `[data-icon]` selector which can override the icons used in the plugin settings menu. namespacing the attribute avoids that


FinalAngel:
LGTM

PR #4903, opened 2016-01-12
Prevent duplicate slugs via wizard

mkoistinen:
Currently, the wizards will auto-generate a unique-slug, unless one is provided, then it will use that verbatim. This PR will force even a provided slug to be unique.


czpython:
:+1:

PR #4906, opened 2016-01-12
add a check for the current page if publishing a page from sideframe

vxsx:
if you've changed the slug - just reloading the page would result in
going to the old slug, which shows 404, which is bad ux, this PR fixes that


FinalAngel:
LGTM

FinalAngel:
travis seems to fail :(

vxsx:
screen shot 2016-01-13 at 09 07 03

@FinalAngel thankfully it was only a stalled job :)

PR #4902, opened 2016-01-11
fixes issue with page menu missing

czpython:
* Added new utility ``has_auth_page_permission``
* Modified toolbar to add ``Pages`` menu when there's no pages.

PR #4908, opened 2016-01-12
Fixes wizard not respecting cms_permissions setting

### LINKED ISSUE #4905: Page wizard does not respect CMS_PERMISSION setting
The page wizard form checks the ``GlobalPermission`` model which should only be checked then the ``CMS_PERMISSION`` setting is set to ``True``.

https://github.com/divio/django-cms/blob/develop/cms/cms_wizards.py#L36

- [x] Merge #4908 into 3.2.x
- [x] Forward port to 3.3.x

czpython:
Fixes #4905

Based off #4902

* Refactored cms permission utilities to be more dry.
* Have cms wizard use these utilities instead of using it's own.
* Start split between function and function_from_request

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -498,7 +498,7 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
         Return true if the current user has permission to add a new page.
         """
         if get_cms_setting('PERMISSION'):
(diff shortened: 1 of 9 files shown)


mkoistinen:
Looks VERY good to me... now we just have to do all the other permission check functions (in a future PR).

PR #4912, opened 2016-01-13
Use session DB backend in manage.py to ease test

yakky:
cache backend is pretty frustrating for testing the cms using the integrated manage.py. This changes the session backend to db

### DIFF
--- a/cms/tests/test_cache.py
+++ b/cms/tests/test_cache.py
@@ -64,12 +64,12 @@ class CacheTestCase(CMSTestCase):
             'django.middleware.cache.FetchFromCacheMiddleware'
         ]
         middleware = [mw for mw in settings.MIDDLEWARE_CLASSES if mw not in exclude]
-        with self.settings(CMS_PAGE_CACHE=False, MIDDLEWARE_CLASSES=middleware, SESSION_ENGINE="django.contrib.sessions.backends.cache",):
+        with self.settings(CMS_PAGE_CACHE=False, MIDDLEWARE_CLASSES=middleware):
             with self.assertNumQueries(FuzzyInt(13, 17)):
                 self.client.get('/en/')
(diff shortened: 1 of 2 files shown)


yakky:
@mkoistinen @czpython what's your opinion?

czpython:
I don't think this would have side-effects so lgtm :+1:

yakky:
@czpython no effect when using an installed version of the CMS as manage.py it's not included

yakky:
@ojii any reason for that setting in the first place? (performance during tests, maybe?)

ojii:
Note: I currently don't have acces to a real dev env, so I'm working off GH/memory here

IIRC django-app-manage doesn't set ``SESSION_ENGINE``. however, using cache instead of db generally results in faster session performance. can you elaborate on the "cache backend is pretty frustrating for testing"?

ojii:
Oh, you mean why set it in ``manage.py`` in django CMS, not app-manage. Again, working in a somewhat suboptimal env right now, so please apologize. I think it's because it was always set to that and/or because cache backend will be faster (especially when testing postgres/mysql). I'm still curious as to what makes using db sessions easier for testing.

mkoistinen:
No opinion here. LGTM.

yakky:
@ojii bad wording on my side. I meant that session cache backend is frustrating when **manually** testing (ie: using ``manage.py runserver`` from cms source root.
Coupled with ``locmem`` cache, it means you're logged out whenever runserver is reloaded.

yakky:
@ojii @czpython @mkoistinen pushing a better approach

ojii:
> I meant that session cache backend is frustrating when manually testing (ie: using manage.py runserver from cms source root.
> Coupled with locmem cache, it means you're logged out whenever runserver is reloaded.

okay that makes sense. the patch as it is now seems fine.

PR #4923, opened 2016-01-15
Fixes issue 4922

### LINKED ISSUE #4922: Unable to restore app hooked pages on 3.2.x
PR #4883 introduced reversion 1.10.x support on the CMS.

Due to the way reversion 1.10.x works, the page object is restored and then saved using the change form logic but the PageAdmin's save_model assumes that the page has not yet been created and so tries to create a temporary page to get the treebeard attributes but this fails because things like application_namespace are unique at the database level.

- [x] Merge #4923 into 3.2.x

Fixed. To be forward ported to CMS 3.3.x

czpython:
Fixes #4922
* Adds test for regression.
* Fix is to create a "blank" page instead of a copy of the recovered page in order to avoid integrity errors.

### DIFF
--- a/cms/tests/test_reversion_tests.py
+++ b/cms/tests/test_reversion_tests.py
@@ -252,8 +252,8 @@ class ReversionTestCase(TransactionCMSTestCase):
                 page = create_page(**page_data)
 
                 # Assert page has apphooks
-                self.assertEqual(page.apphook, 'SampleApp')
-                self.assertEqual(page.apphook_namespace, 'SampleApp')
+                self.assertEqual(page.application_urls, 'SampleApp')
+                self.assertEqual(page.application_namespace, 'SampleApp')
 
                 # Create revision
                 make_revision_with_plugins(page, user=None, message="Initial version")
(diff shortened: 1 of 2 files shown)


cms/tests/test_reversion_tests.py
```diff
@@ -231,6 +232,61 @@ def test_recover(self):
...
+        self.assertEqual(Page.objects.count(), 2)
+
+        with self.login_user_context(self.user):
+
+            page_data = self.get_new_page_data_dbfields()
+            page_data['apphook'] = 'SampleApp'
+            page_data['apphook_namespace'] = 'SampleApp'
+
+            with create_revision():
+                # Need to create page manually to add apphooks
```
yakky: why not using the API?

yakky:
Apart for the clarification LGTM

cms/tests/test_reversion_tests.py
czpython: I am using API below in ``page = create_page(**page_data)``.
that's what I meant by "manually".

yakky: Why do I do review in the evening?
Disregard my comment

PR #4926, opened 2016-01-16
Forwardport 3.1.x code

yakky:


czpython:
@yakky all my commits from 3.1.x are already in 3.2.x.
I backported from 3.2.x to 3.1.x.

Some commits here like 3941e426e4c71800841fb33d30e72e486a2a1032 are not in 3.2.x because they're specific to 3.1

yakky:
@czpython yeah, but porting the entire history is a much better approach (until is feasible) as it makes maintenance much easier and keep the history coherent.
This PR takes into account the small differences between 3.1 and 3.2

czpython:
@yakky good to know. @mkoistinen had mentioned that we had to port our prs manually backwards and forwards.

yakky:
@mkoistinen we used to forwardport everything. Has the policy changed?

yakky:
@czpython @mkoistinen are you to merge this?

mkoistinen:
@yakky, the conversation that @czpython and I had related to things that were substantially different between branches, but in general, I recommend to people who know something should be back- or forward-ported to just do it to alleviate issues and time-/effort-required later.

PR #4932, opened 2016-01-18
Revision issues with placeholders from multiple sources

czpython:
cmsplaceholders

Breakdown:

**Feature** - Native cms placeholder (created using the ``placeholder`` tag)
**Newsblog_Base_Top** - Manual placeholder (created using a ``PlaceholderField``)
**Newsblog_Base_Sidebar** - Manual placeholder (created using a ``PlaceholderField``)
**Footer** - Static placeholder (created using the ``staticplaceholder`` tag)

This pr fixes the following issues:

* No revision history created when moving a plugin from **Newsblog_Base_Top** (manual) to **Feature** (native).
* No revision history created when moving a plugin from **Feature** (native) to  **Newsblog_Base_Top** (manual) or to **Footer** (static).
* Revision history marked a plugin as moved when copying a plugin.
* Moved plugin action revision creation to the post_ methods to avoid clashes with other
``create_revision`` decorators being used.
* Improved revision message when moving plugin.
* Improved code in move a plugin logic to avoid issues with overriding variables.

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -609,6 +609,7 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
                 return False
         return True
 
+    @create_revision()
     def post_add_plugin(self, request, placeholder, plugin):
         if is_installed('reversion') and placeholder.page:
             plugin_name = force_text(plugin_pool.get_plugin(plugin.plugin_type).name)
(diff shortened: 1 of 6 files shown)


yakky:
@czpython do you think it's feasible to come up with a test for this? I overlooked this path as there are very few tests for reversion and logic changed a lot.
Does this work in both 1.8 and 1.10?

czpython:
@yakky yes I'm currently working on tests that will test for all the use cases I listed above.
And yes I've tested manually on both 1.8 and 1.10

yakky:
@czpython thanks for the clarification

PR #4933, opened 2016-01-19
do not force origin to get pagetree styles if staticUrl is remote host

### LINKED ISSUE #4929: cms.changelist.js thinks my static files are local (?)
I just attempted to upgrade from Django-CMS 3.0.3 to 3.2 and hit a snag. When I try to load my page, the toolbar looks like it has no styling applied to it, and Chrome's console tells me that `bundle.admin.changelist.min.js` is trying to load some CSS from the URL:

http://my-origin-dns-name.comhttp//my-s3-bucket/cms/js/jstree/themes/default/style.css

Digging into the javascript, I *think* the problem is line 379 of cms/static/cms/js/modues/cms.changelist.js, which reads:

`path: origin + that.options.settings.staticPath + 'cms/js/jstree/',`

I think that `staticPath` gets set to my `STATIC_URL` in `cms/templates/admin/cms/page/tree/base.html`. Since I set my `STATIC_URL` to my S3 bucket, the line 

`path: that.options.settings.staticPath + 'cms/js/jstree/',`

would work just fine. But the actual code prepends `origin` to it, even though I don't host my static files on my origin server.

I freely admit that I don't know much about the internals of Django CMS, and so my be being an idiot. A I doing something wrong, or have I found a problem?

vxsx:
Should fix #4929


FinalAngel:
LGTM

PR #4947, opened 2016-01-23
Refactor commands

### LINKED ISSUE #4679: Refactor the CLI commands
Currently they have inconsitent code and interface (e.g.: options/arguments confusion)
We can refactor them to use the new argparse interface of django commands (optparse will be dropped in 1.10 anyway) and provide a better help support.

At this stage does not involve any functional change

yakky:
This refactoring uses argparse subcommands and makes command more consistent in terms of usage and arguments.
Move to argparse is also required as optparse support will be removed in 1.10.

Currently all arguments are in ``--arg=...`` form to make them self documenting. In some cases they are required, so the positional arguments form would be more correct. I welcome any opinion on this topic (and the PR in general)

* [x] Base commands on argparse
* [x] Improve help output
* [x] Update documentation
* [x] Fix test

Fix #4679


yakky:
Ping @ojii @mkoistinen @evildmp @czpython  for comments

mkoistinen:
I'm a fan of this effort! Thank you for adding consistency to the commands too. +100!

yakky:
@mkoistinen what's your opinion on named / positional arguments?

ojii:
I like the idea of moving to argparse. `foo required-arg --optional=arg` makes so much more sense than `foo --required=arg --optional=arg`.

yakky:
argparse has makes no real differences between arguments and options in terms of definition and behavior and I always find the positional behavior a bit confusing
@mkoistinen @evildmp @czpython @FinalAngel @vxsx  do you have any opinion on the positional / non-positional arguments?

FinalAngel:
@yakky +1 on `foo required-arg --optional=arg`

evildmp:
I prefer everything to be a named argument. Then, examples you give to other people make sense, because it's explicit.

yakky:
@divio/django-cms-core this is complete from my side
yay or nay?

ojii:
Is the `--from=lang` style decided upon now?

yakky:
Given the split votes on the approaches and that even Django project has no best practice to lean for one or the other approach, I decided to keep them as ``--args``. (I refactored ``copy-lang/copy-site`` as sub-sub commands of ``copy``)

PR #4772, opened 2015-11-26
#4771 Adjusted get_object_public_url and get_object_draft_url to use …

marijngoedegebure:
…language attribute.

PR #4950, opened 2016-01-24
Merge #4772

### LINKED ISSUE #4771: django.middleware.locale.LocaleMiddleware required for cms_toolbar
From CMS version 3.1 two functions have been added to toolbar.py. These functions insert a bug for when you render the cms toolbar in a single language project.
When rendering django cms_toolbar it requires that the LANGUAGE_CODE is in the request, like this:

```
def get_object_draft_url(self):
        if self.obj:
            with force_language(self.request.LANGUAGE_CODE):
```

But as is stated in the code at the top of toolbar.py, whether language is in the request depends on whether multiple languages are enabled:

```
if settings.USE_I18N:
    self.language = get_language_from_request(request)
else:
    self.language = settings.LANGUAGE_CODE
```

I propose to adapt this same technique for both get_object_draft_url and get_object_public_url and let them rely on the language attribute of the toolbar.

yakky:
Merges #4772  with test

Fix #4772 
Fix #4771

### DIFF
--- a/cms/tests/no_i18n.py
+++ b/cms/tests/no_i18n.py
@@ -2,15 +2,20 @@
 from django.contrib.auth import get_user_model
 from django.core.urlresolvers import clear_url_caches
 from django.template import Template
+from django.test import RequestFactory
 from django.test.utils import override_settings
+
 from djangocms_text_ckeditor.models import Text
 
 from cms.api import create_page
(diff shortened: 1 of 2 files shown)


yakky:
Synced with upstream to let tests pass

PR #4955, opened 2016-01-26
Improve CMSPluginBase.render documentation

yakky:
I think that suggesting to call the render of the plugin super class should be encouraged and documented

### DIFF
--- a/docs/how_to/custom_plugins.rst
+++ b/docs/how_to/custom_plugins.rst
@@ -265,7 +265,7 @@ Now we need to change our plugin definition to use this model, so our new
         cache = False
 
         def render(self, context, instance, placeholder):
-            context['instance'] = instance
+            context = super(HelloPlugin, self).render(context, instance, placeholder)
             return context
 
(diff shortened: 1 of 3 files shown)


yakky:
@mkoistinen @czpython @evildmp do you agree on this?

docs/how_to/custom_plugins.rst
```diff
@@ -265,7 +265,7 @@ Now we need to change our plugin definition to use this model, so our new
         cache = False
 
         def render(self, context, instance, placeholder):
-            context['instance'] = instance
+            context = super(HelloPlugin, self).render(context, instance, placeholder)
```
mkoistinen: Is this necessary/advised? Isn't the context already passed in as a parameter?

yakky: the context passed as parameter does not have instance / placeholder: they are added in the parent render method

mkoistinen: OK, thats fine then =)

PR #4958, opened 2016-01-26
copy paste fixes

vxsx:
1. Fixes an issue when "Copy all" menu item was available for empty placeholder (clicking caused 500 error)
2. Fixes an issue when placeholder contents copied by "Copy all" button couldn't be pasted anywhere because of [custom hacky parent restriction](https://github.com/divio/django-cms/blob/6918793ef0342e44e33a37ba950d18d1593e3d18/cms/cms_plugins.py#L18)


FinalAngel:
the code is LGTM but can you add the comment into the code why that was necessary?

czpython:
:+1: from our dicussion

PR #4959, opened 2016-01-26
Fix wizard template

yakky:
Why not use shipped jquery?


yakky:
@mkoistinen @FinalAngel @vxsx do you agree with this change?
This is required anyway as in Django 1.9 jquery has been moved

vxsx:
:+1:

mkoistinen:
Its its OK by @vxsx, its OK by me.  Merging.

Issue #4931, opened 2016-01-18
Impossible to override CMS_CACHE_DURATIONS


czpython:
The [docs](http://docs.django-cms.org/en/develop/reference/configuration.html#cms-cache-durations) point out that it's possible to override the ``CMS_CACHE_DURATIONS`` setting for more control over the CMS caching. This unfortunately does not work.

The CMS completely ignores the ``CMS_CACHE_DURATIONS`` setting defined at the project level and instead will use the following defaults:

```python
{
    'menus': getattr(settings, 'MENU_CACHE_DURATION', 60 * 60),
    'content': get_cms_setting('CONTENT_CACHE_DURATION'),
    'permissions': 60 * 60,
}
```

As shown, the CMS incorrectly looks at the deprecated ``MENU_CACHE_DURATION`` setting for the menu cache.
Then it correctly looks at ``CONTENT_CACHE_DURATION`` for the content cache and last it just hard codes a default for the permissions which means it's impossible to override the permissions timeout.

Credits to @vxsx for the first report.

yakky:
Fixed in #4986

PR #4960, opened 2016-01-27
Update changelog and other updates for 3.1.5

yakky:

PR #4962, opened 2016-01-28
Use CMS_CACHE_DURATIONS, when configured

mkoistinen:
Should address #4931.

### DIFF
--- a/cms/apps.py
+++ b/cms/apps.py
@@ -10,5 +10,3 @@ class CMSConfig(AppConfig):
 
     def ready(self):
         setup()
-        from cms.plugin_pool import plugin_pool
-        plugin_pool.set_plugin_meta()
--- a/cms/plugin_pool.py
+++ b/cms/plugin_pool.py
@@ -38,9 +38,66 @@ class PluginPool(object):
         self.plugins = {}
         self.patched = False
 
+    def validate_templates(self, plugin=None):
+        """
+        Plugins templates are validated at this stage
+
+        """
+        if plugin:
+            plugins = [plugin]
+        else:
+            plugins = self.plugins.values()
+        for plugin in plugins:
+            if (plugin.render_plugin and not type(plugin.render_plugin) == property
+                    or hasattr(plugin.model, 'render_template')
+                    or hasattr(plugin, 'get_render_template')):
+                if (plugin.render_template is None and
+                        not hasattr(plugin.model, 'render_template') and
+                        not hasattr(plugin, 'get_render_template')):
+                    raise ImproperlyConfigured(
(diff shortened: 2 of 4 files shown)


cms/utils/conf.py
```diff
@@ -67,9 +66,13 @@ def wrapper():
...
-    return {
+    """
+    Returns the setting: CMS_CACHE_DURATIONS or will attempt to build it from
+    the deprecated settings or will return the defaults.
+    """
+    return getattr(settings, 'CMS_CACHE_DURATIONS', {
         'menus': getattr(settings, 'MENU_CACHE_DURATION', 60 * 60),
-        'content': get_cms_setting('CONTENT_CACHE_DURATION'),
+        'content': getattr(settings, 'CMS_CONTENT_CACHE_DURATION', 60),
         'permissions': 60 * 60,
```
yakky: This should be fixed too, right?

yakky:
Shouldn't this be something like:

    return getattr(settings, 'CMS_CACHE_DURATIONS', {
        'menus': 60 * 60,
        'content': 60,
        'permissions': 60 * 60,
    }

Individual settings are no longer available, and we should just drop them

mkoistinen:
I was just adding a layer of convenience for migrating older projects... but, I'm not passionate about it.

yakky:
setting changed in 2.3 or 2.4 (IIRC) so I guess we can just drop the "compatible" setting

cms/utils/conf.py
```diff
@@ -67,9 +66,13 @@ def wrapper():
...
 def get_cache_durations():
-    return {
+    """
+    Returns the setting: CMS_CACHE_DURATIONS or will attempt to build it from
+    the deprecated settings or will return the defaults.
+    """
+    return getattr(settings, 'CMS_CACHE_DURATIONS', {
         'menus': getattr(settings, 'MENU_CACHE_DURATION', 60 * 60),
-        'content': get_cms_setting('CONTENT_CACHE_DURATION'),
+        'content': getattr(settings, 'CMS_CONTENT_CACHE_DURATION', 60),
```
czpython: why not use ''get_cms_setting''?

czpython:
:+1:

cms/utils/conf.py
mkoistinen: Because that setting is no longer supported. Doesn't matter though, I've removed it altogether based on @yakky's note.

yakky:
Linking to #4960 to update the changelog

yakky:
@mkoistinen I see that tests fail: do we want this in 3.1.5/3.2.1?

PR #4965, opened 2016-01-28
Fixes reload on publish

FinalAngel:
Certain actions do not reload the page when using the pagetree. This adds the last else fallback.


vxsx:
:+1:

PR #4975, opened 2016-02-01
Fix user lookup in page permission admin

### LINKED ISSUE #4842: User lookup not working for view restrictions/page permissions when CMS_RAW_ID_USERS = True
Tested on 3.1.4 and 3.2.0

When trying to add a single user based view restriction or page permission when via the lookup button when ```CMS_RAW_ID_USERS = True```, the 'Select user to change' popup is shown. However, if a user is selected you are taken to the edit page of the user, rather than the user pk being inserted in to the user input field.

From djangoCMS tool bar select Page->Permissions...+ Add another view restriction and select a user

mbambas:
Make sure that related lookup fields get initialized in page permission admin. This fixes #4842.

### DIFF
--- a/cms/templates/admin/cms/page/change_form.html
+++ b/cms/templates/admin/cms/page/change_form.html
@@ -103,6 +103,10 @@
 </form>
 </div>
 
+{% block admin_change_form_document_ready %}
+{{ block.super }}
+{% endblock %}
+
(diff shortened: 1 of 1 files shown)


yakky:
@mkoistinen should we apply this in 3.1.x and the port to the others version

yakky:
@mbambas thanks for spotting this.
Merged through #4989

PR #4969, opened 2016-01-29
Use CMS_CACHE_DURATIONS, when configured

mkoistinen:
Should address #4931.

PR #4986, opened 2016-02-07
Fix cache settings

### LINKED ISSUE #4931: Impossible to override CMS_CACHE_DURATIONS

yakky:
Fix #4931 
Fix #4969 

I set non-zero cache values into ``cli.py`` as due to the bug, cache was active through all the tests, so the correct cache values must be set to get the tests running

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -404,3 +404,10 @@ Please see Install/2.4 release notes *before* attempting to upgrade to version 2
 - Fix JS error when using PageSelectWidget
 - Fix whitespace markup issues in draft mode
 - Detect plugin migrations layout in tests
+
+
+=== 3.0.17 (unreleased) ===
+
+- Fix ExtensionToolbar when language is removed but titles still exists…
+- Fix PageSelectWidget JS syntax
+- Fix cache settings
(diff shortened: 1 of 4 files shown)


yakky:
I set non-zero cache values into ``cli.py`` as due to the bug, cache was active through all the tests, so the correct cache values must be set to get the tests running

czpython:
:+1:

yakky:
@mkoistinen any comment on this vs #4969 ?

mkoistinen:
+1 from me. Feel free to close my earlier attempt.

yakky:
thanks

PR #4994, opened 2016-02-11
Move urls.SLUG_REGEXP to constants

mikek:
This fixes the problem with Django 1.9 system checks. In a few words:
importing from cms.urls triggers app resolver mechanism that fails
if there were no database tables for `contrib.sites` created yet.
It catches `DoesNotExist` but not `OperationalError`. Moving this
constant out to a different location seems to be a better fix.


cms/urls.py
```diff
@@ -4,10 +4,9 @@
 
 from cms.apphook_pool import apphook_pool
 from cms.appresolver import get_app_patterns
+from cms.constants import SLUG_REGEXP
 from cms.views import details
 
-# This is a constant, really, but must live here due to import order
```
mikek: @yakky can you, please check if this is really the issue or move this constant to some other place instead?

Chive:
:+1:

yakky:
LGTM. A lot of circular import fixes since that comments.
Merged through #4995

yakky:
Fixed by #4995

PR #4995, opened 2016-02-11
Fix some early imports issues with Django 1.9

yakky:
Fix #4994


yakky:
@mkoistinen @czpython would you give a second look at this?

Chive:
:+1:

cms/appresolver.py
```diff
@@ -209,6 +211,8 @@ def get_app_patterns():
         current_site = Site.objects.get_current()
     except Site.DoesNotExist:
         current_site = None
+    except OperationalError:
+        return []
```
czpython: isn't this too general? I do think I've gotten this error though..

yakky: It's the only error we get if the database is not ready (e.g.: with no tables in it)

czpython:
:+1:

PR #4996, opened 2016-02-11
Update extending_page_title.rst

retailify:
self._page().get_title() should be self._get_page().get_title()


yakky:
@retailify nice, thanks for the fix

PR #4999, opened 2016-02-12
urlconf during startup: also catch db ProgrammingError

Chive:
The fix introduced in https://github.com/divio/django-cms/pull/4995 only solved my problem half-way, just now I had this error:

```bash
django.db.utils.ProgrammingError: relation "django_site" does not exist
```

### DIFF
--- a/cms/appresolver.py
+++ b/cms/appresolver.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import with_statement
 
-from django.db import OperationalError
+from django.db import OperationalError, ProgrammingError
 
 from cms.utils.compat import DJANGO_1_8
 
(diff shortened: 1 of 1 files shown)


Chive:
Not too sure about the exception I'm catching here. I know for sure that both ``OperationalError`` and ``ProgrammmingError`` happened before. Maybe just catch those two. Or, their base "django.db.utils.DatabaseError"

stefanfoulis:
:+1:

yakky:
@Chive I'd pick ``DatabaseError``: as urlconf maybe loaded when database is not ready in general, I'd go with the superclass for this kind of errors

stefanfoulis:
@yakky ``DatabaseError`` is probably a bit to broad. @Chive looked it up and it seems even a misconfigured field on a Model would be a ``DatabaseError``. If we prevent reporting of that to developers it could be very hard to debug a bug ``models.py``.

yakky:
@stefanfoulis @Chive ok. this is good for merge, then

Issue #5000, opened 2016-02-12
Allow to hide / deregister plugins from configuration


yakky:
Sometimes, even if a plugin is available in the codebase, you don't want it to be available in any place.
Currently ``CMS_PLACEHOLDER_CONF`` only allow to whitelist plugins in specific placeholders.
**Blacklisting** plugins should be available too (either globally or per-placeholder)

See also http://stackoverflow.com/questions/35044928/django-cms-placeholder-plugin-avaliable-only-to-a-specific-placeholder

yakky:
Linked to https://github.com/divio/django-cms/issues/4979

yakky:
Fixed by #5412 
To implement blacklist, add the plugin to be blacklisted in ``CMS_PLACEHOLDER_CONF[None]['excluded_plugins']`` list

marcosguedes:
Started from #5000, now we here :) Updated stackoverflow answer above

yakky:
@marcosguedes awesome!

Issue #4979, opened 2016-02-04
Exclude (Blacklist) Plugins from all placeholders


marcosguedes:
Hi there. I wanted to suggest the possibility of excluding a certain plugin from all placeholders. According to the docs it's only possible if I explicitly mention every other plugin in every placeholder.

It's fine if I have a small number of placeholders but it's an inglorious task on more complex templates (in my case, 12 with a great amount of different placeholders and static_placeholders each).

Thank you

(Any tips are greatly appreciated)

yakky:
Fixed by #5412 
To implement blacklist, add the plugin to be blacklisted in ``CMS_PLACEHOLDER_CONF[None]['excluded_plugins']`` list

marcosguedes:
Thank you

PR #5006, opened 2016-02-14
[WiP] Implement four-level CMS_PLACEHOLDER_CONF

yakky:
Currently ``CMS_PLACEHOLDER_CONF`` only allows values for placeholder and template+placeholder keys
Sometimes a global configuration would make things much nicer / simpler
@ojii suggested a for level structure like (in increasing order of precedence)

    CMS_PLACEHOLDER_CONF['*'] (global)
    CMS_PLACEHOLDER_CONF['template'] (if template is given)
    CMS_PLACEHOLDER_CONF['placeholder']
    CMS_PLACEHOLDER_CONF['template placeholder'] (if template is given)

This is backward-compatible and allow a few neat tricks.

We can also add wildcards to placeholder configuration:

    CMS_PLACEHOLDER_CONF['placeholder-*']

this will recognize ``placeholder-foo`` and ``placeholder-bar``.
We need to make ``*`` (or ``%``) reserved to avoid clashes (this will also solve the @ojii's comment regarding '*' being a valid placeholder name). We can use wildcards also for templates.

Changes needed
 * [x] 4-level configuration
 * [x] tests
 * [ ] documentation

On top of this is easy to implement plugin blacklisting (see #5000, #4979), list of plugins to show in text editor and other plugin configuration options


ojii:
Thinking about this a little more, I think the "global" tag should be `None`, not `'*'`, as `'*'` is technically a legal placeholder name.

yakky:
any comments @divio/django-cms-core  ?

ratmcarlo:
May be an idea to use regexp as CMS_PLACEHOLDER_CONF keys?

yakky:
@ratmcarlo It may look weird, but it totally makes sense to me

czpython:
@yakky I like the idea, however I share agree with @ojii that * can be problematic.

yakky:
Using regexp as suggested by carloratm can solve this as we can use None as default, and rely on regexp to handle wildcards

yakky:
@czpython @ojii have a look at this proposal from @fmarco https://github.com/yakky/django-cms/pull/16
It uses general purpose regexp and it does not dictate any specific syntax or reserved keyword

yakky:
@divio/django-cms-core can I have a feedback on the proposed implementation? It still has rough edges, but before proceeding I'd like a yay / nay if it's the way to go

czpython:
I'm a bit skeptical about supporting regex, I'm not sure if this feature balances out with the extra technical implications.
I think the original idea (4 level conf) is good just not sure about regex.
If anything I'd say that regex in the template might be a bit too much.

ojii:
> I'm a bit skeptical about supporting regex, I'm not sure if this feature balances out with the extra technical implications.
> I think the original idea (4 level conf) is good just not sure about regex.
> If anything I'd say that regex in the template might be a bit too much.

I agree. regex seems like overkill and a solution looking for a problem to solve.

yakky:
Regexp are actually a special case. You can easily use it without them and falling back to the "basic" 4 level structure
This allows to cover cases like "use a specific set of plugins in all the placeholders that starts with 'sidebar-'"
In simple websites this is an overkill, in reasonably sized projects it's almost essential. Especially if you do configure carefully the plugins in placeholders (I may share the regular hundred-lines long CMS_PLACEHOLDER_CONF we are forced to write, sometimes...)

mkoistinen:
The implementation looks great and while I support the effort here, having *this much* complexity here, in a setting, makes me wonder if we're going about this all wrong. I don't have any proposed solutions to change from `CMS_PLACEHOLDER_CONF`, but this seems to be expanding into a lot more than originally intended.

yakky:
@divio/django-cms-core restructured to use simple key parsing, allowing to customize it

yakky:
@mkoistinen @czpython could you please provide a comment on this last approach?

PR #5009, opened 2016-02-16
allow plugins to disable caching dynamically

jrief:
Currently one has to decide whether to cache the plugin's content while declaring the plugin, but sometimes, some of my plugins must disable caching depending on the given context.

I therefore have added a method ``get_use_cache(context, instance, placeholder)`` which can be overridden by a plugin and be used to disable plugin caching dynamically.

This PR is prove of concept, if it will be accepted, I'll add documentation and a unit test.

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -329,6 +329,9 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
             reload_required = options.get('requires_reload', False)
         return reload_required
 
+    def get_use_cache(self, context, instance, placeholder):
+        return self.cache
+
     def get_plugin_urls(self):
         """
         Return URL patterns for which the plugin wants to register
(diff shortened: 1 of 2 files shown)


yakky:
Nice one, I think it should go into 3.3 as 3.2 is only open for bugfixing

cms/plugin_base.py
```diff
@@ -329,6 +329,9 @@ def requires_reload(self, action):
             reload_required = options.get('requires_reload', False)
         return reload_required
 
+    def get_use_cache(self, context, instance, placeholder):
+        return self.cache
```
yakky: I'll rename the attribute as "use_cache" and add a property to handle the old setting

yakky:
@jrief See my previous comment. Also docs and tests needed. But it's good for me

mkoistinen:
This PR is superseded by #5095.

yakky:
Closing due to the merge of #5095

PR #5010, opened 2016-02-16
fixes integrity errors in revert plugins

o-zander:
While trying to recover a deleted page I encountered several integrity errors.

First error was raised because the plugin list wasn't ordered by path so child plugins were inserted before parent plugins (parent id constraint failed). I solved this by sorting the list by path before inserting.

The second problem occurred while inserting the "others"-list. I didn't really get why this is happening and don't know whether this it is the right way to catch the error and just ignore it.

I would appreciate if someone would review my pull request. It doesn't need to be merged but it was easier to show the problems by solving it than by just explaining it.

Problem was found in django-cms 3.1 on django 1.7 but checked against django-cms 3.2. It happened there too. But the data was from django-cms 3.1 so maybe the problem has its source in the way the versions are created and this will be solved by 3.2.

### DIFF
--- a/cms/admin/views.py
+++ b/cms/admin/views.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from django.db import transaction, IntegrityError
 from django.http import Http404
 from django.shortcuts import get_object_or_404
 
@@ -16,6 +17,7 @@ def revert_plugins(request, version_id, obj):
     titles = []
     others = []
     page = obj
+
     for rev in revs:
         obj = rev.object
         if obj.__class__ == Placeholder:
@@ -30,9 +32,12 @@ def revert_plugins(request, version_id, obj):
             titles.append(obj)
         else:
             others.append(rev)
+
     if not page.has_change_permission(request):
         raise Http404
+
     current_plugins = list(CMSPlugin.objects.filter(placeholder__page=page))
+
     for pk, placeholder in placeholders.items():
         # admin has already created the placeholders/ get them instead
         try:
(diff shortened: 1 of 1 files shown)


yakky:
@o-zander have you checked django cms 3.2.1 with reversion 1.10?
Reversion improved a lot their transaction handling and we added some fixes in 3.2.1

o-zander:
I checked it and yes they improved their handling: The error message is way more friendly now :D `Could not revert revision, due to database integrity errors.` The actual error isn't gone so I can assume the reversion data in the database is already corrupted by previous versions. But their new code looks very reasonable to me. I have very many plugins (50+) with a long chain of relations (e.g. cms plugin -> slide -> image) so I wouldn't exclude I've made some mistakes in registering my models with reversion. Maybe you could catch the error in the admin and show a proper error message instead of causing a server error 500.

cms/admin/views.py
```diff
@@ -54,6 +66,10 @@ def revert_plugins(request, version_id, obj):
             if old.pk == plugin.pk:
                 plugin.save()
                 current_plugins.remove(old)
+
+    # TODO: the same rule for the next block (others) applies here, this should be wrapped in an atomic block
+    # see https://docs.djangoproject.com/en/1.9/topics/db/transactions/#controlling-transactions-explicitly
+    # block "Avoid catching exceptions inside atomic!"
```
yakky: The exception is used here exactly to prevent the transaction to fail, and to provide an alternate way to recover the situation

yakky:
Assigning myself for review. @o-zander I'll eventually merge this manually in 3.2.x branch

o-zander:
After seeing the new implementation of reversions revert maybe this should be solved differently. I could add a solution if you are interested.

Another and maybe better approach for both reversion and django-cms will be the mechanism used in loaddata: Deactivate all constraint checks, insert all data, manually check constraints.

yakky:
@o-zander if you have any improved solution, I'd happily have a look at this.
There's a lot of room for improvements in the reversion code, and this is definitely an area we want to invest on

yakky:
@o-zander do you have an improved solution or should I review this as is?

o-zander:
@yakky Sorry, I didn't had time to implement a better approach. I think you need not review the current solution since it could lead to data loss. The new reversion code is sufficient.

Issue #5013, opened 2016-02-17
django-cms middleware breaks django-rest-framework multipart POST data


richardasaurus:
When I have either of:
```
 'cms.middleware.toolbar.ToolbarMiddleware',
 'cms.middleware.language.LanguageCookieMiddleware',
```
in `MIDDLEWARE_CLASSES` settings list. 

this assert will raise in a django-rest `ListCreateAPIView`. 
```python
class ImageList(generics.ListCreateAPIView):
    permission_classes = (DjangoModelPermissions,)
    queryset = models.Image.objects.all()
    serializer_class = serializers.ImageSerializer
    parser_classes = (
        parsers.JSONParser,
        parsers.MultiPartParser,
        parsers.FormParser,
    )

    def create(self, request, *args, **kwargs):
        assert 'image_file' in request.data
       # more code...
```

richardasaurus:
Adding the setting `CMS_TOOLBAR_HIDE = True` "fixes" this for me

richardasaurus:
Fix here https://github.com/divio/django-cms/pull/5014

PR #5014, opened 2016-02-17
Fixes ToolbarMiddleware.is_cms_request logic

richardasaurus:
`is_cms_request` was checking on app_name instead of url_name, which caused this issue (https://github.com/divio/django-cms/issues/5013), and many more I assume.


richardasaurus:
Will add some tests

richardasaurus:
new pr against correct branch https://github.com/divio/django-cms/pull/5015

PR #5015, opened 2016-02-17
Fixes ToolbarMiddleware.is_cms_request logic

richardasaurus:
`is_cms_request` was checking on app_name instead of url_name, which caused this issue (https://github.com/divio/django-cms/issues/5013), and many more I assume.

Will add tests.

### DIFF
--- a/test_requirements/requirements_base.txt
+++ b/test_requirements/requirements_base.txt
@@ -26,4 +26,4 @@ sphinx
 sphinxcontrib-spelling
 pyflakes
 pyenchant
-
+mock>=1.3.0,<2.0.0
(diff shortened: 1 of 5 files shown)


evildmp:
Thanks for this.

Can you also add a single line to the [changelog](https://github.com/divio/django-cms/blob/develop/CHANGELOG.txt) please, so that we have a record in there too?

richardasaurus:
@evildmp done, hopefully correctly

test_requirements/requirements_base.txt
```diff
@@ -26,4 +26,4 @@ sphinx
 sphinxcontrib-spelling
 pyflakes
 pyenchant
-
+mock>=1.3.0,<2.0.0
```
yakky: I respectfully reject the use of mock in the django CMS testsuite
Especially for everything using requests we definitely need **real** object to have useful tests

richardasaurus: I wouldn't call these tests useless, feel free to update them to real objects. It's not simple to figure out how to do so with your test layout in that module.

yakky:
Alternative tests on #5022 . Thanks for contributing this

PR #4967, opened 2016-01-28
Tuck the language toolbar into sub menus so smaller screens can access

doctormo:
On inkscape.org we have many languages active at the same time. The menu for language is very long and often can not fit on a screen 900px high. This patch tucks away each of the functions into it's own sub-menu with the break appearing if any of the menus will be there.

Issue #3742, opened 2015-01-10
Toolbar CSS: many languages make the bottom languages unaccessible


Moini:
Hi,
when the languages dropdown gets too long, the bottom languages flow over the bottom of the browser window and are only accessible if I zoom out (see screenshots).

![normal](https://cloud.githubusercontent.com/assets/3240233/5691777/665b3f80-98d7-11e4-8115-ce5db623ff94.png)
![zoomed_out](https://cloud.githubusercontent.com/assets/3240233/5691778/66608346-98d7-11e4-9ca3-15bc519b3c43.png)

mkoistinen:
Seems like moving the "Delete..." and "Copy all..." into sub-menus would go a long way to solving this.

mkoistinen:
Fixed by #5026

Moini:
Thanks to everyone involved! Is there a screenshot available somewhere? I'd love to know what the description means :)

vxsx:
@Moini there is a screencast available http://d.pr/v/1brhI 
additionally - multiple copy all and delete translations are now grouped in the nested submenus
<img width="392" alt="screen shot 2016-05-16 at 18 30 57" src="https://cloud.githubusercontent.com/assets/366813/15296194/7df46eb4-1b94-11e6-9e78-1f11bdd90ba0.png">

Moini:
I had watched the screencast, but that didn't show what it would look like in edit mode (which is what my main problem was with), there weren't any submenus. 
So thank you very much for that screenshot @vxsx! :D

Does 'Copy all plugins', when you hover over it, tell the user if it's 'to' or 'from'? (I know it's from, of course ;) )

PR #5026, opened 2016-02-24
Long menus fix

vxsx:
this PR fixes an issue where if toolbar menu has too many items and some of them are not accessible because they don't fit. now if the menu doesn't fit the screen the toolbar becomes non-sticky and menus can be scrolled

video: http://d.pr/v/1brhI

should fix this issue: https://github.com/divio/django-cms/issues/3742 
https://github.com/divio/django-cms/pull/4967 added a nice usablility touch by grouping language items together so they don't take as  much space, but this solves the problem for any generic menu


FinalAngel:
LGTM

PR #5029, opened 2016-02-25
Pagetree refresh node

vxsx:
After moving we have to refresh nodes, since the conditions for publishing/unpublishing can be changed
video: http://d.pr/v/1ayaS


FinalAngel:
LGTM /cc @mkoistinen

PR #5031, opened 2016-02-25
Refresh the state of the moved node and descendants...

mkoistinen:
...  to reflect the publishing state of the new parent (or root)


stefanfoulis:
:+1:

PR #5032, opened 2016-02-26
[WiP] Skip reverse descriptor

### LINKED ISSUE #5030: prevent reverse relation clash for plugins (back-references)
CMSPlugins are built with multi table inheritance. Every Custom Plugin inherits from the base ``CMSPlugin`` model.
Django automatically creates a reverse relationship from the base model to the custom plugin model, named the same as the model. It is the class name in lowercase.

So for example the ``djangocms_file.File`` Plugin will result in ``CMSPlugin`` having an attribute called ``file``, which is also available on all other plugins in the project.

If someone else now creates a Plugin with a field called ``file``, there is a naming clash. In previous django versions this was sort of ignored and continued to work. But as of Django 1.9 this results in a hard error.

The error will look something like this:
```
cmsplugin_filer_file.FilerFile.file: (models.E006) The field 'file' clashes with the field 'file' from model 'cms.cmsplugin'.
cmsplugin_filer_link.FilerLinkPlugin.file: (models.E006) The field 'file' clashes with the field 'file' from model 'cms.cmsplugin'.
```

### proposed fix

Explicitly set the reverse relation for the back-reference to be ``'+'`` (meaning no back-reference) by default. If a developer needs the back reference, it can be added back on a per plugin basis and it should be encouraged to add the app name as a prefix.

Currently you can do this on a Plugin-by-Plugin basis like this:

```python
class FoobarPlugin(CMSPlugin):
    cmsplugin_ptr = models.OneToOneField(CMSPlugin, related_name='+', parent_link=True)
    foo = models.BooleanField()
```
But it really ought to be the default.

### Related previous discussions:

* https://github.com/divio/django-cms/issues/3204 (Prefixing the Plugin model name would no longer be necessary)
* http://stackoverflow.com/questions/20334029/django-app-cannot-assign-imagetext-link-must-be-a-link-instance/20346355#20346355
* field renaming workarounds:

  * https://github.com/divio/cmsplugin-filer/pull/219
  * https://github.com/divio/djangocms-file/pull/13

* Similar, but different issue. This one is about the Plugin class, not the Plugin model class: https://github.com/divio/django-cms/issues/462

yakky:
Fix #5030 by setting ``related_name='+'`` on the OneToOne relation between ``CMSPlugin`` superclass and its child classes

Currently one tests fails which uses the ``search`` function in ``PageManager`` https://github.com/divio/django-cms/blob/develop/cms/models/managers.py#L50-L81 which is apparently built on top of  reverse descriptor... Is this really something in use nowadays?
If not we can deprecate the this in 3.2.2 and remove in 3.3

### DIFF
--- a/cms/models/managers.py
+++ b/cms/models/managers.py
@@ -68,7 +68,9 @@ class PageManager(PublisherManager):
         plugins = plugin_pool.get_all_plugins()
         for plugin in plugins:
             cmsplugin = plugin.model
-            if hasattr(cmsplugin, 'search_fields'):
+            if (hasattr(cmsplugin, 'search_fields') and
+                    cmsplugin.cmsplugin_ptr.field.rel.related_name and
+                    cmsplugin.cmsplugin_ptr.field.rel.related_name != '+'):
                 for field in cmsplugin.search_fields:
                     qp |= Q(**{
                         'placeholders__cmsplugin__{0}_{1}__{2}__icontains'.format(
--- a/cms/models/managers.py
+++ b/cms/models/managers.py
@@ -70,8 +70,10 @@ class PageManager(PublisherManager):
             cmsplugin = plugin.model
             if hasattr(cmsplugin, 'search_fields'):
                 for field in cmsplugin.search_fields:
-                    qp |= Q(**{'placeholders__cmsplugin__%s__%s__icontains' % \
-                               (cmsplugin.__name__.lower(), field): q})
+                    qp |= Q(**{
+                        'placeholders__cmsplugin__{0}_{1}__{2}__icontains'.format(
+                            cmsplugin._meta.app_label, cmsplugin._meta.model_name, field
+                        ): q})
         if language:
             qt &= Q(title_set__language=language)
             qp &= Q(cmsplugin__language=language)
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -55,7 +55,9 @@ class PluginModelBase(ModelBase):
         if 'cmsplugin_ptr' not in attrs:
             for field  in new_class._meta.fields:
                 if field.name == 'cmsplugin_ptr':
-                    field.rel.related_name = '+'
+                    field.rel.related_name = '{0}_{1}'.format(
+                        new_class._meta.app_label, new_class._meta.model_name
+                    )
 
         # if there is a RenderMeta in attrs, use this one
         # else try to use the one from the superclass (if present)
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -48,6 +48,15 @@ class PluginModelBase(ModelBase):
         # create a new class (using the super-metaclass)
         new_class = super(PluginModelBase, cls).__new__(cls, name, bases, attrs)
 
+        # Set the one-to-one parent-child field to not create the reverse
+        # related descriptor to avoid name clashes for plugins with the same
+        # fieldnames. Skip this check if the parent-child field has been explicitly
+        # defined in the plugin subclass
+        if 'cmsplugin_ptr' not in attrs:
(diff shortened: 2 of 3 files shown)


yakky:
Ping @mkoistinen @stefanfoulis

stefanfoulis:
Would it make sense to only override ``reverse_name`` if the Plugin author did not set anything custom? E.g if someone absolutely needs a reverse name on a specific Plugin, he could define 
```python
cmsplugin_ptr = models.OneToOneField(CMSPlugin, related_name='my_reverse_because_i_need_it_dont_ask_why', parent_link=True)
```

The current implementation always overwrites ``related_name`` with ``'+'``.

yakky:
@stefanfoulis https://github.com/divio/django-cms/pull/5032/files#diff-499eb37151a864b95e0f1e9e4f7ac050R55 already skips redefining it if the model does define a custom cmsplugin_ptr
We can even ovverride it with a reverse with a different name instead of just removing the reverse

stefanfoulis:
@yakky :+1: ok. I didn't realise that the custom field on the subclass would override this.

yakky:
@stefanfoulis providing a default related_name should be a more compatible solution.
This is going to break some code, but we have little choice. This will need proper documentation

stefanfoulis:
@yakky yes, I agree. I'll bake up a pull-request that is a bit smarter about figuring out the ``reverse_name`` in the search method and does not try to filter by it, if there is no reverse.

And a reverse name (that also works when we eventually change to full notation of the plugin type).

mkoistinen:
I think this is a sensible implementation. Using the app_label_model_name should significantly reduce collisions and if we add documentation of the issue and illustrate best practices we'll be fine.

yakky:
Changing related_name on the fly it's more complicated than this. I need to go back to the drawing board to find a better solution

stefanfoulis:
@yakky what is the problem with the current implementation?

yakky:
the reverse descriptor it's not always there in some code paths.

cms/models/managers.py
```diff
@@ -68,10 +68,14 @@ def search(self, q, language=None, current_site_only=True):
...
+            if (hasattr(cmsplugin, 'search_fields') and
+                    cmsplugin.cmsplugin_ptr.field.rel.related_name and
+                    cmsplugin.cmsplugin_ptr.field.rel.related_name != '+'):
                 for field in cmsplugin.search_fields:
-                    qp |= Q(**{'placeholders__cmsplugin__%s__%s__icontains' % \
-                               (cmsplugin.__name__.lower(), field): q})
+                    qp |= Q(**{
+                        'placeholders__cmsplugin__{0}_{1}__{2}__icontains'.format(
+                            cmsplugin._meta.app_label, cmsplugin._meta.model_name, field
+                        ): q})
```
stefanfoulis: here we could dynamically use ``cmsplugin.cmsplugin_ptr.field.rel.related_name``, so it does not break if we change the recipe to build related name later (or the user overrides it).

stefanfoulis:
@yakky do you mean ``cmsplugin_ptr`` is not always available?
Maybe if the base model of the plugin is used instead of the subclass. I guess in that case there is no ``cmsplugin_ptr``. I'll do some tinkering.

stefanfoulis:
@yakky I'm experimenting here: #5048

stefanfoulis:
@yakky I was unable to find the case where your approach does not work. For me all seems fine with this implementation.

yakky:
@stefanfoulis is this needed anymore?

stefanfoulis:
@yakky I have incorporated (and extended) the fix for the tests and the search method. Those changes are now in ``3.2.2`` and on ``develop``. I'll work on refining the automatic renaming fix next. But I'll cherry-pick what I need from here. So this PR can be closed.

PR #5035, opened 2016-02-26
Improved menu modifiers/page extension docs

evildmp:
The page extension documentation provided misleading information about using menu modifiers; the modifiers documentation needed some improvements too.

### DIFF
--- a/docs/how_to/extending_page_title.rst
+++ b/docs/how_to/extending_page_title.rst
@@ -10,31 +10,47 @@ an icon for every page) by using the extension models:
 respectively.
 
 
-What's the difference?
-======================
+************************
+Title vs Page extensions
+************************
 
 The difference between a **page extension** and a **title extension** is related to the difference
-between the ``Page`` and ``Title`` models. Titles support pages by providing a storage mechanism,
-amongst other things, for language-specific properties of ``Pages``. So, if you find that you need
-to extend the page model in a language-specific manner - for example, if you need to create
-language-specific keywords for each language of your pages - then you may need to use a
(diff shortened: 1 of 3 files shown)


docs/how_to/extending_page_title.rst
```diff
@@ -264,6 +203,23 @@ relationship available, hence the use of the ``{% if ... %}...{% endif %}``
 above.
 
 
+With menus
+===========================
+
+Like most other Page attributes, extensions are not represented in the menu ``NavigationNodes``,
```
ojii: Maybe emphasis/bold the "not" here.

ojii:
Section about using extensions in menus is a lot clearer now. Great work.

docs/how_to/menus.rst
```diff
@@ -175,26 +178,31 @@ To make your modifier available, it then needs to be registered with
...
                 return nodes
-            count = 0
+            # otherwise loop over the nodes
             for node in nodes:
-                node.counter = count
-                count += 1
+                # does this node represent a Page?
+                if node.attr["is_page"]:
+                    # if so, put its changed_by attribute on the node
+                    node.changed_by = Page.objects.get(id=node.id).changed_by
```
czpython: not sure if it's a good example, given that it can result in many queries. Also shouldn't we encourage use of the attrs attribute of a node? so ``node.attrs['changed_by']``?

evildmp: @czpython This is nothing compared to the queries that I've put on menus! When you need an attribute, you need it.

Why should we encourage the use of ``attr``? Is it just to avoid potential conflicts with the names of existing attributes?

PR #5042, opened 2016-02-29
fix test using reverse descriptor

stefanfoulis:
cherry-pick of yakky/django-cms@0db791489e2666de505d8e943c8300c42ed65401 

It fixes an error in the testsuite resulting from this change in djangocms-file: divio/djangocms-file@a67c4e4c4edb8734c64ad866036a45315e1020e2


yakky:
@stefanfoulis I think this test https://github.com/divio/django-cms/blob/f1e04a539d9df7df19fc4412b8f80f3ea471ea3e/cms/tests/test_plugins.py#L926 will also fail (due this this https://github.com/divio/django-cms/blob/6918793ef0342e44e33a37ba950d18d1593e3d18/cms/models/managers.py#L50) which exploits the reverse relations for basic search

stefanfoulis:
replace by #5048

PR #5043, opened 2016-02-29
Paste helpers

vxsx:
after pasting the helpers were showing up again ( because now we refresh nodes, and refreshing the nodes would trigger check helpers)
also reproducible if you paste and then open close some other node - they would appear again
clearing cache helps to remove leftover state


FinalAngel:
LGTN

vxsx:
Merging because tests are failing anyway for the moment

PR #5044, opened 2016-02-29
don't fail on other tests if test po fails

stefanfoulis:
previously if this test failed, it would leave the current directory
in a inconstant state (pointing to a temporary directory that is
deleted).
This caused the following error on many tests in the testsuite:

cwd = os.getcwd()
OSError: [Errno 2] No such file or directory

After this fix the testsuite still fails, but only on the actual tests that still have problems.


yakky:
LGTM

PR #5047, opened 2016-03-01
clearer error reporting for po tests (gettext)

stefanfoulis:
* show a hint if gettext is not installed
* just let the exception bubble up if something goes wrong

### DIFF
--- a/cms/tests/test_po.py
+++ b/cms/tests/test_po.py
@@ -7,12 +7,19 @@ import os
 import shutil
 import subprocess
 import sys
+import time, warnings
 
 THIS_DIR = os.path.dirname(__file__)
 SOURCE_DIR = os.path.abspath(os.path.join(THIS_DIR, '..', 'locale'))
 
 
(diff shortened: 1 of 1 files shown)


cms/tests/test_po.py
```diff
@@ -7,12 +7,19 @@
...
 THIS_DIR = os.path.dirname(__file__)
 SOURCE_DIR = os.path.abspath(os.path.join(THIS_DIR, '..', 'locale'))
 
 
 def compile_messages():
+    # check if gettext is installed
+    try:
+        pipe = subprocess.Popen(['msgfmt', '--version'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    except OSError as e:
+        raise CommandError('unable to run msgfmt (gettext) command. you probably don\'t have gettext installed. {}'.format(e))
```
evildmp: Capital ``U`` for ``Unable``?

PR #5048, opened 2016-03-01
[WIP] namespaced plugin reverse name

stefanfoulis:
this is #5032 rebased on ``release-3.2.x`` and with some additional tweaks.

### DIFF
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -55,7 +55,9 @@ class PluginModelBase(ModelBase):
         if 'cmsplugin_ptr' not in attrs:
             for field  in new_class._meta.fields:
                 if field.name == 'cmsplugin_ptr':
-                    field.rel.related_name = '+'
+                    field.rel.related_name = '{0}_{1}'.format(
+                        new_class._meta.app_label, new_class._meta.model_name
+                    )
 
         # if there is a RenderMeta in attrs, use this one
         # else try to use the one from the superclass (if present)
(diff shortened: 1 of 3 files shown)


cms/models/pluginmodel.py
```diff
@@ -48,6 +48,17 @@ def __new__(cls, name, bases, attrs):
...
         new_class = super(PluginModelBase, cls).__new__(cls, name, bases, attrs)
 
+        # Set the one-to-one parent-child field to not create the reverse
+        # related descriptor to avoid name clashes for plugins with the same
+        # fieldnames. Skip this check if the parent-child field has been explicitly
+        # defined in the plugin subclass
+        if 'cmsplugin_ptr' not in attrs:
+            for field  in new_class._meta.fields:
+                if field.name == 'cmsplugin_ptr':
+                    field.rel.related_name = '{0}_{1}'.format(
```
czpython: wouldn't it be better to leverage django's supported syntax https://docs.djangoproject.com/en/1.7/topics/db/models/#be-careful-with-related-name.

I wonder if it would work at this stage though..

mkoistinen: +1 @czpython.

czpython: I retract my comment.
Django's %(app_name) %(model_name) is broken on 1.6 :cry:

stefanfoulis: @czpython in django-cms 3.3 we're dropping Django 1.6 support. So if it works in Django 1.8+, you can do this :-)

mkoistinen:
@stefanfoulis do we need to pull some of this into #5145?

czpython:
This was fixed by b8bfcf3263a33b2fbd4f54c75eae2355f5040cb5 and dfe49744ca36d853f4cb33284f41506dcc7916b8

PR #5050, opened 2016-03-01
fix page search queries

stefanfoulis:
>This contains the parts of divio/django-cms#5032 and divio/django-cms#5048 that don't cause backwards incompatibility, but do fix bugs related to working around the problem (so you can manually set ``related_name`` on your plugins).

``Page.objects.search()`` dynamically queries all the plugins that define
``search_fields``. The old code did not handle cases where the ``related_name``
of ``cmsplugin_ptr`` was changed, omitted or did not match its
``related_query_name``.

This changes makes the method ignore any plugins that have removed
``related_name`` and handles different ``related_query_names`` correctly.


cms/models/managers.py
```diff
@@ -68,10 +68,23 @@ def search(self, q, language=None, current_site_only=True):
         plugins = plugin_pool.get_all_plugins()
         for plugin in plugins:
             cmsplugin = plugin.model
-            if hasattr(cmsplugin, 'search_fields'):
+            if not (
+                hasattr(cmsplugin, 'search_fields') and
+                hasattr(cmsplugin, 'cmsplugin_ptr')
```
czpython: do we need to check if plugin has cmsplugin_ptr? cms assumes it's there in many parts of code.

stefanfoulis: yes, it's needed. A plugin can also be the bare base model (e.g a Template only plugin). In that case there is no ``cmsplugin_ptr``.

stefanfoulis:
testsuite is failing on Django 1.6 and 1.7. It seems the example models I added for the tests don't work like this.

stefanfoulis:
test models fixed, commits squashed, tests pass.

PR #5052, opened 2016-03-02
fix page search queries

stefanfoulis:
Page.objects.search() dynamically queries all the plugins that define
search_fields. The old code did not handle cases where the related_name
of cmsplugin_ptr was changed, omitted or did not match its
related_query_name.

This changes makes the search method ignore any plugins that have
removed related_name and handles different related_query_names
correctly.


yakky:
@stefanfoulis how about merging this in develop by forward porting from 3.2.x?
There's a lot of things to port from 3.2 to develop

stefanfoulis:
@yakky this PR is cherry-picked from the commits already on ``release/3.2.x``

yakky:
As martin released 3.2.2 I merged release/3.2.x in develop, so this is not needed anymore

PR #5067, opened 2016-03-04
Fixes issue 5054

### LINKED ISSUE #5054: Reversion not created through API calls
Creating pages through the API does **not** create revision history.
This is not really a bug as the API simply never implemented any kind of revision support.

I don't agree with adding revision logic to the ``cms.api`` but I think we can expose ``make_revision_with_plugins`` from [helpers](https://github.com/divio/django-cms/blob/develop/cms/utils/helpers.py#L39) and use it the api provided that the user explicitly wants to create a revision for a page with a new option like ``create_page(with_revision=True)``.

@divio/django-cms-core What do you guys think?

czpython:
Fixes #5054 
Added new ``with_revision`` parameter to both ``create_page`` and ``create_title``.
By default this parameter is ``False``.
Raises ``ImproperlyConfigured`` error if ``with_revision`` is ``True`` but reversion is not enabled.

@yakky @mkoistinen please review


yakky:
LGTM
@evildmp I think we can safely add 'clickjacking' to dictionary (the only failing test)

mkoistinen:
LGTM

PR #4930, opened 2016-01-18
Small documentation fixes

mcavigelli:
Some fixes I have found while reading the docs.
Unfortunately I was not able to run the *make spelling*.

Make with virtualenv caused some access denied errors.

Matthias

PR #5077, opened 2016-03-09
Fixed numerous small documentation issues

evildmp:
* Fixed link to anchor in same document
* Fixed sentence
* cd into doc before pip
* Changed link to new Sphinx homepage
* Following once should be sufficient
* Removed formatting for code


evildmp:
Supersedes #4930

PR #5080, opened 2016-03-10
Use a method to get CMSApp.urls, to allow passing a page object to it

yakky:
This a proposal to change the CMSApp a bit, to allow defining urlconfs base on the page the apphook is attached to
As of now code does not work :), but I'm interested on the feedback over the general idea


yakky:
Ping @divio/django-cms-core for comments

stefanfoulis:
@yakky great idea. What are some example use-cases?

yakky:
@stefanfoulis when an application needs to use different urlconf when apphooks-config / it's attached multiple times : blog uses a set of urlconfs, news a different one. By  checking the instance_name the "concrete" get_urls() method of the apphook can provide a different set of urls.
The method is obviously meant to be redefined in child classes to use the page argument.

yakky:
@divio/django-cms-core any more comment about this?

cms/app_base.py
```diff
@@ -32,3 +43,17 @@ def get_config(self, namespace):
...
+        warnings.warn('Accessing CMSApp.urls directly is deprecated, '
+                      'and it will be removed in version 3.5; CMSApp.get_urls method',
+                      DeprecationWarning)
+        return self._urls
+
+    @legacy_urls.setter
+    def urls(self, value):
+        self._urls = value
+
+    def get_urls(self, page=None):
```
mkoistinen: Since this is all about flexibility, perhaps it makes sense to include `**kwargs` here?

mkoistinen:
I'm a fan of the concept and the implementation. I made a small suggestion too. =)

yakky:
@stefanfoulis @mkoistinen one step beyond. At this point only tests should be missing

yakky:
@divio/django-cms-core complete on my side, please review

mkoistinen:
@yakky any chance this can get passing tests and another approval asap?

czpython:
Ok to merge

PR #5095, opened 2016-03-21
Better cache control

mkoistinen:
In earlier versions of the CMS a plugin author had only one mechanism to affect the cache-ability of the rendered output: setting the CMSPluginBase attribute `cache` to False. The result of this would be that any placeholder, and thus page that contained the plugin would emit the HTTP header `Cache-Control: no-cache, max-age=0, must-revalidate, no-store`.

While this was useful, it is essentially the "nuclear option"; only really useful when the plugin is delivering unique content for every request. In any other scenario, this is overkill and simply degrades the performance of the resulting site.

A useful scenario to imagine is a CalendarPlugin which always displays the current month. In earlier versions of the CMS, in order to achieve this, the plugin would have to set `cache = False` rendering the entire page never cached, just so that it would be updated once a month.

In order to address this sort of use-case, this PR augments the CMSPluginBase with a default method `get_cache_expiration(request, instance)` which, by default, returns `None`. The CalendarPlugin developer, (for example) however, can override this to return the last second of the current month as a (timezone-aware) `datetime` instance. This tells the CMS' caching sub-system that any page containing this plugin is cache-able––at most––until the last second of the current month.

If the `CMS_CACHES['content']` setting is greater than the time remaining until the end of the month and there are no other plugins, then indeed, the page will expire at the end of the month.

The `get_cache_expiration()` may also return a `datetime.timedelta` instance or just an integer number of seconds, which will serve the same purpose, but in a manner relative to the response time of the page.

As `get_cache_expiration()` is a method, with access to `instance` (the CMSPlugin instance), a full range of options are available to the plugin developer, allowing expiration that is determined at runtime, if required.


cms/models/placeholdermodel.py
```diff
@@ -299,4 +305,79 @@ def actions(self):
             self._actions_cache = getattr(field, 'actions', PlaceholderNoAction())
         return self._actions_cache
 
+    def get_expiration(self, request, response_timestamp):
```
czpython: please rename to ```get_cache_expiration```

cms/plugin_base.py
```diff
@@ -174,6 +174,42 @@ def get_require_parent(cls, slot, page):
     def parent(self):
         return self.cms_plugin_instance.parent
 
+    def get_expiration(self, request, instance):
```
czpython: please rename to ```get_cache_expiration```

cms/tests/test_cache.py
```diff
@@ -73,51 +84,311 @@ def test_no_cache_plugin(self):
...
+            'django.middleware.cache.UpdateCacheMiddleware',
+            'django.middleware.cache.CacheMiddleware',
+            'django.middleware.cache.FetchFromCacheMiddleware',
+        ]
+        mw_classes = [mw for mw in settings.MIDDLEWARE_CLASSES if mw not in exclude]
+
+        # from django.core.cache import cache; cache.clear()
+
+        with self.settings(MIDDLEWARE_CLASSES=mw_classes, CMS_IGNORE_PLUGIN_CACHE_ATTRIBUTE=['LegacyCachePlugin', ]):  # noqa
+                page1.publish('en')
```
czpython: indentation is off here

yakky:
I'll check this tomorrow

yakky:
Why having an additional option? We can simply check the return value of ``get_cache_expiration`` if None, it's not implemented in the plugin and we can safely use the ``cache`` attribute, if it's anything else the new interface it's supported and wins over the attribute

docs/how_to/custom_plugins.rst
```diff
@@ -198,9 +195,23 @@ is ``True`` (the default):
...
+period of validity of the content rendered by the plugin. A value of 0 means
+"do not cache". This method can return a future (timezone-aware) date and time,
+a ``timedelta`` or can simply return the number of seconds that the content
+is cache-able.
+
+This method is not required. By default, a plugin will not affect the cache-
+ability of the page it is on. In the `HelloPlugin` above,
+:meth:`get_cache_expiration` was used to return 0 ("no-cache") because we want
+each visitor to see output that is specific to him or her, we need to tell the
+cms to not cache any pages where this plugin is used.
```
yakky: I'd still document cache here, and eventually mark as deprecated

yakky:
I'm not 100% sure we should deprecate the ``cache`` attribute: it's an easy way to implement an "always fresh" plugin state (say for example a plugin rendering a form with POST method): it's rude, but if you need it fresh, you can't have a cached page, and it may save a few queries

czpython:
I agree with @yakky, we should continue to support an easy way to disable the cache for the plugin.

czpython:
Will continue my review tomorrow :sleepy:

mkoistinen:
@yakky re:
1. keeping the `cache` attribute on the CMSPluginBase and
2. not bothering with CMS_IGNORE_PLUGIN_CACHE_ATTRIBUTE

Agreed. Will adjust PR.

mkoistinen:
@django-cms-core - I'd like to start seeing some :+1: or "LGTM" or reasons why we shouldn't move forward on this.

CHANGELOG.txt
```diff
@@ -4,6 +4,7 @@
 * Set the default value of CMSPlugin.position to 0 instead of null
 * Refactor language menu to allow for better integration with many languages
 * Management commands has been completely refactored for better consistency
+* Provides the ability to declare expiration periods on a per-plugin basis
```
czpython: to declare **cache** expiration periods

cms/cache/page.py
```diff
@@ -29,41 +34,68 @@ def _page_cache_key(request):
...
+            not hasattr(request, 'toolbar') or (
+                not request.toolbar.edit_mode and
+                not request.toolbar.show_toolbar and
+                not request.user.is_authenticated()
+            )):
+
+        # This *must* be TZ-aware
+        timestamp = now()
+
+        placeholders = request.current_page.get_placeholders()
```
czpython: I would add a check above to make sure request.current_page points to an actual page `bool(request.current_page)` because it can be None

cms/models/placeholdermodel.py
```diff
@@ -299,4 +306,81 @@ def actions(self):
...
+        Returns the number of seconds (from «response_timestamp») that this
+        placeholder can be cached. This is derived from the plugins it contains.
+
+        This method must return: EXPIRE_NOW <= int <= MAX_EXPIRATION_IN_SECONDS
+
+        :type request: HTTPRequest
+        :type response_timestamp: datetime
+        :rtype: int
+        """
+        lang = getattr(request, 'lang', None)
```
czpython: where is this `lang` attribute set?

cms/models/placeholdermodel.py
```diff
@@ -299,4 +306,81 @@ def actions(self):
...
+
+        min_ttl = MAX_EXPIRATION_TTL
+        for plugin_item in self.get_plugins(lang):
+            instance, plugin = plugin_item.get_plugin_instance()
```
czpython: a placeholder with 20 plugins will result in 20 queries to the db because of this line.
I recommend we don't get the plugin instance unless it's **really** necessary.
This means, we can pass the plugin id and leave it to the plugin authors to get an instance if they needed for computing the expiration

cms/cache/page.py
```diff
@@ -29,41 +34,68 @@ def _page_cache_key(request):
...
+
+        # Checks if there's a plugin using the legacy "cache = False"
+        if all(ph.cache_placeholder for ph in placeholders):
```
czpython: I think this might not work.. `placeholders` come from a call to `get_placeholders()` which will fetch all placeholders for a page at the db level and return fresh instances unless it's called twice on the same page instance.
The only place in the CMS where the `cache_placeholder` attribute is set is in the `downcast_plugins` utility, this means that `cache_placeholder` will always be True here.

Also as mentioned above, `get_placeholders` only looks at the db vs `rescan_placeholder` which actually looks at the template.

mkoistinen: Ooh, you're quite right.

cms/models/placeholdermodel.py
mkoistinen: That would seem logical, but I can't seem to demonstrate that this is quite correct. I've tried with both caching enabled and disabled but the number of queries seem does grow with the number of plugins on the page when in edit mode, but the exact same number of queries are executed with `return 0` inserted before like 323 above (disabling the line in question), and only when the user is logged-in and in edit mode. Rolling back to `develop` and trying again, exhibits the same behavior.

Even if I was adding these queries, which I don't think I am, they are impacting performance by about 225 µsecs per plugin on my local system. (NOTE: that is **micro**seconds, not **milli**seconds).

Considering the fact that this line is clearly _not_ adding any queries, I think its is worth it to keep them in, to maintain the same(-ish) signature as the render() method.

cms/models/placeholdermodel.py
mkoistinen: Changed to `getattr(request, 'LANGUAGE_CODE', None)`

cms/cache/page.py
mkoistinen: Irrelevant due to other changes.

cms/models/placeholdermodel.py
czpython: @mkoistinen I'm a bit curious to know how you measured the performance here.
I've used the django debug toolbar's `sqldebugshell` to isolate the performance for this specific method on the Placeholder model.

Here's an example with the current line:

https://gist.github.com/czpython/32d639e3cd17da64f7ba

And this one is with that line refactored to match my suggestion:

https://gist.github.com/czpython/cd3b52e298e424b15de2

mkoistinen: I was using DDT too, but rechecked things after seeing your results. I blew away my *.pyc files, and I get the results you see. (I did say it seemed logical! =)).

Anyway, digging into it a little, I think I've now got a better solution. a lot of great work is done in `cms.utils.plugins.downcast_plugins()` and I've now tweaked the code to leverage this to not re-query what we've already queried. The result is the performance I was seeing above, but with fresh `pyc` files =).

cms/models/placeholdermodel.py
```diff
@@ -299,4 +306,97 @@ def actions(self):
...
+            """
+            if hasattr(self, '_all_plugins_cache'):
+                for instance in self._all_plugins_cache:
+                    plugin = instance.get_plugin_class_instance()
+                    yield instance, plugin
+            else:
+                for plugin_item in self.get_plugins(lang):
+                    yield plugin_item.get_plugin_instance()
+
+        language = getattr(request, 'LANGUAGE_CODE', None)
```
czpython: use ``get_language_from_request`` from ``cms.utils``

czpython:
:+1:

mkoistinen:
This was rebased and then committed manually.

PR #5097, opened 2016-03-21
Fix toolbar order

### LINKED ISSUE #5023: Clipboard items are not necessarely the first
There is a wrong assumption here: https://github.com/divio/django-cms/blob/release/3.2.x/cms/static/cms/js/modules/cms.plugins.js#L21-L22
Clipboard items are not necessarly the first items of their kind: e.g. when multiple toolbars register for ``render_addons`` (https://github.com/divio/django-cms/blob/develop/cms/toolbar/toolbar.py#L144)

yakky:
and fail gracefully if clipboard plugin is not the topmost plugin

Fix #5023 

If clipboard plugins are not initialized before all the other plugins ``CMS.Plugin._checkIfPasteAllowed`` fails as clipboard plugin settings are not initialized yet.
This is normally not an issue as clipboard plugins are initialized when rendering toolbars addons, which is run before the "normal" placeholder plugins.
In case another toolbar addon defines plugins, is no longer guaranteed that clipboard plugins are rendered first.

This PR addresses the order issue by using a custom toolbar ordering (ie: ``BasicToolbar`` is always rendered first in ``render_addons``), and in ``_checkIfPasteAllowed`` clipboard plugins are checked before reading settings; in case clipboard plugins are not initizialized yet, clipboard is skipped for the plugins that are initialized before the clipboard.


yakky:
@vxsx / @FinalAngel please review the frontend part

yakky:
Few unrelated commits added to fix tests

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -327,17 +327,22 @@ var CMS = window.CMS || {};
                 }
 
                 var bounds = this.options.plugin_restriction;
-                var type = clipboardPlugin.data('settings').plugin_type;
-                var parent_bounds = $.grep(clipboardPlugin.data('settings').plugin_parent_restriction, function (r) {
-                    // special case when PlaceholderPlugin has a parent restriction named "0"
-                    return r !== '0';
-                });
-                var currentPluginType = this.options.plugin_type;
+                if (clipboardPlugin && clipboardPlugin.data('settings')) {
```
vxsx: the case when clipboardPlugin doesn't exist is already covered on line 317

vxsx:
I still don't understand how would the order of toolbars affect the clipboard. As long as the clipboard is rendered in any place it should still work, no?

cms/static/cms/js/modules/cms.plugins.js
vxsx: can just check for `clipboardPlugin.data('settings')`, clipboardPlugin is always truthy since it's a jquery object

yakky: Me and my js skills ... :)
Code fixed

yakky:
@vxsx thanks for the review

vxsx:
np, frontend lgtm

PR #5098, opened 2016-03-22
sane add plugin cleanup

czpython:
Fixes for #4381

* modify the signature for post_add_plugin to not require a placeholder
* removed force_unicode as it's no longer needed after 1.8
* fixes bug on plugin position calculation
* fixes some failing tests
* fixes potential corrupt parent/child relationships
* moved add check logic to form

**FE**
@vxsx so seems I can't get this to work with ckeditor.
We use a ``POST`` request [here](https://github.com/divio/djangocms-text-ckeditor/blob/master/djangocms_text_ckeditor/static/djangocms_text_ckeditor/ckeditor_plugins/cmsplugins/plugin.js#L197) instead of  ``GET`` request.
I tried changing it but then another issue comes up which is that the js expects the response to have things like url key and such, instead the response is a full html page that should be rendered in the modal.

**BE**
@divio/django-cms-core I would like your feedback on this, specially @ojii.

### DIFF
--- a/cms/tests/test_plugins.py
+++ b/cms/tests/test_plugins.py
@@ -40,7 +40,7 @@ from cms.test_utils.project.pluginapp.plugins.validation.cms_plugins import (
 from cms.test_utils.testcases import (
     CMSTestCase, URL_CMS_PAGE, URL_CMS_PLUGIN_MOVE, URL_CMS_PAGE_ADD,
     URL_CMS_PLUGIN_ADD, URL_CMS_PLUGIN_EDIT, URL_CMS_PAGE_CHANGE,
-    URL_CMS_PLUGIN_REMOVE, URL_CMS_PAGE_PUBLISH, URL_CMS_PLUGIN_DELETE, URL_CMS_PLUGINS_COPY)
+    URL_CMS_PLUGIN_REMOVE, URL_CMS_PAGE_PUBLISH, URL_CMS_PLUGINS_COPY)
 from cms.test_utils.util.fuzzy_int import FuzzyInt
 from cms.toolbar.toolbar import CMSToolbar
 from cms.utils.conf import get_cms_setting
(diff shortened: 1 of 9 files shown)


czpython:
@vxsx this is the only test failing https://travis-ci.org/divio/django-cms/jobs/120096594, please have a look.

czpython:
@vxsx I've added the fix for the popup error but still this fails https://travis-ci.org/divio/django-cms/jobs/120152910

vxsx:
@czpython fixed

czpython:
Thanks. This is history @vxsx , for the first time in over a year sane-add-plugin passes tests! :tada: :dancers:

czpython:
Tests are passing!! @ojii

ojii:
whatyearisit

ojii:
Seriously though, great work.

Without me having to go through months of commits, what caused the tests to fail (only on CI) in the end?

cms/tests/test_plugins.py
```diff
@@ -280,47 +280,6 @@ def test_plugin_breadcrumbs(self):
         self.assertTrue('/edit-plugin/%s/'% column.pk, text_breadcrumbs[1]['url'])
         self.assertTrue('/edit-plugin/%s/'% text_plugin.pk, text_breadcrumbs[2]['url'])
 
-    def test_add_cancel_plugin(self):
```
vxsx: btw maybe it makes sense to do the test that the old behavior doesn't actually happen?

czpython: don't think so.. it's very straightforward so should be fine with testing the new behavior

czpython: this test should actually be in ckeditor

czpython:
@ojii honestly I'm not sure.
The way I approached it was to read all your changes and simply refactor what I saw could cause issues.
If I had to guess I'd say it was https://github.com/divio/django-cms/pull/5098/commits/ba1db7057d6ff87fd0fe5481b24f19c09ca7d59d or maybe https://github.com/divio/djangocms-text-ckeditor/commit/65934f6cfa1801c31c6ff45a9bb202bf36ade0ba

ojii:
@czpython okay, thanks. can't believe this might eventually get merged :D

PR #5103, opened 2016-03-24
Unit tests for cms.navigation

vxsx:
22% → 100%

reopen from https://github.com/FinalAngel/django-cms/pull/177

additionally changes the location for downloading phantomjs (reliable one)
and splits up integration tests into 3 buckets so it's faster and less error prone

### DIFF
--- a/cms/tests/frontend/unit/fixtures/toolbar.html
+++ b/cms/tests/frontend/unit/fixtures/toolbar.html
@@ -5,6 +5,61 @@
                 <a href="/" title="cms_version"><span>django CMS</span></a>
             </div>
             <ul class="cms-toolbar-item cms-toolbar-item-navigation">
+                <li>
+                    <a href="">
+                        <span>example.com<span class="cms-icon cms-icon-arrow"></span></span>
+                    </a>
+                    <ul>
+                        <li class=" ">
+                            <a href="/en/admin/cms/page/?page_id=1&amp;language=en" data-rel="sideframe">
+                                <span>Pages...<span class="cms-icon cms-icon-arrow"></span></span>
+                            </a>
+                        </li>
+                        <li class=" ">
+                            <a href="/en/admin/auth/user/" data-rel="sideframe">
+                                <span>Users...<span class="cms-icon cms-icon-arrow"></span></span>
+                            </a>
+                        </li>
+                        <li>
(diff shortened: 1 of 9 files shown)


gulpfile.js
```diff
@@ -175,36 +175,57 @@ gulp.task('tests:unit:watch', function () {
...
+            'dragndrop',
+            'history',
+            'revertLive',
+            'narrowScreen'
+        ]
     ];
+
+    var files = [];
+
+    switch (process.env.INTEGRATION_TESTS_BUCKET) {
```
FinalAngel: this probably might need some comments

gulpfile.js
```diff
@@ -175,36 +175,57 @@ gulp.task('tests:unit:watch', function () {
...
-        'disableToolbar',
-        'dragndrop',
-        'history',
-        'revertLive',
-        'narrowScreen',
-        'clipboard',
-        'modal',
-        'pagetree'
+    var buckets = [
+        [
```
FinalAngel: dunno here what bucket is used for what case

cms/tests/frontend/unit/fixtures/toolbar.html
```diff
@@ -5,6 +5,61 @@
                 <a href="/" title="cms_version"><span>django CMS</span></a>
             </div>
             <ul class="cms-toolbar-item cms-toolbar-item-navigation">
+                <li>
+                    <a href="">
+                        <span>example.com<span class="cms-icon cms-icon-arrow"></span></span>
+                    </a>
+                    <ul>
+                        <li class=" ">
```
FinalAngel: I assume the space in class is cause of copy & paste?

FinalAngel:
:+1: nice work

cms/tests/frontend/unit/fixtures/toolbar.html
vxsx: yup. didn't think it would make sense to prettify fixture

gulpfile.js
vxsx: what do you mean? just three buckets, ordered list of buckets

gulpfile.js
vxsx: added comment, is it more clear that way?

PR #5104, opened 2016-03-24
Added page plugin support.

farooqaaa:
Page plugins can override core page functionalities like page rendering and context.

### DIFF
--- a/cms/page_plugin.py
+++ b/cms/page_plugin.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+from operator import attrgetter
+
+from django.core.exceptions import ImproperlyConfigured
+from cms.exceptions import PluginAlreadyRegistered, PluginNotRegistered
+
+
+class PagePluginBase(object):
+    name = ""
+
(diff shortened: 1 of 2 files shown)


farooqaaa:
Here's an example of updating context:

`from django.template.response import TemplateResponse
from cms.utils import get_template_from_request

class MyPagePlugin(PagePluginBase):
    name = ""

    def render_page(self, request, page, current_language, slug, response):
        context = response.context_data
        context['new_item'] = 'test'

        template_name = get_template_from_request(request, page, no_current_page=True)
        response = TemplateResponse(request, template_name, context)

        from cms.page_rendering import add_xframe_options
        response = add_xframe_options(page, response)

        return response`

yakky:
@farooqaaa could you provide some use cases for which this is needed?
I'm not convinced at all by the design, but knowing a bit the scope of this feature would help clarify it
thanks

czpython:
I think this can be something integrated outside of the core.
Please reopen if you have more info as to what's the scope, we can maybe make it easier to implement outside.

PR #5108, opened 2016-03-28
Bf/raw id fields pages tree

Air-Mark:
There is a bug - when we want to user Page model in a raw_id_fields property, if we open up the dialog to choose a Page and click on any row, nothing happens because there is no JS handler from the "opener" object on Page rows. 
That handler was in the menu_item.html file
https://github.com/divio/django-cms/commits/101fe24d7727756e80ebadbdd137400e022c9aaa/cms/templates/admin/cms/page/tree/menu_item.html
But after the Tree refactoring, that file was removed and it looks like that we forgot to add than handler in current version.


yakky:
@vxsx could you review this? it's better to apply this manually to 3.2.x and then forward port it?

vxsx:
codewise LGTM

https://github.com/divio/django-cms/commit/6216bcc2182ef346f7f79032b4550270585b44a0 original commit (for reference)

maybe we should have a title tooltip as well, or even replace "view" icon with "pick" icon. 
not really sure if it should be onclick on the row since we no longer use click on the page name as means to view it

wdyt @FinalAngel ?
@yakky for me no difference regarding forward-/back-porting, whatever would make most sense to you

yakky:
@vxsx thanks. I'll when everything is ok here I'll merge manually in 3.2 branch
Thanks @Air-Mark for the fix

Air-Mark:
@yakky welcome. Are there any news about the merge?

yakky:
@Air-Mark merged via #5125 thanks for your contribution

PR #5109, opened 2016-03-29
Render Menus from current site fix

berezowski:
current_site(request) returns the site selected in the page admin
dropdown.
This leads to a bug where for example site 1 is rendered
with the menu from site 2, if site 2 is selected in the
field-site-select dropdown


yakky:
@berezowski could you provide more context for the original bug and why this fix is needed?

czpython:
@berezowski Thanks for the report.
I'll submit a separate PR given that the menu system has changed a bit on 3.3

PR #5110, opened 2016-03-29
CMS Toolbar unit tests

vxsx:
49.82% → 92.28%

also some fixes to the toolbar itself

PR #5114, opened 2016-03-30
CMS Sideframe unit tests

vxsx:
67.42% → 100%

waits for https://github.com/divio/django-cms/pull/5110


FinalAngel:
nice work, LGTM

PR #5126, opened 2016-04-01
Port more 3.2.x code

yakky:


docs/conf.py
```diff
@@ -30,7 +30,7 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 #extensions = ['sphinx.ext.autodoc']
 
-extensions = ['djangocms', 'sphinx.ext.intersphinx', 'sphinx.ext.todo']
+extensions = ['djangocms', 'sphinx.ext.intersphinx']
```
vxsx: please don't do this, sphinx.ext.todo is actually useful
this is the correct fix for the sphinx issue https://github.com/divio/django-cms/commit/7979a0620248bc3d82dd35414a342cf7062e001b and it's already in develop

yakky: I'm backporting the fix in 3.2
I think that such test-related fixes should be first applied to earlier versions and then  forwardported, or at least they should be advertised as needing for backporting, otherwise they get lost

PR #5127, opened 2016-04-01
Updated tutorial, added new tutorial sections

evildmp:


evildmp:
Merged after discussion with @Chive

PR #5130, opened 2016-04-02
Include all optional parameters of the show_menu tag

### LINKED ISSUE #5056: Documentation about 'show_menu' templatetag missing two arguments
[Docs](http://docs.django-cms.org/en/develop/reference/navigation.html) about `show_menu` are a bit confusing. It's stated that `show_menu` takes 4 arguments:

>  show_menu takes four optional parameters: start_level, end_level, extra_inactive, and extra_active

When in fact `show_menu` [takes 6 arguments](https://github.com/divio/django-cms/blob/develop/menus/templatetags/menu_tags.py#L108). This is especially confusing, when overriding `menu/menu.html` [template](https://github.com/divio/django-cms/blob/develop/menus/templates/menu/menu.html) where all 6 arguments are used, but docs talk about only 4.

Therefore the docs should probably be supplemented

xpostudio4:
Fixes #5056

### DIFF
--- a/docs/reference/navigation.rst
+++ b/docs/reference/navigation.rst
@@ -30,8 +30,8 @@ show_menu
 The ``show_menu`` tag renders the navigation of the current page.
 You can overwrite the appearance and the HTML if you add a ``menu/menu.html``
 template to your project or edit the one provided with django CMS.
-``show_menu`` takes four optional parameters: ``start_level``, ``end_level``,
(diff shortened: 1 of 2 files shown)


yakky:
Nice, thanks. Will merge manually into 3.2 and then forwardport to develop

yakky:
Merged through abc66b08724e1e2abbce4e8fc1f924f863fe8d60
Thanks for the contribution

Issue #5111, opened 2016-03-29
Can't copy page in admin if CMS_PERMISSION = False


wullerot:
I can't if copy a page if CMS_PERMISSION is set to false. I tested with all defaults from djangocms-installer.

`djangocms -p . mysite`
django==1.8.11
django-cms==3.2.3

In settings I changed CMS_PERMISSION to:
`CMS_PERMISSION = False`

![cantcopypage2](https://cloud.githubusercontent.com/assets/699130/14108687/c94418b6-f5be-11e5-8099-891f9f1e8bca.png)

![cantcopypage3](https://cloud.githubusercontent.com/assets/699130/14108688/c9459376-f5be-11e5-865d-dbd787cb87e4.png)

On page reload the copy disappears

![cantcopypage1](https://cloud.githubusercontent.com/assets/699130/14108686/c9439a12-f5be-11e5-8f28-a6b26b23b31a.png)

pip freeze:
dj-database-url==0.4.0
Django==1.8.11
django-classy-tags==0.7.2
django-cms==3.2.3
django-formtools==1.0
django-reversion==1.8.7
django-sekizai==0.9.0
Django-Select2==4.3.2
django-treebeard==4.0
djangocms-admin-style==1.1.0
djangocms-column==1.6.0
djangocms-file==1.0
djangocms-googlemap==0.4.0
djangocms-inherit==0.1.1
djangocms-installer==0.8.7
djangocms-link==1.7.2
djangocms-picture==0.2.0
djangocms-style==1.7.0
djangocms-teaser==0.2.0
djangocms-text-ckeditor==2.9.1
djangocms-video==1.0.0
html5lib==0.9999999
Pillow==3.1.1
pytz==2016.3
six==1.10.0
tzlocal==1.2.2

czpython:
Verified regression introduced in 3.2

@FinalAngel @vxsx 
In [3.1](https://github.com/divio/django-cms/blob/support/3.1.x/cms/static/cms/js/modules/cms.changelist.js#L592) we do a check for ```that.options.settings.permission``` and only render the copy dialog if this is True, otherwise we proceed with the copy operation.
In [3.2](https://github.com/divio/django-cms/blob/release/3.2.x/cms/static/cms/js/modules/cms.pagetree.js#L442) the conditional is not there and so we try to load the dialog every time even when permissions are off.

vxsx:
Hi @wullerot, can you try to reproduce with latest `release/3.2.x` branch?

[...]

PR #5131, opened 2016-04-02
Bugfix copy page with permissions off

vxsx:
first 3.2.x, needs to be forwardported to develop

supposed to fix #5111

/cc @czpython

PR #5132, opened 2016-04-04
Bugfix pagetree copy page when no permissions

vxsx:
forwardport of https://github.com/divio/django-cms/pull/5131, to fix https://github.com/divio/django-cms/issues/5111


FinalAngel:
LGTM

yakky:
LGTM too. Thanks @vxsx for fixing into 3.2 and forwardporting :)

Issue #5053, opened 2016-03-02
Can't add django cms plugin, haven't necessary permissions.


timur-orudzhov:
Hello. I've got some problems with plugins. I'm a superuser and have all rights, which I could need. But when I try to add a new plugin, there is an error 403 with message:"You have no rights for that". I created a "pages global permissions" for my user, I tryed to create "user-page" user. I uses two django cms versions 3.1.3 and 3.1.5.  However, it doesn't help. And I have no idea what to do.

yakky:
Does this happens on plain CMS pages on apphooks?

timur-orudzhov:
@yakky on plain cms pages

Unysova:
@yakky And there is an interesting thing.The problem occurs only when we use tag "placeholder", when we use "static placeholder", everything is fine

czpython:
@timur-orudzhov @Unysova Can you verify that the page in question does not have any apphook attached to it? There's a known issue where using a `{% placeholder %}` tag inside an apphook rendered template will throw the misleading permission error when trying to add a plugin to it.

Apphook rendered templates are not allowed to have `{% placeholder %}` tags as of 3.0.0

czpython:
Closing due to inactivity.
Issue is likely caused by using a placeholder inside of an apphooked page.

PR #5133, opened 2016-04-04
Add missing permission name

piotrjakimiak:
Permission cms.change_page is missing name when created in migration 0010.
Added migration which update value of that Permission.

### DIFF
--- a/cms/migrations/0015_auto_20160404_1908.py
+++ b/cms/migrations/0015_auto_20160404_1908.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.contrib.auth.models import Permission
+from django.contrib.contenttypes.models import ContentType
+from django.db import migrations, models
+
+
+def forwards(apps, schema_editor):
+    page_model = apps.get_model('cms', 'Page')
+    page_ctype = ContentType.objects.get_for_model(page_model)
+    Permission.objects.filter(
+        name='',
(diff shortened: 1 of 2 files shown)


cms/migrations/0015_auto_20160404_1908.py
```diff
@@ -0,0 +1,34 @@
...
+        name='',
+        codename='change_page', content_type=page_ctype).update(name='Can change page')
+
+
+def backwards(apps, schema_editor):
+    page_model = apps.get_model('cms', 'Page')
+    page_ctype = ContentType.objects.get_for_model(page_model)
+    Permission.objects.filter(
+        name='Can change page',
+        codename='change_page', content_type=page_ctype).update(name='')
```
piotrjakimiak: I'm not sure if backward migration should set empty name.

timur-orudzhov:
@piotrjakimiak Thank you ! Your commit fixed my bug - https://github.com/divio/django-cms/issues/5053

cms/migrations/0015_auto_20160404_1908.py
yakky: @piotrjakimiak Nope, I'd just leave as is when doing backward

yakky:
LGTM. I'll apply this to 3.1 (when first error appeared) and then I'll forward port this

yakky:
@piotrjakimiak thanks for sharing this. I applied to 3.1, will forward port in the next days

PR #5141, opened 2016-04-06
Improve wizard support for i18n projects

mkoistinen:


cms/wizards/views.py
```diff
@@ -28,14 +28,12 @@ class WizardViewMixin(object):
 
     @transaction.atomic
     def dispatch(self, request, *args, **kwargs):
-        self.language_code = get_language_from_request(request, check_path=True)
         response = super(WizardViewMixin, self).dispatch(
             request, *args, **kwargs)
         return response
 
     def get_form_kwargs(self):
```
czpython: you can remove this if we no longer add any kwargs here

czpython:
:+1:

PR #5142, opened 2016-04-06
fixes #5140

### LINKED ISSUE #5140: Page marked as "pending" on move even if unpublished
1. Create root page "sections"
2. Create root page "section 1"
3. Publish both pages
4. Unpublish "section 1"
5. Move "section 1" to be a child of "sections"
6. "section 1" will now be marked as pending

https://drive.google.com/file/d/0B9SwgyV0g1hSc1RSLVoxcVJFNDQ/view

Related to #5107
Verified as a regression of 3.2.
Regression came from https://github.com/divio/django-cms/commit/ec1d6c3c1d666efe3cb6eec5a13f59d81f0fbc42.

czpython:
* Fixed #5140
* Fixed bug where an unpublished child page would be set to pending if it's parent was unpublished.
* Refactored move_page to no longer publish pages.
  instead it now calls ``mark_as_published`` and ``mark_descendants_as_published``.
  This is to avoid executing the whole publication logic on page move which can result in many
  inconsistencies.
* Refactored the logic from ``publish()`` that updates the state of descendant pages
  into a new method called ``mark_descendants_as_published``.
  This is to allow other actions like moving a page to benefit from this operation without
  calling ``publish()``. The refactor removed a lot of dead code.

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -840,14 +840,9 @@ class Page(six.with_metaclass(PageMetaClass, MP_Node)):
                 if not publisher_public.parent_id:
                     publisher_public = page._publisher_save_public(publisher_public)
 
-                # query and caching optimization
-
-                if publisher_public.parent_id and not publisher_public.parent:
-                    page.publisher_public.parent = Page.objects.get(pk=page.publisher_public.parent_id)
-
                 # Check if the parent of this page's
                 # public version is published.
-                if page.publisher_public.parent.is_published(language):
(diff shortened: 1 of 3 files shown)


czpython:
@yakky @mkoistinen please review.
I will backport this into 3.2.x once merged.
This fixes a few critical issues on 3.2.x and thus a new release there should be planned for no later than next week.

mkoistinen:
This looks really good, @czpython. Also, thanks for both retaining existing and writing new comments. Very clear code, easy to follow and thorough. 👍

PR #5145, opened 2016-04-07
Improve plugin related names

mkoistinen:
This is the continuing saga of Django 1.9 vs. cmsplugin_ptr. The changes in this PR were initially contributed by @yakky's #5032.


czpython:
why cms/tests/test_frontend.py → cms/tests/DISABLED_test_frontend.py ?

cms/models/pluginmodel.py
```diff
@@ -48,6 +48,20 @@ def __new__(cls, name, bases, attrs):
...
+        # related descriptor to avoid name clashes for plugins with the same
+        # fieldnames. Skip this check if the parent-child field has been
+        # explicitly defined in the plugin subclass
+        #
+        # NOTE: We only do concrete classes, skipping the abstract ones
+        if 'cmsplugin_ptr' not in attrs and not new_class._meta.abstract:
+            for field in new_class._meta.fields:
+                if field.name == 'cmsplugin_ptr':
+                    app_label = new_class._meta.app_label
+                    model_name = new_class._meta.concrete_model._meta.model_name
```
czpython: huh? are you not already skipping non-concrete classes?

cms/models/pluginmodel.py
```diff
@@ -48,6 +48,20 @@ def __new__(cls, name, bases, attrs):
         # create a new class (using the super-metaclass)
         new_class = super(PluginModelBase, cls).__new__(cls, name, bases, attrs)
 
+        # Set the one-to-one parent-child field to not create the reverse
+        # related descriptor to avoid name clashes for plugins with the same
+        # fieldnames. Skip this check if the parent-child field has been
+        # explicitly defined in the plugin subclass
+        #
+        # NOTE: We only do concrete classes, skipping the abstract ones
+        if 'cmsplugin_ptr' not in attrs and not new_class._meta.abstract:
```
czpython: we should test how this works with proxy models

cms/models/pluginmodel.py
mkoistinen: Yes. strangely, _meta.model_name still comes back with the _deferred_path bit... This fixed it.

czpython: that's very strange.. I advise to look into why it happens.

cms/models/pluginmodel.py
mkoistinen: I've yet to see a CMSPlugin that uses a proxy model... have you?

cms/models/pluginmodel.py
yakky: @mkoistinen could you clarify more this point? What do you mean with ``..._deferred_path`` bit?

cms/models/pluginmodel.py
yakky: I agree, we should explicitly skip proxied models too

yakky:
Currently there is an issue with MTI: the plugins that inherit for a parent plugin overwrites the related_name, which must be kept to the parent one.
I'll send a PR with a test for this, and eventally a solution

yakky:
@mkoistinen @czpython I opened a PR to your branch https://github.com/mkoistinen/django-cms/pull/2

mkoistinen:
This is failing integration tests inside of the call to Django's `get_deleted_objects()`:

https://gist.github.com/mkoistinen/028dd9690d19f6ac1f86a6f791fa4545

Any ideas how to make this work?

mkoistinen:
Django 1.8 and Python 3.4

PR #5147, opened 2016-04-08
Migrate toolbar tests

vxsx:
covers apphook selection
removes already covered tests from test_frontend.py
adds `--no-server` option for local development
adds more info about installation


FinalAngel:
LGTM

PR #5153, opened 2016-04-11
[WiP] Fix position field management

yakky:
Still WiP, but when testing some API operations i found that position and path gets out of sync when not using the frontend editor.
Analysis is not completed yet, I guess there are better strategies to fix this

### DIFF
--- a/cms/admin/placeholderadmin.py
+++ b/cms/admin/placeholderadmin.py
@@ -283,7 +283,6 @@ class PlaceholderAdminMixin(object):
         plugin = CMSPlugin(language=language, plugin_type=plugin_type, position=position, placeholder=placeholder)
 
         if parent:
-            plugin.position = CMSPlugin.objects.filter(parent=parent).count()
             plugin.parent_id = parent.pk
         plugin.save()
         self.post_add_plugin(request, placeholder, plugin)
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -254,9 +254,9 @@ class CMSPlugin(six.with_metaclass(PluginModelBase, MP_Node)):
         super(CMSPlugin, self).move(target, pos)
         self = self.reload()
         try:
-            new_pos = max(CMSPlugin.objects.filter(parent_id=self.parent_id,
+            new_pos = max(CMSPlugin.objects.filter(parent_id=self.parent_id, path__lte=self.path,
                                                    placeholder_id=self.placeholder_id,
-                                                   language=self.language).exclude(pk=self.pk).order_by('depth', 'path').values_list('position', flat=True)) + 1
+                                                   language=self.language).exclude(pk=self.pk).order_by('path').values_list('position', flat=True)) + 1
         except ValueError:
             # This is the first plugin in the set
             new_pos = 0
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -8,7 +8,7 @@ import warnings
 from django.conf import settings
 from django.core.urlresolvers import NoReverseMatch
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
-from django.db import models
+from django.db import models, transaction
 from django.db.models import signals, Model, ManyToManyField
 from django.db.models.base import model_unpickle, ModelBase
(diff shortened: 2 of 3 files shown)


yakky:
IDK why commit 	b37dd94 is in this PR as it's already been merged ...

czpython:
@yakky just so you know I'm working on a refactor in the ``reorder_plugins`` utility. Behavior will stay the same.

yakky:
@czpython ok. What's your opinion on this?

cms/models/pluginmodel.py
```diff
@@ -250,25 +251,23 @@ def save(self, no_signals=False, *args, **kwargs):
...
-                                                   language=self.language).exclude(pk=self.pk).order_by('depth', 'path').values_list('position', flat=True)) + 1
-        except ValueError:
-            # This is the first plugin in the set
-            new_pos = 0
-        self.position = new_pos
-        self.save()
+        plugins = CMSPlugin.objects.filter(
+            parent_id=self.parent_id, placeholder_id=self.placeholder_id, language=self.language
+        ).order_by('path').values_list('pk', flat=True)
+        reorder_plugins(self.placeholder, self.parent_id, self.language, plugins)
```
czpython: On the move endpoint, we already to a calculation of positions so this would be redundant.
I think the logic should be here but then we should add an order parameter to ``move()`` or stick with the utility to not touch an internal of treebeard.
Also ordering by path and then supplying that as order can backfire because they can be out of sync.

cms/admin/placeholderadmin.py
```diff
@@ -283,7 +283,6 @@ def add_plugin(self, request):
         plugin = CMSPlugin(language=language, plugin_type=plugin_type, position=position, placeholder=placeholder)
 
         if parent:
-            plugin.position = CMSPlugin.objects.filter(parent=parent).count()
```
czpython: I guess this relies on the logic being done on the save() method?

czpython:
@yakky I have some concerns but the main one is I don't think this should be done on 3.2.x
Given the refactors being done on 3.3 I suggest we do it there instead.

cms/admin/placeholderadmin.py
yakky: As it's written there it's just plain wrong, and yes, it relies on position being calculated on save

cms/models/pluginmodel.py
yakky: path should be the only relevant indicator of the correct order. In fact, I wonder if we still need position at all

yakky:
I'm fine with moving this to 3.3

cms/models/pluginmodel.py
czpython: I don't think we do but there's the issue of backwards compatibility at the db level (plugins that are out of sync) and also path is a char field and will most likely take longer to sort by than the position field.

yakky: Currently FE editor relies on position to order plugins, while plugin rendering relies on path :)
treebeard apparently does a good job at optimizing indexes and queryes, anyway

czpython: rendering relies on position too https://github.com/divio/django-cms/blob/develop/cms/utils/plugins.py#L132

mkoistinen:
@yakky please re-try some of this now that #4381 has landed.

yakky:
Closing this for now, I'll reopen on develop after further checks

melbic:
We just ran into the same problem working with child plugins. While everything worked fine when adding the plugins for the first time, changing the order would result in correct positions, but not adjusting the  path values. See screenshot.
moka_ _appbuilders_signup_ _public_ _cms_cmsplugin

Nirazul:
Any updates on this? This issue seems to persist throughout v3.4.2.

chschuermann:
We can reproduce the problem even with plugins on the first level. If we add 3 plugins and move the order in the structure mode, the path gets out of sync with the position.

benzkji:
same on 3.4.4. plugin tree operations that are done with the cms api seem to use the treebeard, and forget about position...what should be ok, but is not right now, as there still is the field `position.` as this one is close: is there a new issue for this bug?

PR #5158, opened 2016-04-12
Fix loader that never hides

vxsx:
fixes 2 minor ui issues:

first one is that sometimes loader can be dark gray even though should always be blue
screen shot 2016-04-12 at 13 16 12

and the one when quickly closed modal will never stop loader from spinning
broken: http://d.pr/v/1ayJ2
fixed: http://d.pr/v/1ah46


FinalAngel:
LGTM

PR #5159, opened 2016-04-12
fix minor jumping in structureboard

vxsx:
there was a slight size difference when adding plugin to an empty placeholder 
and slight jump when dragging plugin from one placeholder to another

before: http://d.pr/v/1ggvj
after: http://d.pr/v/13oFT


FinalAngel:
nice :

PR #5162, opened 2016-04-13
Copy page with overwrite url cause maximum recursion depth exceeded in get_available_slug

idogrosberg:


yakky:
I think we definitely needs a test here and / or more context regarding the original error.
It should not be required to reset path when copying a page, it might be some underlying tree error and resetting the path may fix the symptom but not the cause

idogrosberg:
Currently when trying to copy page with overwrite url got this traceback:
https://gist.github.com/idogrosberg/bf4a448d9779579095091f2259364d66

This recursion loop happens because it fails on `is_valid_slug` because there already a title with that path (overwrite url), and we continue calling the function `get_available_slug` without any change.

I think its reasonable that if we got a page with overwrite url and we want to copy it, we should assign new overwrite url either way.

czpython:
Closing as duplicate of https://github.com/divio/django-cms/issues/5311.

PR #5166, opened 2016-04-14
Address security concerns relating to render_model, et al and mark_safe

mkoistinen:


GaretJax:
LGTM, this follows what we discussed on Friday. Nonetheless, somebody from the CMS team shall have a look at this.

(I have the impression that there is a lot of duplicated code in this part of the codebase)

PR #5168, opened 2016-04-17
Address security concerns in render_model* tags

mkoistinen:


GaretJax:
See #5166

stefanfoulis:
@mkoistinen technically LGTM.
I'd try to avoid introducing a new way of describing the problem in the setting name. ``CMS_UNESCAPED_RENDER_MODEL_TAGS`` is not exactly the thing I'd look for when dealing with a problem that is dominated by the term ``mark_safe``.
How about calling it ``CMS_RENDER_MODEL_TAGS_WITH_MARK_SAFE`` or 
``CMS_RENDER_MODEL_TAGS_WITH_MARK_SAFE_BY_DEFAULT``?

stefanfoulis:
@mkoistinen does this implementation also support this notation?
``{% render_model_block mymodel "myfield" mark_safe=False %}``

FinalAngel:
what is the implication if I do {% render_model_block mymodel "myfield" "" "" "safe" %} ?

mkoistinen:
@stefanfoulis re: `CMS_RENDER_MODEL_TAGS_WITH_MARK_SAFE`, that sounds like a safe thing to leave set to True, whereas `CMS_UNESCAPED_RENDER_MODEL_TAGS` sounds unsafe.

I don't think this implementation supports "kwargs" style parameters. I'm not sure its possible without overhauling these template tags.

yakky:
@mkoistinen why adding "mark_safe" when you can use the ``filters`` argument for this?

stefanfoulis:
re ``{% render_model_block mymodel "myfield" mark_safe=False %}``: the reason I'd like it to be possible to explicitly say: "don't mark_safe!" even if ``CMS_UNESCAPED_RENDER_MODEL_TAGS=True`` is so that someone can already fix a template to do the secure thing, even if the global setting in a project is still the insecure option.

yakky:
IIRC classy tags does not support that syntax, but I might be wrong
You can achieve that with ``{% render_model_block mymodel "myfield" "" "" "escape" %} ``

stefanfoulis:
@yakky ok 👍 (``"escape"``)

cms/templatetags/cms_tags.py
```diff
@@ -1058,7 +1064,10 @@ def render_tag(self, context, **kwargs):
         context.push()
         template = self.get_template(context, **kwargs)
         data = self.get_context(context, **kwargs)
-        data['content'] = mark_safe(kwargs['nodelist'].render(data))
+        if get_cms_setting('UNESCAPED_RENDER_MODEL_TAGS'):
```
yakky: @mkoistinen have you considered encapsulating this if to make the code more maintenable?

mkoistinen: Thanks, @yakky turns out, we only need this in one place anyway. =)

cms/templatetags/cms_tags.py
```diff
@@ -1058,7 +1061,7 @@ def render_tag(self, context, **kwargs):
         context.push()
         template = self.get_template(context, **kwargs)
         data = self.get_context(context, **kwargs)
-        data['content'] = mark_safe(kwargs['nodelist'].render(data))
```
czpython: should this not respect the setting? or does it only apply to render_model?

yakky: CMSEditableObjectAdd will only add the cms frontend editor markup, it does not render any user content thus we can just skip it

docs/reference/configuration.rst
```diff
@@ -643,6 +643,33 @@ The path to the media root of the cms media files.
...
+    True
+
+.. warning::
+    In this version of django CMS, this setting has a default value of
+    ``True`` to provide behavior consistent with previous releases.
+    However, all developers are encouraged to set this value to ``False`` to
+    help prevent a range of security vulnerabilities stemming from
+    HTML, Javascript, and CSS Code Injection.
+
+.. important::
```
yakky: Advise here that when removed, the behavior will be the opposite of the current default

yakky:
LGTM

czpython:
Ok to merge

PR #5175, opened 2016-04-19
use `html` instead of `text` method in the message open

vxsx:
modern browsers will use escaped version of the text, meaning
&gt;script&lt;alert(1);&gt;/script&lt; would actually render script
tag with an alert in the message

3.2 version of https://github.com/mkoistinen/django-cms/pull/3


FinalAngel:
LGTM

vxsx:
obsolete as of https://github.com/divio/django-cms/pull/5186

PR #5176, opened 2016-04-19
use `html` instead of `text` method in the message open

vxsx:
modern browsers will use escaped version of the text, meaning
&lt;script&gt;alert(1);&lt;/script&gt; would actually render script
tag with an alert in the message

3.3 version of https://github.com/divio/django-cms/pull/5175 and https://github.com/mkoistinen/django-cms/pull/3


FinalAngel:
LGTM

PR #5180, opened 2016-04-19
Add copy-content option to copy lang command

yakky:
This allows to skip copying content when duplicating a language
~~A test is needed for this~~
Test added

### DIFF
--- a/docs/reference/cli.rst
+++ b/docs/reference/cli.rst
@@ -124,6 +124,8 @@ It accepts the following options
   if set, copied content will be appended to the original one;
 * ``--site``: specify a SITE_ID to operate on sites different from the current one;
 * ``--verbosity``: set for more verbose output.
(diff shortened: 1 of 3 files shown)


czpython:
Ok to merge

docs/reference/cli.rst
```diff
@@ -124,6 +124,8 @@ It accepts the following options
   if set, copied content will be appended to the original one;
 * ``--site``: specify a SITE_ID to operate on sites different from the current one;
 * ``--verbosity``: set for more verbose output.
+* ``--skip-content``: if set content is not copied, and the command  will only
+  create titles in the given language.
```
mkoistinen: ``if set, content is not copied, and the command will only create titles in the given language.``

Remove extra space and add comma for better readability.

mkoistinen:
OK to merge with changes as per above.

PR #5181, opened 2016-04-19
Feature/plugin move speedup

czpython:
* Refactored plugin reordering to avoid unnecessary ``save()`` calls.
* Reduced queries in ``move_plugin``.
* Introduced new plugin method called ``update()`` to update a plugin's data without calling ``save()``.
* Moved placeholder dirty logic to placeholder method ``mark_as_dirty``.
  This is to explicitly mark the placeholder as dirty when needed vs a signal.

I used a placeholder that contained over 20 plugins as my test case.
The load time of moving a plugin in this placeholder was about **3 minutes**.
This has now been lowered to an average of **4 seconds** with this PR.

BEFORE

speedtest-before

AFTER

speedtest-2

### DIFF
--- a/cms/admin/placeholderadmin.py
+++ b/cms/admin/placeholderadmin.py
@@ -509,7 +509,9 @@ class PlaceholderAdminMixin(object):
         move_a_copy = (move_a_copy and move_a_copy != "0" and
                        move_a_copy.lower() != "false")
 
+        source_language = plugin.language
         source_placeholder = plugin.placeholder
+
         if not language and plugin.language:
             language = plugin.language
         order = request.POST.getlist("plugin_order[]")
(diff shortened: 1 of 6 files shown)


cms/admin/placeholderadmin.py
```diff
@@ -609,6 +643,7 @@ def move_plugin(self, request):
         target_placeholder_admin = self._get_attached_admin(placeholder)
 
         if move_a_copy:  # "paste"
+            plugins = [plugin] + list(plugin.get_descendants())
```
yakky: any reason not to use ``plugin.get_tree()``? this will return a queryset with all descendants and the current plugin

cms/admin/placeholderadmin.py
```diff
@@ -579,24 +594,43 @@ def move_plugin(self, request):
                         return HttpResponseBadRequest(force_text(
                             _('parent must be in the same language as '
                               'plugin_language')))
-                    plugin.parent_id = parent.pk
-                    plugin.language = language
-                    plugin.save()
+                    plugin = plugin.update(refresh=True, parent=parent, **plugin_data)
```
yakky: isn't ``parent_id=parent.pk`` quicker? I haven't checked how the ORM translates this in a while, though, I might be wrong

cms/admin/placeholderadmin.py
```diff
@@ -579,24 +594,43 @@ def move_plugin(self, request):
...
+            # order should be a list of plugin primary keys
+            # it's important that the plugins being referenced
+            # are all part of the same tree.
+            order = [int(pk) for pk in order]
+            plugins_in_tree = CMSPlugin.objects.filter(
+                parent=parent_id,
+                placeholder=placeholder,
+                language=language,
+                pk__in=order,
+            )
```
yakky: Isn't this the same as ``parent.get_descendants().filter(pk__in=order)``? This might be quicker

cms/admin/placeholderadmin.py
czpython: afaik doesn't make much difference, unless necessary I prefer to not use the more internal `_id` approach.

cms/admin/placeholderadmin.py
czpython: `parent_id` can be none

cms/admin/placeholderadmin.py
czpython: none, was not aware of `get_tree()`. Thanks!

mkoistinen:
Looks awesome, can't wait for this to be in. OK to merge.

yakky:
Great job

PR #5187, opened 2016-04-20
Fix bad git merge

mkoistinen:
Not really sure how long this has been in there, but it will have affected almost no one.


mkoistinen:
Self-committing, this is meta

PR #5189, opened 2016-04-21
Add documentation on binary libraries

yakky:
@divio/django-cms-core plz review


yakky:
@evildmp could you check if it's the best solution to make it clear the preliminary steps to use the installer?

yakky:
@evildmp could you review this?

PR #5190, opened 2016-04-21
Dropdowns icon

vxsx:
Changes the icon that looks more like "text aligned right" to be a real hamburger 
adds real active state to dropdowns in pagetree

before:
screen shot 2016-04-21 at 08 22 11
screen shot 2016-04-21 at 08 22 17

after:
screen shot 2016-04-21 at 08 21 50
screen shot 2016-04-21 at 08 21 59


Brankub:
👍

PR #5195, opened 2016-04-21
Increments migration number, fixes dependencies

mkoistinen:

PR #5196, opened 2016-04-21
Replaces missing migration, fixes up dependencies

mkoistinen:
Also, see #5195 against the 3.2 branch.

### DIFF
--- a/cms/migrations/0013_auto_20160404_1908.py
+++ b/cms/migrations/0013_auto_20160404_1908.py
@@ -1,0 +1,31 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.contrib.auth.models import Permission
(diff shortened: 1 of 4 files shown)


czpython:
I don't agree here, why is the permissions migration not being set as 0015?

mkoistinen:
To maintain the same order as older versions?

czpython:
@mkoistinen this commit https://github.com/divio/django-cms/commit/85cd73b4798898a0b68b3997edaf702e331aa7d7 introduces the migration as 0013 when it should be 0015, once that's changed then here it would match

mkoistinen:
@czpython Good catch. How's this?

czpython:
Ok to merge

stefanfoulis:
👍

PR #5197, opened 2016-04-21
Adds get_vary_cache_on() support and...

mkoistinen:
... and significantly enhances the placeholder cache.

In particular, the placeholder cache is now decoupled from the page cache. This means that now, when rebuilding a page, it can pull each placeholder from cache independently. This further means that if a single plugin has a short cache expiration (or cache=False), then the page render only needs to build that placeholder, and can use the cached versions of the others.

The plugins now have an additional method: get_vary_cache_on() that allows the developer to vary the plugin's cache on different HTTP headers, this in turn means that the PH cache that the plugins is part of will also vary on the(se) header(s). This ultimately means the whole page will vary on the provided header(s).


yakky:
I'm ok with this implementation

mkoistinen:
Rebased

CHANGELOG.txt
```diff
@@ -23,6 +23,11 @@
 * CMS handles uninstalled apphooks better
 * Added the ability to make pagetree actions work on currently picked language
 * Removed deprecated CMS_TOOLBAR_SIMPLE_STRUCTURE_MODE setting
+* Introduction of the method `get_cache_expiration` on CMSPluginBase to be used
+  by plugins for declaring their rendered content's period of validity.
+* Introduction of the method `get_vary_cache_on` on CMSPluginBase to be used
+  by plugins for declaring ``VARY`` headers.
+* Substantial improvements to the placeholder cache
```
evildmp: Wherever it makes sense, I would like all lines in the changelog to start with a verb - _added_, _removed_, _improved_, _fixed_, _introduced_ etc.I think it makes it easier to scan and understand what has been done.

cms/cache/placeholder.py
```diff
@@ -1,67 +1,182 @@
 # -*- coding: utf-8 -*-
-from django.conf import settings
-from django.utils.encoding import force_text
-from django.utils.timezone import get_current_timezone_name, now
-from cms.cache import _get_cache_version, _set_cache_version, _clean_key, _get_cache_key
-from cms.utils import get_cms_setting
 
+"""
+This module manages placeholder caching. We use a cache-versioning strategy
+with each (placeholder x lang) manages its own version. The actual cache
```
evildmp: ``with`` > ``in which``

docs/reference/api_references.rst
```diff
@@ -368,6 +368,25 @@ cms.plugin_base
         :rtype: ``None`` or ``datetime`` or ``int``
 
 
+    .. method:: get_vary_cache_on(request, instance, placeholder):
+
+        Provides ``VARY`` header strings to be considered by the placeholder
+        and in turn the page.
```
evildmp: "in turn by the page" would be clearer

mkoistinen:
@divio/django-cms-core I'd really like to get this reviewed asap.

cms/templatetags/cms_tags.py
```diff
@@ -184,11 +183,12 @@ def _get_placeholder(current_page, page, context, name):
     placeholders = page.rescan_placeholders().values()
     fetch_placeholders = []
     request = context['request']
+    site_id = get_site_id(getattr(page, 'site_id', None))
```
czpython: why ``getattr``? Page objects always have a not nullable site field so you can just use page.site_id

cms/plugin_rendering.py
```diff
@@ -123,6 +123,8 @@ def render_placeholder(placeholder, context_to_copy, name_fallback="Placeholder"
     if hasattr(placeholder, 'content_cache'):
         return mark_safe(placeholder.content_cache)
     page = placeholder.page if placeholder else None
+    site_id = get_site_id(getattr(page, 'site_id', None))
```
czpython: please use a conditional, page can be none but if not then it will have a site_id attr.

``python
if page:
    site_id = page.site_id
else:
    site_id = get_site_id(None)
``

cms/plugin_rendering.py
```diff
@@ -187,8 +191,9 @@ def render_placeholder(placeholder, context_to_copy, name_fallback="Placeholder"
     context['edit'] = edit
     result = render_to_string("cms/toolbar/content.html", flatten_context(context))
     changes = watcher.get_changes()
-    if placeholder and not edit and placeholder.cache_placeholder and get_cms_setting('PLACEHOLDER_CACHE') and use_cache:
-        set_placeholder_cache(placeholder, lang, content={'content': result, 'sekizai': changes}, request=request)
+    if use_cache and not edit and placeholder.cache_placeholder and get_cms_setting('PLACEHOLDER_CACHE'):
```
czpython: ``use_cache`` will only be true if the user is not logged in, this means that is not necessary to check ``edit`` here

cms/models/placeholdermodel.py
```diff
@@ -429,5 +431,48 @@ def mark_as_dirty(self, language, clear_cache=True):
...
+                    plugin = instance.get_plugin_class_instance()
+                    yield instance, plugin
+            else:
+                for plugin_item in self.get_plugins(lang):
+                    yield plugin_item.get_plugin_instance()
+
+        if not self.cache_placeholder or not get_cms_setting('PLUGIN_CACHE'):
+            return list()
+
+        vary_list = list()
```
czpython: please use a set() to avoid all the checks if an item is in or not.
I recommend just storing them as lower cased values because anyways the check is insensitive.

cms/plugin_base.py
```diff
@@ -219,6 +219,22 @@ def get_cache_expiration(self, request, instance, placeholder):
...
         return None
 
+    def get_vary_cache_on(self, request, instance, placeholder):
+        """
+        Provides hints to the placeholder, and in turn to the page for
+        determining VARY headers for the response.
+
+        Must return one of:
+            - None (default),
+            - String of a case-sensitive header name, or
```
czpython: While it's nice to return a string, I think we should stick to lists or tuples (or other no string iterables).
I think supporting both just adds extra overhead, developers can just return ['some-header']

cms/models/placeholdermodel.py
```diff
@@ -429,5 +431,48 @@ def mark_as_dirty(self, language, clear_cache=True):
...
+            if hasattr(self, '_all_plugins_cache'):
+                for instance in self._all_plugins_cache:
+                    plugin = instance.get_plugin_class_instance()
+                    yield instance, plugin
+            else:
+                for plugin_item in self.get_plugins(lang):
+                    yield plugin_item.get_plugin_instance()
+
+        if not self.cache_placeholder or not get_cms_setting('PLUGIN_CACHE'):
+            return list()
```
czpython: use ``[]`` instead of ``list()`` when returning empty. a bit faster.

cms/models/placeholdermodel.py
```diff
@@ -429,5 +431,48 @@ def mark_as_dirty(self, language, clear_cache=True):
...
+
+        vary_list = list()
+        language = get_language_from_request(request, self.page)
+        for instance, plugin in inner_plugin_iterator(language):
```
czpython: instance can be None for ghost plugins, maybe just check for that and continue if it is None since they're not really considering in rendering.

cms/models/placeholdermodel.py
```diff
@@ -429,5 +431,48 @@ def mark_as_dirty(self, language, clear_cache=True):
...
+        """
+        def inner_plugin_iterator(lang):
+            """See note in get_cache_expiration.inner_plugin_iterator()."""
+            if hasattr(self, '_all_plugins_cache'):
+                for instance in self._all_plugins_cache:
+                    plugin = instance.get_plugin_class_instance()
+                    yield instance, plugin
+            else:
+                for plugin_item in self.get_plugins(lang):
+                    yield plugin_item.get_plugin_instance()
```
czpython: this adds one query to fetch all plugins and then one query per plugin to fetch its instance.
can easily result in over crazy query counts, so this raises big red flag for me.

cms/models/pagemodel.py
```diff
@@ -354,7 +354,7 @@ def _copy_contents(self, target, language):
                 new_phs.append(ph)
                 # update the page copy
             if plugins:
-                copy_plugins_to(plugins, ph, no_signals=True)
```
czpython: why remove ``no_signals``?

cms/cache/placeholder.py
```diff
@@ -1,67 +1,183 @@
...
+def clear_placeholder_cache(placeholder, lang, site_id):
+    """
+    Invalidates all existing cache entries for (placeholder x lang x site_id).
+    We don't need to re-store the vary_on_list, because the cache is now
+    effectively empty.
+    """
+
+    current_version, _ = _get_placeholder_cache_version(placeholder, lang, site_id)
+    version = int(time.time() * 1000000)
+    if current_version >= version:
```
czpython: I know is just in case but this is a relatively "impossible" scenario, don't think is necessary to have this extra code.

cms/cache/placeholder.py
```diff
@@ -1,67 +1,183 @@
...
-                     version=_get_cache_version())
+    key = _get_placeholder_cache_version_key(placeholder, lang, site_id)
 
+    if not version or version < 1:
+        version = int(time.time() * 1000000)
 
-def clear_placeholder_cache(placeholder, lang):
-    from django.core.cache import cache
+    if vary_on_list is None:
+        vary_on_list = list()
```
czpython: use [] instead of list()

cms/models/pagemodel.py
mkoistinen: When signals are suppressed, the placeholder cache doesn't get marked dirty when plugin's are changed.

cms/cache/placeholder.py
mkoistinen: Hmmm, I'm reluctant, but ok.

cms/models/placeholdermodel.py
mkoistinen: This gets all the information it needs from the existing _all_plugins_cache. It is, in fact, query-free to do this in my tests (and the unit tests). Only if _all_plugins_cache isn't available, will it "fallback" to the case that requires queries. I haven't seen this case happen yet, in my testing, but if it does, I've got it covered.

cms/plugin_base.py
mkoistinen: 95% of the time, it'll be a single header, so, I'm not sure I agree.

stefanfoulis:
👍 LGTM. There is nothing I can add after @czpython s thorough review.

cms/models/placeholdermodel.py
czpython: Cool, as long as it's not on every request then is ok.

cms/models/pagemodel.py
czpython: @mkoistinen Good point. I will change ``CMSPlugin.copy_plugin`` to remove the no_signal and instead always explicitly mark the placeholder cache as dirty. Will do so in a separate pr. Ok with you?

cms/plugin_base.py
czpython: Yeah but adding the brackets around it won't really add much vs out overhead of maintaining both ways. Up to you though.

czpython:
Ok to merge.

PR #5198, opened 2016-04-22
Travis cache

vxsx:
Enables travis caching for pip and npm, presumably will speed up build times (fingerscrossed)
Adds a script that caches npm-shrinkwrap as it's the most reliable way to check if the dependencies changed

Sadly it seems can only be battle tested after this is merged and at least one build (the merge one) have been ran, because cache is taken from the original repo


intellectronica:
LGTM

cyberstar:
LGTM

PR #5201, opened 2016-04-24
Update fonts query string automatically

vxsx:
In 3.2.x we introduced `{% static_with_version %}` template tag for cache-busting purposes. Same needed to be done for static assets referenced in css, such as an icon font. Since then it was forgotten, so this PR introduces a postcss plugin to do that automatically.

Version will be read from `cms/__init__.py`


FinalAngel:
LGTM

PR #5204, opened 2016-04-25
Disable sideframe resizing

vxsx:
- sideframe always 90% of the screen except when on screens with width < 768px
- sideframe is no longer resizable

video:
http://d.pr/v/1iHGw


FinalAngel:
LGTM

PR #5210, opened 2016-04-26
Sanitize AppConfig Widget

mkoistinen:


czpython:
Ok to merge

cms/forms/widgets.py
```diff
@@ -291,11 +292,11 @@ def __init__(self, attrs=None, choices=(), app_configs={}):
         super(ApplicationConfigSelect, self).__init__(attrs, choices)
 
     def render(self, name, value, attrs=None, choices=()):
-        output = [super(ApplicationConfigSelect, self).render(name, value, attrs, choices)]
+        output = list(super(ApplicationConfigSelect, self).render(name, value, attrs, choices))
         output.append('<script>\n')
         output.append('var apphooks_configuration = {\n')
         for application, cms_app in self.app_configs.items():
-            output.append("'%s': [%s]," % (application, ",".join(["['%s', '%s']" % (config.pk, force_text(config)) for config in cms_app.get_configs()])))
+            output.append("'%s': [%s]," % (application, ",".join(["['%s', '%s']" % (config.pk, escape(config)) for config in cms_app.get_configs()])))  # noqa
```
yakky: maybe ``escapejs`` here?
This is added to javascript code

yakky:
@mkoistinen See my comment above. Other than that LGTM

cms/forms/widgets.py
stefanfoulis: yes, probably both ``escape`` and ``escapejs``, since ``escapejs`` does not actually make it safe from a html standpoint (it only avoids syntax errors in js strings).

mkoistinen: @yakky, @stefanfoulis good points. I'll check to see which is best (using just escapejs or both), and patch as required.

PR #5229, opened 2016-04-28
removes class cms-loader when cms-admin & cms-admin-modal have been added 

genti94:
If a user doesn't have the rights to edit a plugin, the loader while get shown all the time.

Wrote unit test to check if it removes the cms-loader class when the iframe is loaded


vxsx:
👍

PR #5233, opened 2016-04-28
Implements CMS_WIZARD_CONTENT_PLACEHOLDER setting

mkoistinen:


cms/forms/wizards.py
```diff
@@ -114,7 +114,7 @@ def __init__(self, instance=None, *args, **kwargs):
         self.instance = instance
         super(BaseCMSPageForm, self).__init__(*args, **kwargs)
 
-        if self.page:
+        if getattr(self, 'page', None):
```
czpython: Please don't use getattr unless it's necessary.
This attribute is (should) always be present (even if none).

cms/forms/wizards.py
```diff
@@ -158,15 +158,23 @@ def create_page_titles(page, title, languages):
...
+        placeholder or None.
         """
-        for placeholder in page.get_placeholders():
-            if not placeholder.is_static and placeholder.is_editable:
-                return placeholder
-        else:
-            return None
+        placeholders = page.get_placeholders()
+
+        if slot:
```
czpython: I think it might be better to do a single loop and tackle the slot condition from there

czpython: Also I think that if placeholder is defined then maybe we shouldn't fallback to the first placeholder if the defined one does not exist.
This is because if defined, it means programmer is being explicit about where to add plugins.

cms/forms/wizards.py
```diff
@@ -160,20 +162,19 @@ def create_page_titles(page, title, languages):
...
     def get_placeholder(page, slot=None):
         """
-        Returns the named placeholder or the first editable, non-static
-        placeholder or None.
+        Returns the named placeholder or, if no «slot» provided, the first
+        editable, non-static placeholder or None.
         """
         placeholders = page.get_placeholders()
 
         if slot:
```
czpython: I would write this as:

```python
placeholders = page.get_placeholders()

if slot:
    placeholders = placeholders.filter(slot=slot)

for ph in placeholders:
    if not ph.is_static and ph.is_editable:
        return ph
```

get_placeholders() returns a queryset

czpython:
Ok to merge.

cms/utils/conf.py
```diff
@@ -67,6 +67,7 @@ def wrapper():
     'WIZARD_DEFAULT_TEMPLATE': constants.TEMPLATE_INHERITANCE_MAGIC,
     'WIZARD_CONTENT_PLUGIN': 'TextPlugin',
     'WIZARD_CONTENT_PLUGIN_BODY': 'body',
+    'WIZARD_CONTENT_PLACEHOLDER': None,  # Use first placeholder it finds.
```
FinalAngel: would it be possible to define 'content' as default and if it doesn't find content, or any provided string, fall back to the first?

yakky: It's too project-layout specific, IMO

yakky:
Ok to merge
and much needed

cms/utils/conf.py
FinalAngel: 👍

mkoistinen:
ping @evildmp.

mkoistinen:
Note the failure on Travis has nothing to do with this PR, but rather some spelling issue from other PRs. Merging.

Issue #3774, opened 2015-01-17
Feature request: prevent concurrent editing


Moini:
Hi again ;)

At inkscape.org, we sometimes have more than one person trying to edit a page, although we know about the problem and try to prevent it.
We also know that there are django apps which are made for preventing this (https://www.djangopackages.com/grids/g/concurrentp-edit/).

But, according to Martin Owens, those apps are problematic for us to use.
Quote: "Most of the models require extra fields to be added, this is hard for models we don't have direct control over and thus is something that upstream django-cms must do [ when we want to use it ]".

Is it possible that you can add functionality for preventing concurrent edits?

evildmp:
Hard to do, and very hard to do well.

The way I would do it is this:

Every admin/frontend editing form that wants to be aware of users will have some JS in it that pings the server once every few seconds with a query that contains the content type and id of the object being edited.

This will include every object that might be saved by the form - including those in inlines, so possibly many objects will be pinging the server with this query. (Of course, some objects will also affect other objects that aren't present in the forms (e.g. in `save()` methods) and it would be extremely hard, sometimes impossible, do know what they all are. But this is rarer, in most cases, especially for the CMS.)

When the server is pinged with these queries it will return information about any other forms that are also pinging it, and this can be displayed in the forms, to alert users - either that someone else is has already has the forms open, or that someone else opened one after you did. 

If necessary, it could even prevent a save until the first person has closed their form.

I don't think this system would require any extra fields o the models, because it's all form-oriented.

The human problems are also hard to solve. It's a pain in the neck if someone forgets to close the form and goes to lunch, for example.

The whole thing looks pretty complex to me.

Moini:
"The human problems are also hard to solve. It's a pain in the neck if someone forgets to close the form and goes to lunch, for example." 
- maybe a timeout that can manually be reset by the person who is editing could help? People will learn fast enough that work is lost if they don't save and leave ;)

mkoistinen:
Another approach could be that on "Save", the Admin (class, not person) first just verifies that the object hasn't already changed since the local change-form was first opened (either by comparing all fields, or just a single timestamp field) and if it has, indicates which fields have "collisions" and presents a dialog warning the operator of the situation allowing them to override, reload the other operator's changes, or simply cancel. Seems that this is an extensible approach that doesn't require a lot of real-time messaging.

evildmp:
See also https://groups.google.com/d/topic/django-developers/mOkvOQQGhJk/discussion

[...]

PR #5234, opened 2016-04-29
Prevent concurrent editing of a page

thoreg:
Hi,

Following the latest 'discussion' under https://plus.google.com/+ThomasRega/posts/CtrBnjKvJzQ - I liked the idea just to inform the user instead of hard rejecting/blocking anybody. Like mentioned in the discussion it is not that easy (it is not possible?) to solve the problem at all without the intelligence of the participating user of the system.

I solved that for with a context_processor which - in case - can be used to show a notification to the user, e.g. within the toolbar.

For the project I am working on - I have written some tests for that 'feature' with pytest-bdd, pytest-splinter. The feature file looks like the following:

---8<---
Feature: Prevent concurrent editing of a page

    The content management system is used by serveral users at the same time.

    If user A wants to edit a page, which is already opened by user B, then user A
    receives a notification that the page is already being edited by user B.

    The users have to dissolve this situation by their own.

    Under the hood a 'PageEditLock' entry is created per page when the page 'enters' to edit
    mode and deleted when the page gets published. The 'PageEditLock' entry could also be
    deleted manually in the admin backend.

    There are the different scenarios 'handled':

    1. The user clicks the 'Edit' button of a page in the top right corner
    2. The user clicks to one of the menu items ('settings', 'advanced settings') within the toolbar
    3. The user clicks on the settings of a page in the page list view in the admin backend

    Background: Preparation
        Given a logged in user and a page
        And a second user

    Scenario: Concurrent read access does not lead to a warning message in Frontend
        When the first user starts to edit the page in the frontend
        And the second user visits the page but does not edit it
        Then the second user does not see the warning that the page is locked in the frontend

    Scenario: Concurrent write access leads to a warning message in Frontend
        When the first user starts to edit the page in the frontend
        And the second user starts to edit the page in the frontend
        Then the second user sees the warning that the page is locked in the frontend

    Scenario: Concurrent write access leads to a warning message in Backend
        When the first user starts to edit the page in the backend
        And the second user starts to edit the page in the backend
        Then the second user sees the warning that the page is locked in the backend

    Scenario: An EditLock entry can be deleted via admin interface
        When the first user starts to edit the page in the backend
        And the first user never ever finishes the edit by save/publish
        Then the EditLock entry can be removed via admin interface

    Scenario: The EditLock entry is removed when the user publishes the page
        When the first user starts to edit the page in the frontend
        And the user adds a chapter to the page
        And the user publishes the page
        Then the EditLock entry for the page is removed
---8<---

To be honest - I did not get it how I could add a selenium test to the current (pre-3.3) django-cms. In the changelog there is a note, that selenium tests have been removed.

It would be nice to get some guidance relating to the tests. Are there any plans to integrate tests like: https://pypi.python.org/pypi/pytest-bdd to the project? or alternatives?


vxsx:
Hi @thoreg 
The selenium tests were removed and instead a test suite written with casper.js was added

There is more information on how to run them here
http://docs.django-cms.org/en/develop/contributing/testing.html#integration-tests

lddubeau:
This PR is related to the request that was made in issue #3774.

Things that immediately come to my mind:

1. The scenario "Alice starts editing the page but goes to lunch, on vacation, just forgets to cleanly exit, etc."  is handled by breaking locks manually. I don't see a system where the admin has to manually break locks for these cases being particularly well received by admins. Ideally, as an admin, I want to never get a call or an email from a user asking me to remove a lock manually. For the really odd, exceptional case, okay, but not for run-of-the-mill cases. I had mentioned lock expiration in my comments in the Google+ discussion.

2. I was under the impression that Django CMS' object model would require locking of multiple different kinds of object. If only pages are locked, does this prevent everything presented in a page editing session from being modified by another user elsewhere? For instance if pages A and B both include element X in their contents, and editing either page allows editing X, then locking only at the page level is going to be misleading. Alice will start editing page A, thinking that everything she sees on the page is locked while she's editing it. Bob will do the same with page B. However, if both decide to edit the part of their page which is X, they'll run into the conflict that locking was meant to prevent in the first place. If the documentation is correct and I understood it correctly, [static placeholders](http://docs.django-cms.org/en/latest/introduction/templates_placeholders.html#static-placeholders) seem to be a candidate for this kind of problem.

3. I would suggest that the first version of locking added to Django CMS should be optional. I'm not sure whether it should be on or off by default but admins should be able to use settings to turn it off, or to make it advisory (you get a warning when someone is editing the document you want to edit, but you can go ahead and pay the price if you made a mistake).

thoreg:
@vxsx thanks for the link - what were the reasons for switching from selenium to casper? Is casper able to handle things like timeouts/ajax etc more appropriately?

yakky:
As a first step, I think we should just notify if someone has the same page opened in edit mode (e.g.: not a lock but just a soft notification).
Real locking it's a lot harder than that: you must take into account static placeholders, placeholderfields, titles (I mean different editors working on the same page in different languages is perfectly fine) down to the plugin level
Also, I don't think a dedicated model is the right place to handle this

thoreg:
Hi - thanks for the feedback

I agree in the point that it should be a notification and no real lock which prevents 'hard' the concurrent edit. This is the way how it is implemented - as notification. The name 'PageEditLock' might be changed into something more reasonable. IMHO this 'locking' / notification on page level is enough for the daily practice, but if you say it is not, then may be it is worth to open a follow/other issue or should it be handled right here?
@yakky - would be a flag on the page model the better place?

I would update this change request with the following points:
- make the whole thing optional (setting)
- the flag / model will be automatically cleared after a configureable timeout expired
- documentation
- integration tests

yakky:
My idea of implementing this much needed feature is:

* instrument the code with appropriate signals (e.g.: trigger a signal when user switch to edit mode, in structure mode etc and signals for exiting the editing modes)
* set some kind of flag on all the involved models (mark in some way relevant placeholders, static placeholders etc). In case we can track all these models, that's a case a dedicated model using generic foreign keys
* show a warning on the toolbar to the next user

vxsx:
@thoreg the existing selenium suite was fairly small, never ran on travis (despite what configuration said) and was a little bit buggy. We've had some positive experience with casper before at Divio so we decided to go with it here as well. It's headless, so we can easily run it everywhere, locally and on CI. So far we've written a lot of tests with it for the cms and it seems to handle things pretty well. And yes, timeouts and waiting for ajax calls are pretty straightforward

FinalAngel:
Hey @thoreg @lddubeau. It would be really great if we can provide visual feedback on who is viewing or editing the current page. The locking/unlocking mechanics could be introduced in a later step, with more discussions and feedback beforehand. We discussed the visual feedback within the team and would suggest the following UI/UX strategy:

|image|image|
|---|---|

The image on the left shows the UI element added when the feature is enabled (via setting). The right image is showing the hover state. Once you click on the icon:

|image|image|
|---|---|

A popup will open as shown on the left. The list will show 2 states ordered by:
- Users editing the page
- Users viewing the page

If there are no editors, the icon stays grey. The color of the icon only changes to yellow once there is someone editing in any language of the current page. Always show the editors of your language first, and then other languages etc. As there can be multiple same users editing (for lack of account management) the entry representing "you" is specifically marked.

When hovering over an entry as shown on the right, we show an additional tooltip on who it is (IP based) and since when he's editing the page. Clicking on the entry redirects to the user in the admin.

It's then very important to find the right balance using signals as @yakky mentioned. So that someone doesn't stay as "in edit" forever.

For the future this can be enhanced with user locking or a simple messaging system to notify each other. One step at-a-time. WDYT?

vxsx:
I would be happy to help with the UI

thoreg:
+1 - I like it

lddubeau:
I'm in favor of having 3 options from the get go: locking, advisories, and nothing (what the CMS currently does).

I have multiple concerns about the idea of a first release with support only for advisories. Let me first reiterate an assertion I made in the Google+ discussion: "[...] by the time you have everything needed to reliably tell user A that some other user B is editing the same object user A wants to edit, you already have 99% of the machinery needed to actually lock objects.﻿" "Reliably" is the keyword here. Users should always be able to reliably tell what is safe and unsafe to do from looking at the UI. The information they get should be clear, and accurate enough to ensure that users know what is safe and unsafe to do. They should not have to know the internals of Django CMS or have to second-guess what they see in the UI in order to be *sure* they are not going to inadvertently wreck someone else's work or have their work wrecked inadvertently by someone else. If users have to second-guess the UI, then the advisories will only be window dressing.

One issue that comes to my mind with what I'm seeing in @FinalAngel's UI is how the following scenario would be handled:

1. Alice looks at page X, sees no one is editing it. So it is safe to edit.

2. At the same time, Bob looks at page X, sees no one is editing it. So it is safe to edit.

3. They both decide to edit it.

With a system of advisories, the CMS won't make an issue of this. After all, it could be that Alice and Bob *know* what the other is doing and *decided* to just go ahead. So they end up in an unsafe situation despite the UI giving them information that led them to figure it was safe to edit.

So ok, let's add a check after they manifest their intent to edit. This way, they'll be informed of the danger. To use a very gross pseudo-code representation:

    if not someone_else_is_editing(X) or \
       (prompt("Another user is editing X. "
               "Do you really want to edit? N/Y") == "Y")
        switch_to_edit_mode(X)

But the problem remains. Their requests could both get to the `someone_else_is_editing(X)` when the database still shows that nobody is editing X. So the test indicates no one is editing and they both switch to edit mode. We can make it so that the check can be performed only by one user at a time. However, it cannot be as simple as this. We have a prompt right in the middle of the code so we cannot just acquire a mutex before the check and release it at the end. Otherwise, if Bob got through first, Alice will own the mutex while she's pondering whether she should go ahead with her edit, and Charlie who may also try to edit the page, will have to wait until Alice is done pondering. So maybe:

    must_prompt = True
    with mutex: # Context manager acquires mutex and releases it
        if not someone_else_is_editing(X):
            switch_to_edit_mode(X)
            must_prompt = False

    if must_prompt and (prompt("Another user is editing X. "
                               "Do you really want to edit? N/Y") == "Y"):
        with mutex:
            switch_to_edit_mode(X)

On the other hand, if the CMS actually locks X, then either Alice or Bob wins the race to grab the lock and gets to edit X, and the other gets a report that X is locked by the other user. There's still a need for a mutex because only one user should grab the lock but there's no need to prompt.

yakky:
The biggest issues with *real* lock are IMHO:
- releasing the lock (user acquire the lock by entering edit mode and never exit edit mode)
- level at which lock operates: there are four objects in the CMS that requires locking: Page, Title, Plugin and static placeholder. The higher level the lock operates, the more stringent is the case for a lock release
For one-person website there is obviously no locking issue 
For large team release lock it's not a big deal, as you will always have a supervisor that can override the lock
For small-medium teams it can be a pain point as coordination between such a small team can be complicated

lddubeau:
The issue of people going into edit mode and never exiting is also present with advisories. If Alice starts an editing session and goes away without signaling that she's done, then Bob will see Alice as someone editing the object and will think, erroneously, that he should wait until she's done. Over time, it is guaranteed that users will once in a while forget to properly signal that they are done editing (or will encounter a life-situation at their end that prevents them from signaling their intent: personal disaster, busy at a conference, etc.), so unless there is something else that reins in this issue, an object will slowly accumulate a list of editors who are not really editing.

To prevent locks being held forever, the locks in my system have a configurable expiry time, and these locks are refreshable. A refresh pushes back the expiry time. A save of the document refreshes the locks but this relies on the autosave feature which saves every 5 minutes. (An autosave happens only if there are modifications to be saved, so just keeping the tab open without doing anything won't hold the lock forever.) Other evidence of activity could be be deemed more appropriate in a different context.

Expiry takes care of most cases of forgetfulness. If there is an emergency and a lock has to be released before it expires it does not necessarily need an admin to release it. Users know who hold a lock (because the site tells them) so the first step is to contact the user who has the lock. The user who forgot to properly exit can go back into editing, which will use the lock the user already has, and then properly signal that they are done editing by clicking the exit button. This will release the lock. It is only in a case of emergency, *and* if the user holding the lock cannot immediately release it, that I have to step in and release the lock manually.

> level at which lock operates: there are four objects in the CMS that requires locking: Page, Title, Plugin and static placeholder. The higher level the lock operates, the more stringent is the case for a lock release For one-person website there is obviously no locking issue For large team release lock it's not a big deal, as you will always have a supervisor that can override the lock For small-medium teams it can be a pain point as coordination between such a small team can be complicated

That's a fair concern, and this bring up an issue that has been bugging me for a while. Suppose I have a static placeholder present on all pages, and we only have advisories, no locking. Then Alice goes to edit page A while Bob edits page B, and no one else is editing. If the UI tells them who is editing what *page*, then what Alice see is that she's the sole editor for page A and Bob sees he's the sole editor for page B. In the mind of users who don't happen to know anything about what goes on behind the scenes, they think that it they do not have to coordinate with anyone else because no one else is editing. Which is okay until they both edit the static placeholder shown to them on each page. One of them will win the conflict, the other will lose, but they'll both blame the UI for being misleading once they realize what happened (which could be days later).

FinalAngel:
@lddubeau the idea of a first step with only "advisories" is important to gather user feedback. Going with a full blown all-in-all system without having it tested out in the world might lead to problems (as we've already had historically – see moderation).

If all the facilities and functions are there and visual feedback works as expected, we can still introduce the locking in a shorter release or even make an addon out of it instead baking it into the core.

Regarding your use-case: If two people are deciding on editing the same page, at the same time – the UI element flashes and notifies that there is someone else also editing the page (real-time feedback). If we cannot provide accurate feedback, we cannot implement locking accurately.

Let's try to first implement the user feedback reliably before we do locking.

FinalAngel:
Closing for further discussions in:
https://groups.google.com/forum/#!topic/django-cms-developers/MIhUzcQpWrg

PR #5243, opened 2016-05-03
remove hiding clipboard-from structureboard hide

genti94:


vxsx:
could you adapt the tests please?

PR #5242, opened 2016-05-02
Language headers

vxsx:
This PR makes languages in pagetree clickable to switch view so it shows the page names (and page actions) in correct language.

- [x] changelog

screen shot 2016-05-02 at 18 07 08

"en" is default and picked, view page and edit will go to english version

screen shot 2016-05-02 at 18 08 11

"de" is picked, view page and edit will go to german version (page names still will be in english)

PR #5245, opened 2016-05-04
Pagetree header alignments

vxsx:
same look and feel as filer
additionally fixes the issue with too many unnecessary box shadows 

waits for https://github.com/divio/django-cms/pull/5242/


FinalAngel:
LGTM

PR #5255, opened 2016-05-07
Fixes move_page issue

mkoistinen:
This is particularly nuanced, so, upon proving the situation, I've added copious comments along with the fix.

To reproduce the issue, start with the commit before this PR.

1. Create a test project.
2. Add 4 pages nested within each other so that Alpha is the parent of Beta which is the parent of Gamma which is the parent of Delta.
3. Publish each page.
4. Move Beta to _beneath_ Alpha but at the root. Boom.

Applying this patch, the issue is gone and pages are again freely moveable in the tree.


czpython:
Fixed on 3.3

PR #5263, opened 2016-05-09
fixes/issue 5260

### LINKED ISSUE #5260: Bug with moving a page
Moving a page to the root of pages can result in a 500 error under the following circumstances reported by @mkoistinen 

* Add 4 pages nested within each other so that Alpha is the parent of Beta which is the parent of Gamma which is the parent of Delta.
* Publish each page.
* Move Beta to beneath Alpha but at the root.

The error is caused by [this line](https://github.com/FinalAngel/django-cms/pull/120/files#diff-c0756a8caedaec348582d90875a15194R896).

Calling ``get_last_root_node`` can result in a draft or live page, so certain operations like moving a page that assume the passed page is a draft, will fail and potentially corrupt the tree because it can result in operations being done directly to the live page vs the intended draft page.

czpython:
Fixes #5260

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -991,15 +991,14 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
             # Special case: If «target» is not provided, it means to let the
             # page become a new root node.
             try:
-                tb_target = Page.get_root_nodes().filter(
-                    publisher_is_draft=True, site=site)[position]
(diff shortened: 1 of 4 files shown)


czpython:
Tests are failing because of the known pyflakes update.

mkoistinen:
LGTM. OK to merge.

stefanfoulis:
other than the tests failing (maybe we can just revert to a pyFlakes that works?):
LGTM 👍 ✅

Issue #5259, opened 2016-05-09
Visualise page publish action


petrklus:
In some events in the CMS, a spinning cog is visible in the top-right corner. However, this does not seem to be the case for page publication. 

The fact that page publication is under-way in the background should be visible to the user, letting them know that the action was indeed submitted.

FinalAngel:
@vxsx did we fix that already for 3.3?

vxsx:
not yet, but soon @FinalAngel

PR #5265, opened 2016-05-10
Publishing loading indicator

vxsx:
should fix https://github.com/divio/django-cms/issues/5259
and speedup initialization a bit


Brankub:
👍

FinalAngel:
LGTM

PR #5266, opened 2016-05-10
highlight both username in password if login error

vxsx:
before

screen shot 2016-05-10 at 10 29 07

after

screen shot 2016-05-10 at 10 28 51


FinalAngel:
LGTM

PR #5268, opened 2016-05-10
Toolbar refresh

mkoistinen:


vxsx:
👍

FinalAngel:
FE/Text LGTM

cms/cms_toolbars.py
```diff
@@ -322,6 +323,18 @@ def in_apphook(self):
...
+    def in_apphook_root(self):
+        """
+        Returns True if the request is for a page handled by an apphook, but
+        is also the page it is attached to.
+        :return: Boolean
+        """
+        with force_language(get_language_from_request(self.request)):
+            page = applications_page_check(self.request)
+            if page:
+                return self.request.path == page.get_absolute_url()
```
czpython: @mkoistinen why not ``request.current_page``?

cms/cms_toolbars.py
```diff
@@ -472,10 +484,11 @@ def add_page_menu(self):
             refresh = self.toolbar.REFRESH_PAGE
 
             # menu for current page
-            current_page_menu = self.toolbar.get_or_create_menu(PAGE_MENU_IDENTIFIER, _('Page'), position=1)
+            current_page_menu = self.toolbar.get_or_create_menu(
+                PAGE_MENU_IDENTIFIER, _('Page'), position=1, disabled=self.in_apphook() and not self.in_apphook_root())
```
czpython: please add comments as to why disabled=(conditions) is needed.

ie. "hiding page menu if page is apphook page view"

cms/toolbar/items.py
```diff
@@ -206,11 +206,13 @@ class SubMenu(ToolbarAPIMixin, BaseItem):
     template = "cms/toolbar/items/menu.html"
     sub_level = True
     active = False
+    disabled = False
```
czpython: no need to set this here since it's set on the __init__

cms/toolbar/items.py
```diff
@@ -222,11 +224,16 @@ def add_break(self, identifier=None, position=None):
         return item
 
     def get_items(self):
-        return self.items
```
czpython: why is it necessary to set this on the item instances? don't they  already have they're own disabled state?

cms/toolbar/items.py
```diff
@@ -236,10 +243,10 @@ def get_context(self):
 class Menu(SubMenu):
     sub_level = False
 
-    def get_or_create_menu(self, key, verbose_name, side=LEFT, position=None):
+    def get_or_create_menu(self, key, verbose_name, disabled=False, side=LEFT, position=None):
```
czpython: Please place the kwarg at the end to avoid any backwards incompatible issues where positional arguments are used

cms/toolbar/toolbar.py
```diff
@@ -176,7 +178,7 @@ def get_menu(self, key, verbose_name=None, side=LEFT, position=None):
             return self.menus[key]
         return None
 
-    def get_or_create_menu(self, key, verbose_name=None, side=LEFT, position=None):
+    def get_or_create_menu(self, key, verbose_name=None, disabled=False, side=LEFT, position=None):
```
czpython: same as above

cms/toolbar/toolbar.py
```diff
@@ -260,11 +262,19 @@ def get_object_draft_url(self):
 
     # Internal API
 
-    def _add_item(self, item, position):
+    def _add_item(self, item, position=None):
```
czpython: not sure I follow on creating separate lists for the right last and left last items?
Why not place the items either on the front of the existing list or at the end? Maybe using something like ``collections.deque``?

mkoistinen: Because items are placed into their position as all the toolbars are loaded. By separating the lists, it allows the beginning and end to remain independent from one another as all the toolbars get fully loaded.

cms/toolbar/toolbar.py
mkoistinen: I thought it'd best to keep a consistent signature on related methods. I'll add a note in the release notes.

cms/toolbar/items.py
mkoistinen: I thought it'd best to keep a consistent signature on related methods. I'll add a note in the release notes.

cms/toolbar/items.py
mkoistinen: Because when a parent is disabled, all items should *appear* disabled, even if they're not.

cms/cms_toolbars.py
mkoistinen: I guess I didn't realize that `request.current_page` is ultimately doing the same thing... and is cached, so, thanks! =)

czpython:
Ok to merge

stefanfoulis:
👍

mkoistinen:
Rebased to develop

PR #5269, opened 2016-05-11
Template rendering performance

vxsx:
This PR introduces several performance improvements

- [x] move breadcrumbs to js
- [x] move plugin_child_classes resolving to js
- [x] @czpython put something here 🐼 
- [x] moved breadcrumbs resolving to js + tests
- [x] moved child classes resolving to js + tests
- [ ] docs (not sure if we need here)
- [x] changelog

Backend modifications
* Added new classmethod on CMSPlugin ``requires_parent_plugin``
* ``get_child_classes`` now will correctly calculate child classes when not configured.
* Removed internal ``ExtraMenuItems`` tag.
* Removed internal ``PluginChildClasses`` tag.
* Modified RenderPlugin tag to no longer render the ``content.html`` template and instead just return the results.
** ``Toolbar()`` main class now has a ``get_cached_template`` method that's used to reuse loaded templates
per request. It works like Django's cached template loader but on a request basis.
* ``cms/toolbar/dragbar.html`` is now ``dragbar_template`` is now cached property on Toolbar instance.
* ``cms/toolbar/dragitem.html`` is now ``drag_item_template `` is now cached property on Toolbar instance.
* Cleaned up internal function ``get_toolbar_plugin_struct`` 

Additionally fixed longstanding bug with breadcrumbs never being updated after plugin was moved

Measurements:
for @vxsx locally it's page time to first byte from ~14s to ~5s on a page with 2 placeholders and ~450 plugins (5 different types)


FinalAngel:
FE LGTM

cms/plugin_base.py
```diff
@@ -431,19 +441,55 @@ def get_fieldsets(self, request, obj=None):
...
+    def get_child_plugin_candidates(cls, slot, page):
+        """
+        Returns a list of all plugin classes
+        that will be considered when fetching
+        all available child classes for this plugin.
+        """
+        # By adding this as a separate method,
+        # we allow other plugins to affect the list of child plugin candidates.
+        # Useful in cases like djangocms-text-ckeditor where only text only
+        # plugins are allowed.
```
mkoistinen: Very minor, but even though we have 120 chars for Python, the comments are most readable if they're formatted to fit 80 columns. In this case, it looks like you chopped out some other text, but could you re-justify it into an 80-col block?

cms/plugin_base.py
```diff
@@ -431,19 +441,55 @@ def get_fieldsets(self, request, obj=None):
...
+        from cms.plugin_pool import plugin_pool
+        return plugin_pool.get_all_plugins(slot, page)
+
```
mkoistinen: What is the point of this class method? Its just a thin wrapper for an existing plugin_pool method? Why?

cms/plugin_base.py
```diff
@@ -431,19 +441,55 @@ def get_fieldsets(self, request, obj=None):
...
+    def get_child_classes(self, slot, page):
+        child_classes = self.get_child_class_overrides(slot, page)
```
mkoistinen: Could be good to state the purpose of this method in a here-doc.

cms/plugin_base.py
czpython: As explained in comments, this method allows other plugins to affect the candidates for a plugin's child classes.
Example of this is ckeditor, instead of performing all steps required to get the child plugins, ckeditor just needs to override this method to return text enabled plugins.

cms/toolbar/toolbar.py
```diff
@@ -329,6 +333,22 @@ def request_hook(self):
...
+        return self._cached_templates[template]
+
+    @cached_property
+    def drag_item_template(self):
+        return get_template('cms/toolbar/dragitem.html')
+
+    @cached_property
+    def dragbar_template(self):
+        return get_template('cms/toolbar/dragbar.html')
+
```
mkoistinen: Why not just use your new `get_cached_template()` method instead of using the decorator? I think I prefer the cache to be all in one place, rather than two.

cms/plugin_base.py
```diff
@@ -170,6 +170,16 @@ def render(self, context, instance, placeholder):
...
     @classmethod
+    def requires_parent_plugin(cls, slot, page):
+        requires_parent = cls.get_require_parent(slot, page)
+
+        if requires_parent:
+            return True
+
+        allowed_parents = cls().get_parent_classes(slot, page) or []
+        return bool(allowed_parents)
+
```
mkoistinen: Just a small suggestion: I don't see what the introduction of extra vars serves, the following is still quite readable.
````
def requires_parent_plugin(cls, slot, page):
    if cls.get_require_parent(slot, page):
        return True
    return bool(cls().get_parent_classes(slot, page))
````

File: cms/toolbar/toolbar.py (hunk 4 above)
Comment (czpython) [inline cms/toolbar/toolbar.py]: I thought about it but decided against it because unlike other templates, we already know these before hand and can avoid the extra call to the get_cached_template method every time. I think better would be to use self.get_cached_template() combined with the decorator so we reuse our cached template logic but only call the method once. Ok for you?

File: cms/plugin_base.py (hunk 5 above)
Comment (czpython) [inline cms/plugin_base.py]: Good point, will adjust.

**Comment by mkoistinen** [comment]
Awesome cleanup to both of you!

@czpython I've made some suggestions, but outside of that I'm a bit nervous about the change to the render_plugin tag from InclusionTag to Tag. In particular I'm concerned about changes to the way it interacts with the existing context. Can you help ease my mind here? @ojii can you have a look too?

**Comment by czpython** [comment]
@mkoistinen InclusionTag subclasses Tag and renders a template with context returned by get_context(). In our case we were returning a flat context of {'content': plugin_content}.

This means that the context does not change in any way when rendering the plugin because both inclusion tag and my implementation of tag will render the plugin with the same context variable.
The difference is that one would render another template with the contents of the first render.

File: cms/templates/cms/toolbar/dragbar.html [hunk 6]
```diff
@@ -14,12 +14,12 @@
...
             <div class="cms-submenu-item"><a data-cms-icon="paste" data-rel="paste" href="#">{% trans "Paste" %}</a></div>
             <div class="cms-submenu-item"><a data-cms-icon="bin" data-rel="modal" href="{{ placeholder.get_clear_url }}?language={{ LANGUAGE_CODE }}" data-name="{{ placeholder.get_label }}">{% trans "Empty all" %}</a></div>
-            {% extra_menu_items placeholder %}
+            {% render_extra_menu_items placeholder %}
         </div>
     </div>
     <div class="cms-plugin-picker" data-touch-action="pan-y">
         <div class="cms-quicksearch"><label><input type="text" placeholder="{% trans 'Filter plugins...' %}" /></label></div>
-        {% plugin_child_classes placeholder %}
+        {# plugin child classes here #}
```
stefanfoulis: Is this template comment intentional?

stefanfoulis:
LGTM

mkoistinen:
Ok to merge.

cms/templates/cms/toolbar/dragbar.html
vxsx: yup, would prefer to leave it for now

PR #5270, opened 2016-05-11
Add generic PluginItem attributes

yakky:
There is no way to add arbitrary attributes to the placeholder / plugin menu items, this means for example, that there is no way to add an icon to it (introduced in version 3.2)
This adds an arbitrary set of attributes that are rendered in the template as ``data-`` attributes

Tests and docs added


yakky:
@divio/django-cms-core any opinion on this?

cms/templates/cms/toolbar/dragitem_extra_menu.html
```diff
@@ -1,3 +1,3 @@
 {% for item in items %}
-<div class="cms-submenu-item"><a href="{{ item.url }}" data-rel="{{ item.action }}"{% if item.data %} data-post="{{ item.data }}"{% endif %}{% if question %} data-text="{{ question }}"{% endif %} data-on-success="REFRESH_PAGE">{{ item.name }}</a></div>
+<div class="cms-submenu-item"><a href="{{ item.url }}" {% for key, attr in item.attributes.items %}data-{{ key }}="{{ attr }}" {% endfor %} data-rel="{{ item.action }}"{% if item.data %} data-post="{{ item.data }}"{% endif %}{% if question %} data-text="{{ question }}"{% endif %} data-on-success="REFRESH_PAGE">{{ item.name }}</a></div>
```
vxsx: potentially possible to override other data attributes we already have, like `post` or `text` since jquery will take the first occurrence, not sure if we really want that or not

vxsx:
otherwise i'm ok

cms/templates/cms/toolbar/dragitem_extra_menu.html
yakky: My fault, I thought that last wins. I'll change the order

yakky:
@divio/django-cms-core tests and docs added. Ready for final review

mkoistinen:
LGTM

yakky:
@czpython you ok with this?

czpython:
@yakky Yup. Ok to merge.

PR #5278, opened 2016-05-12
Fixed type

chasmani:
Unnecessary "for" - bad grammar/typo


czpython:
Ok to merge. /cc @evildmp

evildmp:
Thanks for the contribution, but this is already corrected in https://github.com/divio/django-cms/pull/5282.

Issue #5112, opened 2016-03-29
[META] Menus cleanup


czpython:
**To Do**
- [ ] Move menus app into cms app.
- [ ] Refactor cache invalidation.
- [ ] Add site awareness to menu rendering.
- [ ] Multiple menu capabilities

**Open menu related tickets**
- [ ] #4440 Additional setting CMS_MENU_CACHE
- [ ] #4626 Detecting if menu item is associated with a page
- [ ] #4668 show_menu_below_id with multiple sites
- [ ] #5056  Documentation about 'show_menu' templatetag missing two arguments
- [ ] #5068 Menu is inconsistent, cache isn't getting invalidated.
- [ ] #5086 Documentation on extending/modifying menus is unclear
- [ ] #3965 Menu shows random entries from past
- [ ] #4118 The menu system sometimes decides that the selected node is the one associated with a draft page, and then "forgets" what is selected.
- [ ] #3432 menu_tags context's stack bug
- [ ] #2521 Custom specific pages
- [ ] #1873 Language chooser needs alignment
- [ ] #1546 show_menu_below_id issues
- [ ] #1774 menu cache invalidation
- [ ] #3598 show_menu offset
- [ ] #3459 large menus
- [ ] #4228 show_menu_below_id wrong rendering
- [ ] #4099 menu modifiers problems
- [ ] #3094 incorrect logging configuration
- [ ] #1206 get_visible_pages in menu.py doesn't scale

Closed menu related tickets
- [ ] #3343 menus.py fails

Fixed menu related tickets
- [ ] #5136 CMSAttachMenus that are not configured should not be processed
- [ ] #4371 CMS shouldn't try to load CMSAttachMenus if they're not attached

PR #5284, opened 2016-05-12
Enhance menu building by filtering pages by current site

yunmanger1:
In case when we have several sites we get NoReverseErrors when trying to
reverse on apps hooked into pages on different site.


yunmanger1:
Anyway chache is done per site. We could exclude the pages from different sites when building the menu.

czpython:
@yunmanger1 
Thanks for the contribution.
We're currently in the process of cleaning up / refactoring the menu code and have already moved menu logic outside of the menu pool. Adding site awareness is on the to-do.
https://github.com/divio/django-cms/issues/5112

Issue #5273, opened 2016-05-11
filter_horizontal seems to no longer work in CMSPlugins


mkoistinen:
Work-around for now is to add this to your CMSPluginBase descendant (tested on `develop` current to *now*):

````
    def get_form(self, request, obj=None, **kwargs):
        from functools import partial
        kwargs['formfield_callback'] = partial(self.formfield_for_dbfield, request=request)
        return super(HeaderControlPlugin, self).get_form(request, obj=obj, **kwargs)
````

And, of course to add `filter_horizontal = ('my_cms_plugin_model_field', )` as usual.

**Update:** this is a regression in `develop` (AKA CMS 3.3)

PR #5287, opened 2016-05-12
fixes/issue #5273

czpython:
The sane-add-plugin branch introduced a regression that would prevent features like formfield callbacks from working correctly. This manifested itself in the filter_horizonal admin attribute not working correctly.

Related ticket #5273 

The reason for this problem is that Django has some special handling for things like formfield callbacks at the meta class level, so when subclassing we would not inherit the callback.

To fix this we now use internal attributes on the form class that will be used to create the plugin when the form is saved.

@mkoistinen @ojii Please have a look.

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -260,12 +260,13 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
 
         # This flag allows us to know when the plugin initial attributes
         # should be set. This currently only happens on plugin creation.
-        form_class._cms_initial_attributes = True
-        form_class._cms_plugin_language = plugin_data['plugin_language']
-        form_class._cms_plugin_placeholder = plugin_data['placeholder_id']
-        form_class._cms_plugin_parent = plugin_data.get('plugin_parent', None)
-        form_class._cms_plugin_plugin_type = plugin_data['plugin_type']
-        form_class._cms_plugin_position = plugin_data['position']
+        form_class._cms_initial_attributes = {
+            'language': plugin_data['plugin_language'],
+            'placeholder': plugin_data['placeholder_id'],
+            'parent': plugin_data.get('plugin_parent', None),
+            'plugin_type': plugin_data['plugin_type'],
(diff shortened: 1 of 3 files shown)


cms/plugin_base.py
```diff
@@ -369,6 +374,18 @@ def save_model(self, request, obj, form, change):
...
+
+        if getattr(form, '_cms_initial_attributes', None):
+            # Form has the initial attribute hooks
+            obj.language = form._cms_plugin_language
+            obj.placeholder = form._cms_plugin_placeholder
+            obj.parent = form._cms_plugin_parent
+            obj.plugin_type = form._cms_plugin_plugin_type
+            obj.position = form._cms_plugin_position
+        return obj
+
```
mkoistinen: Perhaps this is a matter of style here, but `_cms_initial_attributes` seems like a superfluous flag. Perhaps instead of containing `True` or not existing at all, perhaps it should be a dict which contains all the other attributes, then you should be able to make these assignments more DRY'er, and the existence of the dict is now more meaningful. Something like:

```` python
def get_form(self, request, obj=None, **kwargs):
    form_class = super(CMSPluginBase, self).get_form(request, obj, **kwargs)

    if obj:
        return form_class

    plugin_data = self.validate_add_request(request)

    form_class._cms_initial_attributes = {
        'language': plugin_data['plugin_language'],
        'placeholder': plugin_data['placeholder_id'],
        'parent': plugin_data.get('plugin_parent', None),
        'plugin_type': plugin_data['plugin_type'],
        'position': plugin_data['position'],
    }
    return form_class


def save_form(self, request, form, change):
    obj = super(CMSPluginBase, self).save_form(request, form, change)

    if hasattr(form, '_cms_initial_attributes'):
        for key, value in form._cms_initial_attributes.items():
            setattr(obj, key, value)
    return obj
````

mkoistinen:
Comments above, otherwise LGTM

cms/plugin_base.py
czpython: Great suggestion! I've pushed the updated version.

stefanfoulis:
LGTM

PR #5292, opened 2016-05-14
fixes #5291 on 3.1.x

### LINKED ISSUE #5291: Plugins can sometimes lose position on CMS upgrade
When upgrading from CMS 3.0.x to 3.1.x or 3.2.x plugins are sometimes losing their position (order inside of the placeholder).

This happens because on CMS 3.0.x it's possible (and not uncommon) that two plugins on the same placeholder have the same value for their position field.

So when a user upgrades to 3.1.x or 3.2.x, the ``fix-tree`` command sorts the plugins by position and then reorders them but because the position is the same, the order is not accurate and thus causes the plugins to be given invalid positions.

As you can see here, plugins ``73161`` and ``72681`` share the same value for position
and ``73161`` shows before ``72681``:
```
In [15]: CMSPlugin.objects.filter(placeholder=1503, parent_id__isnull=True).order_by('position').values_list('pk','position')
Out[15]: [(41355, 0), (73161, 1), (72681, 1), (41370, 2)]
```

I've now changed the query a bit to target those two plugins explicitly and for some crazy reason
plugin ``72681`` shows up before ``73161``
```
In [17]: CMSPlugin.objects.filter(placeholder=1503, pk__in=[73161, 72681], parent_id__isnull=True).order_by('position').values_list('pk', 'position')
Out[17]: [(72681, 1), (73161, 1)]
```

czpython:
fixes #5291
* adjusted test to use different query generation.
* sort by path when rebuilding plugin positions.

### DIFF
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -345,7 +345,7 @@ class CMSPlugin(six.with_metaclass(PluginModelBase, MP_Node)):
                 order = CMSPlugin.objects.filter(
                         placeholder_id=placeholder.pk, language=language,
                         parent_id__isnull=True
-                    ).order_by('position').values_list('pk', flat=True)
+                    ).order_by('position', 'path').values_list('pk', flat=True)
                 reorder_plugins(placeholder, None, language, order)
 
                 for plugin in CMSPlugin.objects.filter(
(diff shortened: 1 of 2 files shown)


yakky:
Ok to merge

mkoistinen:
Ok to merge.

PR #5294, opened 2016-05-14
More fixes related to #5260

czpython:
A picture is worth a thousand words.. so they say.

So here's a picture of a broken tree:

broken-tree-after-move

And here's a picture of a healthy tree:
healthy-tree-after-move

The problem with the broken tree is that the second "root" node (parentless) which has id 3
happens to have a path less than the path of the first "root" node's public page which has id 2.

The reason why the tree breaks is because before this change, we would only set the moved page to be a sibling of the old closest root page live version if it had inhered the template, otherwise we would set it as sibling of the draft version which would then change the path of the live version.

@divio/django-cms-core Please review. This is more or less an urgent fix.

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -253,7 +253,7 @@ class Page(six.with_metaclass(PageMetaClass, MP_Node)):
         self.save()
 
         if target_public_page:
-            moved_page = self.move(target, pos=position)
+            moved_page = self.move(target_public_page, pos=position)
         else:
             moved_page = self.move(target, pos=position)
 
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -217,36 +217,38 @@ class Page(six.with_metaclass(PageMetaClass, MP_Node)):
 
         is_inherited_template = (
             self.template == constants.TEMPLATE_INHERITANCE_MAGIC)
+        target_public_page = None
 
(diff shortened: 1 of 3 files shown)


mkoistinen:
LGTM, great commenting! RE:  Asserts, perhaps these should be warnings instead? I worry that an Assert would stop a project in its tracks with no way to recover. Anyone else share this concern?

cms/models/pagemodel.py
```diff
@@ -211,36 +211,51 @@ def move_page(self, target, position='first-child'):
...
         else:
             self.parent_id = target.parent_id
 
         self.save()
-        moved_page = self.move(target, pos=position)
+
+        if target_public_page:
+            moved_page = self.move(target, pos=position)
+        else:
+            moved_page = self.move(target, pos=position)
```
thoreg: IMHO the if / else is not needed because the same statement is used

czpython:
@mkoistinen it's on purpose. Warnings get ignored, these corruptions should now be pretty much gone and if not then they need to be loud so we take them out. They wouldn't stop a project completely, they stop an action from taking place and so you'd have to fix the tree with the command to continue.

cms/models/pagemodel.py
czpython: @thoreg Good catch, here is meant to be self.move(target_public_page, pos=position).

mkoistinen:
OK to merge

PR #5296, opened 2016-05-15
catch TypeError as well

jrief:
While copying a plugin, I had a situation where arithmetic would have given ``None + 1``. Therefore we should also catch a ``TypeError``.

@yakky looking at the query above, why do we need to ``.order_by('depth', 'path')`` the query result, and why don't we use an aggregate for it? You where the last one touching that line of code.


yakky:
@jrief please target release/3.2.x if this affects 3.2.x release. That part of the code is rather old. I agree that a refactoring to use aggregation would help. The ordering is due to the still a bit confusing role of path / position. This is going to be addressed in future releases

PR #5299, opened 2016-05-15
Fixes #5298 on 3.3

### LINKED ISSUE #5298: Attached menus load even when not attached
On addons we have ``menu.py`` files which has a declaration for attach menu, which can be attached to a page.
The issue is that even if those menus are not attached to any page - they are executed and processed on each request for each page.

This issue was introduced in ba0dcad05baae65199303aea0839a78b8bb8f99c

czpython:
Fixes #5298

Basically because the menu logic was coupled with the rendering logic, there was an issue where all attached menu classes would be loaded on every request and all their nodes fetched, regardless of whether the menu class was attached or not. This was a regression and never a feature.

This pull request does a couple of things:

* Decouple the menu rendering logic from the menu discovery pool.
* Bind menu rendering to the request, so during a request the menu entries, modifiers, etc. are all constant.
* Expose a helper method on the menu pool to get all registered menu classes and allow specifying if it's for rendering or not. An example of this is on the admin we want to display all menu classes regardless of whether they are attached menus and have no instances, this is to allow the user to select the menu class but when rendering we **don't** want to load the menu classes that are not attached to an instance and are attached menu subclasses.

I introduced a new class called the ``MenuManager``.
This class is bound to a request and the menu pool instance.
The class exposes the familiar ``get_nodes()`` and ``apply_modifiers()`` methods.

The menu tags have been modified to first get the menu manager and then get the nodes using the manager, the idea is to **not** use the menu pool directly.

Both Menu and Modifier classes now require a ``manager`` instance on initialization.

The menu pool still defines ``get_nodes()`` and ``apply_modifiers()`` for backwards compatibility and will be deprecated in 3.4.

The ``menu_pool.menus`` attribute should **not** be used anymore, instead a new method has been introduced called ``get_registered_menus()``.

``get_registered_menus``
This method makes a query for each attached menu that is registered with the menu system.

Before this change, menus would fetch the instances of attached menus only once in a process and so this causes an issue where attached menus are set to pages but the menu pool would not know about it because it was expanded already.

Sadly moving this calculation to happen at the request/response cycle meant that we would have to make the attached menu queries on every request and once per menu tag!

So I made the menu_manager a context processor variable that only fetches the registered menus when an attribute of the manager is accessed (lazy) and it uses a meomized approach to only do it once, so this way we only get the registered menus literally once per request and only if needed.
If the cms settings context processor is not set then it makes a query per menu tag!

I adjusted the performance tests that would check the loading without context processors, as mentioned above the cms context processor is required and if not added then menu performance does decrease.

TODO:

- [x] Add just a few more tests
- [x] Add settings context processor to cms check
- [x] Add old ``get_nodes`` and ``apply_modifiers()`` to menu pool and deprecate them.

/cc @divio/django-cms-core

### DIFF
--- a/cms/tests/test_views.py
+++ b/cms/tests/test_views.py
@@ -231,7 +231,7 @@ class ContextTests(CMSTestCase):
         context._standard_context_processors = None
         # Number of queries when context processors is not enabled
         with self.settings(**override):
-            with self.assertNumQueries(FuzzyInt(0, 12)) as context:
+            with self.assertNumQueries(FuzzyInt(0, 15)) as context:
                 response = self.client.get("/en/plain_view/")
                 num_queries = len(context.captured_queries)
                 self.assertFalse('CMS_TEMPLATE' in response.context)
@@ -258,7 +258,10 @@ class ContextTests(CMSTestCase):
         # Number of queries when context processors is not enabled
         with self.settings(**override):
             # Baseline number of queries
-            with self.assertNumQueries(FuzzyInt(13, 25)) as context:
+            # It's clear that without the context processors
+            # the number of queries go up.
+            # This is because of the menu manager handling.
+            with self.assertNumQueries(FuzzyInt(20, 47)) as context:
                 response = self.client.get("/en/page-2/")
                 num_queries_page = len(context.captured_queries)
         cache.clear()
@@ -268,10 +271,11 @@ class ContextTests(CMSTestCase):
         with self.settings(**original_context):
             # Exactly the same number of queries are executed with and without
             # the context_processor
-            with self.assertNumQueries(num_queries_page):
+            with self.assertNumQueries(FuzzyInt(13, 25)) as context:
                 response = self.client.get("/en/page-2/")
                 template = Variable('CMS_TEMPLATE').resolve(response.context)
                 self.assertEqual(template, page_template)
+                num_queries_page = len(context.captured_queries)
         cache.clear()
         menu_pool.clear()
         page_2.template = 'INHERIT'
--- a/cms/tests/test_views.py
+++ b/cms/tests/test_views.py
@@ -213,8 +213,6 @@ class ContextTests(CMSTestCase):
 
         page_template = "nav_playground.html"
         original_context = {'TEMPLATES': settings.TEMPLATES}
-        override = {'TEMPLATES': deepcopy(settings.TEMPLATES)}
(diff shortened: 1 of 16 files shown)


cms/tests/test_views.py
```diff
@@ -255,23 +245,15 @@ def test_context_current_page(self):
...
-            with self.assertNumQueries(FuzzyInt(13, 25)) as context:
-                response = self.client.get("/en/page-2/")
-                num_queries_page = len(context.captured_queries)
-        cache.clear()
-        menu_pool.clear()
-
         # Number of queries when context processors is enabled
         with self.settings(**original_context):
             # Exactly the same number of queries are executed with and without
             # the context_processor
```
mkoistinen: It would seem this comment is no longer relevant.

cms/tests/test_views.py
```diff
@@ -229,19 +226,12 @@ def test_context_current_page(self):
...
-                response = self.client.get("/en/plain_view/")
-                num_queries = len(context.captured_queries)
-                self.assertFalse('CMS_TEMPLATE' in response.context)
-        cache.clear()
-        menu_pool.clear()
+
         # Number of queries when context processor is enabled
         with self.settings(**original_context):
             # no extra query is run when accessing urls managed by standard
             # django applications
```
mkoistinen: Also here, we *were* comparing the number of queries to something else, which is now gone. Is this test still meaningful?

mkoistinen:
Looks real nice. I tested on my local machine as well with a small, test project, seems to work well. Please address comments above, otherwise, **OK to merge**.

mkoistinen:
@yakky can you have a look at this, we'd like to squeeze this in, if possible.

cms/tests/test_views.py
czpython: yes the still is still meaningful but comment is not.

eimsbush:
Can this fix also be apllied to the 3.2.x-Release?

FinalAngel:
@eimsbush please see https://github.com/divio/django-cms/pull/5395

czpython:
@eimsbush 
Hello,
We decided to target 3.3+ only because this PR introduced a lot of changes to the menu internals and as such did not fit our criteria for a minor release (3.2.6).

There's no incompatibilities between 3.3 and 3.2 so I highly advise to upgrade.
Please have a look at the upgrade notes http://docs.django-cms.org/en/release-3.3.x/upgrade/3.3.html#upgrading-to-3-3

PR #5300, opened 2016-05-16
Eslint integration

vxsx:
this PR introduces ESLint as a replacement for JSHint + JSCS (which is now part of eslint) we were using before.

New tool is better, faster and smarter.


vxsx:
Canceled the build not to stall the release process ;)

PR #5303, opened 2016-05-17
removes resize when there is a screenblock

genti94:
before 
before

after
after


lory87:
👍

vxsx:
👎

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -420,11 +420,6 @@ var CMS = window.CMS || {};
                 });
             }
 
-            // if there is a screenblock, do some resize magic
-            if (this.ui.screenBlock.length) {
-                this._screenBlock();
```
vxsx: if method is no longer used, remove the method as well (and related tests if required)

PR #5308, opened 2016-05-17
Move DEFAULT_X_FRAME_OPTIONS in cms.utils.conf

### LINKED ISSUE #5163: When changing CMS_DEFAULT_X_FRAME_OPTIONS new migration need to be created
When changing `CMS_DEFAULT_X_FRAME_OPTIONS` in `settings` django wants to create new migration for the `cms` app.

The reason is that initial migration does not get this value dynamically.

One possible solution is to double the code in models and to use it in the migration.

    getattr(settings, 'CMS_DEFAULT_X_FRAME_OPTIONS', Page.X_FRAME_OPTIONS_INHERIT)

Other possible solution will be to move this configuration to `cms.utils.conf` and to get it with `get_cms_setting('DEFAULT_X_FRAME_OPTIONS')`.

Second variant can cause moving definition of: `X_FRAME_OPTIONS_INHERIT`, `X_FRAME_OPTIONS_DENY`, `X_FRAME_OPTIONS_SAMEORIGIN` and `X_FRAME_OPTIONS_ALLOW` from `cms.models.Page` to `cms.constants`

This also will allow overwriting `CMS_DEFAULT_X_FRAME_OPTIONS` in settings to look more explicit:

    CMS_DEFAULT_X_FRAME_OPTIONS = cms.constants.X_FRAME_OPTIONS_SAMEORIGIN

instead of

    CMS_DEFAULT_X_FRAME_OPTIONS = 2

Who will know what `2` means?

I can prepare a Pull Request but let me know which variant you prefer.

vstoykov:
- Make it possible to get DEFAULT_X_FRAME_OPTIONS with get_cms_settings.
- Use get_cms_setting('DEFAULT_X_FRAME_OPTIONS') in initial migration to prevent creating of new migration when change the value in project's settings (Fixes #5163)


docs/reference/configuration.rst
```diff
@@ -1027,10 +1027,10 @@ CMS_DEFAULT_X_FRAME_OPTIONS
 ===========================
 
 default
-    ``Page.X_FRAME_OPTIONS_INHERIT``
+    ``constants.X_FRAME_OPTIONS_INHERIT``
```
mkoistinen: This should simply be ``CMS_DEFAULT_X_FRAME_OPTION``

cms/migrations/0001_initial.py
```diff
@@ -104,7 +104,7 @@ class Migration(migrations.Migration):
                 ('publisher_is_draft', models.BooleanField(db_index=True, default=True, editable=False)),
                 ('languages', models.CharField(max_length=255, null=True, blank=True, editable=False)),
                 ('revision_id', models.PositiveIntegerField(default=0, editable=False)),
-                ('xframe_options', models.IntegerField(default=0, choices=Page.X_FRAME_OPTIONS_CHOICES)),
+                ('xframe_options', models.IntegerField(default=get_cms_setting('DEFAULT_X_FRAME_OPTIONS'), choices=Page.X_FRAME_OPTIONS_CHOICES)),
```
mkoistinen: Should be `get_cms_setting('DEFAULT_X_FRAME_OPTION')`

czpython: The old setting was called with the S at the end.
We don't want to rename it just yet (on the rc).

docs/reference/configuration.rst
czpython: same comment as above.

mkoistinen:
LGTM with comments above.

czpython:
Ok to merge.

czpython:
Thanks @vstoykov

PR #5314, opened 2016-05-18
rename MenuManager to MenuRenderer

czpython:
A change will come in the future to integrate renderers into the cms.
This is a step towards having a similar "api".
The menu renderer is the first of these renderers.

Note, this is not a feature, just renaming.


mkoistinen:
LGTM

stefanfoulis:
:+1:

Was this api in any way public?

mkoistinen:
@stefanfoulis not really, and it was introduced only a couple of commits ago.

PR #5316, opened 2016-05-18
Pagetree without admin style

vxsx:
fixes styles of the pagetree on django only installs


cms/static/cms/sass/components/pagetree/_node-state.scss
```diff
@@ -7,14 +7,15 @@
...
+    vertical-align: top;
 }
 .cms-pagetree-node-state {
     &-empty {
         border-color: $gray-light;
         background-color: $white;
     }
     &-unpublished,
-    &-unpublishedparent {
+    &-unpublished-parent {
```
Brankub: Do not see any BE or FE changes related to this. :-| It was done before?

Brankub:
overall LGTM

cms/static/cms/sass/components/pagetree/_node-state.scss
vxsx: that is something that was overlooked way before

Brankub: 👍

PR #5318, opened 2016-05-18
fixes bug where users can create empty title/slug

czpython:
* covers cases where users try to modify advanced settings without creating title.
* covers cases where users have already created empty title/slug and try to modify advanced settings.


mkoistinen:
LGTM!

PR #5320, opened 2016-05-19
Show a message to a user instead of throwing an error

genti94:
Currently if page publishing fails, user gets no feedback, only `Uncaught error: [object Object]` in the console. This PR fixes that


vxsx:
👍

PR #5331, opened 2016-05-23
Fix a typo in tutorial docs.

WYSIATI:
The appending `You can` in `user/tutorial/page_settings.html` doesn't make any sense, which should be removed. Or it is a sentence to be finished. The best choice currently is to remove these two words.


FinalAngel:
LGTM

mkoistinen:
LGTM

PR #5333, opened 2016-05-23
feature/publish wizard page if first

czpython:
When creating page through admin, the first page is always published automatically.
This PR adds logic to the page wizard to do the same.

### DIFF
--- a/cms/forms/wizards.py
+++ b/cms/forms/wizards.py
@@ -295,9 +295,7 @@ class CreateCMSPageForm(BaseCMSPageForm):
                         add_plugin(**opts)
 
         # is it home? publish it right away
-        pages_in_site = Page.objects.filter(site_id=page.site_id)
-
-        if not self.page and pages_in_site.count() == 1:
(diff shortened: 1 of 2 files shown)


cms/forms/wizards.py
```diff
@@ -294,6 +294,12 @@ def save(self, **kwargs):
                         }
                         add_plugin(**opts)
 
+        # is it home? publish it right away
+        pages_in_site = Page.objects.filter(site_id=page.site_id)
+
+        if not self.page and pages_in_site.count() == 1:
```
piotrjakimiak: I think `not pages_in_site.exclude(pk=page.pk).any()` will have better performance.

czpython: @piotrjakimiak Do you mean ``.exists()``?

piotrjakimiak: Yes, I mean `exists()`

mkoistinen:
LGTM

PR #5339, opened 2016-05-24
Added cachebusting to pagetree menu endpoints

vxsx:
The issue was encountered when some browsers were caching the
output of `get-tree` endpoint which resulted in new pages created
by e.g. wizard not being shown in the pagetree.


FinalAngel:
LGTM

Brankub:
👍

PR #5345, opened 2016-05-27
fixes #4599

### LINKED ISSUE #4599: Decorator csrf_exempt won't work with CMS urls
Hi,

at some point I include the cms urls like this
```python
url(r'^', include('cms.urls'))
```
and I have a view:
```python
@csrf_exempt
def postFromExternal(request):
    pass
```

This view was included in the cms and reachable. In my case this was the callback url from a payment provider, which would send confirmation data via POST. The problem is that it has no CSRF token (and should not have?!), but the CSRF cookie from the client is set and also submitted. So the check from the CSRF middleware will fail.

As far as I can see from the debugger, it seems that the cms also adds a decorator to the already decorated `postFromExternal`, so it is double-wrapped. This leads to the problem that the CSRF middleware won't see the `csrf_exempt = True`, as this is only set for the inner wrapped view.

I was able to do a workaround by adding the url right before including the cms urls:
```python
url(r'^url/to/postFromExternal$', postFromExternal, name="postFromExternal"),
url(r'^', include('cms.urls')),
```

This works, but doesn't feel "clean" to me. So maybe the CMS wrapper could take a look at certain attributes of the wrapped view and copy them?

czpython:
Allow apphook endpoints to disable csrf checks
Fixes #4599

### DIFF
--- a/cms/tests/test_apphooks.py
+++ b/cms/tests/test_apphooks.py
@@ -529,6 +529,28 @@ class ApphooksTestCase(CMSTestCase):
         self.assertIsNotNone(TestApp2)
 
     @override_settings(ROOT_URLCONF='cms.test_utils.project.second_urls_for_apphook_tests')
+    def test_apphook_csrf_exempt_endpoint(self):
+        self.create_base_structure(NS_APP_NAME, 'en', 'instance_ns')
+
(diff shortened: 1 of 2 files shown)


vxsx:
changelog?

Issue #3748, opened 2015-01-11
Misleading error when server not available for AJAX requests


mkoistinen:
With runserver off, and attempting to edit a plugin on a page, the user gets the error message:

"SecurityError: Failed to read the 'contentDocument' property from 'HTMLIFrameElement': Blocked a frame with origin "http://127.0.0.1:8000" from accessing a cross-origin frame."

pajusmar:
Related issue #3028

vxsx:
Should be fixed by #5360

kalombos:
I have  the same error in chromium when i use S3 as storage for media files. I created simple plugin with ImageField and tried to view the image in edit mode and got the error. Does any solution exists?

czpython:
Fixed by https://github.com/divio/django-cms/pull/5360

Issue #3028, opened 2014-04-08
Missing error handling when plugin changes are "saved"


intgr:
In CMS 3.0, when I edit the contents of a CMS plugin via frontend editing and "save" it, the AJAX request is fired off without properly checking its status:
1. If I stop the web server before clicking Save, then the spinner is displayed forever even though the request fails immediately.
2. If the request fails with an HTTP error, for example, missing permissions in the database, the frontend acts as if the save was successful, but when I open it again, it wasn't actually changed.

I haven't yet tracked down the code that is causing this.

mkoistinen:
@intgr Is this still an issue for you in recent builds (3.0.10, in particular)? Perhaps this was satisfied already with some commit for 3.0.6 but we forget to close this issue?

intgr:
@mkoistinen Yes, still a problem. In the 1st case I now get some cross-origin error and in the 2nd case it spins forever.

mkoistinen:
@intgr thanks. We'll investigate.

stefanfoulis:
I can confirm as well, that this is still a problem.

FinalAngel:
This requires more steps within the backend to give me this feedback

yakky:
@FinalAngel let's talk about the support you need on the backend

czpython:
@FinalAngel Is this still an issue on 3.2/3.3?

FinalAngel:
we have a messaging system but we are not checking for imminent failures there /cc @vxsx

vxsx:
I believe both cases are still not covered in 3.3
The first one stops the spinner but doesn't really show correct error message (and error timeouts in e.g. reloadBrowser are not handled at all) 
second one is also not handled sadly

vxsx:
Should be fixed by #5360

PR #5360, opened 2016-06-02
Fixed save error handling in the modal.

vxsx:
Whenever form is submitted in the modal and the reponse returns
http error the cms assumed everything went correctly and closed the
modal thus not showing an error to an unsuspecting user. This
commit changes that and requires the success message to be there to
properly trigger closing of the modal in this case.

Same goes for the error that happens when iframe couldn't be loaded
at all, previously it was throwing DOM Exception and that DOM
exception was being displaying to the user. Now it shows the actual
error that is happening.

Ref: #3028, #3748


cms/templates/cms/toolbar/toolbar_javascript.html
```diff
@@ -34,7 +34,8 @@
             'publish': '{% filter escapejs %}{% trans "Are you sure you want to publish this page?" %}{% endfilter %}',
             'addPluginPlaceholder': '{% filter escapejs %}{% trans "Plugin will be added here" %}{% endfilter %}',
             'confirmDirty': '{% filter escapejs %}{% trans "You have unsaved changes." %}{% endfilter %}',
-            'confirmDirtyESC': '{% filter escapejs %}{% trans "Are you sure you want to abandon these changes?" %}{% endfilter %}'
+            'confirmDirtyESC': '{% filter escapejs %}{% trans "Are you sure you want to abandon these changes?" %}{% endfilter %}',
+            'errorLoadingEditForm': '{% filter escapejs %}{% trans "Could not load the form. Please check if the server is running." %}{% endfilter %}'
```
vxsx: @evildmp could you please check if the message is ok to you?

FinalAngel:
LGTM

cms/templates/cms/toolbar/toolbar_javascript.html
evildmp: "The form could not be loaded. Please check that the server is running correctly."

PR #5364, opened 2016-06-02
Webpack

vxsx:
smarter bundling,
deterministic dependecy resolution
makes way for future possible es6 support, better libraries integration
removes the need for globally exposed jquery

- [x] changelog
- [x] docs


cms/templates/cms/wizards/base.html
```diff
@@ -15,8 +15,5 @@
 
 {% block extrahead %}
     {{ block.super }}
-    <script type="text/javascript" src="{% static_with_version "cms/js/modules/jquery.noconflict.pre.js" %}"></script>
```
FinalAngel: I'm not sure if we can remove it here, I remember that dependencies such as select2 required the namespaced django jQuery?

vxsx: addressed in a different way, please re-review :)

CHANGELOG.txt
```diff
@@ -1,3 +1,9 @@
+=== 3.4.0 (unreleased) ===
+
+* Changed CMS JavaScript bundling from simple concatenation to webpack-based.
```
FinalAngel: might we want to add it to 3.3.1?

cms/static/cms/js/libs/jstree/jstree.grid.min.js
```diff
@@ -9,6 +9,815 @@
  * Works only with jstree version >= 3.0.0
  *
  * $Date: 2015-11-29 $
- * $Revision:  3.4.2 $
+ * $Revision:  3.4.2 $ !!! IMPORTANT it says 3.4.2 but it's actually 3.4.4, and it's modified!
```
FinalAngel: pls make this comment far bigger :D

FinalAngel: can we also mention which lines where modified?

cms/templates/cms/wizards/base.html
FinalAngel: 👍 though needs testing :)

FinalAngel:
Overall LGTM, nice work!

CHANGELOG.txt
vxsx: we can, however it would require some coordination on addons (e.g. djangocms-text-ckeditor needs to be released and pinned in appropriate places)

PR #5373, opened 2016-06-06
Fix plugins that require widgets

vxsx:
will be backported to 3.3.1


cms/templates/admin/cms/page/plugin/change_form.html
```diff
@@ -1,5 +1,17 @@
 {% extends "admin/change_form.html" %}
-{% load cms_admin %}
+{% load cms_admin cms_static %}
+
+{% block extrahead %}
+    {# in case plugins require widgets, they need to have bundle here #}
+    <script src="{% static_with_version "cms/js/dist/bundle.admin.base.min.js" %}" type="text/javascript"></script>
```
FinalAngel: remove type

FinalAngel:
LGTM

Issue #5085, opened 2016-03-15
Plugin edit buttons not working when render_plugin is not called in the template


petrklus:
I have created very simple plugin to allow per-block translation of a page. It works by creating Translated Block Holder, which can then contain multiple TranslatedBlock children. Each child can have

I am having problems when using the structure editor. I can add and edit Holder objects without any problems, however, their children do not seem to be editable. Nothing happens when I click the "pencil" icon, and double-click shows the menu of the parent, not the actual object.

In the screenshot below, I can click the "pencil" for the translated block holder and the first block underneath - however, I cannot even expand the second (de) block - nothing happens on click. No errors are reported in console and I've tested this both on Firefox and Chrome.

![image](https://cloud.githubusercontent.com/assets/1115667/13788647/e942fd96-eae0-11e5-9814-2882515ea8a4.png)

[cnk_djangocms_translated_block.zip](https://github.com/divio/django-cms/files/174522/cnk_djangocms_translated_block.zip)

petrklus:
Just upgraded to 3.2.3 and still seeing the issue.

petrklus:
Furthermore, the plugins that are "broken" also cannot be dragged anywhere - it looks like they are missing some init parameters.

petrklus:
I think have have found the problem, though not sure what the solution may be. In my template, I've got the following:

    {% if LANGUAGE_CODE == translatedblock.show_on_lang %}
    <div class="">
    {% for plugin in translatedblock.child_plugin_instances %}
    	{% render_plugin plugin %}
    {% endfor %}
    </div>
    {% else %}
    
    {% endif %}

Which means that the render_plugin is only called when the language matches. On the plugins where the render_plugin was omitted, the frontend admin did not work. Any tips how to make a render_plugin "dry run"?

vxsx:
It looks like this problem happens because some plugins have no markup in the DOM ), and historically we rely on the markup to store the data about the plugin, and if we cannot access this data the plugin is considered immovable (and it's also not initialized correctly).

It is a problem that is on our list, it's going to be addressed in it's current state soon.

petrklus:
@vxsx crossed wires! :) 

Any tips how to effectively run {% render_plugin plugin %} without the output being shown? My use case is to allow simple customisation of a page without the need to translate/copy all of the contents.

vxsx:
Well, ugly quickfix would be to check if cms user is logged in and render everything, but non-matching ones in invisible divs, and if not to not render.

[...]

PR #5375, opened 2016-06-06
Fixed an issue where plugins with no DOM weren't editable.

vxsx:
We relied on the data stored in the cms-plugin divs, now we rely on
data stored on cms-draggable divs for moving/editing.

Also fixed small bug with order of doubleclick handlers for structureboard. Fixes segmentation plugin and similar issues

Ref #5085


FinalAngel:
LGTM 👍

PR #5378, opened 2016-06-06
Fix #5371. Properly check parent_classes in get_child_classes

### LINKED ISSUE #5371: parent_classes in CMSPlugin
I found possible problem in aldryn_segmentation but IMHO it's a bug in CMS.

Aldryn Segmentation plugins have parent_classes property set.

1. Create parent plugin (SegmentLimitPlugin)
2. `render_plugin` for instance is called with placeholder passed as argument.
3. Child plugins are computed (https://github.com/divio/django-cms/blob/854a4c245d77f7cf26f167409c292cca5d78fcd3/cms/models/pluginmodel.py#L185)
4. `CMSPluginBase.get_child_classes` call `plugin_pool.get_all_plugins(slot, page)`
5. Placeholder is passed so https://github.com/divio/django-cms/blob/854a4c245d77f7cf26f167409c292cca5d78fcd3/cms/plugin_pool.py#L185 is used.
6. Aldryn Segmentation child classes has `plugin.requires_parent_plugin(placeholder, page) == True` so plugins are filtered out.
7. I cannot create any SegmentPlugin inside SegmentLimitPlugin.

I think that `plugin_pool.get_all_plugins` should take plugin instance (SegmentLimitPlugin in my case) and do not filter out plugins that have parent_classes equal to given instance.

I can create PR if i'm right.

piotrjakimiak:
Fixes #5371

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -499,7 +499,7 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
         # Useful in cases like djangocms-text-ckeditor
         # where only text only plugins are allowed.
         from cms.plugin_pool import plugin_pool
-        return plugin_pool.get_all_plugins(slot, page)
+        return plugin_pool.get_all_plugins()
 
     def get_child_classes(self, slot, page):
         """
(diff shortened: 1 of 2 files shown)


czpython:
Perfect. Thanks!
Ok to merge

mkoistinen:
LGTM, thanks @piotrjakimiak!

PR #5384, opened 2016-06-08
corrupted context in render_plugin_toolbar_config tag

Air-Mark:
There is a problem in render_plugin_toolbar_config tag because we call update but we never call pop of context.
I use render_plugin_toolbar_config to enable editing of child plugins without rendering them, but after toolbar refactoring, that tag got a very strange behaviour. So I found that when we call render_plugin we fill up some variables in context like "allowed_child_classes", and then if we use render_plugin_toolbar_config in the template of current plugin to render it's children we fill up exactly the save variables. So when the context comes to the toolbar_plugin_processor it's corrupted with values of it's own children plugins.

### DIFF
--- a/cms/templatetags/cms_admin.py
+++ b/cms/templatetags/cms_admin.py
@@ -6,6 +6,7 @@ from cms.constants import PUBLISHER_STATE_PENDING
 from cms.utils import get_cms_setting
 from cms.utils.admin import get_admin_menu_item_context
 from cms.utils.permissions import get_any_page_view_permissions
+from copy import copy
 from django import template
 from django.conf import settings
 from django.utils.encoding import force_text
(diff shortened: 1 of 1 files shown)


czpython:
@Air-Mark 
Hello, thanks for your contribution.

I would rather use Django's ``.push()`` context manager for this.
An added benefit is that we can leverage the cached template mechanism from the toolbar, like so:

```python
def render_plugin_toolbar_config(context, plugin, placeholder_slot=None,
                                 template='cms/toolbar/plugin.html'):
    request = context['request']

    if toolbar:
        template = toolbar.get_cached_template(template)
    else:
        template = context.template.engine.get_template(template)

    page = context['request'].current_page
    cms_plugin = plugin.get_plugin_class_instance()

    if placeholder_slot is None:
        placeholder_slot = plugin.placeholder.slot

    child_classes = cms_plugin.get_child_classes(placeholder_slot, page)
    parent_classes = cms_plugin.get_parent_classes(placeholder_slot, page)

    updated = {
        'allowed_child_classes': child_classes,
        'allowed_parent_classes': parent_classes,
        'instance': plugin
    }

    with context.push(**updated):
        return template.render(context)
```

Also please include some regression tests.

czpython:
As we're nearing release, replaced by https://github.com/divio/django-cms/pull/5523

PR #5393, opened 2016-06-10
forward port of #2212 to fix #3547

### LINKED ISSUE #3547: We should redirect when serving a "guessed" page
Suppose we have a page with English and French translations:

    /en/car
    /fr/voiture

If a user visits `/fr/car`, the French version will be served *from the given URL*.

It seems:

* **correct** that the French version is served, even though the page slug (`car`) is wrong - it's a good guess by the CMS
* **incorrect** that it doesn't redirect to `/fr/voiture` when it does this (bad for SEO, for sane-looking URLs, for common-sense)

So on detecting this behaviour, we should issue a `301 permanent redirect`.

czpython:
This is a really simple forward port of #2212 to fix #3547
I explicitly left all code improvements out of this PR.
The idea is to reduce the amount of changes because 3.0.x is phasing out.

### DIFF
--- a/cms/tests/views.py
+++ b/cms/tests/views.py
@@ -9,7 +9,7 @@ from django.contrib.auth.models import Permission
 from django.core.urlresolvers import clear_url_caches
 from django.http import Http404
 from django.template import Variable
-from cms.api import create_page
+from cms.api import create_page, create_title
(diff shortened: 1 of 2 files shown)


yakky:
LGTM

stefanfoulis:
LGTM

PR #5397, opened 2016-06-12
docs:intro:apphooks - The "next" section isn't Third-Party Apps

k-funk:


evildmp:
Thanks @k-funk. Would you rebase this on the release/3.3.x branch? That is where this patch should land.

See http://docs.django-cms.org/en/release-3.3.x/contributing/index.html#branch-policy.

k-funk:
I've rebased against `3.3.x`. Should this also be merging into `3.3.x` instead of `develop`? If so, I'll make a new PR, since I don't think there's a way to change that.

czpython:
@k-funk 
Yes, please open PR against 3.3.x and then we'll forward port to develop.
Thanks :)

Issue #5403, opened 2016-06-13
Alias plugin is always cached


sunfirester:
I have been using the alias plugin to duplicate two different headers across pages on my site. I have user specific data in the header itself, which means I can't cache the header placeholder. It seems that when I create an alias to my header placeholder, all the user data is cached.

I can confirm this is happening on pages with the alias plugin, and not on pages with original placeholders.

Thanks in advance.

sunfirester:
I created a PR for this: #5404

FinalAngel:
@divio/django-cms-core is this okay to merge?

yakky:
I think there might be a way to "inherit" caching setting from contained plugins, as a more general solution

mkoistinen:
@sunfirester I'm delighted that you're providing a proposed fix for your issue! Thank you! However, please note that 3.2.5 introduced the get_cache_expiration() method that works along side the cache attribute (which was retained for convenience and backwards compatibility. I've made a similar note on your PR, but perhaps we can work together to find the best solution here. I was planning to do some work on the Alias plugin soon anyway. =)

PR #5404, opened 2016-06-13
disable caching for alias plugins

sunfirester:
#5403


yakky:
As a stop gap looks good

sunfirester:
would something like cache = cmsplugin_ptr.cache work?

yakky:
@sunfirester the cache setting is anyway on the plugin class, not the plugin model. 
Probably defininig ``get_cache_expiration`` (see https://github.com/divio/django-cms/blob/53fd43ec4b383d11e7fa258b130efb33263d6c73/cms/plugin_base.py#L194) in the alias plugin class (https://github.com/divio/django-cms/blob/develop/cms/cms_plugins.py#L30) will do.

Something like this should work::

    def get_cache_expiration(self, request, instance, placeholder):
        if instance.plugin.get_plugin_class().cache:
            return 0
        return None

mkoistinen:
Both the `cache` attribute and the `get_cache_expiration` value should be inherited from the primary plugin. I think if we just set these to False/zero, we'll be trading one issue for another.

sunfirester:
Thank you for your replies guys, I didn't even know about get_cache_expiration.
I agree about inheriting these values from the primary plugin. I will adjust my PR accordingly later today for you to review, and use yakky's suggestion in my project in the meanwhile.

sunfirester:
I changed my PR to try to inherit caching. I see now why my PR was incorrect. If you guys have time, please let me know if you think this makes any sense. Thanks again for your help :)

sunfirester:
My second attempt was wrong as well... I'm not sure how I could grab the cache attribute of the CMSPlugin inside of AliasPlugin.

czpython:
Closing as inactive.
Feel free to open a new pull request based off the latest develop branch.

PR #5412, opened 2016-06-14
Implement four-level CMS_PLACEHOLDER_CONF

### LINKED ISSUE #4979: Exclude (Blacklist) Plugins from all placeholders

### LINKED ISSUE #5000: Allow to hide / deregister plugins from configuration

yakky:
This replaces #5006 

Currently ``CMS_PLACEHOLDER_CONF`` only allows values for placeholder and template+placeholder keys
Sometimes a global configuration would make things much nicer / simpler
@ojii suggested a for level structure like (in increasing order of precedence)

    CMS_PLACEHOLDER_CONF[None] (global)
    CMS_PLACEHOLDER_CONF['template'] (if template is given)
    CMS_PLACEHOLDER_CONF['placeholder']
    CMS_PLACEHOLDER_CONF['template placeholder'] (if template is given)

This is backward-compatible and allow a few neat tricks.

This also implement ``excluded_plugins`` key to exclude plugins from the list of available one

Changes needed
 * [x] 4-level configuration
 * [x] tests
 * [x] documentation

On top of this is easy to implement plugin blacklisting (see #5000, #4979), list of plugins to show in text editor and other plugin configuration options

A step further is to implement more flexible way to select configuration for each placeholder based on different attributes than name and template

Fix #5000 
Fix #4979

### DIFF
--- a/cms/tests/test_plugins.py
+++ b/cms/tests/test_plugins.py
@@ -219,6 +219,42 @@ class PluginsTestCase(PluginsTestBaseCase):
                 FilteredSelectMultiple,
             )
 
+    def test_not_add_plugin(self):
+        """
+        Test that you can't add a text plugin
+        """
+
+        CMS_PLACEHOLDER_CONF = {
+            'body': {
+                'excluded plugins': ['TextPlugin']
+            }
+        }
+
+        # try to add a new text plugin
+        with self.settings(CMS_PLACEHOLDER_CONF=CMS_PLACEHOLDER_CONF):
+            page_data = self.get_new_page_data()
+            self.client.post(URL_CMS_PAGE_ADD, page_data)
+            page = Page.objects.all()[0]
+            installed_plugins = plugin_pool.get_all_plugins('body', page)
+            installed_plugins = [cls.__name__ for cls in installed_plugins]
+            self.assertNotIn('TextPlugin', installed_plugins)
+
+        CMS_PLACEHOLDER_CONF = {
+            'body': {
+                'plugins': ['TextPlugin'],
+                'excluded plugins': ['TextPlugin']
+            }
+        }
+
+        # try to add a new text plugin
+        with self.settings(CMS_PLACEHOLDER_CONF=CMS_PLACEHOLDER_CONF):
+            page_data = self.get_new_page_data()
+            self.client.post(URL_CMS_PAGE_ADD, page_data)
+            page = Page.objects.all()[0]
+            installed_plugins = plugin_pool.get_all_plugins('body', page)
+            installed_plugins = [cls.__name__ for cls in installed_plugins]
+            self.assertNotIn('TextPlugin', installed_plugins)
+
     def test_plugin_edit_marks_page_dirty(self):
         page_data = self.get_new_page_data()
         response = self.client.post(URL_CMS_PAGE_ADD, page_data)
--- a/cms/tests/test_plugins.py
+++ b/cms/tests/test_plugins.py
@@ -219,14 +219,14 @@ class PluginsTestCase(PluginsTestBaseCase):
                 FilteredSelectMultiple,
             )
 
-    def test_not_add_plugin(self):
+    def test_excluded_plugin(self):
         """
         Test that you can't add a text plugin
         """
 
         CMS_PLACEHOLDER_CONF = {
             'body': {
-                'excluded plugins': ['TextPlugin']
+                'excluded_plugins': ['TextPlugin']
             }
         }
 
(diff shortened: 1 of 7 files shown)


cms/tests/test_plugins.py
```diff
@@ -219,6 +219,42 @@ def test_plugin_add_form_integrity(self):
                 FilteredSelectMultiple,
             )
 
+    def test_not_add_plugin(self):
```
czpython: please rename to `test_excluded_plugin`

cms/utils/conf.py
```diff
@@ -254,6 +255,12 @@ def get_unihandecode_host():
         return host + '/'
 
 
+def get_placeholder_config():
+    name = 'PLACEHOLDER_CONF'
+    placeholder_conf = getattr(settings, 'CMS_%s' % name, DEFAULTS[name])
+    return OrderedDict(placeholder_conf)
```
czpython: If `CMS_PLACEHOLDER_CONF` is not an `OrderedDict` when defined in the settings, casting it to an `OrderedDict` afterwards will not keep the ordering.

```python
>>> d = {'1': 1, '2': 2, '3': 3}
>>> from collections import OrderedDict
>>> OrderedDict(d)
OrderedDict([('1', 1), ('3', 3), ('2', 2)])
>>> d
{'1': 1, '3': 3, '2': 2}
```

yakky: Dictionaries are not ordered anyway.
And in the current implementation the order does not matter at all.
But casting to an OrderedDict makes easier to implement things in the future without having to change the data structure.

cms/utils/placeholder.py
```diff
@@ -39,32 +40,44 @@ def get_placeholder_conf(setting, placeholder, template=None, default=None):
...
-            inherit = conf.get('inherit')
-            if inherit:
-                if ' ' in inherit:
-                    inherit = inherit.split(' ')
-                else:
-                    inherit = (None, inherit,)
-                value = get_placeholder_conf(setting, inherit[1], inherit[0], default)
-                if value is not None:
-                    return value
+            for conf_key, conf in iteritems(placeholder_conf):
```
mkoistinen: We should write new code for Py3 and shim it for Py2. By using six.iteritems, we're actually doing the opposite. I suggest to just use plain items(). If this was in a tight loop of many items, there would be grounds for doing something else, but this will not be *that* inefficient in Py2 to make it worthwhile.

mkoistinen:
LGTM!

czpython:
Ok to merge

yakky:
@mkoistinen @czpython PR completed

cms/utils/placeholder.py
```diff
@@ -39,32 +39,44 @@ def get_placeholder_conf(setting, placeholder, template=None, default=None):
...
+
+    CMS_PLACEHOLDER_CONF[None] (global)
+    CMS_PLACEHOLDER_CONF['template'] (if template is given)
+    CMS_PLACEHOLDER_CONF['placeholder']
+    CMS_PLACEHOLDER_CONF['template placeholder'] (if template is given)
     """
+
     if placeholder:
         keys = []
+        placeholder_conf = get_cms_setting('PLACEHOLDER_CONF')
```
czpython: minor question, why call `force_text` on placeholder, template and the keys below?

yakky: A remnant of previous implementations, they can be safely removed

czpython:
Looks good! Ok to merge :shipit:

cms/utils/placeholder.py
yakky: removed

mkoistinen:
LGTM

jrief:
Great feature!

... to just add my two cents:
It would be great to also have a per placeholder configuration. Here we could reuse the existing, but unused field ``Placeholder.default_width``, rename and replace it against a ``JSONField``. Then arbitrary information could be attached to each placeholder object.

PR #5413, opened 2016-06-14
Fixed #3414 -- Allow non-superusers to copy all plugins.

### LINKED ISSUE #3414: Function "Copy all" not usable for users without superuser rights
For placheolders from placeholder model fields the function is not usable. The has_copy_plugin_permission from PlaceholderAdmin and PageAdmin differs. The target placeholder for copy all is the user clipboard. The PageAdmin function takes into account that the clipboard placeholder has no page and doesn't check the permissions for it. The function from PlaceholderAdmin does not and checks them. But there couldn't be found any permissions for the clipboard and the copy permission is always denied!

Permission check for target placeholder @PlaceholderAdmin: https://github.com/divio/django-cms/blob/develop/cms/admin/placeholderadmin.py#L151 vs permission check @PageAdmin https://github.com/divio/django-cms/blob/develop/cms/admin/pageadmin.py#L549

Target Placeholder permission check leads to: https://github.com/divio/django-cms/blob/develop/cms/models/placeholdermodel.py#L103-L114 no superuser, no page, no attached objects -> return None

czpython:
Fixes #3414

Also, adapted tests to no longer use a super user to tests permissions.

### DIFF
--- a/cms/admin/settingsadmin.py
+++ b/cms/admin/settingsadmin.py
@@ -15,6 +15,7 @@ from cms.utils.urlutils import admin_reverse
 
 
 class SettingsAdmin(ModelAdmin):
+
     def get_urls(self):
         def wrap(view):
             def wrapper(*args, **kwargs):
--- a/cms/models/placeholdermodel.py
+++ b/cms/models/placeholdermodel.py
@@ -182,7 +182,7 @@ class Placeholder(models.Model):
         """
         Returns an ITERATOR of all non-cmsplugin reverse foreign key related fields.
         """
-        from cms.models import CMSPlugin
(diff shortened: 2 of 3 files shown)


mkoistinen:
LGTM

yakky:
I admit I don't get this :)
Could you elaborate the reason why https://github.com/divio/django-cms/pull/5413/files#diff-0aa10673163e9997607d1d78edd0c030R202 fixes the issue?
It's not a critic, it's just my brain which apparently can't understand the change

czpython:
@yakky 
Thanks for the feedback.
I've updated the line in question with comments to explain what's going on.

yakky:
@czpython thansk for the comments LGTM

PR #5429, opened 2016-06-17
No-markup-edit-mode*

vxsx:
essentially removes the `div class="cms-plugin"` in favor of handling all the necessary stuff with javascript

- [x] changelog
- [x] more js unit tests
- [ ] ~~more js integration tests~~ not now, but soon
- [x] screenshots / videos

----

#### Before
http://d.pr/v/16z3b &larr; VIDEO
screen shot 2016-06-27 at 11 46 22

#### After
http://d.pr/v/1kl3T &larr; VIDEO
screen shot 2016-06-27 at 11 45 58

Essentially what happens is we have invisible nodes before and after plugin contents which are removed by js and their data is put onto plugin contents. In case the plugin contents have top level non-whitespace text nodes they are wrapped with a custom `<cms-plugin>` tag that has no styles and should not interfere with anything. 

There are still extreme cases where things could go wrong, e.g. plugin markup is like this:

`text node <!-- comment node --> and text node`

which would be renreded then as `<cms-plugin ...>text node </cms-plugin><!-- comment node --><cms-plugin ...>and text node</cms-plugin>` which could interfere with styles like `* + * { margin-top: 10px; }` but I believe this is rare enough case :)

--
<sup>*</sup> there's still a little bit of markup


cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -48,6 +48,8 @@ var Plugin = new Class({
     },
 
     initialize: function initialize(container, options) {
+        var that = this;
```
FinalAngel: why do we need ``that`` here?

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -94,9 +106,39 @@ var Plugin = new Class({
      */
     _setupUI: function setupUI(container) {
         var wrapper = $('.' + container);
+        var contents;
+
+        if (wrapper.length > 1) {
```
FinalAngel: we should probably show the loader as long as this is processing?

czpython:
@vxsx develop was force pushed to be clone of release/3.3.x just FYI

cms/static/cms/js/modules/cms.plugins.js
vxsx: not really. all of this is synchronous and is blocking the main thread anyways
in 99.9999% of the cases it won't be noticeable

vxsx:
@divio/django-cms-core please review, quite a change :)

CHANGELOG.txt
```diff
@@ -1,3 +1,12 @@
+=== 3.4.0 (unreleased) ===
+
+* Changed the way cms plugins are rendered. The div with `cms-plugin` class is
+  no longer rendered around every plugin, instead a combination of `tempalte`
```
FinalAngel: typo ``template``

CHANGELOG.txt
```diff
@@ -1,3 +1,12 @@
+=== 3.4.0 (unreleased) ===
+
+* Changed the way cms plugins are rendered. The div with `cms-plugin` class is
```
FinalAngel: CMS

cms/static/cms/js/modules/cms.plugins.js
FinalAngel: 👍

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -176,51 +223,55 @@ var Plugin = new Class({
...
-            var placeholderId = that._getId(that.ui.dragitem.closest('.cms-dragarea'));
-            var placeholder = $('.cms-placeholder-' + placeholderId);
+        if (!Plugin._isContainingMultiplePlugins(this.ui.container)) {
+            this.ui.container.on(this.doubleClick, dblClickHandler);
+            this.ui.container.on(this.pointerOverAndOut + ' ' + this.touchStart, function (e) {
+                // required for both, click and touch
+                // otherwise propagation won't work to the nested plugin
+                e.stopPropagation();
+                if (e.type === 'touchstart') {
+                    CMS.API.Tooltip._forceTouchOnce();
```
FinalAngel: do we really want to call a private method here? Should we make it public?

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -297,9 +348,10 @@ var Plugin = new Class({
...
 
-        if (clipboardPlugin.data('settings')) {
-            var type = clipboardPlugin.data('settings').plugin_type;
-            var parent_bounds = $.grep(clipboardPlugin.data('settings').plugin_parent_restriction, function (r) {
+        if (clipboardDraggable.data('cms')) {
+            var clipboardPluginData = clipboardDraggable.data('cms');
+            var type = clipboardPluginData.plugin_type;
+            var parent_bounds = $.grep(clipboardPluginData.plugin_parent_restriction, function (r) {
                 // special case when PlaceholderPlugin has a parent restriction named "0"
                 return r !== '0';
```
FinalAngel: what is "r" would rather have it written out

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -673,8 +733,18 @@ var Plugin = new Class({
 
         // set new setting on instance and plugin data
         this.options = settings;
-        plugin.data('settings', settings);
-        draggable.data('settings', settings);
+        if (plugin.length) {
+            var index = plugin.data('cms').findIndex(function (pluginData) {
```
FinalAngel: index should be defined before the if or?

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -711,33 +781,156 @@ var Plugin = new Class({
...
+                        plugin: plugin
+                    });
+                    return false;
+                }
+            } else if (parent.length && child.length) {
+                // if we can't find a plugin on the same level, we need to travel higher
+                // for this we need to find the deepest child
+
+                // TODO this is _just_ an assumtion because in reality we do not know where
+                // would the children be positioned in case the plugin contains multiple divs
```
FinalAngel: ``where the children would be``

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -711,33 +781,156 @@ var Plugin = new Class({
...
+        }
+
+        // if we did not found any match, require reload
+        return true;
+    },
+
+    /**
+     * If a plugin is moved to the the plugin that has no DOM representation
+     * we need to add the data about this plugin to the freshly moved node.
+     * Adds the appropriate classes and pushes data into "cms". The parent data
```
FinalAngel: ``CMS``

vxsx: it's `data('cms')`, i would leave lowercase

cms/static/cms/js/modules/cms.plugins.js
vxsx: not really. we don't enforce 'vars-on-top' in cms, too much stuff would need updating

cms/static/cms/js/modules/nextuntil.js
```diff
@@ -0,0 +1,24 @@
+var $ = require('jquery');
```
FinalAngel: shouldn't it be called jquery.nextuntil?

cms/static/cms/js/modules/cms.plugins.js
vxsx: restriction, fair enough

cms/tests/frontend/integration/editContentTools.js
```diff
@@ -80,11 +80,13 @@ casper.test.begin('Edit utils page content', function (test) {
         })
         // wait till the paste actually succeeds
         .waitForResource(/move\-plugin/)
-        .reload()
-        // check that number of content plugins has been indeed increased
         .then(function () {
+            this.reload();
+        })
+        // check that number of content plugins has been indeed increased
```
FinalAngel: not sure about bit it might be ``have been``

cms/static/cms/js/modules/cms.plugins.js
vxsx: yes we do need to call it, since it won't propagate and won't be catched by the tooltip itself, and no, we shouldn't make it public

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -340,8 +336,8 @@ var StructureBoard = new Class({
         this.ui.container.show();
         this.ui.dragareas.css('opacity', 1);
 
-        this.ui.plugins.not(this.ui.render_model).hide();
-        this.ui.placeholders.show();
+        // this.ui.plugins.not(this.ui.render_model).hide();
```
vxsx: should remove these

FinalAngel:
Overall really great work. I'm a bit concerned about performance here. Do you have any indication on the gains / loses compared to the current version?

cms/static/cms/js/modules/cms.plugins.js
FinalAngel: alright, though I'm not a fan of calling private API methods.

cms/static/cms/js/modules/nextuntil.js
vxsx: jquery.nextuntil won't be linted :)
don't think it's really necessary to indicate the dependency in the filename, now that it's explicit in the source

cms/static/cms/js/modules/cms.plugins.js
FinalAngel: alright, fair enough, I'll keep that in mind 👍

cms/static/cms/js/modules/cms.plugins.js
FinalAngel: 👍

vxsx:
Performance is a thing, yes, it is slower, given that it has to do more stuff.

Here is the comparsion of plugins initialization on a 450 plugins tree:
old [old-init.cpuprofile.zip](https://github.com/divio/django-cms/files/336625/old-init.cpuprofile.zip) 1.76s
new [new-init.cpuprofile.zip](https://github.com/divio/django-cms/files/336626/new-init.cpuprofile.zip) 2.24s

So it's slower for about 1.07ms per plugin on a good machine. Pretty sure it's fine :)

vxsx:
Ok, so after few iterations i've made it go down to 1.48s for 450 plugins, some parts are going to be here, some will be opened against 3.3.x and forwardported, so everyone can benefit

FinalAngel:
LGTM awesome!

PR #5433, opened 2016-06-18
We don't need placeholder edit permissions checking in view mode

yakky:
Currently we calculated edit permissions every time a placeholder is rendered, but that does not make any sense if we're not in edit mode.
Probably we can avoid putting placeholders in the request in that case, but this needs a few more changes, that need to be investigated

Thanks @vad for spotting this

### DIFF
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -125,12 +125,30 @@ def render_placeholder(placeholder, context_to_copy, name_fallback="Placeholder"
     request = context['request']
     if not hasattr(request, 'placeholders'):
         request.placeholders = {}
-    perms = (placeholder.has_change_permission(request) or not placeholder.cache_placeholder)
-    if not perms or placeholder.slot not in request.placeholders:
-        request.placeholders[placeholder.slot] = (placeholder, perms)
+
+    # Prepend frontedit toolbar output if applicable
+    toolbar = getattr(request, 'toolbar', None)
+    if (getattr(toolbar, 'edit_mode', False) and
+            getattr(toolbar, "show_toolbar", False) and
+            getattr(placeholder, 'is_editable', True) and editable):
+        from cms.middleware.toolbar import toolbar_plugin_processor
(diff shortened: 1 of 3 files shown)


yakky:
@divio/django-cms-core any comment?

mkoistinen:
Well, we were effectively checking perms twice, which is no good. So, nice one!

cms/plugin_rendering.py
```diff
@@ -125,12 +125,30 @@ def render_placeholder(placeholder, context_to_copy, name_fallback="Placeholder"
     request = context['request']
     if not hasattr(request, 'placeholders'):
         request.placeholders = {}
-    perms = (placeholder.has_change_permission(request) or not placeholder.cache_placeholder)
-    if not perms or placeholder.slot not in request.placeholders:
-        request.placeholders[placeholder.slot] = (placeholder, perms)
+
+    # Prepend frontedit toolbar output if applicable
```
czpython: maybe write this as :

```python

try:
except AttributeError:
	toolbar = None

if (toolbar and toolbar.edit_mode and toolbar.show_toolbar
     and placeholder.is_editable and editable):
     from cms.middleware.toolbar import toolbar_plugin_processor
     processors = (toolbar_plugin_processor, )
     edit = True
 else:
     processors = None
     edit = False
```

`Placeholder` model has `is_editable` as an attribute so no need to use `getattr`

czpython:
Minor comment. Looks good, thanks :)

cms/plugin_rendering.py
yakky: Yeah, just avoived cluttering the PR  with other optimization at first try
Code reworked now

czpython:
Ok to merge.

yakky:
@mkoistinen @czpython rebased

yakky:
@mkoistinen @czpython any more fixes needed to merge  this?

czpython:
@yakky Nope. Go for it.

yakky:
@czpython :+1:

PR #5436, opened 2016-06-20
Fixes #5405 -- Let Django handle pickling

### LINKED ISSUE #5405: CMSPlugin reduce function breaks in django 1.8
I have been working with pickling/caching models on my site recently. I am forced to work with django 1.8 since it is the TLS version. It seems that "model_unpickle" was changed in django 1.9, and djangocms changed to reflect that change. The problem is that pickling models now breaks with django 1.8. I have overwritten the "__reduce__" function in my models, but perhaps a better solution is needed.

I figured I would at least create an issue so other people would be aware. Here is what I did with my reduce function:

# overwritten the way a model is pickled so that we can pass a deffered_class_factory
    from django.db.models.query_utils import DeferredAttribute, deferred_class_factory
    def __reduce__(self):
        """
        Provide pickling support. Normally, this just dispatches to Python's
        standard handling. However, for models with deferred field loading, we
        need to do things manually, as they're dynamically created classes and
        only module-level classes can be pickled by the default path.
        """
        data = self.__dict__
        # The obvious thing to do here is to invoke super().__reduce__()
        # for the non-deferred case. Don't do that.
        # On Python 2.4, there is something wierd with __reduce__,
        # and as a result, the super call will cause an infinite recursion.
        # See #10547 and #12121.
        deferred_fields = [f for f in self._meta.fields
                           if isinstance(self.__class__.__dict__.get(f.attname),
                                         DeferredAttribute)]
        model = HeaderLink
        return (model_unpickle, (model, deferred_fields, deferred_class_factory), data)

czpython:
Fixes #5405

### DIFF
--- a/cms/models/pluginmodel.py
+++ b/cms/models/pluginmodel.py
@@ -10,8 +10,7 @@ from django.core.urlresolvers import NoReverseMatch
 from django.core.exceptions import ValidationError, ObjectDoesNotExist
 from django.db import models
 from django.db.models import signals, Model, ManyToManyField
-from django.db.models.base import model_unpickle, ModelBase
-from django.db.models.query_utils import DeferredAttribute
+from django.db.models.base import ModelBase
 try:
     # Django >= 1.8, < 1.9
     from django.db.models.fields.related import (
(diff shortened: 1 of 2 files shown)


yakky:
@czpython so pickle basically do the right thing without further hints from us, right?

czpython:
@yakky 
Django does https://github.com/django/django/blob/stable/1.8.x/django/db/models/base.py#L524
We overrode it because back then it wasn't fully implemented (2010).

yakky:
LGTM

mkoistinen:
LGTM. Its always nice when we can improve the code by reducing it =) (pun intended)

PR #5452, opened 2016-06-22
Fixes for CMSApp

jsma:
This adds a missing deprecation warning.

I also changed the docs to encourage the use of `get_menus()` vs setting `_menus` directly.

However, the API docs document both. Should developers only use `get_menus()` and `get_urls()` and treat `_menus` & `_urls` as internal APIs? Or should the docs encourage two different ways to configure the menus and urls for an Apphook?

### DIFF
--- a/cms/app_base.py
+++ b/cms/app_base.py
@@ -75,6 +75,9 @@ class CMSApp(object):
 
     @legacy_menus.setter
     def menus(self, value):
+        warnings.warn('Accessing CMSApp.menus directly is deprecated, '
+                      'and it will be removed in version 3.5; CMSApp.get_menus method',
+                      DeprecationWarning)
         self._menus = value
 
     def get_menus(self, page=None, language=None, **kwargs):
--- a/docs/how_to/apphooks.rst
+++ b/docs/how_to/apphooks.rst
@@ -204,8 +204,10 @@ To disable this behaviour set ``permissions = False`` on your apphook::
 
     class SampleApp(CMSApp):
         name = _("Sample App")
-        _urls = ["project.sampleapp.urls"]
         permissions = False
+
+        def get_urls(self, page=None, language=None, **kwargs):
+            return ["project.sampleapp.urls"]
 
 If you still want some of your views to use the CMS's permission checks you can enable them via a decorator, ``cms.utils.decorators.cms_perms``
 


yakky:
``_urls`` and ``_menus`` are still available for use. They are not going away, and for most use cases it's simpler to just use the attributes: function and attribute way to define urls and menus are both equally viable solutions.
I think that documentation example should use both styles to give them the same weight

czpython:
@yakky 
I think that I misinterpreted this.
Talking with @evildmp, I mentioned that  `_urls` and `_menus` are fully internal and only there to provide backwards compatibility with `urls`.

Exposing `_urls` and `_menus` as public apis feels strange because they start with `_`

jsma:
If there's consensus, I can remove the references from the API documentation as well. Let me know.

yakky:
To speak in Java terms they are meant to be ``private`` and not ``protected``
They must not be accessed directly from outside the CMSApp instances, but the classes themselves are free to override them. They are documented exactly to make it clear they are fully usable. I think it does not make any sense to use a function when you can use an attribute. The ``_`` prefix is only to discourage to access them from outside the class.

czpython:
@yakky If this is the case then why not stick with `urls` and `menus`?

yakky:
Because the only public API (I mean, they way you retrieve menus and urls) is the ``get_menus`` / ``get_urls`` functions
The rest is just implementation details

czpython:
@yakky 
I see your point but even in implementation there's internal (attributes internal to the core like `_state` on models) and external (attributes that can be overridden by subclasses like admin subclasses).

I'd like to stick with Django's approach, we provide a non `_` prefixed attribute (`menus` instead of `_menus`) for overriding the values of this attribute and then a getter method like `get_menus()` for fetching the attribute.

yakky:
``_urls`` and ``_menus`` **still** are private.
keeping the unprefixed version will communicate the wrong API. They **cannot** be accessed directly from outside the class

czpython:
@yakky 
I understand but still I don't think it's necessary. Using the admin as an example, developers know that they shouldn't access `fieldsets` directly, instead use `get_fieldsets()`.

yakky:
But django does not have a legacy api to deprecate which will do the wrong thing

czpython:
I'm ok with using `_` for an attribute that the user overrides and has it's own getter method if that becomes the standard in the available apis like `CMSPluginBase`, `CMSApp`, etc..

In my interpretation, using `_` as a prefix means that this attribute is private to the class, meaning that all interactions with this attribute will happen within the definition of the base class (in this case) and all usage of this attribute (read or write) from outside the base class is discouraged.

That said, what I'm advocating for is consistency. How about we get together and decide on the use of `_` as an attribute prefix in base classes?

/cc @divio/django-cms-core @timgraham

timgraham:
If we were designing the admin options from scratch, I don't think we would prefix them with an underscore. Rewriting [the fieldsets example](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets), I think it would be odd to have users write code that looks like:
```python
class FlatPageAdmin(admin.ModelAdmin):
    _fieldsets = (
        (None, {
            'fields': ('url', 'title', 'content', 'sites')
        }),
    )
```

yakky:
@timgraham If I were designign the CMSApp api from scratch I would do the same :)
The problem here is that we have to clearly separate the deprecated API from the new one

timgraham:
It might not be trivial, but I'd be surprised if there isn't some way to clean it up. Something like `django.utils.deprecation.RenameMethodsBase` that works with properties is what you're looking for, correct?

mkoistinen:
I think there are plenty of examples of things that SHOULD be prefixed with '_' but aren't, and there are even examples of things that shouldn't be, but are (`._meta`, for example), but when you use '_' it should be clear (outside of the legacy exceptions), that it is private, do not touch it. In this case, this is the desire, and that is how it is communicated.

jsma:
As an end user, I can live with moving to the `get_menus` style since it adds flexibility (although an example use case in the docs would be illuminating). Assigning to an `_` attribute feels like I'm doing something I shouldn't be doing.

Based on the original 3.3 docs, I updated a project to rename `menus` to `_menus`, `urls` to `_urls` etc but had no clue *why* this was necessary other than the docs said I had to because "it's the future".

Then there was a recent update to the docs that said emphasized using `get_menus` & `get_urls` instead. I tried to follow these evolving instructions and submitted this PR in a selfish effort to minimize confusion for my future self when creating `CMSApp` subclasses.

If I understand this thread correctly, there's no objection to the PR as it is, but sounds like there isn't consensus yet on what the documented API should be. If there's anything specific that needs to be fixed in this PR, let me know.

Thanks!

czpython:
@jsma 
Thanks for the follow up.
We've not reached a conclusion yet on which direction to go.
I'll update the ticket once something is decided.

Your feedback is greatly appreciated, thanks for that :)

czpython:
@jsma 
Thanks for the follow up.
We've not reached a conclusion yet on which direction to go.
I'll update the ticket once something is decided.

Your feedback is greatly appreciated, thanks for that :)

vThaian:
I'm closing because we don't support 3.3 anymore.

Issue #5449, opened 2016-06-22
Django CMS 3.3: Page tree burger icon has no effect if your page permissions are restricted


bartTC:
If your user is restricted to edit only a subset of pages, the burger icon has no effect on click. The 'Legend" icon too. 

![screen shot 2016-06-22 at 18 11 02](https://cloud.githubusercontent.com/assets/1896/16274245/105aadea-38a5-11e6-84e4-b42060301919.png)

I can see the HTML in the source code related to this, it's rendered there, it just doesn't pop up.

The icon works fine if I have permissions to edit all pages, or if I'm a superadmin. A popup pops up with meta info and copy/paste buttons. 

Django-CMS 3.3, Django 1.9.6

czpython:
Thanks for the report.
I noticed this as well but missed to create ticket 👍

vxsx:
@bartTC could you please check if https://github.com/divio/django-cms/pull/5461 fixes this for you?

bartTC:
@vxsx this fixed the bug, thank you all!

PR #5461, opened 2016-06-28
Fixed dropdowns not showing up when user has limited permissions.

vxsx:
The dropdowns show up as a result of a combination of css classes
added to correct nodes. This functionality relies on having the
same amount of triggers and dropdowns on the page, since they are
referenced by their index and not by parent.

Should fix https://github.com/divio/django-cms/issues/5449


FinalAngel:
LGTM

PR #5462, opened 2016-06-28
Added minor improvements for Plugin initialization.

vxsx:
When _setAddPluginModal is called it was previously creating Modal
instances that weren't likely to be used until later, so now we
instantiate them on demand. This saves about 1-2 ms per plugin,
which doesn't seem like much, but makes a big difference on a page
with 450 plugins, there initialization went down by ~400ms.

Also caching some jQuery variables, and getting rid of unnecessary
loops helps to improve the speed as well.


FinalAngel:
LGTM

PR #5463, opened 2016-06-28
Moved child-plugin-classes templates out of "js" sekizai block

vxsx:
In commit 4f2ec23, a ``<script type="text/cms-template">`` (for
holding client-side content that is not to be rendered until
runtime, and which we will use until the HTML element
``<template>`` is fully supported) was placed inside a Sekizai
block where it would not interfere with page rendering.

Unfortunately, this turned out to be incompatible with Django
Compressor divio#5454, which
raises an error on the (deliberately invalid)
``type="text/cms-template"``.

It has been moved out of the Sekizai block to fix this problem. The
render-blocking effects of the script are negligible.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -32,6 +32,8 @@
   mode.
 * Fixed a bug where users with limited permissions could not interact with
   page tree dropdowns.
+* Fixed a bug where Django Compressor could not be used on the sekizai ``js``
+  block.
 
 
 === 3.3.0 (2016-05-26) ===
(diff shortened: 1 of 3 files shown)


FinalAngel:
LGTM

CHANGELOG.txt
```diff
@@ -32,6 +32,8 @@
   mode.
 * Fixed a bug where users with limited permissions could not interact with
   page tree dropdowns.
+* Fixed a bug where django compressor could not be used on the sekizai ``js``
```
evildmp: ``Django Compressor``

czpython:
👍

PR #5466, opened 2016-06-29
Changed cache-busting implementation for static files

vxsx:
Since 3.2.1 django CMS has used the ``GET`` parameter to reset the cache of its
static files. This doesn't play well with proxies, CDNs and so on.

This patch invalidates caching by changing the paths of static files instead.

The django CMS release process now requires additional steps to account for the
change (documented).


RELEASE_INFO
```diff
@@ -5,8 +5,10 @@
 - Build newest english JS translations (django-admin.py makemessages -l en -d djangojs)
 - Pull translations (tx pull -f -a)
 - Compile translations (django-admin.py compilemessages)
+- Make sure icons are recreated in the new place (gulp icons)
 - Make sure css files are compiled from latest sass source (gulp sass)
 - Make sure js files are bundled and minified propery from latest source (gulp bundle)
+- Make sure to remove js and css files from the older version of CMS
```
evildmp: ``JS`` and ``CSS``

evildmp:
Can you explain in the commit message how it will make the release process more complicated?

More significantly, does this require any documentation changes?

vxsx:
@evildmp well yes, now the old files have to be removed when the cms version is bumped, and then regenerated with new version.

documentation wise i think it's worth adding that one should not load bundles from cms in 3rd party apps, and if they have to - they should use static_with_version templatetag. but not really sure where exactly should it be added

FinalAngel:
How will we fix this in djangocms-text-ckeditor?

vxsx:
@FinalAngel by using static_with_version there https://github.com/divio/djangocms-text-ckeditor/pull/327

czpython:
👍

FinalAngel:
LGTM :)

Issue #5471, opened 2016-06-30
Migration 0010 imports ContentType directly


doctormo:
ContentType model should be gotten via get_model (no models should be used directly in data migrations)

PR #5472, opened 2016-06-30
Fix migration 0010 use of ContentType model

doctormo:
This should fix issue #5471


czpython:
@doctormo 
Hi there,
Thanks for your contribution.
Please rebase and reopen against the `release/3.3.x` branch.

`develop` is only used for approved new features.

PR #5483, opened 2016-06-30
translation of missing stuff just added for persian language.

pouya-abbassi:
Hi!
Many things were left without translating and we are going to use django-cms as our main cms from now on, so we had to translate django-cms completly.

I'll work more on django-cms and I'll help translation and bug-fix (and maybe adding features) in future.
I'm not sure if I was supposed to commit to the develop branch or not, actually it's my first contribute to any public github project. Correct me if I was wrong on anything!


czpython:
@pouya-abbassi 
Hello there,
Thank you for your contribution.
And congratulations on your first contribution (to a public repo) :)

We use transifex to manage all translations for djangoCMS.
You can read more [here](http://docs.django-cms.org/en/release-3.3.x/contributing/translations.html#contributing-translations).

Once your translations have been approved on transifex, we'll take care of merging them in the core prior to releasing.

czpython:
T

pouya-abbassi:
Should I do it again using transifex or is it fine the way it is?

Plus, I found some of these translations (that has been done before) are confusing due to miss-matching situations.
You know! some words have different meaning on different situations!!!
Should I re-translate them like what I did right now or should I do it in transifex???

------------------------------
ps:
I need more time and more playing around to find out where should be edited so I can't do it right now.

czpython:
@pouya-abbassi 
Yes, please use transifex.

PR #5484, opened 2016-07-01
Renamed copy/pasting/updating events

vxsx:
In the past we were using namespaced events for moving/pasting and
updating structureboard events. Due to the nature of jQuery events
namespaces `cms.update` event could trigger `cms.plugins.update`
where it wasn't required, which in turn led to incorrect tree when
pasting plugins. This pr replaces custom events to be unique
strings so they don't interfere with each other.

requires forwardporting


lory87:
👍

FinalAngel:
LGTM

PR #5485, opened 2016-07-01
Fixed an issue when link plugin couldn't be edited

vxsx:
Since #5429 the plugin markup has `cms-plugin` class, which means
that link plugins markup would look like `<a class="cms-plugin" ...>`
which wasn't handled by that doubleclick handler that previously
allowed editing of plugins that contain links.

does not require backporting


FinalAngel:
LGTM

czpython:
@vxsx Is there a test for this?

vxsx:
@czpython nope

vxsx:
@czpython now there is :)

vxsx:
screen shot 2016-07-05 at 09 23 58

Issue #5382, opened 2016-06-08
Contextual menu not working when alias present on the same page


petrklus:
Hi,

I am having issue when using alias to the content on the same page as the content itself. 

The overview of the content structure is following:

Tabs container
 - tab1
   - alias to contentrow2
   - alias to contentrow3
 - tab2
   - contentrow2
 - tab3
  - contentrow3

However, as soon as I create the aliases, I can no longer expand/collapse the content row via the arrow (the expand all option still works), neither can I use the contextual menu (I can use the edit / + buttons).

As soon as the alias is deleted, the contextual buttons start working again for the particular content row.

I can PM login details for a stage server with the issue present on one of the pages. I can also easily reproduce it on my development machine. Sample screenshot:

![image](https://cloud.githubusercontent.com/assets/1115667/15897234/fe795e76-2d93-11e6-8748-674a483182ab.png)

The offending row is the "Content row" plugin, which cannot be expanded when the alias is present. Works just fine when the alias is removed.

czpython:
@petrklus 
Thanks for reporting.
I believe this is a known issue with the Alias plugin /cc @vxsx 
We're planning to refactor this plugin because of this and other issues.

vxsx:
It is a known issue sadly. This happens because the "Content row" plugin is initialized twice on the same page, and so it has 2 click handlers that toggle it, so first one opens it and second one closes. I can provide a small patch for this specific issue for 3.3.1 (as i've sorted it out while working on something else), but all the other problems with it (like when you try to move it when alias plugin is on same page) will stay unsolved until we refactor.

FinalAngel:
@petrklus are you eventually able to help us out there?

petrklus:
@FinalAngel I am happy to test / evaluate a fix, however, I am not yet at 3.3 and it may take a week or two to upgrade.

FinalAngel:
@petrklus helping on testing/evaluating will definitely be a big plus. Also general improvements on the frontend would be very welcome.

vxsx:
@petrklus could you please check if #5486 fixes things for you?

czpython:
Fixed by https://github.com/divio/django-cms/pull/5486

PR #5486, opened 2016-07-01
Fixed alias plugin preventing operations on the original plugin

vxsx:
When alias plugin is initialised it also initialises it's children.
If the alias plugin is on the same page as the plugin that is being
aliased this leads to the event handlers bound twice. So if you
try to toggle context menu or nested plugins in structure mode the
operation would be ran twice and would technically fail. This
commit stops initialisation of the plugin that already has been
initialised.

Essentially this is a backport of some of the changes from #5429
Ref: #5382


FinalAngel:
LGTM

PR #5489, opened 2016-07-01
Permissions cleanup (part 1)

czpython:
This PR introduces multiple bug fixes for the permissions system, as well as small cleanups and a complete overhaul of permission related tests.

As the title suggests, this is not a finished product but it's the "core" of the cleanup, future efforts will be based on this.

High level overview:

* Moved plugin add request validation logic to placeholder admin.
  This is because plugins should have no logic aside from the builtin admin logic
   to validate and allow an add request. Placeholders on the other hand provide a way
   to extend these checks in third party apps.
* Moved all (I hope) admin related tests from `test_placeholder`, `test_admin` and `test_page`
  into the more specific `test_page_admin` module.
* Created a `test_page_user_admin` module for tests that interact with the `PageUser` admin
* Created a `test_page_user_group_admin` module for tests that interact with the `PageUserGroup`
  admin.
* Created a `test_placeholder_app_admin` module for tests that interact with the `PlaceholderAdminMixin` class.
* Refactored page & placeholder tests to account for permissions.
* Removed logic to display unsaved (ghost) plugins in structure board.

Most of the additions/changes are tests, there's several bug fixes but usually one liners.
/cc @divio/django-cms-core

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -630,6 +630,19 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
             return False
         return True
 
+    def has_paste_plugin_permission(self, request, plugins, target_placeholder):
+        for plugin in plugins:
+            if not permissions.has_plugin_permission(request.user, plugin.plugin_type, "add"):
+                return False
+
+        page = target_placeholder.page
+
+        if page and not page.has_change_permission(request):
+            return False
+        if page and not page.publisher_is_draft:
+            return False
+        return True
+
     def has_move_plugin_permission(self, request, plugin, target_placeholder):
         if not permissions.has_plugin_permission(request.user, plugin.plugin_type, "change"):
             return False
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -1150,7 +1150,7 @@ class PageAdmin(PlaceholderAdminMixin, ModelAdmin):
                 return HttpResponseBadRequest(force_text(_("Language must be set to a supported language!")))
             for placeholder in placeholders:
                 plugins = list(
-                    placeholder.cmsplugin_set.filter(language=source_language).order_by('path'))
+                    placeholder.get_plugins(language=source_language).order_by('path'))
                 if not self.has_copy_plugin_permission(request, placeholder, placeholder, plugins):
                     return HttpResponseForbidden(force_text(_('You do not have permission to copy these plugins.')))
                 copy_plugins.copy_plugins_to(plugins, placeholder, target_language)
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -36,9 +36,7 @@ from cms.admin.dialog.views import get_copy_dialog
 from cms.admin.forms import (
     PageForm, AdvancedSettingsForm, PagePermissionForm, PublicationDatesForm
 )
-from cms.admin.permissionadmin import (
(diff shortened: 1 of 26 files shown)


cms/admin/pageadmin.py
```diff
@@ -540,36 +520,40 @@ def has_add_permission(self, request):
         """
         Return true if the current user has permission to add a new page.
         """
-        if get_cms_setting('PERMISSION'):
+        can_add = super(PageAdmin, self).has_add_permission(request)
+
+        if can_add and get_cms_setting('PERMISSION'):
```
mkoistinen: Does it make sense to put something like: `CMS_PERMISSION = get_cms_setting('PERMISSION')` at the top of this file and then use the "constant": `CMS_PERMISSION` everywhere instead? Its unlikely to change between requests =)

czpython: Yup. I'll keep it in mind for the next PR, to avoid re-running tests, etc.. on this one.

cms/models/placeholdermodel.py
```diff
@@ -111,52 +111,35 @@ def _get_url(self, key, pk=None):
             model_name = model.__name__.lower()
             return admin_reverse('%s_%s_%s' % (app_label, model_name, key), args=args)
 
-    def _get_permission(self, request, key):
+    def _has_permission(self, user, key):
```
mkoistinen: This looks like a new method, yet if declares params that will never be used. Intentional?

cms/models/placeholdermodel.py
```diff
@@ -111,52 +111,35 @@ def _get_url(self, key, pk=None):
...
+
+    def _get_object_permission(self, obj, user, key):
         opts = obj._meta
         perm_code = '%s.%s' % (opts.app_label, get_permission_codename(key, opts))
-        return request.user.has_perm(perm_code) or request.user.has_perm(perm_code, obj)
+        return user.has_perm(perm_code) or user.has_perm(perm_code, obj)
 
     def has_change_permission(self, request):
-        return self._get_permission(request, 'change')
+        return self._has_permission(request.user, 'change')
```
mkoistinen: I'm confused how these 3 methods work, since the key param doesn't appear to do anything.

cms/plugin_base.py
```diff
@@ -268,98 +260,6 @@ def has_change_permission(self, request, obj=None):
         return False
     has_delete_permission = has_change_permission
 
-    def get_form(self, request, obj=None, **kwargs):
```
mkoistinen: Lotsa deletions here... is this all dead code?

czpython: It's not, got moved to placeholder admin.

cms/models/placeholdermodel.py
czpython: The methods are not new, I renamed _get_permission to _has_permission but decided to leave the parameters for now until the part two of the cleanup (coming soon).

The changes are in the logic inside, before we were using those parameters but soon I'll remove them.

mkoistinen:
Great work... and long overdue!

I see now why you had to write so many tests. In the end, to we go up or down in tests count and test coverage?

mkoistinen:
OK to merge. Looking forward to part 2.

czpython:
Thanks, test count has gone up over 100 new tests. We'll have to wait on coverage to run and give us the verdict.

czpython:
@yakky Please review :)

yakky:
Sorry, terrible week. Ok to merge

czpython:
The next PR will have the changelog entries.
Another PR will be created in cooperation with @evildmp to update any outdated documentation.

PR #5495, opened 2016-07-04
Fixed #5493 -- Allow extension_pool.register to be used as a decorator

### LINKED ISSUE #5493: Register Page and Title extensions into the pool with decorator.
I think that will be more convenient instead of using

    class MyPageExtension(PageExtension):
        my_field = models.CharField()

    extension_pool.register(MyPageExtension)

Registering to happen with decorator

    @extension_pool.register
    class MyPageExtension(PageExtension):
        my_field = models.CharField()

The easiest way to achieve this is by adding returning the object in `register` (https://github.com/divio/django-cms/blob/release/3.3.x/cms/extensions/extension_pool.py#L12) after everything is done.

vstoykov:
Fixes #5493

### DIFF
--- a/cms/extensions/extension_pool.py
+++ b/cms/extensions/extension_pool.py
@@ -12,6 +12,20 @@ class ExtensionPool(object):
     def register(self, extension):
         """
         Registers the given extension.
+
+        Example::
+
+            class MyExtension(PageExtension):
+                pass
+
+            extension_pool.register(MyExtension)
+
+        or as decorator::
+
+            @extension_pool.register
+            class MyExtension(PageExtension):
+                pass
+
         """
 
         if issubclass(extension, PageExtension):
@@ -23,6 +37,7 @@ class ExtensionPool(object):
                 'Extension has to subclass either %r or %r. %r does not!' % (PageExtension, TitleExtension, extension)
             )
         self._activate_signaling()
+        return extension
 
     def unregister(self, extension):
         """
--- a/cms/test_utils/project/extensionapp/models.py
+++ b/cms/test_utils/project/extensionapp/models.py
@@ -19,7 +19,7 @@ class MyPageExtension(PageExtension):
 extension_pool.register(MyPageExtension)
 
 
+# extension_pool.register can be used also as a decorator
+@extension_pool.register
 class MyTitleExtension(TitleExtension):
     extra_title = models.CharField(blank=True, default='', max_length=255)
-
-extension_pool.register(MyTitleExtension)


cms/tests/test_extensions.py
```diff
@@ -31,8 +31,9 @@ def test_register_extension(self):
         page_extension = self.get_page_extension_class()
 
         # register first time
-        extension_pool.register(page_extension)
+        register_extension_result = extension_pool.register(page_extension)
```
yakky: Maybe a better test is just to use the decorator to register one of the test extensions and check if it's registered in the test.
This looks a bit "hackish"

czpython: I agree. @vstoykov you can use the `test_utils` for that.

vstoykov: The reasons I created the test like that are:

1. By definition decorators are functions which accept function/class as argument and returns function/class. It can return new function/class which inherit the functionality of the original or simply modify the original and return it. The syntax with `@` is syntax sugar from Python 2.6 and up. It's not required to use always a decorator with this syntax.
2. If decorate one of the Extension models in `test_utils/project/extensionapp/models.py` with @-syntax and the decorator functionality is broken (`register` does not return class at all) then when tests are executed, following error will be raised `TypeError: 'NoneType' object is not iterable` even before the tests can be started. This is not useful for developer who accidentally broke that functionality.

If you still prefer one (or both) of the extension models to be decorated with @-syntax in the test app and to revert the change in `test_extensions.py`,  I will do it.

yakky: The more we test the code in the way it's used in practice, the better.
I guess 99% of the decorator usage is with ``@``-syntax, and in the test you can just check that the extension has been properly registered.

vstoykov: As I said the test can't run because of the error. Before test runner even start finding tests, when django is trying to register the model into administration the error is raised.
This means that there is no need to change the tests (compared to code before my PR), because they already test if the models are registered as extensions. And if someone broke the functionality to use `register` as decorator instead of one test failing the whole test suite will not run.

Are you ok with this situation?

yakky: That's what I meant. Just change the way the extension is registered in the test application, and remove the register part in the test, and we're done, no need for more explicit testing of the decorator implementation details

vstoykov:
Ok I've made the changes that you wanted and updated my PR.

czpython:
Thanks @vstoykov.
Ok to merge.

yakky:
Thanks @vstoykov.
Ok to merge.

yakky:
@czpython @mkoistinen do we want this on 3.3 or it's just for 3.4?

czpython:
@yakky I've thought about it and I'm leaning towards 3.4 to be consistent with the policy.

PR #5497, opened 2016-07-04
Fixed #5494 Make it possible to use multi-table inheritance for extensions

### LINKED ISSUE #5494: Make it possible to use multi-table (concrete) inheritance for Page and Title extensions
I just realized that when I have multi-table inheritance for TitleExtension (I guess is the same for PageExtension) and try to publish the page django-cms fails to copy the extension instance and instead save the same instance.

The reason behind this is because `BaseExtension.copy_to_public` method (https://github.com/divio/django-cms/blob/release/3.3.x/cms/extensions/models.py#L64-L87) does not nullify pk fields of parent model(s).

I'm using Django 1.8 and DjangoCMS 3.3.

I created a workaround in my project which involves overwriting whole  `copy_to_public` method without calling `super` which is totally bad thing (If I want to upgrade the CMS in the future).

I can prepare e PR with proposed solution.

vstoykov:
Fixes: #5494

I have a little problem with creating the test for this (creating a migration for test project for `extensionapp`).

You can mark this PR as "work in progress" until I'm ready with the tests. I just wanted to push it right now if you have some objections or recommendations.

### DIFF
--- a/cms/extensions/models.py
+++ b/cms/extensions/models.py
@@ -44,6 +44,12 @@ class BaseExtension(models.Model):
         clone.pk = None
         clone.public_extension = None
         clone.extended_object = target  # set the new public object
+
+        # Nullify all concrete parent primary keys. See issue #5494
+        for parent, field in clone._meta.parents.items():
+            if field:
+                setattr(clone, parent._meta.pk.attname, None)
+
         clone.save(mark_page=False)
 
         # If the target we're copying already has a publisher counterpart, then
@@ -75,8 +81,19 @@ class BaseExtension(models.Model):
         if public_extension:
             this.pk = public_extension.pk  # overwrite current public extension
             this.public_extension = None  # remove public extension or it will point to itself and raise duplicate entry
+
+            # Set public_extension concrete parents PKs. See issue #5494
+            for parent, field in this._meta.parents.items():
+                if field:
+                    setattr(this, parent._meta.pk.attname, getattr(public_extension, parent._meta.pk.attname))
         else:
             this.pk = None  # create new public extension
+
+            # Nullify all concrete parent primary keys. See issue #5494
+            for parent, field in this._meta.parents.items():
+                if field:
+                    setattr(this, parent._meta.pk.attname, None)
+
             this.save(mark_page=False)
             self.public_extension = this
             self.save(mark_page=False)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -13,6 +13,7 @@
 * Fixed a regression when standalone CMS Widgets wouldn't work due to
   non-existing JavaScript dependencies.
 * Fixed a possible recursion error when using the ``Alias`` plugin.
+* Make it possible to use multi-table inheritance for Page/Title extensions.
 
 
 === 3.3.2 (unreleased) ===
--- a/cms/test_utils/project/extensionapp/models.py
+++ b/cms/test_utils/project/extensionapp/models.py
@@ -23,3 +23,24 @@ extension_pool.register(MyPageExtension)
 @extension_pool.register
 class MyTitleExtension(TitleExtension):
     extra_title = models.CharField(blank=True, default='', max_length=255)
+
+
+class MultiTablePageExtensionParent(models.Model):
+    extension_parent_field = models.CharField(blank=True, default='', max_length=255)
+
+
+class MultiTablePageExtension(MultiTablePageExtensionParent, PageExtension):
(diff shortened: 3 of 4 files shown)


vstoykov:
Hmm now when I'm preparing the tests I see that with normal multi-table inheritance it will be impossible to be fixed. Actually my case is with django-polymorphic where related manager returns instance of the child class (the real type) and not the parent (the base).

Are you ok If I include django-polymorphic in tests or this is out of scope for Django CMS?

yakky:
I'm not really in favor of adding yet another test dependency. It's a real pain to maintain them and I don't know if worth the hassle in this case.
@divio/django-cms-core any comments?

vstoykov:
Actually now I think of another approach which will test the problem without using django-polymorphic.

I will give you an example.

First how the models look like if django-polymorphic is used:

    class BaseExtension(PolymorphicModel, PageExtension):
        base_extension_field = models.CharField(max_length=255)

    class RealExtension(BaseExtension):
        real_extension_field = model.CharField(max_length=256)

By this way if you never create instances directly from `BaseExtension` then every time you access the related descriptor in the `Page` object you will have instance of the child class `RealExtension`.

In this case when the page is published extension must be copied the code in `copy_to_public`. The basic logic for doing that is by nullify the PK but with multi table inheritance when you nullify `instance.pk` the `instance.id` is not nullified. In this case when the instance is saved django first tries to save the parent instance and because it's PK is not nullified the instance is updated instead of created and then assigned to the OneToOne relation of the child instance.

If we want to test the same situation but without django-polymorphic we need to have reverse relation between page an child class of multi table inheritance. For that reason I can try to define the models as follows:

    class BaseExtension(models.Model):
        base_extension_field = models.CharField(max_length=255)

    class RealExtension(BaseExtension, PageExtension):
        real_extension_field = model.CharField(max_length=256)

By this way I will still have an extension which can't be copied by current code but will with my proposed changes and related description in the Page object will return directly child instance.

 I will test this tomorrow to see if it will work.

vstoykov:
@yakky I updated my PR with tests which do not include django-polymorhpic but still test the issue that I'm trying to fix.

There will be a possible conflict with #5495 because of `cms/test_utils/project/extensionapp/models.py`. If you decide that this PR can be backported as a bugfix to 3.3 then probably it can be merged first and then I need to rebase to other. If you first merge the other I need to rebase this one.

czpython:
@vstoykov I've merged the other PR into develop.

@yakky I think this fits the criteria for 3.4 as multi-table inheritance on extensions wasn't officially supported.

vstoykov:
@czpython I rebased my branch against develop.

Did you guys think that this PR is ready for review?

czpython:
@vstoykov Hey, I think I misunderstood. Is this PR ready? If so we'll as ready for review.

vstoykov:
@czpython Yes, I think that it is ready.

yakky:
@vstoykov thanks. This LGTM

cms/extensions/models.py
```diff
@@ -75,8 +81,19 @@ def copy_to_public(self, public_object, language):
         if public_extension:
             this.pk = public_extension.pk  # overwrite current public extension
             this.public_extension = None  # remove public extension or it will point to itself and raise duplicate entry
+
+            # Set public_extension concrete parents PKs. See issue #5494
+            for parent, field in this._meta.parents.items():
+                if field:
+                    setattr(this, parent._meta.pk.attname, getattr(public_extension, parent._meta.pk.attname))
```
czpython: nit picking:

```python
if field:
    parent_pk_field = parent._meta.pk.attname
    setattr(this, parent_pk_field, getattr(public_extension, parent_pk_field))
```

vstoykov: Because  I saw the long line above, but ok I will fix it.

vstoykov: Actually @czpython your proposal is also more than 80 chars (95), and I'm not sure what is the benefit.

Instead i can change it like that:

    if field:
        setattr(this,
                parent._meta.pk.attname,
                getattr(public_extension, parent._meta.pk.attname))

But I'm not sure if you like this way of calling functions.

What you think?

vstoykov:
@yakky I rebased my branch against develop and added changelog entry.

I was not made a change yet for the comment from @czpython because I wanted first he to say that it will be ok.

cms/extensions/models.py
czpython: @vstoykov Let's leave it as is for now to be consistent then.

czpython:
Ok to merge

czpython:
Thanks!

PR #5458, opened 2016-06-26
Apphook config multisite

yakky:
Admittedly this was a choice made upon introduction of apphook config, but I can't see a good reason to not allow to reuse the same apphook config on different sites. ``application_namespace`` is already marked as unique only for each site (https://github.com/divio/django-cms/blob/release/3.3.x/cms/models/pagemodel.py#L124) thus I don't think apphook config should be more restrictive. I'd backport this at least to 3.2.x branch too

PR #5511, opened 2016-07-06
Removed globally unique constraint for Apphook configs

yakky:
Backport of #5458


yakky:
@czpython yeah, just waited for tests to finish

PR #5521, opened 2016-07-08
Fixed tests to match updates on ckeditor

czpython:
Some tests have also been moved out of the cms because they tests djangocms-text-ckeditor specific logic.

These tests will fail until the new ckeditor is released.


yakky:
:+1:

PR #5523, opened 2016-07-08
Fixed #5522 -- Don't override global context in render_plugin_toolbar_config tag

### LINKED ISSUE #5522: Corrupted context in render_plugin_toolbar_config tag
Reported by @Air-Mark in https://github.com/divio/django-cms/pull/5384
> There is a problem in render_plugin_toolbar_config tag because we call update but we never call pop of context.
> I use render_plugin_toolbar_config to enable editing of child plugins without rendering them, but after toolbar refactoring, that tag got a very strange behaviour. So I found that when we call render_plugin we fill up some variables in context like "allowed_child_classes", and then if we use render_plugin_toolbar_config in the template of current plugin to render it's children we fill up exactly the save variables. So when the context comes to the toolbar_plugin_processor it's corrupted with values of it's own children plugins.

czpython:
Fixed #5522


mkoistinen:
LGTM!

yakky:
LGTM

PR #5525, opened 2016-07-10
Fixed gulp watch task breaking on lint errors.

vxsx:
The default gulp task starts webpack, eslint and sass watchers. If
ever there was a linting error, linting and sass watchers would
stop because of the unhandled ESLintPlugin error. Gulp-plumber is a
tool to swallow unhandled errors and display them in the console
without breaking the gulp streams.


FinalAngel:
LGTM

PR #5528, opened 2016-07-11
Fixed plugin form false positive "edited" state

vxsx:
When plugin is being edited we track the changes in order to have
the mechanism for saving user's unsaved content. When the
"trackChange" is triggered it didn't take into account actual
values and always considered the field as changed thus prompting
the user even if the form wasn't really changed.


FinalAngel:
LGTM

PR #5530, opened 2016-07-11
Improve warnings

vstoykov:
By default stacklevel argument is `1` which will print the line where the warning was raised instead of the line which is calling deprecated method/property.

If stacklevel is set to 2 then more meaningfull information is shown to the developer.

This probably need to be backported because some of the warnings will be removed in the next release.


czpython:
@vstoykov 
Thanks. Please re-open against the `release/3.3.x` branch, we'll take care of forward port to develop.

vstoykov:
@czpython  #5533

Issue #5432, opened 2016-06-18
Invalid permissions required to paste from the clipboard 


czpython:
Pasting plugins from the clipboard requires users to have `change_placeholderreference` permissions.

czpython:
Fixed by https://github.com/divio/django-cms/pull/5532

Issue #5488, opened 2016-07-01
Invalid permissions required to copy into clipboard


czpython:
A non superuser can't copy plugins from a model which uses a `PlaceholderField` into the clipboard unless he has **change** permissions on the model and change permissions on the `UserSettings` model.

czpython:
Fixed by https://github.com/divio/django-cms/pull/5532

Issue #5451, opened 2016-06-22
Can't "Add" a subpage even though I gave users permission to.


bartTC:
*I am not sure if this is misconfiguration on my side or missing functionality.*

I gave my users 'Add/Edit/Delete' permission to a sub-page of the page tree, in this case the "New York" page. 

![screen shot on 2016-06-22 at 18_05_55](https://cloud.githubusercontent.com/assets/1896/16274550/61bf1b16-38a6-11e6-921e-4bd7cf87d788.png)

My understanding is that the user can add pages below the "New York" page but I'm missing the Plus button there.

![](http://d.pr/i/1dpsJ+)

Workaround I found:

If I'm on the New page itself and in edit mode, I can add a sub-page via the menu "Create page > New sub page". This works fine, as expected.

bartTC:
Note that the user can also click the menu item "Create page > New page" which would create a page of at the same page level, which is not allowed. This results in a popup with a response '403 Forbidden' which is correct behavior. However I'd propose to disable that "New page" button since the user has no permission to do so in first place.

czpython:
@bartTC 
Thanks for reporting.
We're currently working in a cleanup of our permissions system for 3.4
I'll keep this open in the meantime.

czpython:
Fixed by https://github.com/divio/django-cms/pull/5532

Issue #5450, opened 2016-06-22
Django CMS 3.3: Advanced Permissions popup is just empty if the user has no permission to edit them


bartTC:
I gave my users access to a subset of pages to edit, but I didn't gave an explicit permission to edit 'advanced settings'.

The menu item for this is disabled (ok)

![](http://d.pr/i/9Qas+)

However the user is still able to access that screen by the bottom of the regular settings

![](http://d.pr/i/7XYk+)

Or by Shift+Click on the settings icon in the page tree.

In both cases the user will just see a blank, gray popup with no information. The backend is triggering a 401 response correctly.

czpython:
@bartTC 
Thanks for reporting.
We're currently working in a cleanup of our permissions system for 3.4
I'll keep this open in the meantime.

czpython:
Fixed by https://github.com/divio/django-cms/pull/5532
The shift+click part will be fixed by https://github.com/divio/django-cms/pull/5617

PR #5532, opened 2016-07-11
Refactored permission internals

czpython:
Part two of three 😈 

This part of the refactor introduces new logic for consistency and efficiency with regards to permissions handling.

One of the main changes is that placeholder permissions are now calculated based on the object containing the placeholder.

It helps to think of placeholders as an attachment which has no individual logic or permissions on it's own, but when attached to a Page, Article, etc. then it needs to respect the permissions of those objects. Placeholders are assumed to always change the contents of the object it's attached to, this means that all plugin operations on a placeholder will effectively only check whether the user can **change** the object attached to the placeholder.

To provide these objects with more control over permission checks, placeholders will call a new `has_placeholder_change_permission` method on the object attached to it if it's available.

So in the case of a Page object which has it's own permissions logic, we define `Page.has_placeholder_change_permission()` and have it check if the user can change the page. If no `has_placeholder_change_permission` method is found, then we fallback to `django.auth` permissions and check if the user has can_change permissions on the object.

Last, if the placeholder has no attached object then by design the user can't perform any operations on this placeholder (add, change, delete).

## Permissions overview

CMS Permissions Off

Page

- Add - Requires user to have both `cms.change_page` and `cms.add_page` permission
- Change - Requires user to have both `cms.change_page` permission
- Delete - Requires user to have both `cms.change_page` and `cms.delete_page` permission

Placeholder

- Add plugin - Requires user to have add permissions on the plugin's model and change permissions on the page.
- Change plugin - Requires user to have change permissions on the plugin's model and change permissions on the page.
- Delete plugin - Requires user to have delete permissions on the plugin's model and change permissions on the page.
- Clear - Requires user to have delete permissions on all plugins inside the placeholder and change permissions on the page.

CMS Permissions On

Page

- Add - Requires global or page specific add and change permissions.
- Change - Requires global or page specific change permissions.
- Change advanced settings - Requires global or page specific change and change advanced settings permissions.
- Change permissions - Requires global or page specific change and change page permissions permissions.
- Publish - Requires global or page specific change and publish permissions.
- Move - Requires global or page specific change and move permissions.
- Delete - Requires global or page specific delete and change permissions.

Fixes:

https://github.com/divio/django-cms/issues/5432
https://github.com/divio/django-cms/issues/5450
https://github.com/divio/django-cms/issues/5451
https://github.com/divio/django-cms/issues/5488
https://github.com/divio/django-cms/issues/4928

All the changes below are for internal permission apis:

Change in signature:
	```cms.utils.permissions.get_subordinate_users(user)``` -> ```get_subordinate_users(user, site)```
	```cms.utils.permissions.get_subordinate_groups(user)``` -> ```get_subordinate_groups(user, site)```
	```cms.utils.permissions.get_user_permission_level(user)``` -> ```get_user_permission_level(user, site)```
```cms.models.managers.PagePermissionManager.subordinate_to_user(user)``` -> ```PagePermissionManager.subordinate_to_user(user, site)```

Functions removed:

`cms.utils.permissions.cache_tree_children`

Classes moved:
```cms.admin.permissionsadmin.GenericCmsPermissionAdmin``` -> ```cms.admin.useradmin.GenericCmsPermissionAdmin```

Classes removed
	`cms.models.managers.PagePermissionsPermissionManager`

Removed templates
	```admin/cms/page/tree/lazy_child_menu.html```
	```admin/cms/page/tree/menu_fragment.html```


stefanfoulis:
LGTM

PR #5539, opened 2016-07-13
Fix selection of page admin languages

nolsto:
Addresses issue #5470

PR #5543, opened 2016-07-14
Fixed #5470 -- Page admin's views' incorrect selection of available languages

### LINKED ISSUE #5470: Languages in tabs of change form from PageAdmin's add_view come from request rather than Page site ID
In `update_language_tab_context`, called from `add_view`, the `obj` is always `None`, and does not exist for languages to be determined by a `site_id`. Instead the tabs are populated from the languages of the site the request is from:

``` python
language = get_language_from_request(request, obj)
languages = self._get_site_languages(obj)
```

This is causing form validation errors without displayed messages when trying to create a page for a site that does not share any languages with the site originating the requests.

nolsto:
Now get languages from site-specific cms admin site rather than the
current site if the obj doesn't exist in the view.

Getting the language from the request now gets available languages from
cms_admin_site in request.session if current_page is not present. If the
language is not in that set, use the first available from the set.


nolsto:
This is the properly-based and formatted PR, originally #5539.

yakky:
LGTM

evildmp:
@nolsto Don't forget to add a changelog entry: http://docs.django-cms.org/en/release-3.3.x/contributing/development-policies.html#changelog

FinalAngel:
@nolsto closing this for now. If still relevant please reopen against develop branch :)

Thank you very much.

Issue #5476, opened 2016-06-30
Jquery error in admin when I use PageSmartLinkField in my models


V1ce:
Hi,

Before upgrade django-cms to 3.3 from 3.2.5 i was able to re-use the PageSmartiLinkField in my own models, but since the upgrade i now get jquery errors when i go to the model edition in the admin. 

This looks like a conflict with my other js files. i got this : 

- select2.js : L22, $ is undefined
- forms.pagesmartlinkwidget.js : L19, $ is not a function
- In the js part of admin "add" page : L339, CMS.$ is not a function

I got this in the popup edition with toolbar and in the admin site directly. In the admin site i don't load js plugins so this is really a conflict?

Other thing, when i go to the "advanced settings" of a page, where the PageSmartlink widget is used, i don't got these errors.

I load jquery in the "head" of my base.html with some other js files, and i load some other js file in the footer too.

Im not very familiar with these history of js conflict, i would appreciate if someone can help for understand this.

Thanks

vxsx:
Hi, I haven't seen this kind of error before. It would be helpful if you could provide the minimal example to replicate the issue.

Also, could you try reproducing the issue on latest `release/3.3.x` branch?

V1ce:
Django 1.9.7
python 3.4.2
django-cms 3.3.0

Here my current use of PageSmartLinkField in my "news" app for the project "bgl".

**bgl/forms.py** : global PageSmartLinkField form extended by my apps

```python
from django import forms
from django.utils.translation import ugettext_lazy as _
from cms.forms.fields import PageSmartLinkField
from cms.utils.i18n import get_language_tuple
from django.forms.widgets import HiddenInput
from django.utils.translation import get_language

class LinkForm(forms.ModelForm):

    lien  = PageSmartLinkField(label=_('Redirect'), required=False,
                                  help_text=_('Entrer une url valide ou sélectionner une page'),
                                  placeholder_text=_('Entrez adresse ou recherche'),
                                  ajax_view='admin:cms_page_get_published_pagelist',
                                  
    )
    language = forms.ChoiceField(label=_("Language"),required=False, choices=get_language_tuple(),
                                 help_text=_('The current language of the content fields.'))
    
    class Meta:
        exclude = ('language', )
        
    def __init__(self, *args, **kwargs):
       super(LinkForm, self).__init__(*args, **kwargs)
       self.fields['language'].widget = HiddenInput()
    
       if not self.fields['language'].initial:
[...]

PR #5546, opened 2016-07-15
Fixed a regression when standalone widgets wouldn't work

vxsx:
The regression happened because in
https://github.com/divio/django-cms/commit/a44cf25d6d5fd63cd2e9bd2ad1b5e14ae15c7d7b
and before that in 3.3.0 the files required for widgets were
removed from media js class on the assumption that the CMS provided
bundles would be loaded anyway. This commit changes the way
requirements are loaded for those widgets so they can be used
standalone and on CMS pages as well.

https://github.com/divio/django-cms/issues/5476


lory87:
lgtm

cms/forms/widgets.py
```diff
@@ -24,8 +24,9 @@ class PageSelectWidget(MultiWidget):
 
     class Media:
         js = (
```
czpython: please change to:

```
js = (
    static_with_version('cms/js/widgets/forms.pageselectwidget.js'),
    static_with_version('cms/js/dist/bundle.forms.pageselectwidget.min.js'),
)
```

cms/forms/widgets.py
```diff
@@ -123,9 +126,9 @@ class Media:
             )
         }
         js = (
-            'cms/js/select2/select2.js',
-            'cms/js/widgets/forms.pagesmartlinkwidget.js',
+            'cms/js/dist/bundle.forms.pagesmartlinkwidget.min.js',
         )
+        js = tuple(map(static_with_version, js))
```
czpython: same as above

cms/forms/widgets.py
```diff
@@ -191,8 +196,9 @@ class AppHookSelect(Select):
 
     class Media:
         js = (
-            'cms/js/widgets/forms.apphookselect.js',
+            'cms/js/dist/bundle.forms.apphookselect.min.js',
         )
+        js = tuple(map(static_with_version, js))
```
czpython: same as above

cms/forms/widgets.py
```diff
@@ -244,8 +250,9 @@ class ApplicationConfigSelect(Select):
...
```
czpython: same as above

cms/forms/widgets.py
vxsx: sure, don't really need both of them though

czpython:
Ok to merge

Issue #5518, opened 2016-07-08
Swipe up or down starting from a disabled menu voice doesn't scroll the menu


itbabu:
A swipe down should scroll the menu up. This doesn't happen if the swipe starts from a disabled menu.
This is very noticeable if there are a lot of disabled menu voices.

![swipe](https://cloud.githubusercontent.com/assets/650691/16691158/be3154f4-452b-11e6-994f-c8b080ab36d3.png)

I can reproduce this on:
- my android phone with Chrome and Firefox
- my laptop with chrome-devtools and the device toolbar

czpython:
Fixed on 3.3.2

PR #5549, opened 2016-07-18
Fixed toolbar menu scrolling on mobile.

vxsx:
It wasn't possible to scroll the menu if it was initiated on
disabled item.

#5518


lory87:
👍

PR #5558, opened 2016-07-20
Plugin rendering refactor

czpython:
The goal of this refactor was to simplify the logic behind plugin/placeholder rendering with speed as the motivator.

A quick test on my machine showed that a page with 300 plugins on CMS 3.3.1 takes about 2 seconds to load, with this branch it takes about 0.4 to 0.5 seconds.

All performance gains are due to the new structuring of the Python/Django code, I was unable to optimize the SQL queries involved in rendering as they were already heavily optimized.

Here's an overview of the changes:

New utilities
	`get_toolbar_from_request`

Removed internal functions
	`cms.middleware.toolbar.toolbar_plugin_processor`
	`cms.plugin_rendering.render_plugin`
	`cms.plugin_rendering.render_plugins`
	`cms.plugin_rendering.render_placeholder`
	`cms.plugin_rendering.render_placeholder_toolbar`
	`cms.templatetags.cms_tags._get_placeholder`
	`cms.templatetags.cms_tags.get_placeholder_content`
	`cms.templatetags.cms_tags._show_placeholder_for_page`

Removed internal attributes
	`request.placeholders`
	`request.toolbar.placeholder_list`
        `plugin_base.CMSPluginBase.frontend_edit_template`

Internal modifications
	`Placeholder.render()` now requires a `cms_content_renderer` instance in the context.
	`CMSPlugin.render()` now requires a `cms_content_renderer` instance in the context.
	`RenderPlugin` now requires a `cms_content_renderer` instance in the context.
	`CMSPluginBase.get_child_classes()` is now a class method for consistency with the other methods.
	`CMSPluginBase.get_parent_classes()`  is now a class method for consistency with the other methods.
	Moved in-memory template caching from the `CMSToolbar` class to the `ContentRenderer` class.

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -8,7 +8,7 @@ from django.shortcuts import render_to_response
 from django import forms
 from django.contrib import admin
 from django.contrib import messages
-from django.core.exceptions import ImproperlyConfigured
+from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.template.defaultfilters import force_escape
 from django.utils import six
 from django.utils.encoding import force_text, python_2_unicode_compatible, smart_str
@@ -96,7 +96,6 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
 
     form = None
     change_form_template = "admin/cms/page/plugin/change_form.html"
-    frontend_edit_template = 'cms/toolbar/plugin.html'
     # Should the plugin be rendered in the admin?
     admin_preview = False
 
@@ -150,11 +149,17 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
             warnings.warn('CMSPlugin.render_template attribute is deprecated '
                           'and it will be removed in version 3.2; please move'
                           'template in plugin classes', DeprecationWarning)
-            return getattr(instance, 'render_template', False)
+            template = getattr(instance, 'render_template', False)
         elif hasattr(self, 'get_render_template'):
-            return self.get_render_template(context, instance, placeholder)
+            template = self.get_render_template(context, instance, placeholder)
         elif getattr(self, 'render_template', False):
-            return getattr(self, 'render_template', False)
+            template = getattr(self, 'render_template', False)
+        else:
+            template = None
+
+        if not template:
+            raise ValidationError("plugin has no render_template: %s" % self.__class__)
+        return template
 
     @classmethod
     def get_render_queryset(cls):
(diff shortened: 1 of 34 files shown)


yakky:
@czpython assingning myself too to review this

cms/plugin_base.py
```diff
@@ -386,22 +391,23 @@ def get_child_plugin_candidates(cls, slot, page):
         from cms.plugin_pool import plugin_pool
         return plugin_pool.get_all_plugins()
 
-    def get_child_classes(self, slot, page):
+    @classmethod
+    def get_child_classes(cls, slot, page):
```
yakky: Why making this a class method?

cms/plugin_base.py
```diff
@@ -422,14 +428,15 @@ def get_child_classes(self, slot, page):
                 continue
         return child_classes
 
-    def get_parent_classes(self, slot, page):
+    @classmethod
+    def get_parent_classes(cls, slot, page):
```
yakky: Why making this a class method?

cms/plugin_base.py
czpython: As with the one below, two reasons:

1. Consistency with the other `get_x` methods related to plugin classes/types.
2. It's much more efficient to call `Class.get_child_classes()` vs creating instance of `Class` every time just to call `get_child_classes`.

yakky: We're closing the door to the possibily for the users to customize the parent/child classes based on the plugin instance. Not sure of the tradeoff

yakky:
Overall looks very good
Could you check this deprecation warning here https://travis-ci.org/divio/django-cms/jobs/147205933#L787?
In https://github.com/divio/django-cms/pull/5558/files#diff-7a16c3a25f2b77fe9faea8717080696cR298 and similar we should flatten the context (using flatten_context from classy tags)

cms/plugin_base.py
czpython: The door was already closed.
A `CMSPluginBase` instance was never guaranteed, and usually it would be created on the fly without any uniqueness to it meaning that there really wouldn't be a difference between instances of the same plugin class except for their memory id.

I'm aiming for decoupling more of these methods from the instance because it's very confusing and inconsistent that for some we need class methods and for other instance methods.

czpython:
@yakky why not use `context.flatten()`?

yakky:
``plugin.render()`` is not guaranteed to return a context, users may override and return a dictionary. ``flatten_context`` has some strategies to do the right thing in any sensible case

cms/plugin_base.py
yakky: Ok

evildmp:
@czpython Impressive! Please don't forget to add a changelog entry. I am assuming that this doesn't have any documentation implications.

czpython:
@yakky I've left the warnings for now to land this. Changing it to flatten_context causes errors when using the `Template()` object directly 😪

PR #5560, opened 2016-07-21
Plugin save callback

vxsx:
Modal has it's won mechanism for reloading when the plugin is
saved. However when the plugin is saved - close_frame template is
rendered where reloadBrowser is called again. So whenever the
plugin is saved - server takes 2x time to process the response. We
cannot remove the template since it can be used by other plugins
for their own means (most notably djangocms-text-ckeditor).

Part of performance improvements for #5558


FinalAngel:
LGTM

Issue #5555, opened 2016-07-19
Make note of need for migrations in plugin tutorial


pdbethke:
Just a suggestion from a user working to develop a new CMS plugin that uses a model -- it may seem intuitive to some django programmers but you might want to add a line in the http://docs.django-cms.org/en/release-3.3.x/how_to/custom_plugins.html#storing-configuration tutorial noting that you should run **manage.py makemigrations** and **manage.py migrate** following your model creation to properly set up the configuration fields -  it tripped me up a little before I realized what was the issue when I went to remove or save the newly created plugin.

evildmp:
Thanks, fixed in https://github.com/divio/django-cms/pull/5566.

PR #5566, opened 2016-07-24
Improved How To: Custom Plugins

an0o0nym:
I found that django migration commands were missing after modifying models.py in Storing Configuration section of How To: Custom Plugins. This may save some time for forgetful people like me who are just starting with django-cms.


evildmp:
Thanks for your contribution!

evildmp:
Fixes #5555.

(Let's see if it closes that automatically.)

PR #5568, opened 2016-07-25
Fixed #5402 -- recursion error when referencing same Alias plugin

### LINKED ISSUE #5402: Copy & Paste Alias Plugin from/into the same placeholder causes infinite loop
... and this seems to make the edit page inaccessible which makes it impossible to fix the thing.

The only solution I found is removing the plugin entry directly on the database level.

On Aldryn it's even more severe because there the db needs to be copied to stage and then to local, then  manually fixing, and then the full round-trip back to production.

I'm not saying copying / pasting an alias into the same placeholder makes much sense but sooner or later somebody does it... dooh.

czpython:
Fixed #5402

This prevents the recursion error by failing silently (not rendering the alias), thus allowing the user to remove the recursive plugin.

A better fix will be to detect this when pasting the plugin but this will come in the future when we add validation hooks for certain actions (pasting, moving, etc.)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,6 +4,7 @@
   started on the disabled menu item on small screens.
 * Fixed a regression when standalone CMS Widgets wouldn't work due to
   non-existing JavaScript dependencies.
+* Fixed a possible recursion error when using the ``Alias`` plugin.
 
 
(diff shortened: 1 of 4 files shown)


yakky:
LGTM

PR #5571, opened 2016-07-27
Fixed race condition in plugin save callback.

vxsx:
PR #5560 introduced plugin save callback, but it reset the
`isReloading` flag right before the actual reload. Murphy's law
kicked in and on the pages with extreme amount of plugins
 the `onPluginSave` callback was fired right after the actual
reload already started. This commit fixes it, by not resetting the
flag at all.

Also, it is not recommended to use Class.js mixins feature anymore,
as it copies the methods and internal state, which can lead to
problems. Won't be a problem longterm since we will want to 
remove class.js from CMS


FinalAngel:
LGTM

Issue #5590, opened 2016-08-05
Plugins using filter_horizontal are broken in 3.3.1


akx:
We have a plugin that has a `filter_horizontal` field, and it's broken since 3.3.1 due to the 1Password workaround introduced in commit 5bcb54a5fca91d5cd20deaad125aa2b34fa68ab9 (PRs #5312 #5367). See below for the gory details.

`filter_horizontal` select boxes get their "currently selected" values select-alled just before form submission by a `submit` event handler on the form by Django's SelectFilter2 javascript.

The [Django CMS modal code in cms.modal.js](https://github.com/divio/django-cms/blob/3d67879973aac61be97ed240262421f35f5d8c4a/cms/static/cms/js/modules/cms.modal.js#L688) copies buttons from the form into the modal's footer, then hooks them up to submit the form.

It can do the submission in two different ways; one is by simulating a click on the original button, the other is by directly submitting the form. The heuristic that leads to direct submission is that there's exactly one button in the form (which is the usual case).

However, when directly submitting a form using `form.submit()`, there's this:

> The form's `onsubmit` event handler (for example, `onsubmit="return false;"`) will not be triggered when invoking this method from Gecko-based applications. In general, it is not guaranteed to be invoked by HTML user agents. 
> — https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit

akx:
I can also confirm that installing the 1Password extension on Firefox fixes this issue (since we then use the old code path that simulates clicking the button), so it's definitely the aforementioned commit.

akx:
A workaround for this is to override the `admin/cms/page/plugin/change_form.html` template to add something like

```django
    <div id="bug-workaround" style="display: none">
        {% comment %}
        Work around Django-CMS issue #5590 (https://github.com/divio/django-cms/issues/5590)
        by adding a dummy button that triggers the old-style submission of forms.
        {% endcomment %}
        <button></button>
    </div>
```

to the `{% block submit_buttons_bottom %}` block.

vxsx:
Hi, thanks for the report @akx . I don't really see how installing the extension would change the code path, it depends on the number of buttons, not on extension being present. Or am I missing something?

It seems though that this happens because we submit the form using jQuery and not using the native event, and the submit listener added by SelectField2 isn't firing because of it.

reduced testase: http://codepen.io/vxsx/pen/vKVpdK (same behaviour in FF and Chrome)

[...]

PR #5312, opened 2016-05-18
add a workaround for 1password on firefox

vxsx:
This PR fixes https://github.com/divio/django-cms/issues/4376 for me locally
I don't fully understand why is this happening and will first contact 1password support to hopefully clarify things.

PR #5594, opened 2016-08-08
Fixed submit handlers not being executed on modal form

vxsx:
This is a regression from the #5312, submitting form with jQuery
doesn't run event handlers that are attached with `addEventListener`,
e.g Django's `filter_horizontal` field.

Ref: #5590


FinalAngel:
LGTM

FinalAngel:
LGTM and Crazy

PR #5604, opened 2016-08-11
Fix example in doc about ModelForm

Flimm:
This fixes the bug that would cause the error:

    django.core.exceptions.ImproperlyConfigured: Please set entry 'model' attribute or use ModelForm subclass as a form


Flimm:
Could this sort of information be included in http://docs.django-cms.org/en/latest/contributing/documentation.html ?

For such a tiny change, this is just needless work.

PR #5605, opened 2016-08-11
Refactored Pagetree dropdowns

vxsx:
This fixes a bug when there is a different number of triggers and
menus on the page (due to missing permissions etc) and click on a
trigger would open wrong menu because indexes wouldn't match.

Don't think it deserves changelog


FinalAngel:
LGTM

PR #5606, opened 2016-08-11
Fix example in doc about ModelForm

Flimm:
This fixes the bug that would cause the error:

    django.core.exceptions.ImproperlyConfigured: Please set entry 'model' attribute or use ModelForm subclass as a form

This is similar to this pull request but on the right branch: https://github.com/divio/django-cms/pull/5604


czpython:
@Flimm Thanks!

Issue #5610, opened 2016-08-14
cms.models.Page.get_absolute_url should really prefix URLs with Site domains.


heppstux:
Let's say we have a main Site ("example.com") and some other site ("sub.example.com"). Both running the same instance of django-cms with settings.SITE_ID set accordingly.

While the cms.models.fields.PageField offers the options to choose Pages from a different Site as target – the generated URL doesn't reflect changing from one Site to another – it'll always output relative URLs which – in the case of linking from a Page on the Site "example.com" to any other Page on "sub.example.com" - just doesn't make any sense. I'd expect the link to show up with the correct hostname prefix. This issue concerns all Plugins that generate links. Fun-Fact: cms.admin.pageadmin seems to generate URLs on it's own respecting the Site. So it seems like there was already some attempt to work around this issues but not in a way, that would solve the problem globally.

Good news is – I think I'm on a good way to supply a fix.

I basically changed the get_absolute_url method of the models.Page – it takes the current SITE_ID and the site.pk of the Page into account when generating the URL. I had to remove some code from the cms.admin.pageadmin that prefixed domains to some links instead of relying on the generic get_absolute_url of the Page model.

However:
I had (and still have) a couple test failing after my edits.

Most of them were just query-count asserts, that failed since my changes force asking the database for the page's site domain every time they generate an URL – sad but what to do. I adjusted those tests accordingly. 

Some other tests failed due to their used of a (strange) RegEx that is supposed to check for "valid URLs". I adjusted this RegEx too – however, I think these test shouldn't really rely on a RegEx – I have a strong feeling, that this can't actually work.

Unfortunately some test are still failing – and I have just some remote idea why: I could really need some help to figure out what's wrong here and how to fix these issues properly. It seems they are related to my changes in cms.admin.pageadmin.

mammuth:
The PageField gives the user the option to link to another site in a multisite environment, but simply doesn't return the correct url via `field.get_absolute_url`.

So I'd like to suggest to treat this issue as a bug and not an enhancement.

czpython:
It is labeled as enhancement because the `PageField` is a field that links and returns a page, anything beyond that (getting any data from that page) is outside of the scope of the field and the responsibility of the app that implements the field.
As discussed in https://github.com/divio/django-cms/pull/5611, we'll just add a method to the `Page` model that provides a standard for third-party apps to get the url from a page.

mammuth:
I see, thanks for pointing me towards the discussion in the PR.

[...]

PR #5611, opened 2016-08-14
[WIP] Page.get_absolute_url should actually return a site domain aware version of the page's url

heppstux:
see issue #5610

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -3,6 +3,7 @@ from logging import getLogger
 from os.path import join
 
 from django.contrib.sites.models import Site
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.urlresolvers import reverse
 from django.db import models
@@ -174,10 +175,15 @@ class Page(six.with_metaclass(PageMetaClass, MP_Node)):
         if not language:
             language = get_language()
         with i18n.force_language(language):
-            if self.is_home:
-                return reverse('pages-root')
             path = self.get_path(language, fallback) or self.get_slug(language, fallback)
-            return reverse('pages-details-by-slug', kwargs={"slug": path})
+            if self.is_home:
+                relative_url = reverse('pages-root')
+            else:
+                relative_url = reverse('pages-details-by-slug', kwargs={"slug": path})
+            if self.site:
+                if self.site.domain and settings.SITE_ID != self.site.pk:
+                    return '//%s%s' % (self.site.domain, relative_url)
+        return relative_url
 
     def get_public_url(self, language=None, fallback=True):
         """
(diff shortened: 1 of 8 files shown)


czpython:
Hello @heppstux ,
Thanks for the patch.
I'd rather we have a different method on the `Page` model that always return a full url w/ domain and all, instead of overloading the `get_absolute_url()` method.

This is for several reasons:

- Backwards compatibility
- Consistency with Django
- Consistency in behavior (always return full url vs only when it doesn't match current site)

We can call the new method `get_full_url()`, takes same parameters as `get_absolute_url()` but returns the url with domain.

Then we can adapt only the parts in the code that need this behavior like the link plugin, etc..

heppstux:
okay, sounds reasonable – however a lot of (third-party) plugins will not benefit automatically from this adjustments. Also I feel, it's the actual job of the get_absolute_url to actually do what it's name suggests and return the real url of a Page.

How can we make sure, all parts of the CMS benefit from this new function other than deprecate the get_absolute_url method...

yakky:
@heppstux Right or wrong ``get_absolute_url`` only return the **path** of a model, not the full URL.
I don't think django CMS should break this in anyway and I'd go for an additional method as suggested by Paulo, without deprecating ``get_absolute_url``

heppstux:
Okay – I can live with that.

Even though I see some more problems and even more confusion coming: Should get_draft_url and get_public_url also get alternative "versions" with explicit support for Sites?

Let's say "get_full_draft_url" and "get_full_public_url".

yakky:
get_draft_url and get_public_url are only meant to be used within the frontend editor, I don't think they  need a full url version

czpython:
I agree, the draft / public methods should only be used internally by the cms and as such, is not needed to create a "full" version for them.

cms/models/pagemodel.py
```diff
@@ -174,10 +175,15 @@ def get_absolute_url(self, language=None, fallback=True):
...
         with i18n.force_language(language):
-            if self.is_home:
-                return reverse('pages-root')
             path = self.get_path(language, fallback) or self.get_slug(language, fallback)
-            return reverse('pages-details-by-slug', kwargs={"slug": path})
+            if self.is_home:
+                relative_url = reverse('pages-root')
+            else:
+                relative_url = reverse('pages-details-by-slug', kwargs={"slug": path})
+            if self.site:
```
vstoykov: This check can be optimized to reduce SQL queries when page's site is the same as configured in settings.

    if self.site_id and settings.SITE_ID != self.site_id and self.site.domain:
        return '//%s%s' % (self.site.domain, relative_url)

Actually I'm not sure if the check for site.domain is needed at all. `domain` field is not only required but also unique and if there is a `site_id` then there will be a `site.domain`.

heppstux: yes, thanks :)

czpython:
Hello @heppstux, any updates?

heppstux:
hi @czpython, sorry, didn't have the chance to do any further investigation/coding on the topic yet – still it remains hot. I'd suggest to reject this PR and I'll try to open another one later.

czpython:
Thanks for the follow-up.
Closing as suggested.

Issue #5310, opened 2016-05-17
Menus show last selected site instead of current site


czpython:
Create two sites:

1 - example-1.com
2 - example-2.com

Run your server with site 1 active.
Create multiple pages on site 1.
Go to the home page and verify that the pages show up in menu.

Run your server with site 2 active.
Create multiple pages on site 2.
Go to the home page and verify that the pages show up in menu.

Now while site 2 is active, switch to site one on the site selector in the admin.
Go back to the homepage and you should see the incorrect menu for site 1.

Video:

https://drive.google.com/file/d/0B9SwgyV0g1hST1ZEbl9qT2M4Mlk/view

Notice that enabling cache brings forth another problem where cache is not refreshing when switching sites.

This bug comes from a deep look at https://github.com/divio/django-cms/pull/5109

razisayyed:
Sorry For Disturb,
But I really need to know if this bug would be fixed in the near future.

PR #5612, opened 2016-08-14
Fixes #5609 -- ToolbarMiddleWare should clear menu-caches respecting current Site and language

### LINKED ISSUE #5609: cms.middleware.toolbar doesn't clear menu cache properly
Using a plugin (with cache=False) in a static placeholder on a multi-Sites instance of DjangoCMS I figured out, that the toolbar-middleware doesn't clear the menu cache respecting the current SITE_ID.

Thus in edit-mode with activated toolbar my menu-plugin would render arbitrary navigation structure from other Sites.

heppstux:
see issue #5609

### DIFF
--- a/cms/middleware/toolbar.py
+++ b/cms/middleware/toolbar.py
@@ -5,6 +5,7 @@ Edit Toolbar middleware
 from django.contrib.admin.models import LogEntry, ADDITION, CHANGE
 from django.core.urlresolvers import resolve
 from django.http import HttpResponse
+from django.utils.translation import get_language
 
 from cms.toolbar.toolbar import CMSToolbar
 from cms.utils.conf import get_cms_setting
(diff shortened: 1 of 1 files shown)


czpython:
Hello @heppstux ,
Thank you very much for the patch.
Please add tests to the `test_menu.py` file
Also don't forget to add a changelog entry under 3.3.3 (unreleased) http://docs.django-cms.org/en/release-3.3.x/contributing/development-policies.html#changelog.
Should be something like "Fixed bug where menu cache was not cleared for the current site".

czpython:
Having a closer look at the patch, it's a bit unclear how it fixes the issue.
menu_pool.clear() will clear the cache for all sites/languages which I think was intended behavior given that you can render menu nodes from different sites.

I think the real issue is #5310.

czpython:
Closing as this issue is covered in https://github.com/divio/django-cms/issues/5310

PR #5615, opened 2016-08-16
Removed revision support from core

czpython:
As announced, we're removing revision support from the core and will re-introduce it as a separate app.

The _History_ menu entry is gone along with the _Undo_ / _Redo_ / _Revert to live_ options.

@divio/django-cms-core


czpython:
Ping @yakky

yakky:
@czpython LGTM

fuzzy76:
Where was this announced? I can't find the rationale anywhere, and I am quite frankly a bit puzzled by anyone shipping a CMS without revision support since it is normally regarded a pretty basic functionality.

bharling:
Blimey, yeah this one caught me out too, wasn't aware it was to be deprecated. Do we have an alternative available yet?

PR #5616, opened 2016-08-16
Fixed a bug when some plugins wouldn't be editable

vxsx:
Since real preview now stores the data about corresponding plugins
on actual plugin markup (in forms of classes and jQuery data) -
plugins that have their markup somehow modified or cloned wouldn't
appear triggering tooltip or being editable by double clicking on
them. That was happening because libraries usually keep classes but
they do not keep the data and event handlers associated with the
plugin, so in case e.g. a slide was cloned - it wouldn't be
editable, you would have to find the original slide and double
click on that. This commit changes the way handlers for plugins are
attached so the only thing that is required from a library
modifying the plugin markup is the `cms-plugin-x` class.

Doesn't need the changelog since real preview was introduced in 3.4


FinalAngel:
nice 👍

PR #5618, opened 2016-08-17
Added fuzzy matching to the plugin picker

vxsx:
Previously you could only find the plugin by a full substring match, now it's fuzzy, which simplifies finding correct plugin.

## before
screen shot 2016-08-17 at 10 02 12

## after
screen shot 2016-08-17 at 10 01 52


lory87:
👍

FinalAngel:
👍

PR #5619, opened 2016-08-17
Improved `shift+space` shortcut UX

vxsx:
Content is now scrolled to the top of the plugin, putting the
plugin at 20% of window height vs putting window center in the
middle of the plugin. In case the plugin is very big you would end
up in the middle of nowhere.

This also increases overlay fade out timeout so it's more visible.


lory87:
👍

FinalAngel:
+1

PR #5621, opened 2016-08-17
Fix example in doc about ModelForm (#5606)

czpython:
Forward port of #5606


czpython:
Merged in 4fd0d66ecdc44fa081b1d19e8550320a00b59b76

PR #5628, opened 2016-08-19
Fixed #5627 Do not set session cookie on every request when `CMS_TOOLBAR_ANONYMOUS_ON` is `False`

### LINKED ISSUE #5627: Set session cookie on every request when `CMS_TOOLBAR_ANONYMOUS_ON` is `False`
I found very strange problem. When I set `CMS_TOOLBAR_ANONYMOUS_ON = False` I see that for anonymous users django creates a session and not only session is created but session cookie is sent from the server on every request.

This behavior is breaking caching trough http proxy because every anonymous visitor has different session created on first load and on every request it will sent different cookies than other visitors and can't share the cache.

I found that the problem is in `ToolbarMiddleware` (https://github.com/divio/django-cms/blob/3.3.1/cms/middleware/toolbar.py#L102-L103). If check is added to see if there is a need of modification of session variables then the session will not be modified and django will not send it to the client.

I will prepare PR with a fix.

vstoykov:
I'm not sure how to test this. If someone can give me a hint I can add tests too.

If this can be backported to stable 3.3.x will be awesome

### DIFF
--- a/cms/middleware/toolbar.py
+++ b/cms/middleware/toolbar.py
@@ -77,8 +77,10 @@ class ToolbarMiddleware(object):
             if build in request.GET and not request.session.get('cms_build', False):
                 request.session['cms_build'] = True
         else:
-            request.session['cms_build'] = False
-            request.session['cms_edit'] = False
+            if request.session.get('cms_build'):
+                request.session['cms_build'] = False
+            if request.session.get('cms_edit'):
+                request.session['cms_edit'] = False
         if request.user.is_staff:
             try:
                 request.cms_latest_entry = LogEntry.objects.filter(
--- a/cms/tests/test_toolbar.py
+++ b/cms/tests/test_toolbar.py
@@ -364,8 +364,8 @@ class ToolbarTests(ToolbarTestBase):
         page = create_page("toolbar-page", "nav_playground.html", "en",
                            published=True)
         request = self.get_page_request(page, self.get_nonstaff(), edit=True)
-        self.assertFalse(request.session.get('cms_build', True))
-        self.assertFalse(request.session.get('cms_edit', True))
+        self.assertNotIn('cms_build', request.session)
+        self.assertNotIn('cms_edit', request.session)
 
     def test_hide_toolbar_disabled(self):
         page = create_page("toolbar-page", "nav_playground.html", "en",


vstoykov:
Ok I found where it fails and there I will put the tests.

vstoykov:
I just looked at these tests and I wonder why by default if some value is missing from the session it will be set to `True`.

Probably the whole method of checking session values need to be revised. Before I made any other changes to tests are you ok with adding new argument to `ToolbarTestBase.get_page_request` that will accept initial session data. If we want to test that `True` became `False` we will set it as `True` in the beginning. If we want to test if it was not set and and middleware are not touching it then we can check if the value is `None`.

What you think?

cms/middleware/toolbar.py
```diff
@@ -77,8 +77,10 @@ def process_request(self, request):
             if build in request.GET and not request.session.get('cms_build', False):
                 request.session['cms_build'] = True
         else:
```
czpython: I suggest using elif in both cases below

czpython:
Thanks for the patch @vstoykov.
Your suggestion for the tests sounds good to me 👍 

Regarding landing this on 3.3.x, my only concern is that now those flags are only set when needed which is fine for the cms but might have some issues on apps that rely on this, I doubt it but is just to keep in mind. 3.4 will be released sooner than 3.3.3.

@yakky thoughts?

cms/middleware/toolbar.py
vstoykov: It's tempting to use `elif` when you see  `else` and then `if` but in this case setting `cms_edit` to `False` is not related to that if `cms_build` is `True` or not.

Both conditions need to be checked if the condition in the upper `if` fails. If I change it like:

    if not request.session.get('cms_toolbar_disabled', False) and (
                request.user.is_staff or (anonymous_on and request.user.is_anonymous())
        ):
        ...
    elif equest.session.get('cms_build'):
        request.session['cms_build'] = False
    elif if request.session.get('cms_edit'):
        request.session['cms_edit'] = False

This will have totally different behavior for `cms_edit`.

Probably I misunderstood your suggestion. What you have in mind?

czpython: Ah you're right, my fault. It's good as it is.

vstoykov:
What are the plans for 3.4, because from tomorrow I will be in vacation without computer and will be unable to finish the changes in the tests.

czpython:
We're planning to release rc1 no later than Wednesday next week.
After that we'll be on code freeze until the stable release hits early next month.

Maybe you can do a test just for this new change and land a separate PR for the proposed changes?

vstoykov:
I updated the test to represent that the value must not be present in the session. When we make it possible to set initial value of the session then we can test setting from `True` to `False`

vstoykov:
@czpython 
The test that is failing is only for Django1.8 with Python3.4 and is related to documentation. It's trange beacuse I'm not changing anything related to documentation.

czpython:
@vstoykov yeah looks like there was a new sphinx release that's breaking things.
I've pinned a working version on 0136232ded0b97605d88df3da0f42a7fa13fcda5.
Waiting on tests to finish to confirm.

czpython:
@vstoykov Please rebase (added a fix for sphinx) and update the changelog under 3.4 with "Changed toolbar middleware to not set a session cookie on every request when `` `CMS_TOOLBAR_ANONYMOUS_ON` `` is set to `` `False` ``"

yakky:
@vstoykov @czpython LGTM

Issue #5631, opened 2016-08-21
Unable to select page with PageField


czpython:
I'm using Aldryn Forms which has a plugin with `PageField`.
If I create a form and select a page as a redirect target, the next time I edit the form I get a JS error:

`Uncaught TypeError: Cannot read property '_pageSelectWidgets' of undefined`

Only after I publish the page, the error **sometimes** goes away.

http://xplorer-1002-stage.us.aldryn.io/en/forms/?edit&language=en

![cms_bug_with_pagefield](https://cloud.githubusercontent.com/assets/994951/17838047/6069e704-6791-11e6-8073-b7f27f88e30f.png)

Reproduced on both 3.3 and 3.4

macolo:
Seems to be related to this here: https://github.com/divio/django-cms/issues/5476

vxsx:
3.2 issue is a little different (and indeed is #5476), 3.3 and 3.4 though break in this particular case because of djangocms-text-ckeditor

czpython:
Fixed by https://github.com/divio/django-cms/pull/5637

PR #5637, opened 2016-08-24
Fixed a bug with multiple admin bundles

vxsx:
In case there were two bundles on the page, all the subsequent ones were
overriding CMS namespace, which could contain widget information, which
would in turn result in errors.

https://github.com/divio/django-cms/issues/5631


FinalAngel:
LGTM

PR #5649, opened 2016-09-01
fixed CMS_TEMPLATES_DIR without __init__.py

fvgoto:
Previous `if config_path:` always holded.
Specifying a `CMS_TEMPLATES_DIR` without having a `__init__.py` file (which would contain `TEMPLATES` dict) in that directory, it resulted in following error
```
  File "/home/ME/venv/lib/python3.4/site-packages/cms/utils/conf.py", line 154, in get_templates
    template_module = load_from_file(config_path)
  File "/home/ME/venv/lib/python3.4/site-packages/cms/utils/django_load.py", line 116, in load_from_file
    with open(module_path, 'r') as openfile:
FileNotFoundError: [Errno 2] No such file or directory: '/home/ME/PRJ/APP/templates/pages/__init__.py'
```

Following doc states `__init__.py` is optional:
- [comment preceding changed lines](https://github.com/divio/django-cms/blob/develop/cms/utils/conf.py#L147)
- [reference/configuration -> CMS_TEMPLATES_DIR](http://djangocms.readthedocs.io/en/latest/reference/configuration/?highlight=%20can%20__init__.py#std:setting-CMS_TEMPLATES_DIR)


czpython:
Thanks for the patch @fvgoto.
The tests are failing because `FileNotFoundError` is only available on python 3 and up.
Please adapt the code to work on both python 2 and 3

czpython:
Thanks!

PR #5664, opened 2016-09-07
Fixed #5651 -- Corectly get language from request

### LINKED ISSUE #5651: Putting language= in querystring breaks CMS pages
Adding a `?edit&language=` on any CMS page leads to a `NoReverseMatch` error. [See here for an example](https://www.django-cms.org/en/blog/?edit&language=).

My best guess why this happens without looking deeper: the toolbar admin language is set to the empty string which breaks language-based reversing of URLs. Solution: better sanitation of the toolbar language input.

vstoykov:
In `get_language_from_request` if for some reason local variable `language` becomes empty string instead of `None` then treat is as `None` (Duck typing).

Fixes #5651

I opened this PR against release/3.3.x because #5651 is marked as bug for 3.3.x. If you prefer I can open it against develop and you can backport it then (also for 3.4), instead of forward porting.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -13,6 +13,7 @@
 * Fixed a bug where moving a published page under a page marked as pending publishing
   is left as published instead of being marked as pending publishing.
 * Fixed AttributeError when using ``create_page`` in management command
+* Fixed a bug in getting the language from current request which can cause error 500
 
 
 === 3.3.2 (2016-08-11) ===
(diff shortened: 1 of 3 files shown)


vstoykov:
I see that tests are failing but this is related to Filer and easy_thumbnails for some reason and I don't think that is related to changes that I've made.

czpython:
Thanks for the patch @vstoykov.
release/3.3.x is perfect, please rebase as I've pushed a fix for the failing tests.

Also please add a test for #5651, as well as a changelog entry under 3.3.3

czpython:
Thanks @vstoykov 👍

PR #5665, opened 2016-09-07
Fixed a bug where pages would appear to be movable without permissions

vxsx:
`Boolean("false") === true`

(facepalm)


FinalAngel:
:D 👍

PR #5672, opened 2016-09-09
Allways escape text passed from django to JS

vstoykov:
This will fix bug when there is a new line (or other non valid characters) in the resulted text which will lead to JavaScript error.

This is the same as CMS.config.lang values in the same file.


FinalAngel:
@vxsx valid or?

vxsx:
yes, 👍

czpython:
Thanks! @vstoykov

vstoykov:
This will be forward ported or I need to open new PR in order this to be also in 3.4 release?

czpython:
@vstoykov 
No worries, I'll land this on 3.4 later today.
Thanks for the followup.

PR #5674, opened 2016-09-10
Allways escape text passed from django to JS (#5672)

czpython:
Forward port https://github.com/divio/django-cms/pull/5672


FinalAngel:
👍

czpython:
Merged in https://github.com/divio/django-cms/commit/a468385b5bc1ad13d3a1c16d75257c0d486efb71

PR #5675, opened 2016-09-12
[WIP] untranslated placeholders

benzkji:
This is a proposal, how a "untranslated" setting could be added to the placeholder config. untranslated placeholders would just ignore the language field on plugins, and always show all plugins. Typical use cases: Header images, and other "non language specific content". I'm aware that almost the same can be achieved with language_fallbacks, but IMO it's worth it, as it makes the UI more clean for those use cases.

Differences to language fallbacks. Fallbacks need to be defined, untranslated just is untranslated. Also, fallback plugins are not shown in edit mode (so one could add a plugin, and no more fallback would be needed), wereas untranslated plugins are visible in any language, also in edit mode.

I'm sure there would be some more changes and tests needed (cms/utils/plugins.py also has get_plugins_for_page and reorder_plugins) - this is a proof of concept, to check if this feature has any chance to get into django-cms...

### DIFF
--- a/cms/tests/test_placeholder.py
+++ b/cms/tests/test_placeholder.py
@@ -624,6 +624,11 @@ class PlaceholderTestCase(CMSTestCase, UnittestCompatMixin):
         placeholder_de = title_de.page.placeholders.get(slot='col_left')
         add_plugin(placeholder_en, TextPlugin, 'en', body='en body')
 
+        context_en = SekizaiContext()
+        context_en['request'] = self.get_request(language="en", page=page_en)
+        context_de = SekizaiContext()
+        context_de['request'] = self.get_request(language="de", page=page_en)
+
         conf = {
             'col_left': {
                 'untranslated': True,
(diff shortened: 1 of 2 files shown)


benzkji:
I'm aware that there might be a small performance loss, because it takes two queries instead of one, for getting all plugins for a page. if there are translated and untranslated placeholders on a page. but, if the untranslated would be configured as fallback, it might even take more queries, for each of the fallback placeholders...so I guess it's not as bad as it looks on first sight.

czpython:
Hello @benzkji,
Thanks for this proposal.

I've thought about this PR for a while and for me its a no-go.
My reasoning is that so far (in my experience), I have not encountered a case where this is truly needed. Because of this, I don't think its worth the complexity/performance hit.

For future reference, please submit any feature proposals to our [Developers](https://groups.google.com/forum/#!forum/django-cms-developers) mailing list.

/cc @yakky

czpython:
Closing to continue the conversation in the developers group.
https://groups.google.com/forum/#!topic/django-cms-developers/bTeaEU2RU44

PR #5679, opened 2016-09-13
Fixed double-escaping when rendering placeholder with dashes

czpython:
Using `escapejs` and `json.dumps()` will double escape certain characters.


stefanfoulis:
👍

PR #5685, opened 2016-09-16
Fixed a regression when static placeholder was uneditable

vxsx:
Plugin code assumed that if there are multiple containers then it is a
plugin wrapped with `<template>` tags. That is not the case with
multiple static placeholders on the same page.


FinalAngel:
LGTM

PR #5686, opened 2016-09-16
Make check unique namespace works in multisites

razisayyed:
Filter based on site_id to allow namespace to be used in multisites.
Already fixed in previous versions of cms 
#5458


czpython:
Thanks for the patch @razisayyed
We'll apply the other PR and backport to 3.4

PR #5689, opened 2016-09-17
Fixed example code for custom `render_placeholder`.

jsma:
Fixed indentation so copy/paste works for end users.


czpython:
Thanks @jsma

PR #5691, opened 2016-09-17
Fix example code in 3.4 release notes

rfleschenberg:
There is a small variable name mismatch (`renderer` vs `content_renderer`)


czpython:
Thanks @rfleschenberg.
This has already been fixed by https://github.com/divio/django-cms/pull/5689.

Issue #5698, opened 2016-09-19
Keyboard shortcuts interfere with form input


jsma:
Sorry for short details at the moment but if the toolbar is enabled, I can't fill out forms on the site. The keyboard shortcut hijacks the clicks and switches to structure mode, etc. Any way to disable shortcuts globally?

vxsx:
There's no way to disable them, but they shouldn't be enabled when a form field is focused, thanks for the report.

PR #5699, opened 2016-09-20
Fixed a bug when keyboard shorcuts were triggered when form fields were focused

vxsx:
Keyboardjs doesn't support this out of the box, so we introduce a patch
to prevent callbacks running when input is focused.

Fixes https://github.com/divio/django-cms/issues/5698


cms/static/cms/js/modules/keyboard.js
```diff
@@ -0,0 +1,32 @@
...
+    return newFn;
+}
+
+/**
+ * Override keyboardjs methods to disallow running callbacks
+ * if input is focused
+ */
+keyboard._applyBindings = override(keyboard._applyBindings, function (originalBind) {
+    return function (event) {
+        if ($(':focus').is('input, textarea, select, [contenteditable]')) {
```
FinalAngel: do we also need to add option or is it automatically handled by select?

vxsx: you can't really focus option

Issue #4512, opened 2015-09-26
Hide plugins (in structure mode) - feature [re]quest(ion)


creimers:
Other CMS provide the user with the possibility to 'hide' plugins, i.e. to not show them to the ordinary site visitor without having to delete them.

Is a similiar functionality in the making for djangoCMS by chance?

This might have been asked here before, in which case I apologize for my insufficient research.

yakky:
What's the use case for this feature? (BTW you can easily create a ``HiddenPluginsPlugin`` which hides the contained plugins according to some specific rules -like: hide plugins for non staff users)

creimers:
Use case: Some banner or image on a page that you periodically want to manually hide and show. Or some content that you need later but now happen to have the time to insert it in the CMS. You could then leave it hidden for the time being.

Can't think of anything more illustrating right now, but I have been asked this by multiple clients who were switching *to* djangoCMS and appreciated such functionality in their old CMS.

Of course I could create a ``HiddenPluginsPlugin``, but a 'hide/show' option along with 'copy', 'edit', 'delete', ... in the structure view would appear much more convenient to me.

yakky:
Looks legit. Thanks for the suggestion @creimers

creimers:
@yakky Yes I'll be able to contribute this with some help. Is there documentation on how the structure mode works/is built?

jrief:
You may have a look at the segmentation plugin.
https://github.com/aldryn/aldryn-segmentation

djangocms-casade also offers segmentation.

yakky:
@creimers no need for structure mode: https://github.com/divio/django-cms/blob/develop/cms/plugin_base.py#L333-L340 use these to add items to the placeholder / plugin menu, a view in ``PlaceholderAdminMixin``, then I guess we could handle this with a field that saves whether to hide the plugin or not.
In edit mode the plugin must be rendered (eventually adding a specific class to make visible they are "hidden"), so you'll need to adjust the ``get_plugins`` function https://github.com/divio/django-cms/blob/develop/cms/utils/plugins.py#L26

evildmp:
This has been discussed before, in fact I discussed it recently with someone but I can't remember where or with whom.

In short, it's a feature I've wished for since 2.3, and it would have enormous practical and time-saving use.

Each plugin that is ``publish``-enabled (this might not be appropriate for every kind of plugin) should have a checkbox, ``Publish``, that is on by default.

[...]

PR #5723, opened 2016-10-10
[WIP] Hide plugins

nmashton:
This is implementation of the feature described in issue #4512, trying to follow the implementation suggestion given by [this comment](https://github.com/divio/django-cms/issues/4512#issuecomment-143528095) (after a false start with making the field editable). Feedback is requested!

To hide a plugin, just choose the "Hide" option in its dropdown menu. The plugin and all aliases dependent on the plugin will be hidden. (The menu item will also become "Show".)

One question for the core developers: *should* all the aliases depending on a hidden plugin be hidden? My client has a use case for which this makes sense (hiding seasonal content); but is that going to the most common situation?

One minor incomplete feature of this branch: a class `cmsplugin-hidden` is added to the `cms-dragitem` element if the plugin's `cmsplugin_hidden` value is `True`, but this added class currently does nothing. What would be an intuitive way to indicate the "hidden" status of the plugin?


czpython:
Hello @nmashton,
Thank you for taking on this task.

I have some reservations regarding the feature itself, we have not discussed it thoroughly with the rest of the core developers or the board. As such, we'll discuss it on our next board meeting (Monday next week).

Just to be clear my reservations are regarding the following:

- Will this be used enough to balance with the extra code added to the core?
- Is it possible to make it into a separate app?
- Can we avoid adding a new field to the `CMSPlugin` class?
- Design-wise we need to come up with something to show that this plugin is hidden.

yakky:
I don't see why you can' implement this with a wrapper plugin where you can implement complex policies to show / hide content instead of a simple boolean

nmashton:
@yakky: thanks for the suggestion; I will take a look at aldryn-segmentation (etc) for inspiration and attempt an implementation along those lines.

@czpython: great, thank you. I look forward to hearing the outcome of that discussion.

czpython:
Hello @nmashton,

As promised, here's the outcome of the board meeting in regards to this feature:

The consensus is that this is good feature to have at the plugin level (vs nesting plugins).
That said, we think that adding this to the core is not ideal as it would cause several issues with existing plugins due to migrations, as well as the fact that we're trying to simplify the core as much as possible and strip it from features that can be done in third party apps (ie. revision history).

We'd like to help with the design of this app (if you're ok with it).

Here's our suggestions:

* Create a new model with a relationship to `CMSPlugin` class.
* Add a `disabled = BooleanField(default=False)` field to this class.
* Create a plugin on this app whose only purpose is to extend the global plugin menu actions, this plugin is not renderable or even shown to the user, it's only used internally. I'm not too fond of this approach (abusing plugin to extend menu only) but I think it's the easiest solution for now, anything else would require a big change on the core to allow non plugins to extend the plugin actions menu. The [alias plugin](https://github.com/divio/django-cms/blob/release/3.4.x/cms/cms_plugins.py#L62) is a good example of how to achieve this.
* An advantage of using an internal plugin is that you can use the `get_plugin_urls()` method to register the ajax endpoint that will be called when a user activates/deactivates a plugin.
* Naming-wise, we suggest "disable" and "enable" as the action names, "hide" and "show" are used by the structure board already and this can confuse some users.

I think that should cover most of the logic/design of the app, the main things missing are:

* Somehow get the cms to know about the disabled field and not render the plugin.
* Performance impact per plugin

I can help extending the cms to tackle those two, just need to figure out what is best generic approach.

If you're still interested, please create a topic in the [django CMS developers group](https://groups.google.com/forum/#!forum/django-cms-developers), we can continue our collaboration there.

PR #5727, opened 2016-10-12
Changed logo on welcome page to svg background-image

vxsx:
There is a bug in Chrome 52+ that renders the glyph we were using for
logo incorrectly on HiDPI screens, given that this bug is fairly rare,
we are switching to background image.


FinalAngel:
LGTM

PR #5731, opened 2016-10-19
Fixes #5093 -- Use passed database alias in migrations

### LINKED ISSUE #5093: Migration 0005 doesn't use selected database
When trying to create a new database via syncdb (or migrate) we get the following error:

  File "/home/doctormo/www.inkscape.org/pythonenv/lib/python2.7/site-packages/MySQLdb/cursors.py", line 205, in execute
    self.errorhandler(self, exc, value)
  File "/home/doctormo/www.inkscape.org/pythonenv/lib/python2.7/site-packages/MySQLdd/connections.py", line 36, in defaulterrorhandler
    raise errorclass, errorvalue
django.db.utils.OperationalError: (1054, "Unknown column 'cms_page.level' in 'field list'"

The command was `manage.py migrate --database=local` and the local database is a postgresql database. But the error above looks like it's trying to contact the mysql database (which is 'default')

czpython:
Fixes #5093

### DIFF
--- a/cms/migrations/0005_auto_20140924_1039.py
+++ b/cms/migrations/0005_auto_20140924_1039.py
@@ -12,6 +12,7 @@ ALPHABET = '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ'
 
 
 class MP_AddHandler(object):
+
     def __init__(self):
         self.stmts = []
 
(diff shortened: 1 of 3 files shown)


czpython:
Ping @yakky

PR #5591, opened 2016-08-05
Improve documentation for `copy_relations` method

vstoykov:
Before coping related objects from old instance current one need to be deleted or at public version of the page can appear duplicates.

PR #5750, opened 2016-10-31
Improve documentation for `copy_relations` method

vstoykov:
Before coping related objects from old instance current one need to be deleted or at public version of the page can appear duplicates.

This is the same PR as #5591 but against `release/3.4.x`

### DIFF
--- a/docs/how_to/custom_plugins.rst
+++ b/docs/how_to/custom_plugins.rst
@@ -361,6 +361,11 @@ new plugin::
         title = models.CharField(max_length=50)
 
         def copy_relations(self, oldinstance):
+            # Before copying related objects from the old instance, the ones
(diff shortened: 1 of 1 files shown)


docs/how_to/custom_plugins.rst
```diff
@@ -361,6 +361,11 @@ new plugin::
         title = models.CharField(max_length=50)
 
         def copy_relations(self, oldinstance):
+            # Before coping related objects from old instance current one need
+            # to be deleted or at public version of the page can appear
+            # duplicate items.
+            self.associated_item.all().delete()
```
evildmp: "Before copying related objects from the old instance, the ones on the current on need to be deleted. Otherwise, duplicates may appear on the public version of the page."

vstoykov:
I fixed the wordings and updated my PR

PR #5753, opened 2016-11-03
bf/is_popup_pages

Air-Mark:
After refactoring is_popup variable is missing in context of render_admin_rows.
So raw_id_field with Page model does not work.

### DIFF
--- a/cms/utils/admin.py
+++ b/cms/utils/admin.py
@@ -125,7 +125,7 @@ def render_admin_rows(request, pages, site, filtered=False, language=None):
             'site_languages': languages,
             'open_nodes': open_nodes,
             'cms_current_site': site,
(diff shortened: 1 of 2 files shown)


cms/utils/admin.py
```diff
@@ -124,6 +125,7 @@ def render_page_row(page):
             'site_languages': languages,
             'open_nodes': open_nodes,
             'cms_current_site': site,
+            'is_popup': IS_POPUP_VAR in request.GET
```
yakky: I'll check ``POST`` too:
```
(IS_POPUP_VAR in request.POST or IS_POPUP_VAR in request.GET),
```

czpython:
@yakky I don't think this needs to be backported, the refactors mentioned where done on 3.4.x only.

PR #5739, opened 2016-10-25
Added cms_path to plugin / placeholder operations

vxsx:
this should be fine @czpython

PR #5758, opened 2016-11-15
Operation hooks

czpython:
**TODO**:
Rebase with 3.4.x once https://github.com/divio/django-cms/pull/5739 is merged


czpython:
Ping @yakky

cms/admin/pageadmin.py
```diff
@@ -841,28 +897,69 @@ def copy_page(self, request, page_id, extra_context=None):
...
         if statics:
-            static_ids = statics .split(',')
-            for pk in static_ids:
-                static_placeholder = StaticPlaceholder.objects.get(pk=pk)
+            static_ids = statics.split(',')
+            static_placeholders = StaticPlaceholder.objects.filter(pk__in=static_ids)
+
+            for static_placeholder in static_placeholders.iterator():
+                # TODO: Maybe only send one signal...
+                # this would break the obj signal format though
```
yakky: It looks ugly, but emitting one signal per static placeholder it's the only sane option

cms/models/pluginmodel.py
```diff
@@ -250,6 +250,23 @@ def get_plugin_instance(self, admin=None):
         self._inst = instance
         return self._inst, plugin
 
+    def get_bound_plugin(self, admin=None):
```
yakky: What's the difference with ``get_plugin_instance`` above?

cms/admin/pageadmin.py
czpython: Agree, we anyways have to iterate through each static placeholder to publish it.

cms/models/pluginmodel.py
czpython: `get_plugin_instance()` unfortunately returns plugin instance and plugin admin instance which I dislike as it forces us to always do `get_plugin_instance()[0]`

Instead, as I've started to introduce the terminology of bound and unbound plugins, I decided to create this more explicit method which will only return the plugin instance.

That said, you found a small bug.. I need to remove the `admin` parameter :man_facepalming:

yakky: Why not makeing ``get_bound_plugin`` a wrapper to ``get_plugin_instance``:

    def get_bound_plugin(self, admin=None):
        return self.get_plugin_instance(admin)[0]

czpython: I don't want get_bound_plugin() to be related to get_plugin_instance() or the concept of plugin admin classes in any way.

The goal is to remove get_plugin_instance() at some point.

yakky: Ok, then how about the other way around? Using ``get_bound_plugin`` in ``get_plugin_instance`` instead of these lines https://github.com/divio/django-cms/blob/b91e58e14c47a00adcf102851a0a51de5d4de544/cms/models/pluginmodel.py#L239-L250
Having two method each one the subset of the other with slightly different implementations can be confusing and it can generate errors

czpython:
Ok to merge @yakky?

yakky:
@czpython LGTM

PR #5759, opened 2016-11-15
Fixed #5562 -- Show welcome screen regardless of DEBUG value

### LINKED ISSUE #5562: Unable to create an initial page when DEBUG = False
With `DEBUG = False` whatever you do if you don't have any content in your database will lead you to a 404 page (even if you go to http://example.com/?edit), making you unable to do anything on your website. I'm not sure how I've never dealt with this issue before, having been a DjangoCMS user for quite some time now, so I might just be missing something. Anyway here's the code that fires up the "new page wizard" if there's no content and you're in DEBUG mode: https://github.com/divio/django-cms/blob/release/3.3.x/cms/page_rendering.py#L59. According to this code there's no way to get this wizard with debug mode disabled and without having any page in your database.

According to the docs (http://docs.django-cms.org/en/release-3.3.x/user/tutorial/index.html):

> When you visit a brand new site for the first time, you will be invited to log in.

I'd suggest to either fix these instructions to explain how to create the first page, or fix the code to make the process more user-friendly (I still haven't found a proper way to create my first page except enabling DEBUG on prod and then create the initial page).

czpython:
Fixed #5562 

* Changed signature of internal `_handle_no_page` function

### DIFF
--- a/cms/page_rendering.py
+++ b/cms/page_rendering.py
@@ -25,7 +25,7 @@ def render_page(request, page, current_language, slug):
     context['has_view_permissions'] = user_can_view_page(request.user, page)
 
     if not context['has_view_permissions']:
-        return _handle_no_page(request, slug)
+        return _handle_no_page(request)
 
     response = TemplateResponse(request, template_name, context)
     response.add_post_render_callback(set_page_cache)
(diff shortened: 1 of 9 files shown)


czpython:
@vxsx would you mind having a look? Seems like one of the FE tests is legitimately failing.

cms/page_rendering.py
```diff
@@ -67,3 +60,10 @@ def _handle_no_page(request, slug):
         raise exc
     raise Http404('CMS Page not found: %s' % request.path)
 
+
+def _render_welcome_page(request):
+    context = {
+        'cms_version': __version__,
+        'cms_edit_on': get_cms_setting('CMS_TOOLBAR_URL__EDIT_ON'),
+    }
+    return TemplateResponse(request, "cms/welcome.html", context)
```
FinalAngel: shouldn't that be single quotes?

czpython:
This will land on 3.4.3

czpython:
Because this is a change in behavior, I'm pushing this to 3.5.0

edevil:
Since this only applies to 3.5.0, which has not been released yet, what is the correct way to bootstramp a CMS project?

Do we have to start it in DEBUG mode to add the first page?

czpython:
You can add the first page directly in the admin.

PR #5761, opened 2016-11-16
Docs: Run the django CMS installer in debug mode

Luke31:
-Since django CMS installer version 0.9, the installer runs by default in batch mode. (Source: https://github.com/nephila/djangocms-installer/issues/291#issuecomment-250995737)
-With the previous tutorial the django CMS installer version >= 0.9 will not ask any questions.
-To run the django CMS installer in wizard mode, one must set the -w flag

### DIFF
--- a/docs/introduction/install.rst
+++ b/docs/introduction/install.rst
@@ -49,11 +49,12 @@ Create a new directory to work in, and ``cd`` into it::
 
 Run it to create a new Django project called ``mysite``::
 
-    djangocms -f -p . mysite
+    djangocms -w -f -p . mysite
 
 This means:
 
 * run the django CMS installer
+* use wizard mode (``-w``) - **required for this tutorial**
 * install Django Filer too (``-f``) - **required for this tutorial**
 * use the current directory as the parent of the new project directory (``-p .``)
 * call the new project directory ``mysite``


czpython:
Hello @Luke31,
Thanks for the PR.

We removed the wizard mode from the installer as a default to make it easier for newcomers to get it up and running. As such, I think we should instead leave the same command but remove the wizard steps from the docs. To be confirmed by @evildmp and @yakky.

evildmp:
I'd prefer the simplest possible options for newcomers' installations, with a clear note that there are other options available. 

@Luke31 Would you like to update the documentation to reflect that?

Luke31:
I'd happily apply these new changes of the new simple batch option. However, I ran exactly into the same issue as the user mentioned in the first post: Incorrect timezone setting: local (Yes for some reason or another, my /etc/timezone file is not only empty, it doesn't even exist)
To overcome this issue, I had to run the installer in wizard-mode.

Now I don't know how many other users will run in similar issues without using the wizard and if so, I think the newcomers' installations will not get any simpler in batch mode. What do you think?

czpython:
I'm not familiar with the timezone issue but I think that its something we should fix on djangocms-installer.

Luke31:
So should I change it or leave it until the timezone issue has been addressed?

czpython:
I suggest fix it and we can land it once the timezone issue is addressed.
@yakky what do you think? Have you had a look at the timezone issue in https://github.com/nephila/djangocms-installer/issues/291?

yakky:
@czpython So far I've been unable to reproduce the issue
However, installer 0.9.2 ships with a workaround that should prevent any error like this fallbacking to UTC

evildmp:
@Luke31 Can you test the new django CMS Installer again please, to see if it trips up on the timezone?

If that seems OK, please go ahead and amend the PR to mention ``-w`` as an alternative option.

evildmp:
Replaced by https://github.com/divio/django-cms/pull/5815.

PR #5777, opened 2016-11-24
Fixed #5663 -- Add name of template tags to be discoverable by IDEs

### LINKED ISSUE #5663: Template tags should be discoverable by IDEs
Maybe add name 'placeholder' to register.tag(Placeholder) in https://github.com/divio/django-cms/blob/develop/cms/templatetags/cms_tags.py#L228 like that:

```
register.tag(Placeholder) -> register.tag('placeholder', Placeholder)
```

It's only for comfort: PyCharm will not be to highlight as warning and we would be jump to source code from template.

gkmngrgn:
Fixed #5663
It's required to be discoverable by PyCharm.

### DIFF
--- a/cms/templatetags/cms_tags.py
+++ b/cms/templatetags/cms_tags.py
@@ -156,7 +156,7 @@ class PageUrl(AsTag):
         return ''
 
 
-register.tag(PageUrl)
+register.tag('page_url', PageUrl)
 register.tag('page_id_url', PageUrl)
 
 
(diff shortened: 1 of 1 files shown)


gkmngrgn:
Additionally, we can remove `name` definition of template tags. But I'm not sure it's true way or not.

czpython:
Thanks @gkmngrgn.
I think its fine to leave the name attribute for now.

PR #5779, opened 2016-11-26
Fixed #5778 -- Don't set cache when cms_edit is True on session.

### LINKED ISSUE #5778: Cache can be set with the toolbar on in certain cases
When `CMS_TOOLBAR_ANONYMOUS_ON` is set to `True`, a user can pass the ?edit GET parameter to display the toolbar but when `CMS_PAGE_CACHE` is enabled, the cms will cache the entire page content including the toolbar. Note that this doesn't happen as soon as ?edit is passed, it will only happen if ?edit is passed and then the same user (same session) requests the page without the ?edit.

czpython:
Fixed #5778


FinalAngel:
@yakky is that also okay for you?

PR #5780, opened 2016-11-26
Fixed #5037 -- Add a simple helper text for inherit placeholder in structure mode.

### LINKED ISSUE #5037: Inherit placeholders should has icon at structure mode

gkmngrgn:
(#5037)

issue-5037

### DIFF
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -262,6 +262,7 @@ class ContentRenderer(object):
             page=current_page,
             editable=True,
             nodelist=nodelist,
+            inherit=inherit
         )
 
         # don't display inherited plugins in edit mode, so that the user doesn't
@@ -283,6 +284,7 @@ class ContentRenderer(object):
                 page=page,
                 nodelist=None,
                 editable=False,
+                inherit=inherit
             )
 
             if inherited_content:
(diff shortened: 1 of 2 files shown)


cms/plugin_rendering.py
```diff
@@ -481,6 +483,7 @@ def _render_page_placeholder(self, context, slot, page, editable=True, nodelist=
                 return nodelist.render(context)
             return ''
 
+        placeholder.is_inherit = inherit
```
czpython: I'm -1 on this as we're trying to avoid these on-demand attributes.
We would have to check what is the best way to pass this information

cms/templates/cms/toolbar/dragbar.html
```diff
@@ -25,6 +25,7 @@
 
     <div class="cms-dragbar-title">
         {{ placeholder.get_label }}
+        {% if placeholder.is_inherit %}<small>{% trans "Inherit Placeholder" %}</small>{% endif %}
```
czpython: I would like to keep consistency with the other placeholder types, as https://github.com/divio/django-cms/issues/5037 suggests we should just display an icon instead of text.

czpython:
Hello @gkmngrgn,
Thanks for this patch.

As suggested in the requested changes, I would like to stick to an icon instead of text to keep consistency.

cms/templates/cms/toolbar/dragbar.html
gkmngrgn: @czpython You're right. I can use the same pin icon like the static placeholders, and additionally we can add a tooltip about the placeholder is inherit. Does it make sense?

czpython: Yes I'm ok with this, would like to get @FinalAngel and @yakky involved.

FinalAngel:
I'd also prefer to stick to an icon, something like http://fontawesome.io/icon/sitemap/

czpython:
Closed by https://github.com/divio/django-cms/pull/5996

PR #5782, opened 2016-11-27
Fixed #5781 -- Don't inherit content if user has overridden it.

### LINKED ISSUE #5781: Inherited placeholders don't show overridden content
When using an inherited placeholder:

`{% placeholder something inherit %}`

This CMS will not display any plugins added to the placeholder in child pages, instead it will always show the inherited content.

czpython:
Fixed #5781

### DIFF
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -264,6 +264,9 @@ class ContentRenderer(object):
             nodelist=nodelist,
         )
 
+        if content:
+            return content
+
         # don't display inherited plugins in edit mode, so that the user doesn't
         # mistakenly edit/delete them. This is a fix for issue #1303. See the discussion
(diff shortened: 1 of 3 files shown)


cms/plugin_rendering.py
```diff
@@ -264,6 +264,9 @@ def render_page_placeholder(self, slot, context, inherit, nodelist=None):
             nodelist=nodelist,
         )
 
+        if content:
+            return content
+
```
jrief: I would merge this with the next line:

```
        if content or not inherit or self.toolbar.edit_mode:
            return content
```

czpython: I thought about it but decided against it because the next line is very different in logic.
Best to keep the lines separate and address the second line directly with the comment

PR #5783, opened 2016-11-28
Fixes #5725 -- Page cache and django.contrib.messages compatibility

### LINKED ISSUE #5725: page cache & django.contrib.messages
The page cache seems not compatible with `django.contrib.messages`. Pages are cached with messages if any and displayed without taking current available messages in account.

As a workaround, we disabled page cache in settings (but we loose cache benefit in all other legitimate cases):

```
CMS_PAGE_CACHE = False
```

Fixes:
- A specific fix could be to check messages in set_page_cache/get_page_cache functions.
- A more flexible fix could be to allow to write a custom `_cache_page_key` function configurable in settings (and also prevent cache use in case of `None` key). And it would be wise to advise developer to only add a suffix in those custom implementations (or return `None`).

Both could be done. What would you say?

ollb:
Fixes #5725

### DIFF
--- a/cms/cache/page.py
+++ b/cms/cache/page.py
@@ -14,9 +14,14 @@ from cms.constants import EXPIRE_NOW, MAX_EXPIRATION_TTL
 from cms.toolbar.utils import get_toolbar_from_request
 from cms.utils import get_cms_setting
 from cms.utils.helpers import get_timezone_name
+from django.contrib.messages import get_messages
 
 
 def _page_cache_key(request):
+    # do not store nor retrieve page in cache if some messages are going to be displayed
+    messages = get_messages(request)
+    if len(messages) or getattr(messages, 'used', False) or getattr(messages, 'added_new', False):
+        return None
     #md5 key of current path
     cache_key = "%s:%d:%s" % (
         get_cms_setting("CACHE_PREFIX"),
@@ -68,8 +73,8 @@ def set_page_cache(response):
             get_cms_setting('CACHE_DURATIONS')['content'],
             min_placeholder_ttl
         )
-
-        if ttl > 0:
+        page_cache_key = _page_cache_key(request)
+        if ttl > 0 and page_cache_key is not None:
             # Adds expiration, etc. to headers
             patch_response_headers(response, cache_timeout=ttl)
             patch_vary_headers(response, sorted(vary_cache_on_set))
@@ -79,7 +84,7 @@ def set_page_cache(response):
             # recomputing it on cache-reads.
             expires_datetime = timestamp + timedelta(seconds=ttl)
             cache.set(
-                _page_cache_key(request),
+                page_cache_key,
                 (
                     response.content,
                     response._headers,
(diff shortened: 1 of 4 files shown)


ollb:
Travis lies ?

cms/cache/page.py
```diff
@@ -14,9 +14,14 @@
 from cms.toolbar.utils import get_toolbar_from_request
 from cms.utils import get_cms_setting
 from cms.utils.helpers import get_timezone_name
+from django.contrib.messages import get_messages
 
 
 def _page_cache_key(request):
+    # do not store nor retrieve page in cache if some messages are going to be displayed
+    messages = get_messages(request)
+    if len(messages) or getattr(messages, 'used', False) or getattr(messages, 'added_new', False):
```
czpython: I would like to avoid bundling the check logic to this function.
This function should always generate the key regardless of conditions.

I would create a function called "can_cache_page" or something like that which encapsulates the logic in [here](https://github.com/divio/django-cms/blob/release/3.4.x/cms/cache/page.py#L38) as well as the messages check.

cms/cache/page.py
```diff
@@ -68,8 +73,8 @@ def set_page_cache(response):
             get_cms_setting('CACHE_DURATIONS')['content'],
             min_placeholder_ttl
         )
-
-        if ttl > 0:
+        page_cache_key = _page_cache_key(request)
+        if ttl > 0 and page_cache_key is not None:
```
czpython: as pointed above cache key should not be none.

cms/cache/page.py
```diff
@@ -95,7 +100,12 @@ def set_page_cache(response):
 
 def get_page_cache(request):
     from django.core.cache import cache
-    return cache.get(_page_cache_key(request), version=_get_cache_version())
+    page_cache_key = _page_cache_key(request)
+    if page_cache_key is None:
+        return None
+    else:
+        return cache.get(page_cache_key,
```
czpython: same as above

cms/tests/test_cache.py
```diff
@@ -451,6 +453,30 @@ def test_cache_page(self):
...
 
             #
+            # Test that a message should prevent cache to be retrieved
+            #
+            messages = get_messages(request)
+            messages.add(40, 'this is a django.contrib.message message')
+            messages.update(response)
+            request.session.save()
+
+            #import ipdb; ipdb.set_trace()
```
czpython: please remove

ollb:
@czpython ok, it seems fair enough to me.

I guess the conditions on which we want to cache page aren't the sames than the one on which we want to retrieve a page from cache. So we probably need two functions : `can_get_page_cache` and `can_set_page_cache`.

Another thing : the purpose of the code in `set_page_cache` is mainly to test if we can cache page but it also provides the ttl to cache page. Then `can_set_page_cache` could return the `ttl` ?

Are you ok with that ?

czpython:
Hello @ollb,
`can_cache_page` can handle both get and set as the conditions are the same.

The `can_` functions should return `True` or `False` no other value.
The `set_page_cache` function has the job of checking if page cache can be set via `can_cache_page` and if it can be set then it calculates the ttl of the cache and sets it.

czpython:
Hello @ollb ,
Would you be able to make these changes before the end of the week?
We're planning to hit code freeze by then.

czpython:
I think we'll have to think about this further.
Other data bound to the request like messages might also need to be excluded from cache.
As a result, we'll likely have to implement a hook to determine if a page can be cached or fetched from cache.

PR #5795, opened 2016-12-08
removed deprecated argument current_app

jrief:
The argument current_app is deprecated since Django-1.9, so it's safe to remove it completety


jrief:
You should ignore the failed test https://travis-ci.org/divio/django-cms/jobs/182426636
This is just one build, which for some reasons just failed to start, thus has nothing to do with this patch. All other tests pass from the matrix passed without problems.

czpython:
Thanks @jrief!

pcolmant:
@czpython, am I wrong or the change made in this #5795 has been "cancelled" when the #5799 has been "committed" ?

PR #5799, opened 2016-12-11
Fixes/cms 3.4 compat

jrief:
PR depends on PR #5795

- added deprecation waring on unused method.

- since ``get_parent_classes()`` became a classmethod, do not instantiate the plugin.


jrief:
@czpython I'm currently porting djangocms-cascade to CMS-3.4.
What I see, is that a lot of code changed, therefore Cascade does not work anymore as it used to. I currently have to reread a lot of the changed code and therefore I stumbled across quite often at statements such as ``parent_classes or []`` or ``child_classes or []``.

I don't know if this has been documented somewhere, but the until version 3.3, if ``parent_classes = None`` meant, that we should not care if parents exists or not. Therefore having ``parent_classes = []`` was fundamentally different from ``parent_classes = None``. I'm unsure, if this is still the case in version 3.4.

Thanks

czpython:
@jrief In this case should not be a different behavior if attribute is set to `None` or empty list, the only reason for the `or []` format is to do `in` operations. In some cases like https://github.com/divio/django-cms/blob/release/3.4.x/cms/plugin_base.py#L171 we can remove it as it will be the same for `bool()`

jrief:
@czpython rebased

czpython:
Hello @jrief,
Please add only the following entries:

* Added a deprecation warning to method ``render_plugin()`` in class ``CMSPluginBase``.
* Since ``get_parent_classes()`` became a classmethod, do not instantiate plugin before invocation.

To the changelog here https://github.com/divio/django-cms/blob/release/3.4.x/CHANGELOG.txt and remove the file docs/upgrade/3.4.2.rst as we only add this file prior to releasing.

czpython:
Hello @jrief,
Would you be able to make the requested changes before the end of the week?
We're planning to hit code freeze by then.

jrief:
@czpython as you requested, I reverted some changes. However there seems to be a conflict in CHANGLELOG.txt, I can't resolve locally.

Also, why did you want me to remove code, which breaks Django-1.10 compatibility?

czpython:
Thanks @jrief.
Which code are you referring to?

jrief:
placeholderadmin.py line 744:

I reverted to this, which is incompatible with Django-1.10:
```
        return TemplateResponse(request, "admin/cms/page/plugin/delete_confirmation.html", context,
                                current_app=self.admin_site.name)
```

czpython:
Ah that is because the PR with that fix was already merged in https://github.com/divio/django-cms/commit/824f389ee4b1907e28463c8b4bfa64742756df7b

jrief:
yes, but that PR (824f389) was from myself, and this PR depended on that old one.

btw. I tried to fix the conflict in CHANGELOG.txt but did not succeed. Can you do that?

czpython:
is why I asked for this one to be rebased with release/3.4.x so it picks up those changes.
No worries though, I will take care of the CHANGELOG.txt conflicts and rebase with release/3.4.x

jrief:
ok, thanks

czpython:
@jrief I've done the rebasing but I'm unable to push to your branch.
Did you enable "Edits from maintainers"? or better yet can you give me access to your fork?

czpython:
Merged in a9e2a8d3b7a6b6bc4f6f0ca3edd492be6ddd05f0

PR #5801, opened 2016-12-12
Fix language used in page settings link

sephii:
The language used for the settings button in the admin pages menu always points to the last language because it's using a loop variable instead of the proper language.

This is already fixed in master but I'm not sure if master is used for bugfix releases so feel free to close this PR if you think it's not necessary.

### DIFF
--- a/cms/utils/admin.py
+++ b/cms/utils/admin.py
@@ -114,7 +114,7 @@ def render_admin_rows(request, pages, site, filtered=False, language=None):
             'filtered': filtered,
             'metadata': metadata,
             'page_languages': page.get_languages(),
-            'preview_language': language,
+            'preview_language': lang,
             'has_add_page_permission': user_can_add(user, target=page),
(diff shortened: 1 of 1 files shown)


FinalAngel:
@vxsx @mkoistinen didn't we had here once a discussion related to the user interface language and how it should behave in this case?

vxsx:
We did, and afaicr we decided to stick with language, because of the pagetree "languages" header
it should pick that instead of the request language

sephii:
That's actually the behaviour of my patch (ie. use `get_language_from_request`), and it matches the refactoring done in `master`. The current implementation in `3.4.1` is buggy because it always uses the last language from the languages list because it incorrectly references a loop variable (https://github.com/divio/django-cms/blob/release/3.4.x/cms/utils/admin.py#L82).

czpython:
@sephii You're right, oversight on my end. Will review/merge a bit later.

Issue #3479, opened 2014-10-01
migration issue after changing template choices [1.7] 


chhantyal:
After adding custom templates in setting file:

```
CMS_TEMPLATES = (
    ('cms/content_even.html', 'Template for even number boxes'),
    ('cms/content_odd.html', 'Template for odd number boxes'),
)
```

When I run makemigrations command, it seems to create separate migration file like this:

```
migrations.AlterField(
            model_name='page',
            name='template',
            field=models.CharField(default=b'INHERIT', help_text='The template used to render the content.', max_length=100, verbose_name='template', choices=[(b'cms/content_even.html', 'Template for even number boxes'), (b'cms/content_odd.html', 'Template for odd number boxes'), (b'INHERIT', 'Inherit the template of the nearest ancestor')]),
        )
]
```

This will force us to create migration script every time we change template choices.

othane:
This still seems to be a problem for the CMS_TEMPLATE_INHERITANCE setting ... If I set this to False I get a spurious migration made against the cms ?

jsma:
Can you put together a pull request for this? You'll just need to modify the initial 0001 migration to make the 'default' choice on the `template` field to dynamically determine the default, just like is done in the model definition `TEMPLATE_DEFAULT = TEMPLATE_INHERITANCE_MAGIC if get_cms_setting('TEMPLATE_INHERITANCE') else get_cms_setting('TEMPLATES')[0][0]`.

You should also check if any subsequent migrations reference this field as well and make sure those point to the dynamic default as well.

othane:
Sure I will take a look this afternoon

othane:
OK I have done the patch, I am not 100% on your repo strucuture, it looks like I should make a pull request into the releases branch? or the 3.4.x branch (does that back pulled into develop I assume)?

PR #5807, opened 2016-12-15
Fix template inheritance spurious migration

othane:
I think this is the only migration that touches the template field, this certainly fixed the problem for me.

Hopefully I am making this pull request into the right branch, if not let me know and ill rebase it to something else

Also I did not see any documentation about running local tests before sending a pull request, so let me know if I have broken something and ill fix it.

see https://github.com/divio/django-cms/issues/3479

Cheers.


othane:
Hmmm I mucked that up some how ... it should only be 1 commit

PR #5808, opened 2016-12-15
fix template inheritance setting creates migration

othane:
I think this is the only migration that touches the template field, this certainly fixed the problem for me.

Hopefully I am making this pull request into the right branch, if not let me know and ill rebase it to something else

Also I did not see any documentation about running local tests before sending a pull request, so let me know if I have broken something and ill fix it.

see #3479

Cheers.

### DIFF
--- a/cms/migrations/0001_initial.py
+++ b/cms/migrations/0001_initial.py
@@ -3,12 +3,15 @@ from __future__ import unicode_literals
 
 from cms.models import ACCESS_CHOICES, Page
 from cms.utils.conf import get_cms_setting
+from cms.constants import TEMPLATE_INHERITANCE_MAGIC
 from django.conf import settings
 from django.db import models, migrations
 import django.utils.timezone
 from django.utils.translation import ugettext_lazy as _
 
 template_choices = [(x, _(y)) for x, y in get_cms_setting('TEMPLATES')]
+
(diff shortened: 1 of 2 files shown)


cms/migrations/0001_initial.py
```diff
@@ -3,13 +3,16 @@
...
 from cms.utils.conf import get_cms_setting
+from cms.constants import TEMPLATE_INHERITANCE_MAGIC
 from django.conf import settings
 from django.db import models, migrations
 import django.utils.timezone
 from django.utils.translation import ugettext_lazy as _
 
 template_choices = [(x, _(y)) for x, y in get_cms_setting('TEMPLATES')]
 
+TEMPLATE_DEFAULT = TEMPLATE_INHERITANCE_MAGIC if get_cms_setting('TEMPLATE_INHERITANCE') else get_cms_setting('TEMPLATES')[0][0]
```
czpython: To keep consistency with the variable above, please lowercase this one. Or maybe uppercase both since they're both constants.

othane:
should all be fixed up now ... cheers :)

czpython:
@othane Sorry I forgot one more thing, please add a changelog entry https://github.com/divio/django-cms/blob/release/3.4.x/CHANGELOG.txt

othane:
no problem, I think I put that in the right place now :)

PR #5809, opened 2016-12-15
Fixes/flexible parent+child classes

jrief:
In django-CMS version 3.4, the method to determine if plugins are allowed as child and/or parent plugins, has changed a lot.

This is causing major problems in djangocms-cascade, since there I have to determine possible relationships dynamically. This means that depending on where inside the placeholder tree a plugin is located, possible parent and/or child relations can be allowed or denied.

Therefore I looked for alternatives and one proposal to fix it, would be to optionally pass in the current plugin instance to the classmethods ``CMSPlugin.get_child_classes(...)`` and ``CMSPlugin.get_parent_classes(...)``. Additionally I have to deactivate the placeholder caching for these plugins, since the list of allowed child- and parent-classes then has to be build dynamically.

This PR depends on #5799. If we can agree on it, I'll add some documentation and tests.


czpython:
Hello @jrief,
I'm not very familiar with djangocms-cascade but this change for me is a no-go as it adds more complexity to this mechanism.
Prior to the changes in 3.3.x, the plugin parent fetching also did not have any kind of awareness of the plugin instance.

That said, let's discuss how we can help with djangocms-cascade and then if it makes sense for other apps/projects then we can look into a change on the django CMS core.

jrief:
@czpython In CMS version 3.3 the plugins had an attribute pointing to the current instance. I used that to determine the dynamic properties of the tree.

Since ``get_parent_classes`` and ``get_child_classes``, now are classmethods, this option is not available anymore.

If this feature request is rejected, how would you propose to implement dynamic parent- and child-relationships? If I can not find an alternative solution for this, it's a huge setback for djangocms-cascade.

czpython:
You mean the `cms_plugin_instance`? This was a really nasty internal.. but I see your point.
Can you explain a bit on what logic you're currently using to determine parent classes and child classes for plugins in djangocms-cascade?

jrief:
Yes, ``cms_plugin_instance`` was the then only working approach until version 3.3.

BTW, did you ever try Cascade? It is the primary plugin ecosystem in django-SHOP, so if you want to get a quick look, please download a fully working Docker container as explained here: http://django-shop.readthedocs.io/en/latest/tutorial/quickstart.html

On the first Django-CMS day in Florence, I gave a presentation about this concept. @yakky and @evildmp were present, so it would be great to hear about their opinions too.

In django-SHOP, if you open the checkout page in structure mode ( http://*my-docker-machine*:9001/en/cart/checkout/?edit ), there are a few Segment plugins. They can be considered as wrappers around other plugins, since they do not add any HTML logic and therefore a Segment-Plugin can be added anywhere in between. Now this would cause trouble for their children, since they must fulfill their constraints. Therefore a Segment plugin tries to bypass those children and parent constraints.

Many other plugins in Cascade use that same approach, for instance the Accordion Panel, the Tabset, the Processbar, etc. I named them "transparent" since they can be moved between two plugins without "disturbing" their natural parent-, child-classes constraints.

czpython:
@jrief I'll try to have a look this weekend but we're hitting code-freeze by the end of next week, so its very likely any change regarding this won't make it into 3.4.2

jrief:
I believe, that with the above patches everything works fins. Anyway, having context sensitive rules, which plugins can – and which can't be added, may be useful for other eco-system too.

With this patch, Cascade currently works very well, even better than before. Therefore it would be really helpful, if it is going to be merged.

jrief:
yesterday I've written a small blog post about this problem: 
https://plus.google.com/+JacobRief/posts/4moFBcDqpT1

czpython:
Hello @jrief,
I've spoken to Iacopo about this.
We agreed on the use-case, so we'll review the implementation once more and get back to you with any adjustments.

jrief:
Great! These are super good news for everybody using DjangoCMS-Cascade and Django-SHOP.

If my design proposal is accepted, I'll add the missing unit tests as soon as possible.

yakky:
Assigning to myself for a more in-depth review

jrief:
shall I start to add some unit tests for this feature request?

czpython:
Hello @jrief, 
We're having a board meeting next week.
I'll review this no later than this week and will bring it up for discussion at the meeting.

cms/plugin_base.py
```diff
@@ -115,6 +115,9 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
 
     disable_child_plugins = False
 
+    child_plugins_cache = True
```
czpython: Please rename these to:

`cache_child_classes`
`cache_parent_classes`

cms/templatetags/cms_js_tags.py
```diff
@@ -45,10 +45,20 @@ def render_placeholder_toolbar_js(placeholder, render_language, content_renderer
     )
 
     def _render_plugin_js(plugin):
+        if plugin.plugin_type in plugin_children:
```
czpython: please use a try/except instead of an if for both:

```python
try:
    child_classes = plugin_children[plugin.plugin_type]
except KeyError:
    # Plugin has opted out of the cached child classes
    # cache_child_classes is set to False
    child_classes = plugin.plugin_class.get_child_classes(slot=slot, page=page, instance=plugin)
```

cms/plugin_base.py
czpython: Also above these two add a comment warning the user that disabling these will result in a performance hit depending on how many of these plugins are found on the page.

cms/templatetags/cms_js_tags.py
```diff
@@ -45,10 +45,20 @@ def render_placeholder_toolbar_js(placeholder, render_language, content_renderer
...
+            child_classes = plugin_children[plugin.plugin_type]
+        else:
+            child_classes = plugin.plugin_class.get_child_classes(slot=slot, page=page, instance=plugin)
```
czpython: I believe this is an `AttributeError`.. plugin instances don't have a `plugin_class` attribute.

jrief:
I hope, that I changed all your annotations.

For CMS plugins, I believe we can rely on ``plugin.plugin_type`` always being there.

jrief:
@czpython any change requests still open?

bittner:
I'm getting a `TypeError: get_parent_classes() missing 1 required positional argument: 'page'` with

- Django==1.9.12
- django-cms==3.4.2
- djangocms-cascade==0.11.3

At first sight the problem seems not directly related to Cascade. According to the 3.4.2 release notes the "CMSPlugin.get_parent_classes() [was changed] from method to classmethod". Now all old cms plugins are potentially broken. Will that stay the same or is there a chance this will be reverted again?

Details
-------
The error is raised by something as simple as an `object` in a template:
``` django
{% for object in objects_list %}
    <li class="wow fadeInUp" data-wow-delay="0.{{ forloop.counter0 }}s">
        <figure><img src="{{ object.image.url }}" alt="{{ object.alt_text }}"></figure>
        <div class="caption">
            {% if object.link %}<a href="{{ object.link.get_public_url }}">{{ object.link_label|safe }}</a>{% endif %}
            {% if object.text %}<p>{{ object.text|safe }}</p>{% endif %}
        </div>
    </li>
{% endfor %}
```

django.template:
``` django
Exception while resolving variable 'object' in template 'holiday_theme/plugin_circle_images_gallery.html'.
```

exception:
``` django
TypeError: get_parent_classes() missing 1 required positional argument: 'page'
```

electroniceagle:
@czpython, Are there any missing changes?  I'm about to give this PR a try.
@bittner, try the https://github.com/jrief/djangocms-cascade/tree/releases/0.12.x branch if you implement this pull request

czpython:
@jrief @electroniceagle 
I've been quite busy so I've not been able to have another look.
That said, my comment about the `AttributeError` has not been addressed.

cms/templatetags/cms_js_tags.py
```diff
@@ -45,10 +45,20 @@ def render_placeholder_toolbar_js(placeholder, render_language, content_renderer
...
     def _render_plugin_js(plugin):
+        try:
+            child_classes = plugin_children[plugin.plugin_type]
+        except KeyError:
+            child_classes = plugin.plugin_class.get_child_classes(slot=slot, page=page, instance=plugin)
+
+        try:
+            parent_classes = plugin_parents[plugin.plugin_type]
+        except KeyError:
+            parent_classes = plugin.plugin_class.get_parent_classes(slot=slot, page=page, instance=plugin)
```
czpython: This throws an `AttributeError`

cms/templatetags/cms_js_tags.py
```diff
@@ -45,10 +45,20 @@ def render_placeholder_toolbar_js(placeholder, render_language, content_renderer
     )
 
     def _render_plugin_js(plugin):
+        try:
+            child_classes = plugin_children[plugin.plugin_type]
+        except KeyError:
+            child_classes = plugin.plugin_class.get_child_classes(slot=slot, page=page, instance=plugin)
```
czpython: This throws an `AttributeError`

jrief: Sorry that slipped under my radar.
The current patch should fix it.

cms/templatetags/cms_js_tags.py
jrief: Sorry that slipped under my radar.
The current patch should fix it.

jrief:
@czpython are there still any open change requests on this pull request?

cms/templatetags/cms_js_tags.py
```diff
@@ -48,12 +48,14 @@ def _render_plugin_js(plugin):
         try:
             child_classes = plugin_children[plugin.plugin_type]
         except KeyError:
-            child_classes = plugin.get_plugin_class().get_child_classes(slot=slot, page=page, instance=plugin)
+            plugin_class = plugin_pool.plugins[plugin.plugin_type]
```
jrief: btw. in order to avoid this, I added a method

```
    @cached_property
    def plugin_class(self):
        return self.get_plugin_class()
```

to my model inheriting from ``CMSPlugin``. Later on, I forgot about it, hence your correct annotation 12 days ago about the ``AttributeError`` when attempting to access the ``plugin_class``. What about adding that method directly to ``CMSPlugin``?

jrief:
Short annotation about the 3 failing unit tests, they all failed because of this error:
``OperationalError: (2006, 'MySQL server has gone away')``
which is not related to buggy code.
@czpython please restart them manually – I have no permissions for this

cms/templatetags/cms_js_tags.py
czpython: Its ok, adding the method to the class won't help us here as I still want to avoid calling method + pool for each plugin.

czpython:
Merged manually in:

https://github.com/divio/django-cms/commit/84c9b1bec721a4011094b467ae101f95b07f9e3b
https://github.com/divio/django-cms/commit/c9c4665b858e3e830bf8a98d3709bdb447171b2b
https://github.com/divio/django-cms/commit/63e51f37eabf08623e14e07d4d920d17640839ba

Thanks!

czpython:
I left https://github.com/divio/django-cms/pull/5809/commits/e639eb692b084a93a113fa5b379a8e004ebf9c1a out as django didn't deprecate `current_app` on reverse function.

PR #5825, opened 2016-12-27
Re-introduced "Revert to live" menu option

czpython:


cms/admin/pageadmin.py
```diff
@@ -840,6 +852,26 @@ def copy_page(self, request, page_id, extra_context=None):
...
+        """
+        page = get_object_or_404(self.model, id=page_id)
+
+        # ensure user has permissions to publish this page
+        if not self.has_revert_to_live_permission(request, language, obj=page):
+            return HttpResponseForbidden(force_text(_("You do not have permission to revert this page.")))
+
+        page.revert_to_live(language)
+
+        messages.info(request, _('The page "%s" was successfully reverted.') % page)
```
evildmp: ``"%s" was reverted to the live version.`` would be better:

* tells us *what* it was reverted to
* "page" and "successfully" are both redundant

cms/models/pagemodel.py
```diff
@@ -321,6 +321,29 @@ def move_page(self, target, position='first-child'):
...
 
+    def revert_to_live(self, language):
+        """Revert the draft version to the same state as the public version
+        """
+        if not self.publisher_is_draft:
+            # Revert can only be called on draft pages
+            raise PublicIsUnmodifiable('The public instance cannot be reverted. Use draft.')
+
+        if not self.publisher_public:
+            raise PublicVersionNeeded('A public version of this page is needed')
```
evildmp: Can we inform the user what to do?

cms/admin/pageadmin.py
czpython: Sure. I'll change it.

cms/models/pagemodel.py
czpython: I would rather not, this is an internal. The exception will never be user facing.

czpython:
Ping @yakky

PR #5826, opened 2016-12-28
Rewrote manual installation how-to documentation

evildmp:
Completely rewrites http://docs.django-cms.org/en/stable/how_to/install.html, removing out-of-date and hard-to-keep-updated sections.

Brings everything else up-to-date, and better in line with http://docs.django-cms.org/en/stable/introduction/index.html.


README.rst
```diff
@@ -56,14 +56,27 @@ More information on `our website <http://www.django-cms.org>`_.
 Requirements
 ************
 
-django CMS requires Django 1.8, and Python 2.7, 3.3 or 3.4.
+See the `Python/Django requirements for the current release version
+<http://http://docs.django-cms.org/en/stable/#software-version-requirements-and-release-notes>`_ in our documentation.
```
czpython: invalid link?

README.rst
```diff
@@ -56,14 +56,27 @@ More information on `our website <http://www.django-cms.org>`_.
...
 
 *************
 Documentation
 *************
 
-Please head over to our `documentation <http://docs.django-cms.org/>`_ for all
-the details on how to install, extend and use the django CMS.
+We maintain documentation for several versions of the project. Key versions are:
+
+* `stable <http://http://docs.django-cms.org>`_ (default), for the **current release** version
```
czpython: shouldn't these links and others be https?

docs/how_to/install.rst
```diff
@@ -1,671 +1,519 @@
...
+Install the django CMS package
+******************************
 
-Recommended
-===========
+Check the :ref:`Python/Django requirements <requirements>` for this version of django CMS.
 
-These packages are not *required*, but they provide useful functionality with
-minimal additional configuration and are well-proven.
+django CMS also has other requirements, which it lists as dependencies in its ``setup.py``.
```
czpython: we shouldn't link to setup.py, the requirements section above should contain all dependencies including the ones in setup.py

docs/how_to/install.rst
```diff
@@ -1,671 +1,519 @@
...
-.. _django-treebeard: http://code.tabo.pe/django-treebeard/src
-.. _django-sekizai: https://github.com/ojii/django-sekizai
-.. _djangocms-admin-style: https://github.com/divio/djangocms-admin-style
+******************************
+Install the django CMS package
+******************************
 
-Recommended
-===========
+Check the :ref:`Python/Django requirements <requirements>` for this version of django CMS.
```
czpython: I would say "Check the .. **section** for this version of django CMS."

docs/how_to/install.rst
```diff
@@ -1,671 +1,519 @@
...
 
-* Creating a separate set of credentials for the django CMS project.
-* Creating a new database for the django CMS project, not reusing an
-  existing one.
+to the list.
 
-.. _PostgreSQL: http://www.postgresql.org/
-.. _MySQL: http://www.mysql.com
+You can also add ``'cms.middleware.utils.ApphookReloadMiddleware'``. It's not absolutely necessary, but it's
+:ref:`useful <reloading_apphooks>`. It needs to be as close to the start of the list as possible.
```
czpython: technically it should be the first one

docs/how_to/install.rst
```diff
@@ -1,671 +1,519 @@
...
-Notice that django CMS v3.2 introduces a new middleware:
-``cms.middleware.utils.ApphookReloadMiddleware``. This should be placed very
-near the top of your middleware classes tuple/list.
+* ``{% load cms_tags sekizai_tags %}`` loads the template tag libraries we use in the template.
+* ``{% page_attribute "page_title" %}`` extracts the page's ``page_title`` attribute.
+* ``{% render_block "css" %}`` and ``{% render_block "js" %}`` are template tags that load blocks of HTML defined by
+  Django applications. django CMS defines blocks for CSS and JavaScript, and requires these two Sekizai tags. We
+  recommended placing ``{% render_block "css" %}`` just before the ``</head>`` tag, and and ``{% render_block "js" %}``
+  tag just before the ``</body>``.
+* ``{% cms_toolbar %}`` renders the django CMS toolbar.
```
czpython: should we point out here that this tag cannot be used inside another tag?
or is it somewhere else?

docs/how_to/install.rst
```diff
@@ -1,671 +1,519 @@
...
 
-Now, on with the actual template files!
+You also need to add::
 
-Fire up your favourite editor and create a file called ``base.html`` in a folder called
-``templates`` in your ``myproject`` directory.
+    THUMBNAIL_HIGH_RESOLUTION = True
 
-Here is a simple example for a base template called ``base.html``:
+    THUMBNAIL_PROCESSORS = (
```
czpython: I don't think we should be this detailed about django-filer or other dependency installations.
Instead we should just link to them because otherwise we'd have to maintain it here too and the cms docs shouldn't be affected too much from a change in third party apps.

README.rst
evildmp: It's invalid right now, but this very pull request will make it valid.

README.rst
evildmp: No - not at present, anyway. If you use https:
screenshot 2017-01-03 12 49 17

docs/how_to/install.rst
evildmp: That's exactly what I want to get away from. 

The requirements listed in ``setup.py`` will always be the actual requirements, whereas the ones listed here will inevitably become out-of-date or incorrect (as they were before). Unless we can guarantee that a list here will be updated whenever someone touches setup.py, I don't want to include it.

czpython: Then shouldn't we just link to the setup.py and not have a `Python/Django requirements <requirements>` section

README.rst
czpython: I meant that http is there twice

docs/how_to/install.rst
evildmp: Now linked to the reference, which does explain this.

docs/how_to/install.rst
evildmp: It's a balancing act - I have removed vast amounts of installation and configuration of dependencies, but kept this because I think not having it would either risk presenting a stumbling block, or significantly reducing the experience of the user.

I added a note to the section to the effect that this is just an example configuration to get the user started.

README.rst
evildmp: When you put it like that...

docs/how_to/install.rst
evildmp: I think that having the Python and Django (and only the Python and Django) version requirements in the home page of the documents is fine, with the README and any other documentation (like the install documentation) pointing at that single authoritative place.

We should not have requirements lists buried away in the documentation - once, on the home page, is OK though, and it would be a bit weird not to have any mention of them there.

README.rst
```diff
@@ -56,14 +56,27 @@ More information on `our website <http://www.django-cms.org>`_.
...
```
czpython: @evildmp looks like another repeated http

evildmp: Squashed!

PR #5832, opened 2017-01-01
WiP: Cache per-site APP_RESOLVERS

yakky:
Plus allow reload_urlconf to not clear cache on every load

Coupled with nephila/djangocms-multisite#18 it fixes the issues with multiste apphooks

This is just a preliminary PR, it needs documentation and tests, but at this stage it should be complete for initial real-world testing

### DIFF
--- a/cms/appresolver.py
+++ b/cms/appresolver.py
@@ -16,12 +16,12 @@ from cms.models.pagemodel import Page
 from cms.utils.compat import DJANGO_1_8, DJANGO_1_9
 from cms.utils.i18n import get_language_list
 
-APP_RESOLVERS = []
+APP_RESOLVERS = {}
 
 
 def clear_app_resolvers():
     global APP_RESOLVERS
-    APP_RESOLVERS = []
+    APP_RESOLVERS = {}
 
 
 def applications_page_check(request, current_page=None, path=None):
@@ -38,7 +38,12 @@ def applications_page_check(request, current_page=None, path=None):
     for lang in get_language_list():
         if path.startswith(lang + "/"):
             path = path[len(lang + "/"):]
-    for resolver in APP_RESOLVERS:
+    try:
+        current_site = Site.objects.get_current(request)
+    except Site.DoesNotExist:
+        return None
+
+    for resolver in APP_RESOLVERS.get(current_site, []):
         try:
             page_id = resolver.resolve_page_id(path)
             # yes, it is application page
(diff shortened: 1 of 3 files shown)


cms/appresolver.py
```diff
@@ -218,6 +223,11 @@ def _get_app_patterns():
         current_site = None
     included = []
 
+    # If APP_RESOLVERS is loaded,
+    if APP_RESOLVERS.get(current_site, []):
+        return copy(APP_RESOLVERS.get(current_site, []))
```
czpython: why copy?

cms/appresolver.py
```diff
@@ -38,7 +39,11 @@ def applications_page_check(request, current_page=None, path=None):
     for lang in get_language_list():
         if path.startswith(lang + "/"):
             path = path[len(lang + "/"):]
-    for resolver in APP_RESOLVERS:
+    try:
+        current_site = Site.objects.get_current(request)
+    except Site.DoesNotExist:
+        current_site = None
```
czpython: we should fail fast here.
I suggest `return None`

czpython:
Thanks @yakky.
Docs and tests will come on this same PR?

cms/appresolver.py
yakky: Done

cms/appresolver.py
czpython: I would prefer to have the copy in the urls.py then, this way is more clear why is needed and logic falls on the caller instead of this function.

yakky: Change done

yakky:
@stefanfoulis @czpython I'll integrate some tests / comments, but feel free to test it together with nephila/djangocms-multisite#18

stefanfoulis:
Local tests were successful (tested with the aldryn-newsblog Apphook). I'll now roll out this change on the test servers of this project and report back if there are any issues.

yakky:
Thanks @stefanfoulis 
In the weekend I should be able to add some tests

stefanfoulis:
@yakky I've done some more testing. Good to merge from my point of view.

czpython:
@yakky 
Just fyi, we're planning the RC for Monday afternoon.

yakky:
@czpython @stefanfoulis I'll have the tests in this PR by tomorrow

stefanfoulis:
ping @yakky :-D

stefanfoulis:
@yakky would be really awesome if we could get this into ``3.4.4``.

yakky:
Oops, somehow I completely forgot this. I should be able to send in the tests tomorrow

yakky:
@stefanfoulis works well in production?

czpython:
@yakky Due to the size of the release, we've decided to go-ahead with it, so this would land on 3.4.5

czpython:
@yakky can you land this by the 3.4.5 release?

czpython:
@yakky 
Please reopen against the develop branch

PR #5835, opened 2017-01-04
Don't reload plugins on move

czpython:
https://github.com/divio/djangocms-history/commit/5bf6a690e8589341e3fba83401908167ccd136be

### DIFF
--- a/cms/plugin_base.py
+++ b/cms/plugin_base.py
@@ -123,7 +123,7 @@ class CMSPluginBase(six.with_metaclass(CMSPluginBaseMetaclass, admin.ModelAdmin)
 
     action_options = {
         PLUGIN_MOVE_ACTION: {
-            'requires_reload': True
+            'requires_reload': False
         },
         PLUGIN_COPY_ACTION: {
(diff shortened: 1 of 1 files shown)


stefanfoulis:
@czpython To understand the implications of this change. Does this mean that moving any plugin that uses its position in the plugin tree or the context as set by parent plugins to influence the way it renders, to not show up correctly until the next full page reload?

e.g Grid/Row/Column plugins that pass width to child plugins. Or a GalleryItem plugin that renders differently depending on the options of the parent Gallery plugin.

So I guess this change is a tradeoff between speed and compatibility. I understand that in the vast  majority of cases there will be no difference.

czpython:
@stefanfoulis 
Correct. If a plugin depends on its parent for rendering then it wouldn't render correctly.
This is already the case though https://github.com/divio/django-cms/blob/3.4.1/cms/plugin_base.py#L125, I had changed this to reload the page for the undo/redo app but decided to change it back so anyone opting out of the undo/redo doesn't get the decreased performance from a reload.

PR #5837, opened 2017-01-05
Fixes a typo in the CMS_TOOLBAR_HIDE reference.

evildmp:
[ci skip]

### DIFF
--- a/docs/reference/configuration.rst
+++ b/docs/reference/configuration.rst
@@ -1014,7 +1014,8 @@ CMS_TOOLBAR_HIDE
 default
     ``False``
 
-If True, the toolbar is hidden in the pages out django CMS.
+If True, the toolbar is not displayed in pages that use the ``{% cms_toolbar %}`` template tag if they are not
+actually being handled by a django CMS view.
 
 .. versionchanged:: 3.2.1: CMS_APP_NAME has been removed as it's no longer required.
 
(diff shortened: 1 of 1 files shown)


docs/reference/configuration.rst
```diff
@@ -1014,7 +1014,8 @@ CMS_TOOLBAR_HIDE
 default
     ``False``
 
-If True, the toolbar is hidden in the pages out django CMS.
+If True, the toolbar is not displayed in pages that use the ``{% cms_toolbar %}`` template tag if they are not
+actually being handled by a django CMS view.
```
czpython: Maybe.. 

> If True, the toolbar is not displayed in non django CMS pages

evildmp: The we end up with weird things like "non-django CMS" (*non-django*? eh? which is simply incorrect) or *non-django-CMS* (which is unaesthetic).

docs/reference/configuration.rst
```diff
@@ -1014,10 +1014,15 @@ CMS_TOOLBAR_HIDE
 default
     ``False``
 
-If True, the toolbar is hidden in the pages out django CMS.
+By default, the django CMS toolbar is displayed to logged-in admin users on all pages that use the ``{% cms_toolbar
+%}`` template tag. Its appearance can be optionally restricted to django CMS pages only (technically, pages that are
+published by a django CMS view).
```
czpython: published seems a bit misleading.. maybe "**rendered** by a django CMS view"

PR #5845, opened 2017-01-10
Improved Page and Title extension documentation

### LINKED ISSUE #5083: Documentation bug in "Extending the page & title models"
In `docs/how_to/extending_page_title.rst` I read:

> Once you’ve registered your admin class, a new model will appear in the top- level admin list.

but since the method from the base class `PageExtensionAdmin.get_model_perms()` returns an empty dict, nothing appears in the top- level admin list.

### LINKED ISSUE #5092: Documentation Page and Title
The example provided for the "populate" method for Title extension is not correct:

```
def populate(self):
    # setup the extension toolbar with permissions and sanity checks
    current_page_menu = self._setup_extension_toolbar()
    # if it's all ok
    if current_page_menu and self.toolbar.edit_mode:
        # create a sub menu
        position = 0
        sub_menu = self._get_sub_menu(current_page_menu, 'submenu_label', 'Submenu', position)
        # retrieves the instances of the current title extension (if any) and the toolbar item URL
        urls = self.get_title_extension_admin()
        # cycle through the title list
        for title_extension, url in urls:
            # adds toolbar items
            sub_menu.add_modal_item('icon for title %s' % self._page().get_title(),
                                    url=url, disabled=not self.toolbar.edit_mode)
```

As there is no self._page(). (error 'HeroImageExtensionToolbar' object has no attribute '_page')

Also, the documentation does not include a template example when using Title instead of Page extension. I have resorted to using context processor to get it to show up in my template:

```
def add_hero(request):
    print request
    return {
        "heroimageextension" : request.current_page.get_title_obj_attribute("heroimageextension")
    }
```

Then in template

```
{% if heroimageextension %}
    {{ heroimageextension.hero_image.url }}
{% endif %}
```

But that is hardly ideal (having to have it in the settings etc.). Please provide a nicer way to do this!

evildmp:
* improved Page example
* corrected Title example so it actually works properly
* removed broken example from cms.toolbar.ExtensionToolbar comments
* bonus: removed out-of-date note about redo/undo from documentation home page

Fixes #5092
Fixes #5083

[ci skip]

### DIFF
--- a/docs/how_to/extending_page_title.rst
+++ b/docs/how_to/extending_page_title.rst
@@ -1,13 +1,10 @@
 #################################
-Extending the page & title models
+Extending the Page & Title models
 #################################
 
-.. versionadded:: 3.0
-
-You can extend the page and title models with your own fields (e.g. adding
-an icon for every page) by using the extension models:
-``cms.extensions.PageExtension`` and ``cms.extensions.TitleExtension``,
-respectively.
+You can extend the :class:`cms.models.Page` and :class:`cms.models.Title` models with your own fields (e.g. adding an
+icon for every page) by using the extension models: ``cms.extensions.PageExtension`` and
+``cms.extensions.TitleExtension``, respectively.
 
 
 ************************
@@ -15,15 +12,13 @@ Title vs Page extensions
 ************************
 
 The difference between a **page extension** and a **title extension** is related to the difference
-between the ``Page`` and ``Title`` models.
+between the :class:`cms.models.Page` and :class:`cms.models.Title` models.
 
-Titles support pages by providing a storage mechanism, amongst other things, for language-specific
-properties of ``Pages``. So, if you find that you need to extend the page model in a
-language-specific manner - for example, if you need to create language-specific keywords for each
-language of your pages - then you may need to use a ``TitleExtension``.
+* ``PageExtension``: use to add fields that should have **the same values** for the different language versions of a
+  page - for example, an icon.
+* ``TitleExtension``: use to add fields that should have **language-specific values** for different language versions
+  of a page - for example, keywords.
 
-On the other hand if the extension you'd like to create is the same for all of the different
-languages of the page, then you may be fine using a ``PageExtension``.
 
 ***************************
 Implement a basic extension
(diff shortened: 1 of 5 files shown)


docs/how_to/extending_page_title.rst
```diff
@@ -134,37 +131,113 @@ Here's a simple version for our example::
...
+
+                # we now also need to get the titleset (i.e. different language titles)
+                # for this page
+                page = self._get_page()
+                titleset = page.title_set.filter(language__in=get_language_list(page.site_id))
+
+                # create a 3-tuple of (title_extension, url, title)
+                nodes = [(title_extension, url, title.title) for (
+                    (title_extension, url), title) in zip(urls, titleset)
+                    ]
```
evildmp: This shouldn't be necessary! There ought to be a much nicer way to get the language versions' titles. I think it will need different behaviour from ``get_title_extension_admin()`` however - what do you think @czpython?

docs/reference/pages.rst
```diff
@@ -6,3 +6,6 @@ Models
 
     A ``Page`` is the basic unit of site structure in django CMS. The CMS uses a hierachical page model: each page
     stands in relation to other pages as parent, child or sibling.
+
+    A ``Page`` also has language-specific properties - for example, it will have a title and a slugfor each language it
```
czpython: and a **slug for**

docs/how_to/extending_page_title.rst
czpython: You can do:

```python
# retrieves the instances of the current title extension (if any)
# and the toolbar item URL
urls = self.get_title_extension_admin()

# cycle through the list of nodes
for title_extension, url in nodes:
    title = title_extension.extended_object
    # adds toolbar items
    sub_menu.add_modal_item('Rate %s' % title, url=url, disabled=not self.toolbar.edit_mode)
```

evildmp: That won't work unfortunately, except on titles that already have been extended. 

Otherwise, you'll get an attribute error when trying to get the ``extended_object`` of  ``None``.

czpython: Ok 😢

evildmp: Really, we need ``get_title_extension_admin()`` to return more information - but that's going to be a breaking change for many existing implementations that use it.

PR #5861, opened 2017-02-01
Fixed a bug where publish cancel wasn't respected

vxsx:
If the page is being published for the first time the confirmation
dialog had no effect on other event handlers, notably the one that
was actually publishing the page.


vxsx:
@FinalAngel please check again (now with tests)

CHANGELOG.txt
```diff
@@ -1,6 +1,6 @@
 === 3.4.3 (unreleased) ===
 
-<new-entries-start-here>
+* Fixed a bug when canceling publishing dialog wasn't respected
```
FinalAngel: I would say ``where canceling`` /cc @evildmp ?

czpython: Fixed a bug where canceling the publishing dialog wasn't respected.

PR #5873, opened 2017-02-08
Pinned intersphinx_mapping for Django to version 1.10

evildmp:
Occasionally, Django documentation references change; this will prevent unexpected breakage of links.

Also, we don't support Django's master branch, so we should refer to the latest supported release instead (currently 1.10).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,7 +33,7 @@ sys.path.append(os.path.join(os.path.abspath('.'), '_ext'))
 extensions = ['djangocms', 'sphinx.ext.intersphinx', 'sphinx.ext.todo', 'sphinx.ext.autodoc']
 intersphinx_mapping = {
     'python': ('http://docs.python.org/3/', None),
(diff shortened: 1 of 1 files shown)


docs/conf.py
```diff
@@ -33,7 +33,7 @@
 extensions = ['djangocms', 'sphinx.ext.intersphinx', 'sphinx.ext.todo', 'sphinx.ext.autodoc']
 intersphinx_mapping = {
     'python': ('http://docs.python.org/3/', None),
-    'django': ('http://docs.djangoproject.com/en/dev/', 'http://docs.djangoproject.com/en/dev/_objects/'),
+    'django': ('https://docs.djangoproject.com/en/1.10/', 'https://docs.djangoproject.com/en/1.10/_objects/'),
```
FinalAngel: can we pin it to the latest version? Otherwise we'd always have to manually change it

czpython: I don't think we can/should as we don't always support the latest django.
Instead we should just update this as we add support for newer versions.

PR #5534, opened 2016-07-11
Fixed #5492 - Use get_placeholder_conf fallback for PlaceholderModel.get_label

yakky:
PlaceholderModel.get_label did not implement the get_placeholder_conf fallback
Fix #5492

PR #5877, opened 2017-02-10
Fixed #5859 - Use get_placeholder_conf fallback for PlaceholderModel.get_label

### LINKED ISSUE #5859: CMS_PLACEHOLDER_CONF[‘template placeholder’]['name'] seems to be ignored in 3.4.x (worked in 3.3.x)
Hey there...
after updating to 3.4.2 from 3.3.x the follwing pattern doesnt result in displaying the given name in dragbar.html.

CMS_PLACEHOLDER_CONF[‘template placeholder’]: {
        'name': u 'some name',
 },

I couldn't find any recent changes causing this but follwing the request for the placeholder label it appears to me that the template name is not being passed as needed:

1. Render the dragbar
https://github.com/divio/django-cms/blob/release/3.4.x/cms/templates/cms/toolbar/dragbar.html#L27

2. Call for the placeholder label (no mention of the template!!!)
https://github.com/divio/django-cms/blob/release/3.4.x/cms/models/placeholdermodel.py#L71

3. Here the template name would be required but isn't passed in step 2.
https://github.com/divio/django-cms/blob/release/3.4.x/cms/utils/placeholder.py#L37

**After reverting back to 3.3.2 the get_label method passes the template name!!!!!**

```    
def get_label(self):
        from cms.utils.placeholder import get_placeholder_conf
        if self.page:
            template = self.page.get_template()
        else:
            template = None
        name = get_placeholder_conf("name", self.slot, template=template, default=title(self.slot))
        name = _(name)
        return name
```

**Somehow this fix got lost**
ISSUE: https://github.com/divio/django-cms/issues/5492

czpython:
Fixed #5859
Forward port of https://github.com/divio/django-cms/pull/5534

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -6,6 +6,7 @@
   CachedStaticFilesStorage or similar is used for STATICFILES_STORAGE
 * Changed the signature for internal ``cms.plugin_base.CMSPluginBase`` methods ``get_child_classes``
   and ``get_parent_classes`` to take an optional ``instance`` parameter.
+* Fixed error in retrieving placeholder label from configuration.
 
 
 === 3.4.2 (2017-01-23) ===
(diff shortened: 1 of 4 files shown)


czpython:
Merged manually in b252f53d1ab38cf2d5947ec9a703982464c306bf

PR #5878, opened 2017-02-14
Started complex plugin integration docs

evildmp:
[ci skip]

### Summary

Improves documentation for complex plugin integration.

### DIFF
--- a/docs/topics/complex_plugins.rst
+++ b/docs/topics/complex_plugins.rst
@@ -1,0 +1,79 @@
+..  _complex_plugins:
+
+
+##########################
+Complex plugin integration
+##########################
+
+****************************************
+Overview of how django CMS wraps plugins
+****************************************
+
+A django CMS plugin editing interface is fundamentally a Django admin interface for the model object in question,
+wrapped with some additional layers to present it in a way that's appropriate to the editing task and workflow.
+
+The wrapping looks a bit like this:
+
+* ``<div class="cms-modal-iframe>...</div>``, a modal dialog box, containing
+
+  * ``<iframe...>``, which contains
+
+    * ``<html>...</html>``, the actual Django admin page
+
+In order to present a consistent interface to the user, django CMS overrides some parts of the interface Django
+provides.
+
+For example, the Django admin provides a ``<div class="submit-row">`` for forms, that contains **Save** and **Delete**
+buttons. In a standard Django admin page, **Save** is actually an ``<input type="submit">`` and **Delete** is simply a
+link.
+
+django CMS hides the ``div.submit-row`` (with ``visibility: none``) and instead provides its own buttons: **Cancel**,
+**Delete** and **Save**.
+
+The plugin wrapper moves the functionality of Django buttons is moved from within the ``<html>...</html>`` to the
+``<div class="cms-modal-iframe>...</div>``.
+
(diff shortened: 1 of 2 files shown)


docs/topics/complex_plugins.rst
```diff
@@ -0,0 +1,79 @@
...
+The textarea is then saved in the normal way; if there are any plugins contained in it they will be saved to the
+database before the operation can finally be considered to have concluded successfully.
+
+.. todo: the saving of the plugins in the text editor is done in Python, not JS, correct?
+
+.. todo:
+
+    If the modal's Save invokes form's submit in such a way that it can successfully run the JS validation on form
+    elements, why does it not also invoke the JS that an editor plugin uses to transfer the editor content back to the
+    textarea? It's now not clear to me why it's necessary for the CKEditor to do this separately in
```
vxsx: the code you reference is not related to saving the TextPlugin in any way
this is for _children_ of the TextPlugin, which we display as ckeditor widgets versus doing what CMS normally does and reloading the page

docs/topics/complex_plugins.rst
```diff
@@ -0,0 +1,79 @@
...
+
+This plugin uses a Django form :class:`Textarea widget <django:django.forms.Textarea>`, which creates an HTML
+``<textarea>`` - *which it hides*. It then creates its own text-editing interface, the familiar CKEditor, and on
+**Save** parses the content in that (some of which may not be text content, such as plugins) and places it in the
+textarea.
+
+The textarea is then saved in the normal way; if there are any plugins contained in it they will be saved to the
+database before the operation can finally be considered to have concluded successfully.
+
+.. todo: the saving of the plugins in the text editor is done in Python, not JS, correct?
```
vxsx: yes it is.

the textarea value is cleaned up by ckeditor immediately before submission and then is parsed on the backend side. everything in between `<cms-plugin>` tags is removed and then in the live mode is replaced by appropriate plugin contents.

however the fun part starts when you create a child plugin in ckeditor but delete it before you save the TextPlugin. in this case the plugin was created but couldn't be removed - in that case the JS makes a separate request to delete that orphaned plugin

docs/topics/complex_plugins.rst
```diff
@@ -0,0 +1,79 @@
...
+
+How wrapping is implemented
+***************************
+
+In order to achieve this, django CMS refers the **Save** and **Delete** actions to the admin form.
+
+Hitting django CMS's **Save** button invokes the ``<input>`` on the admin form, as though it had been invoked in the
+browser by the user, by assigning its own **Save** button to the form's ``submit`` event.
+
+.. todo: Anything else we should note?
```
vxsx: probably shouldn't, but here are few more things we do:

the cms does some hidden things like this for example https://github.com/divio/django-cms/blob/release/3.4.x/cms/static/cms/js/modules/cms.modal.js#L820-L849

or for example there are certain conditions for the modal to be closed after "saving" that may seem non-obvious https://github.com/divio/django-cms/blob/release/3.4.x/cms/static/cms/js/modules/cms.modal.js#L972-L979

or that we inject ctrl+enter shortcut to submit the modal from the keyboard even if you're not inside the form field (cmd+enter for mac)

FinalAngel:
@evildmp I'm closing this for now. If it's still relevant please re-open with the necessary adaptions.

PR #5889, opened 2017-03-04
Improved Documentation on Navigation Modifiers

exploide:
### Summary

When I wrote a navigation modifier for the first time, I was inspired by the example from the documentation and I just had to adapt it to my needs. Unfortunately, this caused major performance issues in production because there were over 20000 executed database queries! I hadn't questioned the way the example modifier was written, so I ran into this issue. I spend a lot of time figuring out what exactly was going on and developed a way more efficient implementation of my modifier.

What was wrong with the documentation? At first, it started with
```python
# if the menu is not yet cut, don't do anything
if post_cut:
    return nodes
```
That does exactly the opposite of what the the comment is indicating. It's executed for the big uncut tree as well as the breadcrumb.

Afterwards, it executes
```python
for node in nodes:
    if node.attr["is_page"]:
        node.attr["changed_by"] = Page.objects.get(id=node.id).changed_by
```
That's a fresh database lookup for each and every node present in the tree.

No wonder that our page took 10 seconds to load. To be honest, we were doing more complex stuff with page and title extensions there. 😋 But using the same structure of how and when the QuerySet operations were performed.

To prevent others from running into this again, I rewrote the example in the documentation and noted a few tips about performance considerations. I hope you find this useful. :)

### DIFF
--- a/docs/how_to/menus.rst
+++ b/docs/how_to/menus.rst
@@ -179,7 +179,7 @@ To make your modifier available, it then needs to be registered with
 Now, when a page is loaded and the menu generated, your modifier will
 be able to inspect and modify its nodes.
 
-Here is an example of a simple modifier that places a Page's attribute in the corresponding
+Here is an example of a simple modifier that places the Page's ``changed_by`` attribute in the corresponding
 ``NavigationNode``::
 
     from menus.base import Modifier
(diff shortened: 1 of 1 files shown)


docs/how_to/menus.rst
```diff
@@ -179,31 +179,36 @@ To make your modifier available, it then needs to be registered with
 Now, when a page is loaded and the menu generated, your modifier will
 be able to inspect and modify its nodes.
 
-Here is an example of a simple modifier that places a Page's attribute in the corresponding
+Here is an example of a simple modifier that places the Page's ``changed_by`` attribute in the corresponding
```
evildmp: *the Page's* > *each Page's*

docs/how_to/menus.rst
```diff
@@ -179,31 +179,36 @@ To make your modifier available, it then needs to be registered with
...
-    class MyMode(Modifier):
+    class MyExampleModifier(Modifier):
         """
-
+        This modifier makes the changed_by attribute of a page
+        accessible for the menu system.
         """
         def modify(self, request, nodes, namespace, root_id, post_cut, breadcrumb):
-            # if the menu is not yet cut, don't do anything
+            # only do something when the menu is already cut
```
evildmp: *is already* > *has already been*

docs/how_to/menus.rst
```diff
@@ -231,7 +236,22 @@ of :class:`~menus.base.NavigationNode` instances.
   the case ``post_cut`` is ``True``.
 
 ``breadcrumb``
-  Is this breadcrumb call rather than a menu call?
+  Is this a breadcrumb call rather than a menu call?
+
+
```
evildmp: Remove additional newline

docs/how_to/menus.rst
```diff
@@ -231,7 +236,22 @@ of :class:`~menus.base.NavigationNode` instances.
...
+Note: Navigation modifiers can quickly become a performance bottleneck. Each modifier is called
```
evildmp: Please remove *Note:* and add a new heading for this section:
```
Performance issues in menu modifiers
====================================
```

Finally, please move the new *Performance issues in menu modifiers* section after the example of the modifier that marks all node levels.

exploide:
@evildmp thanks for your review! I incorporated the changes you requested :)

evildmp:
Thanks @exploide

PR #5898, opened 2017-03-16
Improved CMSApp.get_urls()

jrief:
### Summary

One would expect that it should also be possible to create an apphook using the ``get_urls()`` method such as:

```
class MyApp(CMSApp):
    name = _("My App")

    def get_urls(self, page=None, language=None, **kwargs):
        return [
            url(r'^$', MyListView.as_view()),
            url(r'^(?P<slug>[\w-]+)/?$', MyDetailView.as_view()),
        ]
```

instead of returning a list of urlpatterns in a file.

Looking at function ``cms.appresolver.get_app_urls()`` this behavior apparently was intended, but not yet documented. Also, the first ``yield`` statement returns a list, while the second returns a single URL pattern.

Now both ``yield`` statements return a list, as intended.

Fixes #5898

### DIFF
--- a/cms/appresolver.py
+++ b/cms/appresolver.py
@@ -164,7 +164,7 @@ def get_app_urls(urls):
                     "URLConf `%s` has no urlpatterns attribute" % urlconf)
             yield getattr(mod, 'urlpatterns')
         else:
-            yield urlconf
+            yield [urlconf]
 
 
 def get_patterns_for_title(path, title):
--- a/cms/appresolver.py
+++ b/cms/appresolver.py
@@ -163,6 +163,8 @@ def get_app_urls(urls):
                 raise ImproperlyConfigured(
                     "URLConf `%s` has no urlpatterns attribute" % urlconf)
             yield getattr(mod, 'urlpatterns')
+        elif isinstance(urlconf, (list, tuple)):
+            yield urlconf
         else:
             yield [urlconf]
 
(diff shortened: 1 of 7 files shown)


jrief:
@yakky If you use the legacy pattern, for instance

```
class MyApp(CMSApp):
    name = _("My App")
    url = ['myapp.urls.my_view']
```

with ``myapp/urls/my_view.py`` containing the url patterns from above, then in the patched function ``get_app_urls()`` only the first ``yield`` is ever executed. Hence that patch in my opinion is safe.
If you agree to merge, I'll add a note to the docs and inform about that alternative possibility.

jrief:
Please ignore the one failed unit test, there was a timeout.
For details look here: https://travis-ci.org/divio/django-cms/jobs/212422539

jrief:
@yakky while running some tests, the database died, so it's not a matter of failing unit tests:

```
_mysql_exceptions.OperationalError: (2006, 'MySQL server has gone away')
```

could you please restart the build or just ignore the failing tests.

jrief:
@yakky added another test, but imho that [yield statement](https://github.com/divio/django-cms/pull/5898/files#diff-f3785671e75d38ac4caccfa09ec8ec61L167) will never be reached anyway

czpython:
Hello @jrief,
Thanks for the contribution.
Please add some unit tests to the `test_apphooks.py` file.

evildmp:
Docs look good, however tests are required. Thanks!

jrief:
This feature now has its own unit test.
So hopefully it can be merged soon.

jrief:
@czpython since tests have been added, label **needs tests** can be removed.

yakky:
Thanks @jrief !

czpython:
Merged manually in c986ee7f6e782a84efdc55cec0333a9b4ebbe688

PR #5908, opened 2017-03-22
Fixed #5900 -- Don't assume a published page has public translation

### LINKED ISSUE #5900: Move pending page throws db error
When as a page is moved, if its moved under another page, the cms will check its publisher status, if the parent page is not published, then the cms will mark all translations for this page as pending.

There's a bug where the mark_as_pending logic assumes that a published draft translation has a public draft translation, normally this is the case but when the translation is pending, its possible for there not to be a public version of it.

czpython:
Fixed #5900

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -819,9 +819,11 @@ class Page(six.with_metaclass(PageMetaClass, MP_Node)):
         return True
 
     def mark_as_pending(self, language):
+        assert self.publisher_is_draft
+
         public = self.get_public_object()
 
-        if public:
+        if public and public.get_title_obj(language, fallback=False):
             state = public.get_publisher_state(language)
             # keep the same state
             # only set the page as unpublished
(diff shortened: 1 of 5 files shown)


cms/models/pagemodel.py
```diff
@@ -833,25 +835,21 @@ def mark_as_pending(self, language):
...
-        if draft and draft.is_published(language) and draft.get_publisher_state(
-                language) == PUBLISHER_STATE_DEFAULT:
+        if self.is_published(language) and self.get_publisher_state(language) == PUBLISHER_STATE_DEFAULT:
             # Only change the state if the draft page is published
             # and it's state is the default (0)
-            draft.set_publisher_state(language, state=PUBLISHER_STATE_PENDING)
+            self.set_publisher_state(language, state=PUBLISHER_STATE_PENDING)
 
     def mark_descendants_pending(self, language):
         assert self.publisher_is_draft
```
czpython: There's a lot of asserts like this for internal page methods.
I decided to leave them there because they prevent anyone using these internals from corrupting the db and also help when reading through the code to understand that these are only meant to work with draft (or public) pages.

PR #5912, opened 2017-03-24
Improved performance of publish operation

czpython:


czpython:
Closing as it will be done in 3.5.x

PR #5918, opened 2017-04-03
Don't override django's fieldset template

czpython:


FinalAngel:
@czpython are there no specific changes to that template from djangos?

PR #5920, opened 2017-04-04
Improved menu performance

czpython:
Fixes https://github.com/divio/django-cms/issues/5310
Fixes https://github.com/divio/django-cms/issues/5947

### DIFF
--- a/cms/cms_menus.py
+++ b/cms/cms_menus.py
@@ -151,9 +151,9 @@ def page_to_node(renderer, page, home, language):
     # Now finally, build the NavigationNode object and return it.
     ret_node = CMSNavigationNode(
         translation.menu_title or translation.title,
-        '',
-        page.pk,
-        parent_id,
+        url='',
+        id=page.pk,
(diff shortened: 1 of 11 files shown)


cms/cms_menus.py
```diff
@@ -145,65 +142,81 @@ def page_to_node(renderer, page, home, cut):
...
-    attr['redirect_url'] = page.get_redirect()  # save redirect URL if any
+    attr['redirect_url'] = translation.redirect  # save redirect URL if any
 
     # Now finally, build the NavigationNode object and return it.
-    ret_node = NavigationNode(
-        page.get_menu_title(),
-        page.get_absolute_url(),
+    ret_node = CMSNavigationNode(
+        translation.menu_title or translation.title,
+        '',
```
yakky: I feel uneasy about this change, but just for symmetry reasons

czpython: I do too :(
Unfortunately, even adding a lazy object causes a crazy speed decrease (about 1s).

PR #5929, opened 2017-04-13
Fixed an issue with JS not being able to determine correct path to async bundle

### LINKED ISSUE #5928: Issue with injected JS file (page select widget)
I noticed the page select widget wasn't showing any pages, and a look at the console showed two static paths merged into one.
I originally raised this as a djangocms-link [issue](https://github.com/divio/djangocms-link/issues/122) but was asked to raise here as it's a cms core issue.

Here's a screenshot to highlight the issue (top line, yellow highlight);

screen shot 2017-04-11 at 10 09 43

This only occurs when `DEBUG = False`

#### Environment

* Python version: 2.7
* Django version: 1.10.6
* django CMS version: 3.4.2

vxsx:
### Summary

Existing assumption was that filenames would never be rewritten, which
resulted in incorrect matches when filenames were hashed, e.g.
`bundle.forms.apphookselect.min.HASH.js` which never matched the
provided regex and resulted in CMS trying to load admin bundle from
incorrect place.

Fixes #5928

https://github.com/divio/djangocms-link/issues/122

Relax the expected filename string


marksweb:
@vxsx do you know when this might get released?

marksweb:
@vxsx do you know when this might get released?

czpython:
Hello @marksweb,
This will be released with 3.4.4 which if everything goes according to plan will be released this coming Thursday.

marksweb:
Wonderful @czpython!

That'll stop my client asking about when they'll be able to use the page select field :+1:

PR #5932, opened 2017-04-16
[3.5.x] Django 1.11 support

yakky:
### Summary

Enabled Django 1.11 compatibility

### TODO

* [x] Basic code running on 1.11
* [x] Tests running on 1.11
* [x] Widgets changes still on the way
* [x] Rendering errors in the wizard form
* [x] FE tests changes (?)

EDIT: 14.06.2017
- [x] Logout from toolbar doesn't work correctly (throws 405 response to a post request to logout)
- [x] since some of the dependencies are not 1.11 compatible (aldryn-bootstrap3 and django-filer) the decision was made to remove them from test requirements and to write our own plugin replacements. @czpython the plugins i need:
    - [x] should be similar to row + col plugins from that package - row plugin with child restriction and column plugin with parent restriction. it should be possible to create row plugin with such params so it would also create a number of columns immediately
    - [x] some kind of link plugin (using djangocms_link currently)


yakky:
@FinalAngel @vxsx there are some js issues in ``forms.apphookselect.js`` in advanced page form which I cannot debug due to my poor frontend skills. Basically the current js does not hide / shows fields according to the selected apphook

.travis.yml
```diff
@@ -135,27 +152,37 @@ matrix:
       env: FRONTEND=1 UNIT=1
     - python: 3.5
       env: FRONTEND=1 LINT=1
+    - python: 3.6
```
vxsx: frontend unit tests and lint aren't python dependent, no need to do them twice

yakky: Indeed, this is the exclude configuration :)

yakky:
@vxsx can't run the FE Tests locally to debug them (not even in plain 3.4.x branch)
BE wise this should be ready

yakky:
@vxsx @FinalAngel any news on the FE tests side?

test_requirements/django-1.11.txt
```diff
@@ -0,0 +1,7 @@
+-r requirements_base.txt
+Django>=1.11,<1.12
```
yakky: @timgraham ahem :blush:

FinalAngel:
@yakky let me check with @vxsx if we can squeeze it in this week

cms/templatetags/cms_tags.py
```diff
@@ -96,9 +95,10 @@ def _get_page_by_untyped_arg(page_lookup, request, site_id):
         if settings.DEBUG:
             raise Page.DoesNotExist(body)
         else:
+            mw = getattr(settings, 'MIDDLEWARE', getattr(settings, 'MIDDLEWARE_CLASSES'))
```
czpython: can you add a utility to `cms.utils.compat.dj` that returns the middleware classes?

cms/templatetags/cms_wizard_tags.py
```diff
@@ -11,6 +12,7 @@
 register = template.Library()
 
 
+@register.tag(name='cms_wizard')
```
czpython: was this template tag documented? If not then no need for deprecation cycle and can be removed.

cms/tests/test_cache.py
```diff
@@ -363,8 +365,10 @@ def test_expiration_cache_plugins(self):
             self.assertNotEqual(response['Cache-Control'], cache_control1)
             # However, the Expires timestamp will be the same
             self.assertEqual(response['Expires'], expires1)
-            # As will the Last-Modified timestamp.
-            self.assertEqual(response['Last-Modified'], last_modified1)
+            # No longer set / needed by Django 1.11
```
czpython: if this is no longer needed on 1.11, can we remove it all-together? Seems like this check was not necessary.

cms/tests/test_page_user_admin.py
```diff
@@ -393,7 +396,13 @@ def test_user_cant_change_self(self):
 
         with self.login_user_context(staff_user):
             response = self.client.post(endpoint, data)
-            self.assertEqual(response.status_code, 404)
+            if DJANGO_1_10:
+                self.assertEqual(response.status_code, 404)
+            else:
+                self.assertRedirects(response, redirect_to)
+                msgs = CookieStorage(response)._decode(response.cookies['messages'].value)
```
czpython: do we need to check for the message? I think the redirect should be enough.

cms/wizards/forms.py
```diff
@@ -43,6 +45,71 @@ def optional_fields(self):
...
+    if DJANGO_1_10:
+        class WizardOptionRenderer(forms.widgets.RadioFieldRenderer):
+            class WizardOptionInput(forms.widgets.RadioChoiceInput):
+
+                def __init__(self, name, value, attrs, choice, index):
+                    super(WizardOptionWidgets.WizardOptionRenderer.WizardOptionInput, self).__init__(
+                        name, value, attrs, choice, index
+                    )
+                    try:
+                        wizard = wizard_pool.get_entry(choice[0])
```
czpython: what would cause a value/key error here? is it necessary to check for this?

cms/wizards/forms.py
```diff
@@ -43,6 +45,71 @@ def optional_fields(self):
...
+                    self.outer_html,
+                    id_attr=format_html(' id="{}"', id_) if id_ else '',
+                    content=mark_safe('\n'.join(output)),
+                )
+
+        renderer = WizardOptionRenderer
+    template_name = 'cms/wizards/wizardoptionwidget.html'
+
+    def create_option(self, name, value, label, selected, index, subindex=None, attrs=None):
+        try:
```
czpython: what would cause a `ValueError` here?

.travis.yml
```diff
@@ -95,6 +101,26 @@ matrix:
...
     - python: 3.3
       env: TEST_DOCS=1 DJANGO=1.10 DATABASE_URL='sqlite://localhost/:memory:' MIGRATE_OPTION='--migrate'
+    - python: 3.4
+      env: TEST_DOCS=1 DJANGO=1.10 DATABASE_URL='sqlite://localhost/:memory:' MIGRATE_OPTION='--migrate'
+    - python: 3.5
+      env: TEST_DOCS=1 DJANGO=1.10 DATABASE_URL='sqlite://localhost/:memory:' MIGRATE_OPTION='--migrate'
+    - python: 2.7
+      env: TEST_DOCS=1 DJANGO=1.10 DATABASE_URL='sqlite://localhost/:memory:' MIGRATE_OPTION='--migrate'
+
+    - python: 3.3
```
yakky: To me we can limit a lot the tests on different python versions: focusing on 2.7/3.5 with limited matrix on other versions

cms/templatetags/cms_wizard_tags.py
yakky: Removing it in a patch release of an LTS looks *risky* to me, but I'm -0 on removing it

cms/tests/test_page_user_admin.py
yakky: The post might be redirected for different reasons, and it would be better to track it

cms/wizards/forms.py
yakky: I added this because of failures in the test suite (can't recall now exactly where, though)

cms/wizards/forms.py
yakky: See above :)

vinitkumar:
@yakky Is Django1.11 support only for 3.5.x and not 3.4.x?

FinalAngel:
@vinitkumar django CMS 3.4.5, which will be released soon, will have Django 1.11 support

vinitkumar:
@FinalAngel Okay, thanks 👍

alvein:
@vinitkumar any Scheduled date when 3.4.5 will be released?

vinitkumar:
@alvein I don't work with Divio, though I contacted them on the IRC and they said it's going to be released sometime this week.

alvein:
@vinitkumar, oh, sorry, want to ping @FinalAngel 
anyway, thank you for information and sorry for disturbing.

PR #5936, opened 2017-04-20
Add documentation RE email config

mcat-ee:
### Summary

This pull request adds some documentation that is present on the Divio support site but should be documented here as well (As there is no documentation present relating configuration the SMTP options for Django-CMS)

Adds an 'EMAIL_URL' section to the configuration documentation


czpython:
Hello @mcat-ee,
Thanks for the PR.

`EMAIL_URL` is not a django CMS setting.
Its specific to Divio cloud.

django CMS itself does no email sending other than django's broken link middleware.

PR #5945, opened 2017-04-21
Added commit message handler to not run all the tests if only docs were changed

vxsx:
correct way would be to check if the changes are only in .rst file, but this could be sufficient


vxsx:
https://travis-ci.org/divio/django-cms/builds/224770195 - here's an example of how it works

PR #5953, opened 2017-04-30
Update extending_page_title.rst (Handling relations)

haricot:
In the doc example of "Handling relations",  we need to change ManyToMany to  ManyToManyField  because that back these errors:
> AttributeError: module 'django.db.models' has no attribute 'ManyToMany'
> project.MyPageExtension.page_categories: (fields.W340) null has no effect on ManyToManyField.


czpython:
Thanks!

PR #5955, opened 2017-05-03
Enhanced copy/paste

vxsx:
It is now possible to paste plugin from a clipboard without reloading
the page after it was copied in a different tab. This also makes
clearing clipboard "optimistic" as it now clears it in the UI
immediately without waiting for ajax call to finish

This is currently open to `release/3.4.x` but it will be rebased to `develop` as soon as `develop` would be set to correct state by @czpython


czpython:
@vxsx Feel free to rebase and merge on develop branch.

PR #5959, opened 2017-05-04
Fixed #5871 -- Do not show plugins the user can't add

### LINKED ISSUE #5871: Plugins are displayed despite the user has no permission to add them
### Summary

In structure mode, if the user clicks `+` to add a plugin to a placeholder, all registered plugins are displayed. Even plugins which the user has no (add) permission for. 

Once clicked, the popup/iframe displays that the user has no permission to add this plugin. This is ok so it does not draw a security issue. 

### Expected behaviour

Do not show plugins the user has no permission to add, in the plugin list.

#### Environment

* django CMS version: 3.3.x and 3.4.2

haricot:
#5871

### DIFF
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -18,6 +18,8 @@ from cms.utils.conf import get_cms_setting, get_site_id
 from cms.utils.django_load import iterload_objects
 from cms.utils.placeholder import get_toolbar_plugin_struct, restore_sekizai_context
 
+from cms.utils.permissions import has_plugin_permission
+from functools import partial
 
 DEFAULT_PLUGIN_CONTEXT_PROCESSORS = (
     plugin_meta_context_processor,
@@ -439,10 +441,12 @@ class ContentRenderer(object):
         return content
 
     def render_editable_placeholder(self, placeholder, context, language):
+        can_add_plugin = partial(has_plugin_permission, user=self.request.user, permission_type='add')
+        plugins = [plugin for plugin in self.registered_plugins if can_add_plugin(plugin_type=plugin.value)]
         plugin_menu = get_toolbar_plugin_struct(
-            plugins=self.registered_plugins,
-            slot=placeholder.slot,
-            page=placeholder.page,
+        plugins=plugins,
+        slot=placeholder.slot,
+        page=placeholder.page,
         )
         new_context = {
             'plugin_menu': plugin_menu,
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -3,7 +3,6 @@ import warnings
 
 from collections import deque
 
-from cms.utils.permissions import has_plugin_permission
 from functools import partial
 
 from classytags.utils import flatten_context
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -1,4 +1,4 @@
-4# -*- coding: utf-8 -*-
+# -*- coding: utf-8 -*-
 import warnings
 
 from collections import deque
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -1,7 +1,10 @@
-# -*- coding: utf-8 -*-
+4# -*- coding: utf-8 -*-
 import warnings
 
 from collections import deque
+
+from cms.utils.permissions import has_plugin_permission
+from functools import partial
 
 from classytags.utils import flatten_context
 from django.template import Context
@@ -16,10 +19,9 @@ from cms.toolbar.utils import get_toolbar_from_request
 from cms.utils import get_language_from_request
 from cms.utils.conf import get_cms_setting, get_site_id
 from cms.utils.django_load import iterload_objects
-from cms.utils.placeholder import get_toolbar_plugin_struct, restore_sekizai_context
-
 from cms.utils.permissions import has_plugin_permission
-from functools import partial
+from cms.utils.placeholder import get_toolbar_plugin_struct, restore_sekizai_context
+ 
 
 DEFAULT_PLUGIN_CONTEXT_PROCESSORS = (
     plugin_meta_context_processor,
@@ -444,9 +446,9 @@ class ContentRenderer(object):
         can_add_plugin = partial(has_plugin_permission, user=self.request.user, permission_type='add')
         plugins = [plugin for plugin in self.registered_plugins if can_add_plugin(plugin_type=plugin.value)]
         plugin_menu = get_toolbar_plugin_struct(
-        plugins=plugins,
-        slot=placeholder.slot,
-        page=placeholder.page,
+            plugins=plugins,
+            slot=placeholder.slot,
+            page=placeholder.page,
         )
         new_context = {
             'plugin_menu': plugin_menu,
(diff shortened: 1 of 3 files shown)


czpython:
Hello @haricot,
Thanks for fixing this! Here is my feedback:

Because the ``get_toolbar_plugin_struct`` function takes a list of plugins and returns their toolbar code, I suggest to not do the permissions check here. The caller is the one responsible for passing the plugins that will actually be displayed. In this case the caller is [render_editable_placeholder](https://github.com/divio/django-cms/blob/release/3.4.x/cms/plugin_rendering.py#L441).

Here it's much easier to check for permissions because you have the user object attached to the renderer, thus no need to use the thread user.

So you can modify the ``render_editable_placeholder`` function like so:

```python
plugins = [plugin for plugin in self.registered_plugins if has_plugin_permission(self.request.user, plugin.value, 'add')]
plugin_menu = get_toolbar_plugin_struct(
    plugins=plugins,
    slot=placeholder.slot,
    page=placeholder.page,
)
```

To avoid calling has_plugin_permission out of scope, you could use functools.partial so:

```python
from functools import partial

can_add_plugin = partial(has_plugin_permission, user=self.request.user, permission_type='add')
plugins = [plugin for plugin in self.registered_plugins if can_add_plugin(plugin_type=plugin.value)]
plugin_menu = get_toolbar_plugin_struct(
    plugins=plugins,
    slot=placeholder.slot,
    page=placeholder.page,
)
```

haricot:
Many thanks @czpython ! :)  It is very instructive. It works well ! 
 
Small reminders regarding the configuration of the rights of the plugins:
For static_placeholder, we must configure them in the site administration interface> Authentication and Authorization> group or user.
On the other hand for {% placeholder "foo"%}, in superuser we must also think of adding in the site administration> DJANGOCMS> Pages global permissions.> rights to a group or user.
Then we have to reload the page manually just after the rights configuration and this will not show the plugins if user does not have the permissions. 
Otherwise the official doc is fine.[ permissions](http://docs.django-cms.org/en/release-3.4.x/topics/permissions.html)

cms/utils/placeholder.py
```diff
@@ -108,7 +107,7 @@ def get_toolbar_plugin_struct(plugins, slot=None, page=None):
 
     # plugin.value points to the class name of the plugin
     # It's added on registration. TIL.
-    for plugin in plugins:
+    for plugin in plugins:
```
czpython: extra space

cms/utils/placeholder.py
```diff
@@ -17,7 +17,6 @@
 from cms.exceptions import DuplicatePlaceholderWarning
 from cms.utils import get_cms_setting
 
-
```
czpython: No need to remove blank line.
Two blank lines should be after the last import.

cms/plugin_rendering.py
```diff
@@ -439,10 +441,12 @@ def render_editable_plugin(self, instance, context, plugin_class,
         return content
 
     def render_editable_placeholder(self, placeholder, context, language):
+        can_add_plugin = partial(has_plugin_permission, user=self.request.user, permission_type='add')
+        plugins = [plugin for plugin in self.registered_plugins if can_add_plugin(plugin_type=plugin.value)]
         plugin_menu = get_toolbar_plugin_struct(
-            plugins=self.registered_plugins,
-            slot=placeholder.slot,
-            page=placeholder.page,
+        plugins=plugins,
```
czpython: please indent these

cms/plugin_rendering.py
```diff
@@ -18,6 +18,8 @@
 from cms.utils.django_load import iterload_objects
 from cms.utils.placeholder import get_toolbar_plugin_struct, restore_sekizai_context
 
+from cms.utils.permissions import has_plugin_permission
```
czpython: please add this import under ``from cms.utils.django_load import iterload_objects`` to keep imports in alphabetical order.

cms/plugin_rendering.py
```diff
@@ -18,6 +18,8 @@
...
+from functools import partial
```
czpython: as above please add this import under ``from collections import deque`` to separate top-level packages.

haricot:
Thank you for your constructive remarks!
I hope this is the proper implementation of the test.
Sorry for the Travis test repeatedly, I should probably test it locally and in a branch that does not bind to the PR.

haricot:
I'm not sure if adding a specific user with a single permission in test_plugins.py is a good idea on my part. ^^

cms/tests/test_plugins.py
```diff
@@ -1228,6 +1234,51 @@ def test_plugin_toolbar_struct(self):
             self.assertEqual(force_text(style_config['module']), expected_struct_de['module'])
             self.assertEqual(force_text(style_config['name']), expected_struct_de['name'])
 
+    def test_plugin_toolbar_struct_permissions(self):
```
czpython: The test needs to be done differently because the permission check is done in the rendering call, so we need to test the function output relative to the placeholder rendering.

Here's what I came up with:

```python
from django.template.context import Context

...


def test_plugin_toolbar_struct_permissions(self):
    page = self.get_permissions_test_page()
    staff_user = self.get_staff_user_with_no_permissions()
    placeholder = page.placeholders.get(slot='body')
    page_url = page.get_absolute_url() + '?' + get_cms_setting('CMS_TOOLBAR_URL__EDIT_ON')
    
    self.add_permission(staff_user, 'change_page')
    self.add_permission(staff_user, 'add_text')

    with self.login_user_context(staff_user):
        request = self.get_request(page_url, page=page)
        request.session['cms_edit'] = True
        request.toolbar = CMSToolbar(request)
        content_renderer = self.get_content_renderer(request=request)
        context = {'request': request, 'cms_content_renderer': content_renderer}
        output = content_renderer.render_placeholder(
            placeholder,
            context=Context(context),
            language='en',
            page=page,
            editable=True,
        )
        self.assertIn('<a data-rel="add" href="TextPlugin">Text</a>', output)
        self.assertNotIn('<a data-rel="add" href="LinkPlugin">Link</a>', output)
```

haricot:
The test function `render_placeholder` had crossed my mind, it's much clearer now.
I hope this is not a problem, i had to change  :
`from django.template.context import Context` 
to 
`from sekizai.context import SekizaiContext`

Because that would return this error :
`django.template.exceptions.TemplateSyntaxError: You must enable the 'sekizai.context_processors.sekizai' template context processor or use 'sekizai.context.SekizaiContext' to render your templates.
`

Nevertheless the context processor 'sekizai' look well loaded :

settings.TEMPLATES[0]['OPTIONS']['context_processors'] 
```
['django.contrib.auth.context_processors.auth', 'django.contrib.messages.context_processors.messages', 'django.template.context_processors.i18n', 'django.template.context_processors.debug', 'django.template.context_processors.request', 'django.template.context_processors.media', 'django.template.context_processors.csrf', 'cms.context_processors.cms_settings', 'sekizai.context_processors.sekizai', 'django.template.context_processors.static'
```]

Thanks again for your help, you did everything, I hope to do better next time.

czpython:
Thanks @haricot!

PR #5966, opened 2017-05-09
Update templatetags.rst

paramono:
### Summary

Replace {% cms_placeholder %} with {% placeholder %}


paramono:
Thanks for accepting, @czpython! This is a very minor fix, but it certainly saves documentation users like me from confusion.

There's another thing I want to point to. Docs claim that you can place `static_placeholder` anywhere in your template, although in reality, it must be somewhere below `cms_toolbar`. I opened an issue #5967 related to that. I want to make a contribution to docs with this issue in mind, but first I would like to have a comment from maintainers who better understand the internals of Django CMS.

Can you please comment on the issue?

PR #5980, opened 2017-06-08
Content / Structure split

czpython:
The main goal of this change is to introduce reload-less operations.
We've separated the content rendering from the structure mode rendering to selectively update parts of the content / structure markup.

Additions:

* `CMSToolbar` subclasses can now define a `Media` class or `media` property to render toolbar specific media.
* Introduced ``TemplatesCache`` class to cache templates per request in case the cached template loader is not used.

Changes:

* The toolbar `build` GET parameter has been renamed to `structure`.
* The `data` parameter in `PluginMenuItem` is now optional, defaults to `None`.
* ``CMSPluginBase.get_extra_placeholder_menu_items `` is now a `classmethod`.
* ``request.toolbar.edit_mode`` is now ``request.toolbar.edit_mode_active``
* ``CMSToolbar.edit_mode`` is now ``CMSToolbar.edit_mode_active``
* The `plugin_meta_context_processor` plugin context processor has been removed.
  If needed, this can easily be re-integrated using the plugin context processors setting.

Removed internals:

* ``render_placeholder_toolbar_js`` template tag.
* ``render_plugin_toolbar_config`` template tag, use ``render_plugin_init_js `` instead.
* ``cms/templates/cms/toolbar/placeholder.html``
* ``cms_content_renderer`` context variable.
* ``constants.PLUGIN_MOVE_ACTION``
* ``constants.PLUGIN_COPY_ACTION ``
* ``Placeholder.render``
* ``CMSPluginBase.get_action_options``
* ``ContentRenderer.drag_item_template``
* ``ContentRenderer.drag_item_menu_template ``
* ``ContentRenderer.dragbar_template ``
* ``CMSToolbar.render_addons``
* ``CMSToolbar.post_template_render_addons``
* ``CMSToolbar.request_hook``

Removed as part of deprecation:

* ``CMSPlugin.render_plugin``
* ``PlaceholderAdminMixin.post_add_plugin``
* ``PlaceholderAdminMixin.post_copy_plugins``
* ``PlaceholderAdminMixin.post_edit_plugin``
* ``PlaceholderAdminMixin.post_move_plugin``
* ``PlaceholderAdminMixin.post_delete_plugin``
* ``PlaceholderAdminMixin.post_clear_placeholder``

Gotchas:

* Adding/Editing/Removing/Moving a plugin no longer reloads the page.
* Calling `self.request.path` or `self.request.path_info` inside of a `CMSToolbar` subclass method (a custom toolbar) is no longer encouraged and can lead to issues.
This is because the toolbar is loaded via AJAX when something changes on the page, as a result
both `request.path` and `request.path_info` will point to the path of the AJAX endpoint instead of the path of the page. To get the correct path, call `self.toolbar.request_path` instead.


cms/static/cms/js/modules/cms.base.js
```diff
@@ -324,6 +333,7 @@ CMS.API.Helpers = {
      * @param {String[]} [params] array of `param=value` strings to update the url
      * @returns {String}
      */
+    // TODO get rid of this
```
FinalAngel: what do you want to replace it with?

cms/templates/admin/cms/page/close_frame.html
```diff
@@ -19,13 +19,11 @@
...
         // the dataBridge is used to access plugin information from different resources
         // Do NOT move this!!!
-        Window.CMS.API.Helpers.dataBridge = {
-            'plugin_id': '{{ plugin_id|unlocalize }}',
-            'plugin_type': '{% filter escapejs %}{{ type }}{% endfilter %}',
-            'plugin_name': '{% filter escapejs %}{{ name }}{% endfilter %}',
-            'plugin_icon': '{% filter escapejs %}{{ icon }}{% endfilter %}',
-            'plugin_desc': '{% filter escapejs %}{{ alt }}{% endfilter %}'
-        };
+        Window.CMS.API.Helpers.dataBridge = {{ plugin_data|safe }};
```
FinalAngel: |safe might cause problems if not correctly treated in the JS, are we sure this will not allow js injecting?

cms/static/cms/js/modules/cms.base.js
vxsx: this PR introducts URL.js for proper url handling, should use that instead of our own

vxsx: we should also remove the `once`, `debounce` and `throttle` as we now have lodash and lodash provides those.

cms/templates/admin/cms/page/close_frame.html
FinalAngel: @czpython can you confirm that this is okay?

czpython: @vxsx @FinalAngel I've switched it to use escapejs instead.

czpython: @vxsx @FinalAngel reverted the change and instead I'm now manually escaping the `plugin_desc`.

PR #5981, opened 2017-06-09
fix 5828: check for missing i18n context processor

### LINKED ISSUE #5828: Create page wizard on new site works inconsistently
When launching a new django CMS project, the content creation wizard **under some circumstances:**

* **does not** work when invoked from the *Add the first page* link on the splash screen
* **does not** work if it opens automatically
* **does** work if you use the **Create** button in the toolbar.

When it fails, it says: "Please choose an option from below to proceed to the next step.".

screenshot 2016-12-27 16 45 49.

I can replicate this on a manually set up project running Django 1.9.12 and django CMS 3.4.1, but not on a project set up using the django CMS installer.

briancappello:
### Summary

Update manual how-to install doc and `cms check` script to require the `django.template.context_processors.i18n` context processor.

Fixes #5828

### DIFF
--- a/cms/tests/test_check.py
+++ b/cms/tests/test_check.py
@@ -61,13 +61,22 @@ class CheckTests(CheckAssertMixin, TestCase):
 
     def test_no_cms_settings_context_processor(self):
         override = {'TEMPLATES': deepcopy(settings.TEMPLATES)}
-        override['TEMPLATES'][0]['OPTIONS']['context_processors'] = ['sekizai.context_processors.sekizai']
+        override['TEMPLATES'][0]['OPTIONS']['context_processors'] = ['django.template.context_processors.i18n',
+                                                                     'sekizai.context_processors.sekizai']
+        with self.settings(**override):
+            self.assertCheck(False, errors=1)
+
+    def test_no_i18n_context_processor(self):
+        override = {'TEMPLATES': deepcopy(settings.TEMPLATES)}
+        override['TEMPLATES'][0]['OPTIONS']['context_processors'] = ['cms.context_processors.cms_settings',
+                                                                     'sekizai.context_processors.sekizai']
         with self.settings(**override):
             self.assertCheck(False, errors=1)
 
     def test_no_sekizai_template_context_processor(self):
         override = {'TEMPLATES': deepcopy(settings.TEMPLATES)}
-        override['TEMPLATES'][0]['OPTIONS']['context_processors'] = ['cms.context_processors.cms_settings']
+        override['TEMPLATES'][0]['OPTIONS']['context_processors'] = ['cms.context_processors.cms_settings',
+                                                                     'django.template.context_processors.i18n']
         with self.settings(**override):
             self.assertCheck(False, errors=2)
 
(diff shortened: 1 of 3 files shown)


czpython:
Hello @briancappello,
Thanks for this fix.
The cms itself does not require i18n to function, so I prefer if we remain free this dependency.

Instead I suggest to rebase against the `develop` branch and modify the `_render_welcome_page` function inside ``cms/page_rendering.py`` to add a context variable called `cms_language` whose value comes from calling `get_language_from_request` which you can import from `cms.utils`

briancappello:
@czpython 
Okay, makes sense. I just started playing around with this cms earlier this week so you'll have to forgive me for not being very familiar with it :)

I did try rebasing against `develop`, but it failed with conflicts unrelated to my changes. If I'm understanding you correctly, it sounds like you're suggesting a new change instead of a modification to the commit in this branch, so maybe it would be easier if I just opened a new PR? 

The change you suggest seems simple enough, however I'm at a loss as to how to actually test its effects, or even what the implications of the change are / how it would fix the referenced bug. (Both `python manage.py runserver` and `python testserver.py` crash with exceptions? Or do I need to install the locally modified cms app code into a separate django project to actually run it?)

czpython:
@briancappello
No worries, glad you're giving this a shot :)

I agree with your suggestion, its easier to just create a new branch/pr from develop.

In regards to running tests, I recommend going through our [guide](http://docs.django-cms.org/en/release-3.4.x/contributing/testing.html#running-tests) and once tests are passing for develop, you can start with the fix.

The fix has two parts, one is adding the `cms_language` context variable when `welcome.html` is rendered but the other part (and root issue) is that the wizard doesn't handle an invalid language correctly. The language is passed via the `language` GET parameter and if its invalid, the wizard fails with this cryptic "Select a step" error. So I suggest to replace:

https://github.com/divio/django-cms/blob/release/3.4.x/cms/wizards/views.py#L11
with `from cms.utils import get_language_from_request`

https://github.com/divio/django-cms/blob/release/3.4.x/cms/wizards/views.py#L88
with `kwargs['wizard_language'] = get_language_from_request(self.request)`

https://github.com/divio/django-cms/blob/release/3.4.x/cms/wizards/views.py#L97
with `get_language_from_request(self.request)`

the test itself can be a call POST request to the cms_wizard enpoint with an invalid language and asserts that the page was created correctly.

edevil:
Did this stall?

I encountered this same problem the first time I started django CMS, having used the manual install. It does not make a good first impression. :)

Shall I give it a go?

briancappello:
@edevil That would much appreciated!

edevil:
I can see that the development version has changed this part of the code. Can it be that this has already been solved for the next release?

czpython:
Hello @edevil,
We have not addressed this yet.
If you would like to tackle it, please use the develop branch as base.

czpython:
@edevil oops.. so we **did** fix this in **develop** but not on 3.4.x. So please use release/3.4.x as base branch. The fix on develop was part of a larger refactor, as a result we didn't backport it.

czpython:
Closed by https://github.com/divio/django-cms/pull/6284

PR #5985, opened 2017-06-13
Added slug generation to the wizard as well

vxsx:
### Summary
The "add page" in the pagetree has slugifying directly in the frontend, wizards for creating pages don't

Add slugifying to wizard as well.


cms/forms/wizards.py
```diff
@@ -21,6 +21,7 @@
     user_can_add_page,
     user_can_add_subpage,
 )
+from cms.utils.urlutils import static_with_version
```
czpython: move this under the cms.utils.conf import

PR #5964, opened 2017-05-07
Fixed #5740 -- Use the correct url when ManifestStaticFilesStorage or similar is used

sephii:
Fixed #5740.

Issue #5907, opened 2017-03-22
Static file loading fails with cached template loader


stefanfoulis:
The change in https://github.com/divio/django-cms/commit/8fabb5c3ef8b9283b535ea1b76f029c51d3e9c29#diff-a0493e27f48b50e18312b9f4508fc29d breaks static file loading if ``django.template.loaders.cached.Loader`` is used.
It appears that the path leaks between requests, adding the django-cms version to the path another time for every request. It results in urls like this:
``/static/cms/css/3.4.2/3.4.2/3.4.2/3.4.2/3.4.2/3.4.2/3.4.2/3.4.2/3.4.2/3.4.2/3.4.2/cms.base.css``

affected:
release/3.4.x branch (not been released on pypi yet)

PR #5993, opened 2017-06-19
Fixes #5740 -- Use the correct url when ManifestStaticFilesStorage or similar is used

### LINKED ISSUE #5740: The file 'cms/css/cms.base.css' could not be found with CompressedManifestStaticFilesStorage
I am stuck with this issues since yesterday, and cannot find a solution. 

I use django-cookiecutter and integrated my django-cms app. Everything works fine, except when I open pages that have django-cms CSS. I get the following error:

```
ValueError: The file 'cms/css/cms.base.css' could not be found with <whitenoise.storage.CompressedManifestStaticFilesStorage object at 0x7f238b1fa610>.
  File "django/core/handlers/exception.py", line 39, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 217, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 215, in _get_response
    response = response.render()
  File "django/template/response.py", line 109, in render
    self.content = self.rendered_content
  File "django/template/response.py", line 86, in rendered_content
    content = template.render(context, self._request)
  File "django/template/backends/django.py", line 66, in render
    return self.template.render(context)
  File "opbeat/instrumentation/packages/base.py", line 63, in __call__
    args, kwargs)
  File "opbeat/instrumentation/packages/base.py", line 222, in call_if_sampling
    return self.call(module, method, wrapped, instance, args, kwargs)
  File "opbeat/instrumentation/packages/django/template.py", line 18, in call
    return wrapped(*args, **kwargs)
  File "django/template/base.py", line 208, in render
    return self._render(context)
  File "django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 994, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 961, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 174, in render
    return compiled_parent._render(context)
  File "django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 994, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 961, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 174, in render
    return compiled_parent._render(context)
  File "django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 994, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 961, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 174, in render
    return compiled_parent._render(context)
  File "django/template/base.py", line 199, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 994, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 961, in render_annotated
    return self.render(context)
  File "django/template/loader_tags.py", line 70, in render
    result = block.nodelist.render(context)
  File "django/template/base.py", line 994, in render
    bit = node.render_annotated(context)
  File "django/template/base.py", line 961, in render_annotated
    return self.render(context)
  File "django/templatetags/static.py", line 104, in render
    url = self.url(context)
  File "cms/templatetags/cms_static.py", line 31, in url
    url = super(StaticWithVersionNode, self).url(context)
  File "django/templatetags/static.py", line 101, in url
    return self.handle_simple(path)
  File "django/templatetags/static.py", line 114, in handle_simple
    return staticfiles_storage.url(path)
  File "django/contrib/staticfiles/storage.py", line 132, in url
    hashed_name = self.stored_name(clean_name)
  File "django/contrib/staticfiles/storage.py", line 292, in stored_name
    cache_name = self.clean_name(self.hashed_name(name))
  File "django/contrib/staticfiles/storage.py", line 95, in hashed_name
    (clean_name, self))
```

Colelctstatic correctly copied the file and whitenoise even `Post-processed` the file:

```
django_1    | Copying '/usr/local/lib/python2.7/site-packages/cms/static/cms/css/3.4.1/cms.base.css'
django_1    | Post-processed 'cms/css/3.4.1/cms.base.css' as 'cms/css/3.4.1/cms.base.1784b3b6c6ab.css'
```

Here are some settings, where I believe they might have an impact:

```
MIDDLEWARE = (
    'django.middleware.security.SecurityMiddleware',
    'cms.middleware.utils.ApphookReloadMiddleware',
    'django.contrib.sessions.middleware.SessionMiddleware',
    'django.middleware.csrf.CsrfViewMiddleware',
    'django.contrib.auth.middleware.AuthenticationMiddleware',
    'django.contrib.messages.middleware.MessageMiddleware',
    'django.middleware.clickjacking.XFrameOptionsMiddleware',
    'django.middleware.locale.LocaleMiddleware',
    'django.middleware.common.CommonMiddleware',
    'cms.middleware.user.CurrentUserMiddleware',
    'cms.middleware.page.CurrentPageMiddleware',
    'cms.middleware.toolbar.ToolbarMiddleware',
    'cms.middleware.language.LanguageCookieMiddleware',
    # Your stuff goes here (save in config.middlewares)
    'config.middlewares.RequireLoginMiddleware',
)
# STATIC FILE CONFIGURATION
# ------------------------------------------------------------------------------
# See: https://docs.djangoproject.com/en/dev/ref/settings/#static-root
STATIC_ROOT = str(ROOT_DIR('staticfiles'))

# See: https://docs.djangoproject.com/en/dev/ref/settings/#static-url
STATIC_URL = '/static/'

# See: https://docs.djangoproject.com/en/dev/ref/contrib/staticfiles/#std:setting-STATICFILES_DIRS
STATICFILES_DIRS = (
    str(APPS_DIR.path('static')),

)

WHITENOISE_MIDDLEWARE = ('whitenoise.middleware.WhiteNoiseMiddleware', )
MIDDLEWARE = WHITENOISE_MIDDLEWARE + MIDDLEWARE
STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
```

Anyone any idea? Any other info you need?

Pankrat:
### Summary

Fixes #5740

This just adds a test for the changes proposed by @sephii in https://github.com/divio/django-cms/pull/5964. I've mocked the staticfiles storage in order to avoid having to collect static files in the test environment.

I tried adding a test for #5907 as well but I could never reproduce the described problem (neither in a test nor in production).

### Alternatives

If the goal of the static files versioning is just cache busting (which is already solved in the `ManifestStaticFilesStorage`), one could also replicate the static files without the version prefix and let `collectstatic` pick it up. This wouldn't require any code changes but a replication of the static files in two locations (with and without the version prefix).

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -28,6 +28,8 @@
 * Fixed a bug in which the placeholder cache was not consistently cleared when a page was published.
 * Enhanced the plugin menu to not show plugins the user does not have permission to add.
 * Fixed a regression which prevented users from setting a redirect to the homepage.
+* Fixed a ``ValueError`` raised when using ``ManifestStaticFilesStorage`` or similar for static files.
+  This only affects Django >= 1.10
 
 
 === 3.4.3 (2017-04-24) ===
--- a/cms/templatetags/cms_static.py
+++ b/cms/templatetags/cms_static.py
@@ -28,5 +28,7 @@ def do_static_with_version(parser, token):
 class StaticWithVersionNode(StaticNode):
 
     def url(self, context):
-        url = super(StaticWithVersionNode, self).url(context)
-        return static_with_version(url)
+        path = self.path.resolve(context)
+        path_with_version = static_with_version(path)
+
+        return self.handle_simple(path_with_version)
--- a/cms/tests/test_templatetags.py
+++ b/cms/tests/test_templatetags.py
@@ -10,12 +10,15 @@ from django.contrib.sites.models import Site
 from django.core import mail
 from django.core.exceptions import ImproperlyConfigured
 from django.test import RequestFactory
+from django.test.utils import override_settings
 from django.utils.html import escape
 from django.utils.timezone import now
 from djangocms_text_ckeditor.cms_plugins import TextPlugin
 
 from sekizai.data import UniqueSequence
 from sekizai.helpers import get_varname
+
+from mock import patch
 
 import cms
 from cms.api import create_page, create_title, add_plugin
@@ -99,6 +102,25 @@ class TemplatetagTests(CMSTestCase):
 
         output = self.render_template_obj(template, {}, None)
         self.assertEqual(expected, output)
+
+    @override_settings(STATICFILES_STORAGE='django.contrib.staticfiles.storage.ManifestStaticFilesStorage')
+    @patch('django.contrib.staticfiles.storage.staticfiles_storage')
+    def test_static_with_version_manifest(self, mock_storage):
+        """
+        Check that static files are looked up at the location where they are
+        stored when using static file manifests.
+        """
+        mock_storage.url.side_effect = lambda x: '/static/' + x
+        expected = 'cms/css/%(version)s/cms.base.css'
+        expected = expected % {'version': cms.__version__}
+
+        template = (
+            """{% load cms_static %}<script src="{% static_with_version "cms/css/cms.base.css" %}" """
+            """type="text/javascript"></script>"""
+        )
+
+        self.render_template_obj(template, {}, None)
+        mock_storage.url.assert_called_with(expected)
 
 
 class TemplatetagDatabaseTests(TwoPagesFixture, CMSTestCase):
--- a/cms/tests/test_templatetags.py
+++ b/cms/tests/test_templatetags.py
@@ -111,16 +111,22 @@ class TemplatetagTests(CMSTestCase):
         stored when using static file manifests.
         """
         mock_storage.url.side_effect = lambda x: '/static/' + x
-        expected = 'cms/css/%(version)s/cms.base.css'
-        expected = expected % {'version': cms.__version__}
 
         template = (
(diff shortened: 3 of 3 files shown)


czpython:
Hello @Pankrat,
Thanks for the test.
I've tried it locally but it passes even without the patch.

Pankrat:
Hi @czpython 

Thanks for taking the time to check that. 

I've changed the test to check that the storage is called with the correct parameters and made sure the test fails without the patch (checked against Django 1.10).

czpython:
@Pankrat looks like the test fails with Django < 1.10

Pankrat:
@czpython Looks like the static files storage is only invoked for this template tag on Django 1.10 and above. I had no better idea than making the test more permissive. It should now pass with or without the patch on Django <1.10 but will require the patch on Django >=1.10.

This means that #5740 should only affect Django 1.10 and above.

Pankrat:
@czpython Is there something that still needs to be done on this PR from your point of view?

czpython:
Thanks for the follow-up @Pankrat.
I've updated the changelog and will land afterwards.

PR #5999, opened 2017-06-22
Removed automatic home-page logic

czpython:
Most of the line changes come from adaptations to existing tests which assumed that the first page created through any means will automatically become the homepage.
This assumption is no longer valid, only pages created through the admin or the wizard will become the homepage **IF** they are the first page to be created.

The new logic is much faster than the existing one because it no longer saves each title for each page when the homepage is set to update its path.
Instead it makes use of sql `CONCATENATE` and `SUBSTRING` to alter the path accordingly.

## Summary
* Introduced a new page burger menu item called "Set home"
* Removed the logic that selected a homepage automatically on save
* Refactored plugin copying into a new function
* Removed some page and plugin signal handlers
* Publishing a page will no longer call save() on the draft page

The following todos will be done in upcoming prs:

* Release note changes
* Tests for new logic


cms/admin/pageadmin.py
```diff
@@ -667,6 +676,48 @@ def change_template(self, request, object_id):
...
     @transaction.atomic
+    def set_home(self, request, object_id):
+        page = get_object_or_404(
+            self.model,
+            pk=object_id,
+            publisher_is_draft=True,
+        )
+
+        if not self.has_change_permission(request, page):
+            message = _("You do not have permission to set 'home'")
```
FinalAngel: missing . at the end of the sentence

czpython: I'll address this on a separate pr.

PR #6007, opened 2017-07-02
supported django versions

alskgj:
### Summary

changed docs to reflect that django'cms supports django 1.9 and 1.10 in addition to django 1.8


czpython:
Thanks!

PR #6015, opened 2017-07-07
Switch md5 for sha1 for caching hash function

shaunbrady:
FIPS[0] is not a support crytographic function on FIPS compliant
systems, and is thus disabled.  Moving md5 hashing to use sha256, we can
use django-cms on FIPS compliant systems.

Test updated to include new function output.

[0] http://csrc.nist.gov/groups/STM/cmvp/documents/140-1/140val-all.htm

### Summary

On FIPS enabled systems, md5() functions are not available.  Moving to sha256, django-cms can be used.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,7 @@
   was rendered before attempting to inherit content from parent pages.
 * Changed page/placeholder cache keys to use sha1 hash instead of md5 to be FIPS compliant.
 
+
 === 3.4.4 (2017-06-15) ===
 
 * Fixed a bug in which cancelling the publishing dialog wasn't respected.
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -9,7 +9,7 @@
   is cached in an ancestor page.
 * Fixed a regression where the code following a ``{% placeholder x or %}`` declaration,
   was rendered before attempting to inherit content from parent pages.
-
+* Changed page/placeholder cache keys to use sha1 hash instead of md5 to be FIPS compliant.
 
 === 3.4.4 (2017-06-15) ===
 
--- a/cms/cache/page.py
+++ b/cms/cache/page.py
@@ -17,11 +17,11 @@ from cms.utils.helpers import get_timezone_name
 
 
 def _page_cache_key(request):
-    #md5 key of current path
+    #sha1 key of current path
     cache_key = "%s:%d:%s" % (
         get_cms_setting("CACHE_PREFIX"),
         settings.SITE_ID,
-        hashlib.md5(iri_to_uri(request.get_full_path()).encode('utf-8')).hexdigest()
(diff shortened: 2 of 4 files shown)


czpython:
Hello @shaunbrady,
Thanks for the fix.
I'll try to get to it by the end of the week.

shaunbrady:
No problem!  Let me know if I can answer any questions or rework any thing.

czpython:
@shaunbrady @yakky 
What do you think about `SHA1`? (security is not a problem here). We use it for the wizard.
My concern with SHA256 is that it is considerably larger in size than md5 and the whole idea behind hashing in this case was to make the cache key smaller.

stefanfoulis:
Comparisons

```
echo "123" | md5
ba1f2511fc30423bdbb183fe33f3dd0f
echo "123" | shasum -a 1
a8fdc205a9f19cc1c7507a60c4f01b13d11d7fd0  -
echo "123" | shasum -a 256
181210f8f9c779c26da1d9b2075bde0127302ee0e3fca38c9a83f5b1dd8e5d3b  -
```

shaunbrady:
@yakky @czpython I went with sha256 because while both sha1 and sha256 are FIPS compliant, sha1 has demonstrated collisions against it.  I very much realize the application here is not one of security, but should the FIPS standard be updated, I suspect we'll be back here in short order.  I'm fine with sha1 or sha256.  Let me know.  Thanks!

PS:  I've said on similar PRs to other Django projects that I wish one could pass in to hashlib functions an "I'm not using this for security" flag to distinguish such use cases as this, but as it stands, the compliance requires the entire library to be disabled.

czpython:
I suggest to use sha1, feels like the least invasive on a patch release (3.4.5).
I'll take out some of those hash calls on 3.5.

shaunbrady:
Cool, I'll push up a new PR tonight.  We'd rather see this sooner than later, as we're forked currently, so I'm especially interested if we can get on a patch release.

Thanks!

stefanfoulis:
I'd imagine that ``sha1`` is faster than ``sha256`` (just guessing, I don't actually know). Since speed is a high concern here, I also prefer ``sha1``.

stefanfoulis:
``sha1`` is even a little bit faster than ``md5`` :-)
``sha256`` is slower.
http://atodorov.org/blog/2013/02/05/performance-test-md5-sha1-sha256-sha512/

shaunbrady:
See?  I'm making your guys code faster :)  Ignore the pedantic compliance issue.  I have tests passing.  I'm going to just force push to this same branch like a rebel.

shaunbrady:
Force pushed and renamed PR.  Let me know if this works.  Thanks!

czpython:
Thanks @shaunbrady,
Please add an entry to the bottom of the 3.4.5 `CHANGELOG` section.
Something like "Changed cache key hashing to use sha1 instead of md5 to be FIPS compliant."

shaunbrady:
@czpython Added a new commit.

czpython:
Thanks!

PR #6034, opened 2017-08-01
Fixes #5989 Adds redirect loop check

### LINKED ISSUE #5989: Creating infinite loop of redirection, gives "0 error" but publishes the page
### Steps to reproduce:
1. Have 2 page
2. Make page Alpha redirect Betta
3. Publish the changes on Alpha
4. Make page Betta redirect Alpha
5. Publish the changes on Betta
6. Press "view published"

### Actual result:
After publishing Betta - the "0 error" is shown. But the changes are published and "ERR_TOO_MANY_REDIRECTS" error page is shown
[video proof](https://www.screencast.com/t/RzYm6AZa)

### Expected result:
1. Make "0 error" message gives more information, for better understanding the cause of the problem
2. Restrict page from publishing

SteinRobert:
### Summary

 - adds redirect loop check util function
 - adds check to AdvancedSettingsForm

I'd like to have some feedback if the approach is fine or if there is a better way of doing the check. If the approach is fine I'll add some tests as well.

Fixes #5989 

#5989 

Added a util - title.py to cms.utils. This module contains a function which is able to detect redirect loops, no matter if it is a redirect of a page to itself or a redirect through various other pages creating a redirect loop.

### DIFF
--- a/cms/tests/test_page_admin.py
+++ b/cms/tests/test_page_admin.py
@@ -366,7 +366,9 @@ class PageTest(PageTestBase):
         with self.login_user_context(superuser):
             data['redirect'] = '/'
             response = self.client.post(endpoint, data)
-            self.assertRedirects(response, redirect_to)
+            validation_error = \
+                '<ul class="errorlist"><li>This redirect configuration may create an infinite redirect loop.</li></ul>'
+            self.assertContains(response, validation_error, html=True)
 
         with self.login_user_context(superuser):
             data['redirect'] = '/hello'
--- a/cms/utils/title.py
+++ b/cms/utils/title.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from cms.models import Title
+
+
+def has_redirect_loop(title, language, redirect_path):
+    redirect_path = redirect_path.strip('/')
+    try:
+        redirected_titles = Title.objects.filter(language=language, path=redirect_path)
+    except Title.DoesNotExist:
+        return False
+    if title.pk in redirected_titles.values_list('pk', flat=True):
+        return True
+    else:
+        for redirected_title in redirected_titles:
+            if redirected_title.redirect:
+                if has_redirect_loop(title, language, redirected_title.redirect):
+                    return True
+        return False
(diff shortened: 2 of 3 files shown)


cms/utils/title.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from cms.models import Title
+
+
+def has_redirect_loop(title, language, redirect_path):
```
czpython: I think the recursive check isn't necessary.
We only care about the title being modified, thus the check can be reduced to the something like:

```python
if redirect and redirect == self.cleaned_data['path']:
    raise ValidationError("You can't redirect to the same url.")
```

in the forms.py

There is still a condition where a loop can be created, redirect to a page whose title of same language redirects back to this one, but lets address that on a separate pull request with its own test.

cms/tests/test_page_admin.py
```diff
@@ -366,7 +366,9 @@ def test_page_redirect_field_validation(self):
         with self.login_user_context(superuser):
             data['redirect'] = '/'
             response = self.client.post(endpoint, data)
-            self.assertRedirects(response, redirect_to)
+            validation_error = \
```
czpython: please move the check to its own test

cms/utils/title.py
SteinRobert: Thank you very much for the feedback! Why do you think a recursive check isn't necessary? It would resolve the addressed issue and even more complicated ones (A -> B -> C -> A).

czpython: I want to tackle the first scenario (redirect to the same page) in this pr and the second scenario (redirect to a page which redirects to current one) in a separate pr, I would like to prevent redirecting to a page with a redirect set, this would avoid having to do the recursive check. I can't think of many use-cases where is needed to have such complex redirect tree.

FinalAngel:
@SteinRobert would you be able to reopen this pr against develop and address the issues mentioned above? Otherwise I will close the PR for now.

FinalAngel:
Closing this for now @SteinRobert, please reopen against develop with the suggested changes. Thank you very much :)

PR #6040, opened 2017-08-08
sort the Sites menu

fp4code:
### Summary

Sort the Sites menu

An unsorted Sites menu is non usable if number of SItes is > 5.


czpython:
Hello @fp4code,
Thanks for your contribution.
To keep consistency, please add the sorting to the `get_user_sites_queryset` function instead.

czpython:
Never mind my last comment.
I think its good where it is, tests are missing though.

czpython:
Fixed on 3.5.

PR #6043, opened 2017-08-13
Fixes #6027 - Removes prevent_descendant_update flag

### LINKED ISSUE #6027: Slug change doesn't propagate to all child pages
### Summary

When changing the slug of the page - not all of it's children update their path.
I create pages like this:
```
- Homepage (completely unrelated)
- page-a
     - page-b
        - page-c
```
So the slugs are:
```
/en/page-a
/en/page-a/page-b
/en/page-a/page-b/page-c
```

After that I publish _every_ page.
After that I change slug of "Page A" to "expected" and publish it.

### Expected behaviour
The resulting pages slugs are:
```
/en/expected
/en/expected/page-b
/en/expected/page-b/page-c
```

#### Actual behaviour
```
/en/expected
/en/expected/page-b
/en/page-a/page-b/page-c
```
If there are children to page-c - their paths aren't updated either.
Also this might be relevant:
```
>>> from cms.models import Page
>>> page = Page.objects.get(pk=18)
>>> page.title_set.values("path")
<QuerySet [{'path': 'page-a/page-b/page-c'}]>
>>> page.publisher_public.title_set.values("path")
<QuerySet [{'path': 'expected/page-b/page-c'}]>
```

#### Environment

* Python version: 3.5
* Django version: 1.10
* django CMS version: 3.4.4 (actually 3.5)

SteinRobert:
### Summary

 - removes prevent_descendant_update flag and functionality
 - post_save_title is now triggered recursively and updates paths properly
 - modifies `test_rename_node` to test grandchild path as well

Fixes #6027 

#6027 

Removed the `prevent_descendant_update` flag and logic which prevented path update from propagating. Seems like this was introduced in commit 89f0974a26ea663deb68d69e40f71b5cdefd62f6. As far as I see there is no other use of this flagging mechanism in the current code.  Also modified a test which now checks grandchild path update in addition to chid path update.

### DIFF
--- a/cms/tests/test_page.py
+++ b/cms/tests/test_page.py
@@ -1044,6 +1044,25 @@ class PageTreeTests(CMSTestCase):
         parent.publish('en')
         child = create_page('child', 'nav_playground.html', 'en', slug='child', published=True, parent=parent)
         child.publish('en')
+
+        page_title = Title.objects.get(page=parent)
+        page_title.slug = "father"
+        page_title.save()
+
+        parent = Page.objects.get(pk=parent.pk)
+        parent.publish('en')
+        child = Page.objects.get(pk=child.pk)
+
+        self.assertEqual(child.get_absolute_url(language='en'), '/en/father/child/')
+        self.assertEqual(child.publisher_public.get_absolute_url(language='en'), '/en/father/child/')
+
+    def test_rename_node_alters_descendants(self):
+        home = create_page('grandpa', 'nav_playground.html', 'en', slug='home', published=True)
+        home.publish('en')
+        parent = create_page('parent', 'nav_playground.html', 'en', slug='parent', published=True)
+        parent.publish('en')
+        child = create_page('child', 'nav_playground.html', 'en', slug='child', published=True, parent=parent)
+        child.publish('en')
         grandchild = create_page('grandchild', 'nav_playground.html', 'en', slug='grandchild', published=True,
                                  parent=child)
         grandchild.publish('en')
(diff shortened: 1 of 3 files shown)


cms/tests/test_page.py
```diff
@@ -1044,6 +1044,9 @@ def test_rename_node(self):
         parent.publish('en')
         child = create_page('child', 'nav_playground.html', 'en', slug='child', published=True, parent=parent)
         child.publish('en')
+        grandchild = create_page('grandchild', 'nav_playground.html', 'en', slug='grandchild', published=True,
```
czpython: Please move the grandchild test to its own method.
Something like: `test_rename_node_alters_descendants`

SteinRobert:
Updated the tests and added the changelog entry.

cms/tests/test_page.py
```diff
@@ -1058,27 +1058,25 @@ def test_rename_node(self):
...
-        parent = Page.objects.get(pk=parent.pk)
-        parent.publish('en')
-        child = Page.objects.get(pk=child.pk)
-
-        self.assertEqual(child.get_absolute_url(language='en'), '/en/father/child/')
-        self.assertEqual(child.publisher_public.get_absolute_url(language='en'), '/en/father/child/')
-        self.assertEqual(grandchild.get_absolute_url(language='en'), '/en/father/child/grandchild/')
-        self.assertEqual(grandchild.publisher_public.get_absolute_url(language='en'), '/en/father/child/grandchild/')
+        self.assertEqual(grandchild_1.get_absolute_url(language='en'), '/en/father/child/grandchild-1/')
+        self.assertEqual(grandchild_1.publisher_public.get_absolute_url(language='en'), '/en/father/child/grandchild-1/')
```
SteinRobert: If this pages aren't published this test won't work.

czpython: I removed it because we're passing `published=True` to the api.. strange that it fails, likely is the title cache. I'll check it in a bit.

czpython:
Thanks!

PR #6046, opened 2017-08-14
Solves #6045 by not throwing an exception when an invalid ip address …

### LINKED ISSUE #6045: 500 is generated when an invalid IP address is provided
### Summary

When a request contains an invalid IP Address in the headers, for example:

`HTTP_X_REAL_IP: '209.141.53.577'`

And `settings.INTERNAL_IPS` uses `IpRange`, then

`cms.middleware.toolbar.ToolbarMiddleware.is_cms_request` invokes `IpRange.__contains__` (the moment it asks `if client_ip not in internal_ips:`

### Expected behaviour

The invalid IP address is not contained among the `IpRange`, and thus `cms.middleware.toolbar.ToolbarMiddleware.is_cms_request` returns `False`

#### Actual behaviour

The moment `IpRange.__contains__` indirectly invokes `IpRange._cast`, an exception is raised:

`TypeError("expected ip address, 32-bit integer or 128-bit integer")`

#### Environment

* Python version: any
* Django version: any
* django CMS version: any

#### Side note:

Even though you do not control the settings, the use of IpRange, it makes sense that IpRange throws an exception and it makes sense to use that package in the settings. And thus, the location where the validation is done, should take that into consideration. That is the reason I sent the patch to you and not someone else.

lgandras:
…is specified

### Summary

Fixes #6045

### DIFF
--- a/cms/middleware/toolbar.py
+++ b/cms/middleware/toolbar.py
@@ -2,7 +2,9 @@
 """
 Edit Toolbar middleware
 """
+from django import forms
 from django.contrib.admin.models import LogEntry, ADDITION, CHANGE
+from django.core.exceptions import ValidationError
 from django.core.urlresolvers import resolve
 from django.http import HttpResponse
 
@@ -27,8 +29,13 @@ class ToolbarMiddleware(MiddlewareMixin):
 
         if internal_ips:
             client_ip = get_request_ip(request)
-            if client_ip not in internal_ips:
+            try:
+                client_ip = forms.GenericIPAddressField().clean(client_ip)
+            except ValidationError:
                 return False
+            else:
+                if client_ip not in internal_ips:
+                    return False
 
         if not toolbar_hide:
             return True


cms/middleware/toolbar.py
```diff
@@ -27,7 +27,11 @@ def is_cms_request(self, request):
 
         if internal_ips:
             client_ip = get_request_ip(request)
-            if client_ip not in internal_ips:
+            try:
+                if client_ip not in internal_ips:
+                    return False
+            except TypeError:
```
czpython: Please use `forms.GenericIPAddressField` to clean/validate the ip value.
Something like:

```python
try:
    client_ip = forms.GenericIPAddressField().clean(client_ip)
except ValidationError:
    return False
else:
    return client_ip in internal_ips
```

lgandras:
The patch makes now a type validation and afterwards checks if the IP is allowed, as requested.

czpython:
Thanks @lgandras.
Because we're about to release 3.4.5, the core is currently on freeze.
I'll land the pr afterwards (~2 weeks).

czpython:
Thanks!

PR #6048, opened 2017-08-18
Updated authorship, copyright information

evildmp:


setup.py
```diff
@@ -34,8 +34,8 @@
 ]
 
 setup(
-    author='Patrick Lauber',
-    author_email='digi@treepy.com',
+    author='Divio AG and contributors',
+    author_email='info@divio.ch',
```
vxsx: `.com` ?

FinalAngel: @vxsx not setup in this case

PR #6072, opened 2017-09-18
Support Content-Security-Policy without `script-src 'unsafe-inline'`

### LINKED ISSUE #5381: Support Content-Security-Policy without "script-src unsafe-inline"
Django 1.10 lands support for running the admin with a Content-Security-Policy without "script-src unsafe-inline" applied.

Django CMS should follow suit and remove all inline JavaScript. Replacing large blocks of JS with `<script src={% static 'path/to/old_inlines/eg.js' %}></script>` and `<script>var foo = {{bar}}></script>` with:
```js
# path/to/fooUsage.js
window.foo = $('#django-cms-foo').data('foo');
```
```html
<script id='django-cms-foo' src={% static 'path/to/fooUsuage' %} data-foo={{ bar }}></script>
```

rspeed:
### Summary

This PR removes inline scripts which trigger errors when using a [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) without `script-src 'unsafe-inline'`.

Fixes #5381

This PR does the following:

* Converts inline data-carrying scripts (such as CMS.config) into JSON structures.
* Modifies scripts to load data from JSON structures.
* Moves inline logic scripts to script files, or removes them outright.
* Misc minor changes to support the above.
* Fixes an unrelated bug in toolbars.js that caused errors (CMS wasn't being imported).


rspeed:
Please note that this *does not* completely eliminate `unsafe-inline` from the CSP. Django-CMS uses an external script from jQuery called `PEP.js` which dynamically generates an inline script block, which violates a CSP without `script-src 'unsafe-inline'`. This could potentially be fixed in-place, as the script is already forked for a different reason.

rspeed:
There are also two places where the JSON is being built by hand in templates, which may lead to maintenance issues in the future:

1. [cms/toolbar/toolbar_javascript.html](/rspeed/django-cms-contrib/blob/c11ef35c990a2332b4d182652d1160e16fc84e2f/cms/templates/cms/toolbar/toolbar_javascript.html)
1. [cms/toolbar/plugin.html](/rspeed/django-cms-contrib/blob/c11ef35c990a2332b4d182652d1160e16fc84e2f/cms/templates/cms/toolbar/plugin.html)

The structures should probably be built in the views and passed through `json.dumps` instead, but I felt it wasn't necessary for this PR.

FinalAngel:
@vxsx is this still relevant, could you have a look? @rspeed and thank you for the contribution. Would you be able to rebase to develop depending on @vxsx feedback?

vxsx:
still relevant, and looks good apart from conflicts

FinalAngel:
@rspeed would be great to get the conflicts resolved and we'll merge it into the next release :)

FinalAngel:
@rspeed we'd be happy to merge your work. Please resolve the conflicts against PR and open it again and I'll make sure someone gets it in. Thank you very much!

PR #6076, opened 2017-09-20
Fixed #6074 -- Added a destroy method for the Plugin javascript class

### LINKED ISSUE #6074: Destroy CMS.Plugin instances
### Summary

At moment we can create new `CMS.Plugin` instances in runtime but it's not clear how to destroy them removing all DOM nodes and all the events creaded with the Plugin.

### Expected behaviour

I expect that a plugin can be created and destroyed for example:
```js
const plugin = new CMS.Plugin(name, options)

// remove DOM and events without side effects
plugin.destroy()
```

### Actual behaviour

A plugin can only be created

### Environment

* Python version: 3.5.4
* Django version: 1.10.7
* django CMS version: 3.4.4

p.s.
[This method](https://github.com/divio/django-cms/blob/release/3.4.x/cms/static/cms/js/modules/cms.plugins.js#L1935-L1939) could also return an array of the plugins created using `$.map` instead of `$.each`

GianlucaGuarini:
This feature is not tested because the integration tests seem to be broken on the `develop` branch 

### Summary

- Add the `uid` helper to properly identify any UI Element
- Create aliases to the wrapped `$(window)` and `$(document)` objects
- Add the `destroy` method to the Plugin class removing DOM and events

Fixes #6074

https://github.com/divio/django-cms/issues/6074

See https://github.com/divio/django-cms/issues/6074

### DIFF
--- a/cms/static/cms/js/modules/cms.plugins.js
+++ b/cms/static/cms/js/modules/cms.plugins.js
@@ -58,6 +58,9 @@ var Plugin = new Class({
         }
     },
 
+    // these properties will be filled later
+    modal: null,
+
     initialize: function initialize(container, options) {
         this.options = $.extend(true, {}, this.options, options);
 
@@ -814,7 +817,12 @@ var Plugin = new Class({
     destroy() {
         // close the plugin modal if it was open
         // TODO: shouldn't the modal be destroyed as well at this point?
-        this.ui.modal.close();
+        if (this.modal) {
+            this.modal.close();
+            // unsubscribe to all the modal events
+            this.modal.off();
+        }
+
         // remove all the plugin UI DOM elements
         // notice that $.remove will remove also all the ui specific events
         // previously attached to them
@@ -1118,8 +1126,8 @@ var Plugin = new Class({
             var isPlaceholder = !dragItem.length;
             var childrenList;
 
-            // make sure also the modal is part of the ui collection
-            this.ui.modal = modal = new Modal({
+            // make sure also the modal is available to this instance
+            this.modal = modal = new Modal({
                 minWidth: 400,
                 minHeight: 400
             });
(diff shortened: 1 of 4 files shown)


GianlucaGuarini:
I have added some basic unit tests but I wasn't able to add integration tests. Please let me know if it's everything fine

GianlucaGuarini:
@vxsx I have rebased against `develop`, added the `.cleanup` method and added a bool param to the `.destroy` method to eventually automatically remove the DOM nodes. I am not able to run the integration tests here though. Please let me know if I can do anything else for you

vxsx:
@GianlucaGuarini it's ok, the integration tests will run on travis
could you change `destroy(mustCleanup)` to `destroy({ mustCleanup })`? would prefer to avoid single boolean params. Also github says there are conflicts in cms.plugins.js (that might be my fault, sorry) - could you rebase again?

GianlucaGuarini:
@vxsx done.. now it should be also in sync with `develop`

GianlucaGuarini:
The unit tests fails again after the merge but I don't know why. 😢

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -811,10 +811,10 @@ var Plugin = new Class({
      * Destroys the current plugin instance removing only the DOM listeners
      *
      * @method destroy
-     * @param {Boolean} mustCleanup - if true it will remove also the plugin UI components from the DOM
+     * @param {Boolean} options.mustCleanup - if true it will remove also the plugin UI components from the DOM
      * @returns {void}
      */
-    destroy(mustCleanup) {
+    destroy({ mustCleanup }) {
```
vxsx: should be`{ mustCleanup = false } = {}`, the unit tests fail when object isn't passed

GianlucaGuarini: ah right i will fix it asap (probably tomorrow) sorry.. I am overloaded of things to do

GianlucaGuarini: ok now it should be fine and the tests are green

vxsx:
Thanks a lot

PR #6079, opened 2017-09-26
Added note about get_absolute_url or get_success_url on Wizard classes

marksweb:
### Summary

I've added a note about `get_success_url` on `Wizard` classes if the model doesn't define a `get_absolute_url` or that method may not return anything. I implemented a `Wizard` in accordance with the docs & the `get_absolute_url` for my model didn't return anything causing an error in the CMS.

Rather than raise an issue, I have simply added a note to the wizard docs which should cover any ambiguity on this.


docs/how_to/wizards.rst
```diff
@@ -49,6 +49,11 @@ do this for ``MyApp``, it might look like this::
 
     wizard_pool.register(my_app_wizard)
 
+.. note::
+
+    if your model doesn't define a ``get_absolute_url`` function then your wizard
```
evildmp: Capitalise *If*.

evildmp: Thanks @marksweb, I've left a number of comments, this seems to be one of those simple-looking fixes that uncovers a number of related issues. Would you care to tackle them?

marksweb: Yeah I'll have a look at it when I get a minute.

On a side note, I remember making that 'if' lowercase as it was consistent with the existing note I checked for reference, so I'll take a look over that to ensure consistency & sentence case.

evildmp:
Thanks @marksweb I will take a look.

evildmp:
Merged in https://github.com/divio/django-cms/commit/fff36adc3e6ae371200696e39ed6a4ba276846c2, thanks for your contribution @marksweb!

PR #6086, opened 2017-09-29
General docs fixes

### LINKED ISSUE #5738: Fix NavigationNode.attr documentation
The part of the documentation about the `attr` attribute of the `NavigationNode` class says that there should be a key `navigation_extenders` in the `NavigationNode.attr` dict. According to the docs, `navigation_extenders` is a list containing navigation extenders **including AppHooks**.

I was trying to find my apphooks in there, but they do not appear. On IRC @evildmp concluded that this is probably a documentation error, as AppHooks should not appear there.

### LINKED ISSUE #5880: ShowMenu template tag - namespace parameter seems to be ignored.
### Summary
The template tag `show_menu` does not seem to take the parameter `namespace` into consideration.

### Expected behaviour
The template tag should cut off all menu nodes, which do not have the namespace of the parameter `namespace`.

#### Actual behaviour
The template tag `show_menu` simply shows all menu nodes and does not cut off the nodes with namespace which differ from the parameter `namespace` passed to the template tag. 

#### Environment

* Python version: 2.7
* Django version: 1.8.16
* django CMS version: 3.4.2

### LINKED ISSUE #5967: Static placeholders inside <head>
I thought it's a good idea to create a static placeholder in `<head>` so that users could easily add snippets to it (for example, Google Analytics code). The docs say it's possible:

>The `static_placeholder` template tag can be used **anywhere** in any template and is not bound to any page or model.

However, if I add `static_placeholder` in `<head>`, it becomes unusable ("add plugin" and "menu" buttons for this placeholder do nothing at all). Further investigation has shown that `cms_toolbar` is the culprit here:

>The `cms_toolbar` template tag has to be placed after the `<body>` tag and before any {% cms_placeholder %} occurrences within your HTML

What is the recommended way to add user-editable Snippets to `<head>`?

### Summary

`static_placeholder` (as well as `placeholder`) is only usable inside `<body>`. If `static_placeholder` is added to head, "add plugin" and "menu" buttons for this placeholder won't work, since placeholder is located before `cms_toolbar`. Is it possible to make `static_placeholder` editable in `<head>`, or is there any other way to add Snippets inside `<head>`?

Docs for static_placeholder template tag should probably be updated, since it's clear it cannot be used "anywhere" and needs to be placed some place after `cms_toolbar`. 

#### Environment

* Python version: 3.5.2
* Django version: 1.10.6
* django CMS version: 3.4.3

evildmp:
Addresses various documentation issues:

* Added notes on placeholder & cms_toolbar tags: Fixes #5967
* Added note on show_menu namespace parameter:  Fixes #5880
* Removed incorrect note from navigation docs: Fixes #5738

### DIFF
--- a/docs/reference/templatetags.rst
+++ b/docs/reference/templatetags.rst
@@ -15,10 +15,14 @@ top of your template::
 
     {% load cms_tags %}
 
+
+Placeholders
+============
+
 ..  templatetag:: placeholder
 
 placeholder
-===========
+-----------
 
 .. versionchanged:: 2.1
 
@@ -67,17 +71,29 @@ pages have plugins that generate content::
 See also the :setting:`CMS_PLACEHOLDER_CONF` setting where you can also add extra
 context variables and change some other placeholder behaviour.
 
+..  important::
+
+    ``{% placeholder %}`` will only work inside the template's ``<body>``.
+
+
 ..  templatetag:: static_placeholder
 
 static_placeholder
-==================
+------------------
+
 .. versionadded:: 3.0
 
-The static_placeholder template tag can be used anywhere in any template and is not bound to any
-page or model. It needs a name and it will create a placeholder that you can fill with plugins
-afterwards. The static_placeholder tag is normally used to display the same content on multiple
-locations or inside of apphooks or other third party apps. Static_placeholder need to be published to
-show up on live pages.
+The ``{% static_placeholder %}`` template tag can be used anywhere in a template element after
+the ``{% cms_toolbar %}`` tag. A static placeholder instance is not bound to any particular page
+or model - in other words, everywhere it appears, a static placeholder will hold exactly the same
+content.
+
+The ``{% static_placeholder %}`` tag is normally used to display the same content on multiple
+locations or inside of apphooks or other third party apps.
+
+Otherwise, a static placeholder behaves like a "normal" placeholder, to which plugins can be added.
+
+A static placeholder needs to be published to show up on live pages, and requires a name.
 
 Example::
 
@@ -88,7 +104,7 @@ Example::
 
 .. warning::
 
-    Static_placeholders are not included in the undo/redo and page history pages
+    Static placeholders are not included in the undo/redo and page history pages
 
 If you want additional content to be displayed in case the static placeholder is
 empty, use the ``or`` argument and an additional ``{% endstatic_placeholder %}``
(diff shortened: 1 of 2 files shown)


docs/reference/templatetags.rst
```diff
@@ -70,14 +74,21 @@ context variables and change some other placeholder behaviour.
...
+the ``{% cms_toolbar %}`` tag. A static placeholder instance is not bound to any particular page
+or model - in other words, everywhere it appears, a static placeholder will hold exactly the same
+content.
+
+The ``{% static_placeholder %}`` tag is normally used to display the same content on multiple
+locations or inside of apphooks or other third party apps.
+
+Otherwise, a static placeholder behaves like a "normal" placeholder, to which plugins can be added.
+
+A static placeholder need to be published to show up on live pages, and requires a name.
```
czpython: **needs**

docs/reference/templatetags.rst
```diff
@@ -851,9 +873,18 @@ Example::
...
-.. note::
+..  note::
 
-    Be aware that you can not surround the cms_toolbar tag with block tags.
+    Be aware that you cannot surround the ``cms_toolbar`` tag with block tags.
     The toolbar tag will render everything below it to collect all plugins and placeholders, before
     it renders itself. Block tags interfere with this.
 
+
+..  warning::
```
czpython: this warning is also above, is it ok to repeat it?

docs/reference/templatetags.rst
```diff
@@ -112,12 +123,22 @@ Example::
 Note that the `Django "sites" framework <https://docs.djangoproject.com/en/dev/ref/contrib/sites/>`_ *is* required and
 ``SITE_ID`` *must* be set in ``settings.py`` for this (not to mention other aspects of django CMS) to work correctly.
 
+..  warning::
+
+    Placeholders may be added inside the template's ``<head>``. There are very few cases where you
```
FinalAngel: I can't imagine one, a specific template tag might be better? @vxsx AFAIK we wouldn't it be a problem to place cms_toolbar in between the body and header?

vxsx: template tag is _always_ better, and placeholders in head will never work since 
you cannot put `{% cms_toolbar %}` in the head, since it produces toolbar markup 
page extension or custom template tag is the solution, not a placeholder

evildmp: Actually I tested with `{% cms_toolbar %}` and placeholder template tag in the `<head>` - it works, but not as expected (they appear in the body in the DOM anyway).

vxsx: that depends on the browser

czpython: I suggest to add a note saying this is not possible.

evildmp:
@czpython Please check updates based on feedback.

PR #6088, opened 2017-10-03
Used plugin_pool.register_plugin() and apphook_pool.register() as decorators in examples.

sir-sigurd:


evildmp:
Thank you @sir-sigurd.

PR #6114, opened 2017-11-08
Bugfix/3.5.x/button group css fix

vxsx:
before:
screen shot 2017-11-08 at 16 01 11

after:
screen shot 2017-11-08 at 16 00 35


vxsx:
no need to wait for ci here

PR #6126, opened 2017-11-13
Fixed #6196 -- Prevent users from passing a public page as parent in create_page api function

### LINKED ISSUE #6196: create_page api function incorrectly accepts public page as parent

jedie:
Fixed #6196
A "published" parent will be accepted but will raise into problems, because of a wrong page tree.

I just add a assertment in `api.create_page()` and fix some tests that have used the published page as parent.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -14,7 +14,7 @@
 * Introduced logic to copy pages to different sites from the admin.
 * Removed "View on Site" button when adding a page
 * Welcome page no longer uses multilingual URLs when not required.
-* Prevent users from passing a public page as parent in create_page api function
+* Prevent users from passing a public page as parent in ``create_page`` api function
 
 
 === 3.4.5 (2017-10-12) ===
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -14,6 +14,7 @@
 * Introduced logic to copy pages to different sites from the admin.
 * Removed "View on Site" button when adding a page
 * Welcome page no longer uses multilingual URLs when not required.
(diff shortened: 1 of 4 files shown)


bmassot:
I recently tried to create a page whose parent was already published and, as you wrote, I experienced weird behaviors: the page wasn't shown in the page tree in the admin interface, but I could find it using the search form. Once published, this page was shown in the tree, but, while actually published, it was shown as unpublished in the admin.

Why is it possible to create a published page's child through the GUI but not through the API? Are there possible workarounds? Would it help to unpublish the parent, create the child page, and then republish the parent?

czpython:
All published pages have a draft and live version.
The cms will always allow you to create a child page under a published parent page.
What should not be allowed (and is fixed by this pr) is to pass a **public** page as the parent page instead of draft page.

bmassot:
Thank you czpython! I got it working using the draft=True parameter in my call to cms.utils.page_resolver.get_page_from_path() to retrieve the parent page.
As it may be not obvious to everyone, I think it's worth writing in the API doc.

jedie:
I "rebase" the patch to development branch.

czpython:
Hello @jedie,
Please add a new test on the tests_api module which just does an `assertRaises` to catch the new assert.

jedie:
> Please add a new test

Good idea. I have done this with https://github.com/divio/django-cms/pull/6126/commits/45ef26406f6a837251eeef90ba65075747354bc2#diff-e1121ea58dc56d8c1fd1a0ef63e599e6R235

czpython:
Thanks!

jedie:
Thank you ;) Hope the next release is coming soon ;)

PR #6128, opened 2017-11-14
Removed outdated deprecation notes in docs.

sir-sigurd:


docs/how_to/menus.rst
```diff
@@ -39,10 +39,6 @@ Create a ``cms_menus.py`` in your application, with the following::
 
     menu_pool.register_menu(TestMenu)
 
-.. note:: Up to version 3.1 this module was named ``menu.py``. Please
```
czpython: Unfortunately, support for the old filename was not removed on 3.4.
Can you change this to say it will be removed on 3.5?

PR #6154, opened 2017-11-23
Fixed a bug when adding multiple plugins

vxsx:
The structure returned into the dataBridge when adding the plugin was
ordered `parent -> child + child` but the frontend expects it to be
`child + child -> parent` because if the plugin doesn't have it's own
markup - the data about the child and parent is added to the same node
and the first one initialized must be the innermost one.

### DIFF
--- a/cms/toolbar/utils.py
+++ b/cms/toolbar/utils.py
@@ -107,6 +107,7 @@ def get_plugin_tree_as_json(request, plugins):
                 'cms_toolbar': toolbar,
             }
             tree_structure.append(template.render(context))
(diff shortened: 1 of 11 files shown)


cms/toolbar/utils.py
```diff
@@ -107,6 +107,9 @@ def collect_plugin_data(plugin):
                 'cms_toolbar': toolbar,
             }
             tree_structure.append(template.render(context))
+
+    tree_data.reverse()
```
czpython: no need for this newline

PR #6167, opened 2017-12-07
Fixed #5952 -- Set touch='selected' in jstree.dnd config.

### LINKED ISSUE #5952: Mobility issues with admin page tree
### Summary

We discovered an usability issue regarding the admin panel on a mobile touch device. When the page tree is opened, it is easy to mess up the ordering of the pages, since each scroll or zoom gesture is interpreted as a "move this page" action. At least as long as the gesture is done over the page title column, which is usually the largest one. I think scroll or zoom gestures are much more frequent on a mobile device than moving a page on purpose.

I can think of two ways to improve the behavior:
- page move is triggered after selecting a page with longpress
- move the page only when the grey-colored grab icon (the left-most thing in a row) is touched and not on every move on the title column

#### Expected behaviour
I can scroll through the page tree without moving pages all the way around.

#### Actual behaviour
The page tree got messed up after scrolling through the page tree.

#### Environment

* Python version: 3.5.3
* Django version: 1.10.7
* django CMS version: 3.4.3

\cc @ObviouslyNot

sir-sigurd:
### Summary
Fixes #5952 

Set touch='selected' in jstree.dnd config to prevent accidental page dragging in admin on touch devices.
See https://www.jstree.com/api/#/?f=$.jstree.defaults.dnd.touch:
> controls whether dnd works on touch devices. If left as boolean true dnd will work the same as in desktop browsers, which in some cases may impair scrolling. If set to boolean false dnd will not work on touch devices. There is a special third option - string "selected" which means only selected nodes can be dragged on touch devices.


sir-sigurd:
@vxsx IMO background-color for .jstree-clicked should be adjusted because currently the selected row is hardly distinguishable from the other ones, I was thinking about setting it to $color-primary but probably that's too much.

vxsx:
I think it's ok for now as is. When we get to fix other contrast issues it will be fixed as well

PR #6176, opened 2017-12-11
Solved PDF documentation issues

fp4code:
### Summary

Solved three PDF documentation issues

The three corresponding errors were logged in http://m.discodop.readthedocs.io/projects/django-cms/builds/6379857/:

#### 1st issue

    ! Package inputenc Error: Unicode char ⨯ (U+2A2F)
    (inputenc)                not set up for use with LaTeX.

Solved by changing character VECTOR OR CROSS PRODUCT by MULTIPLICATION X and
added ``tex_replacements.append(('✕', r'\(\times\)'))`` to conf.py

#### 2nd issue

    ! LaTeX Error: Too deeply nested.

End of file docs/topics/menu_system.rst is rewrited. The style is not perfect, but somewhat better.

#### 3rd issue

    l.11736 ...Japanese as \sphinxquotedblleft{}日本docs/index.rst
                                                  語\sphinxquotedblright{} ...

This is solved by three tex_replacements.append, one for each character. Not very elegant, but that's working.


evildmp:
Thanks @fp4code for this.

**1st issue** I'm quite happy with the solution for the crosses in the table.

**2nd issue** I am not able to build the documentation such that I see the error myself locally. I am not in favour of the solution proposed however, to remove the nested lists that make it possible to follow the explanation.

I'd like to know: how nested is too deep nesting? For example, u we reduced the level by one, would that solve the problem? Otherwise, I would want to look carefully at the list to find a way to simplify it further,

**3rd issue** It looks a bit hacky, but t's OK by me. However, please add a comment in the `conf.py` to explain why those changes are there, with a link to this PR so that in future the history is available.

docs/conf.py
```diff
@@ -246,3 +246,16 @@
 spelling_word_list_filename = 'spelling_wordlist'
 
 spelling_ignore_pypi_package_names = True
+
+latex_elements = {
+    'preamble': r'''\usepackage{CJKutf8}''',
+}
+
+from sphinx.util.texescape import tex_replacements
```
evildmp: Please can we have a comment here indicating why we're doing this.

docs/topics/menu_system.rst
```diff
@@ -153,36 +153,38 @@ Many of the methods below pass this list of nodes to the ones it calls, and retu
...
-        * :py:meth:`menus.templatetags.menu_tags.cut_levels()` removes nodes from the menu according to the arguments provided by the template tag
-        * :py:meth:`menus.menu_pool.MenuPool.apply_modifiers()` with ``post_cut = True`` loops over all the Modifiers again
-            * :py:class:`cms.menu.NavExtender`
-            * :py:class:`cms.menu.SoftRootCutter`
-            * :py:class:`menus.modifiers.Marker`
-            * :py:class:`menus.modifiers.AuthVisibility`
-            * :py:class:`menus.modifiers.Level`:
-                * :py:meth:`menus.modifiers.Level.mark_levels()`
-        * return the nodes to the context in the variable ``children``
+.. |_| unicode:: 0xA0 0xA0 0xA0
```
evildmp: Not in favour of this solution, it would be better to find a way to reduce the nesting if necessary.

fp4code: Do you really want the nesting structure or just a tabular appearance? I have added dots to give a better look. I will try to show the result on readthedocs.

fp4code: Please, compare even the first nested levels here:
https://media.readthedocs.org/pdf/django-cms/latest/django-cms.pdf (p.110) with there:
https://media.readthedocs.org/pdf/patched-djangocms-doc/latest/patched-djangocms-doc.pdf (p.109)

Also, here:
http://docs.django-cms.org/en/latest/topics/menu_system.html and there:
http://patched-djangocms-doc.readthedocs.io/en/latest/topics/menu_system.html

I do agree, my solution is not perfect yet, by using nested levels will be never good I'm afraid.

evildmp: @fp4code I feel the nesting is necessary, because the levels of nesting represent the logic of the system. It's true that it doesn't look good in the PDF version, but on the other hand, far more people will use the HTML, which is more important.

What is the maximum level of nesting that the PDF renditions can handle? It may be possible to break apart the monolithic list into smaller, shallower ones.

evildmp: @fp4code Re the third issue (Japanese), I'm also happy with that.

FinalAngel:
@fp4code would you be able to re-open the PR against develop?

@evildmp what is still missing from your end to get it merged into the next release?

FinalAngel:
Closing this for now @fp4code, please re-open on develop with the suggested edits if still relevant. Thank you very much!

PR #6177, opened 2017-12-11
Minor bugfixes for 3.5

vxsx:
1. made "view on site" links close the sideframe (and removed bunch on unused code)
2. fixed a bug where welcome page (and wizard) wouldn't respect toolbar language but instead would use the page language


cms/templates/cms/welcome.html
```diff
@@ -1,4 +1,4 @@
-<!doctype html>{% load i18n staticfiles cms_tags cms_static sekizai_tags %}
+<!doctype html>{% load i18n l10n staticfiles cms_tags cms_static sekizai_tags %}
```
czpython: we don't need the l10n

vxsx: right, fixed

Issue #6179, opened 2017-12-11
Rendering the menu with an un-configured language raise LanguageError


czpython:
Given the following settings:

```python
LANGUAGE_CODE = 'en'
CMS_LANGUAGES = {
    1: [
            {
                'code': 'en',
                'name': gettext('English'),
                'fallbacks': ['fr', 'de'],
                'public': True,
            },
        ],
    2:  [
            {
                'code': 'de',
                'name': gettext('German'),
                'fallbacks': ['fr'],
                'public': True,
            },
            {
                'code': 'fr',
                'name': gettext('French'),
                'public': True,
            },
        ],
}
```

If a user on site 2 requests a url which renders the menu, while using "en", the menu code will raise a `LanguageError`.

czpython:
Fixed by https://github.com/divio/django-cms/pull/6182

Issue #6173, opened 2017-12-09
Rendering a homepage with unconfigured language as staff-user throws exception


czpython:
Given the following settings:

```python
LANGUAGE_CODE = 'en'
CMS_LANGUAGES = {
    1: [
            {
                'code': 'en',
                'name': gettext('English'),
                'fallbacks': ['fr', 'de'],
                'public': True,
            },
        ],
    2:  [
            {
                'code': 'de',
                'name': gettext('German'),
                'fallbacks': ['fr'],
                'public': True,
            },
            {
                'code': 'fr',
                'name': gettext('French'),
                'public': True,
            },
        ],
}
```

If a staff user requests the homepage for site 2 using "en" as the language (``/en/``),
the cms will fail to redirect to one of the configured languages and instead raise
a `LanguageError`.

The reason for this is that the cms sets the redirect to an [attribute](https://github.com/divio/django-cms/blob/2d0778c2bfc6dd250e74353429baec95d29595d2/cms/views.py#L97) in the toolbar (since user is staff) instead of redirecting immediately, this causes the code to continue to the other conditions and raise a `LanguageError`.

czpython:
Fixed by https://github.com/divio/django-cms/pull/6182

Issue #6172, opened 2017-12-09
Rendering a homepage with unconfigured language as anon-user throws exception


czpython:
Given the following settings:

```python
LANGUAGE_CODE = 'en'
CMS_LANGUAGES = {
    1: [
            {
                'code': 'en',
                'name': gettext('English'),
                'fallbacks': ['fr', 'de'],
                'public': True,
            },
        ],
    2:  [
            {
                'code': 'de',
                'name': gettext('German'),
                'fallbacks': ['fr'],
                'public': True,
            },
            {
                'code': 'fr',
                'name': gettext('French'),
                'public': True,
            },
        ],
}
```

If the user requests the homepage for site 2 using "en" as the language (``/en/``),
the cms will fail to redirect to one of the configured languages and instead raise
a `LanguageError`.

The reason for this is that ``get_default_language()`` uses ``LANGUAGE_CODE`` regardless
of whether that language is configured for the site.

czpython:
Fixed by https://github.com/divio/django-cms/pull/6182

Issue #5841, opened 2017-01-06
LanguageError is thrown when user.is_staff and toolbar.edit_mode are True


nolsto:
### Summary

A condition in my multisite app causes me to have arrived at a cms view with `current_language = 'en-us'` and `user_languages = ['en-gb', 'fr']` at [cms/views.py#L83](https://github.com/divio/django-cms/blob/release/3.3.x/cms/views.py#L83).

When the user is unauthenticated, or the user is staff but the toolbar is not in edit mode, the elif condition of [cms/views.py#L96-L99](https://github.com/divio/django-cms/blob/release/3.3.x/cms/views.py#L96-L99) is reached and an HttpResponseRedirect to /en-gb/ is returned. All good.

When the user is staff, and the toolbar is in edit mode, `request.toolbar.redirect_url` is set, but a LanguageError is thrown at [cms/views.py#L110](https://github.com/divio/django-cms/blob/release/3.3.x/cms/views.py#L110) in `get_redirect_on_fallback` before a response can be returned.

### Expected behaviour

I expect to be redirected to /en-gb/ regardless of the properties of the user or the toolbar state.

#### Environment

* Python version: 3.5.2
* Django version: 1.9.10
* django CMS version: 3.3.3

czpython:
Hello @nolsto,
Can you post your language settings for both django and django-cms?

czpython:
Fixed by https://github.com/divio/django-cms/pull/6182

PR #6182, opened 2017-12-13
Refactored the fallbacks handling logic

czpython:
Fixes https://github.com/divio/django-cms/issues/6183
Fixes https://github.com/divio/django-cms/issues/6179
Fixes https://github.com/divio/django-cms/issues/6173
Fixes https://github.com/divio/django-cms/issues/6172
Fixes https://github.com/divio/django-cms/issues/5841

**BACKARDS INCOMPATIBLE**
The `language` parameter is no longer used as a way to get the current language.
From now on, the request language (or default language) is the one used.

### DIFF
--- a/cms/cms_menus.py
+++ b/cms/cms_menus.py
@@ -9,7 +9,12 @@ from cms.models import EmptyTitle
 from cms.utils import i18n
 from cms.utils.compat import DJANGO_1_9
 from cms.utils.conf import get_cms_setting
-from cms.utils.i18n import get_fallback_languages, hide_untranslated
+from cms.utils.i18n import (
+    get_default_language_for_site,
+    get_fallback_languages,
+    hide_untranslated,
+    is_valid_site_language,
+)
 from cms.utils.permissions import get_view_restrictions
 from cms.utils.page import get_node_queryset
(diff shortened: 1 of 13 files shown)


vxsx:
I'm afraid the integration test failure is real. Previously in edit mode you could switch to a language that's not yet created and you would see the screen block with text "This page has no preview! It is being redirected to $url". Now it just redirects to the language that exists

cms/cms_menus.py
```diff
@@ -137,23 +145,27 @@ def get_menu_node_for_page(renderer, page, language):
...
-        id=page.pk,
-        parent_id=parent_id,
-        attr=attr,
-        visible=page.in_navigation,
-        path=translation.path or translation.slug,
-        language=(translation.language if translation.language != language else None),
-    )
-    return ret_node
+    for lang in [language] + fallbacks:
+        translation = page.title_cache[lang]
```
stefanfoulis: possible ``KeyError``?

czpython: The title_cache is populated with ``EmptyTitle`` instances for all fallbacks, these instances are considered ``False`` when evaluated below.

PR #6184, opened 2017-12-13
Fixes #5297: Allows unicode slugs

### LINKED ISSUE #5297: django CMS should support Unicode slugs
Django has recently implemented support for Unicode slugs.

See https://github.com/django/django/pull/4518 for the implementation and https://code.djangoproject.com/ticket/16501 for the original ticket.

Alir3z4:
### Summary

At the moment only ASCII page slugs are accepted and user cannot enter Unicode character as slugs.
Unicode Slugs are very important when it comes to CMS.

This pull requests makes Django CMS PageTitle model to Allows Unicode slugs.

* Fixes #5297

* https://github.com/divio/django-cms/issues/5297
* http://support.divio.com/project-types/django-cms/unicode-characters-in-url-slugs

The `cms.models.Title.slug` gets a `allows_unicode` parameter. (since https://github.com/django/django/pull/4518)

### DIFF
--- a/cms/models/titlemodels.py
+++ b/cms/models/titlemodels.py
@@ -32,7 +32,7 @@ class Title(models.Model):
                                   help_text=_("overwrite the title in the menu"))
     meta_description = models.TextField(_("description"), max_length=155, blank=True, null=True,
                                         help_text=_("The text displayed in search engines."))
-    slug = models.SlugField(_("slug"), max_length=255, db_index=True, unique=False)
+    slug = models.SlugField(_("slug"), max_length=255, db_index=True, unique=False, allow_unicode=True)
     path = models.CharField(_("Path"), max_length=255, db_index=True)
     has_url_overwrite = models.BooleanField(_("has url overwrite"), default=False, db_index=True, editable=False)
     redirect = models.CharField(_("redirect"), max_length=2048, blank=True, null=True)
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -9,6 +9,7 @@
 * Fixed "Expand all / Collapse all" not reflecting real state of the placeholder tree
 * Fixed a bug where Aliased plugins would render even if their host page was unpublished.
 * Fixed a bug where focusing inputs in modal would require 2 clicks in some browsers
+* Allowing Unicode Slugs for Pages.
 
 
 === 3.4.5 (2017-10-12) ===


cms/models/titlemodels.py
```diff
@@ -32,7 +32,7 @@ class Title(models.Model):
                                   help_text=_("overwrite the title in the menu"))
     meta_description = models.TextField(_("description"), max_length=155, blank=True, null=True,
                                         help_text=_("The text displayed in search engines."))
-    slug = models.SlugField(_("slug"), max_length=255, db_index=True, unique=False)
+    slug = models.SlugField(_("slug"), max_length=255, db_index=True, unique=False, allow_unicode=True)
```
Alir3z4: This will fail on Django 1.8.

I can check for the django version and if it's higher than 1.8 then apply the `allow_unicode=True` unless leave the slug field untouched.

Any idea how to handle @czpython ?

vstoykov: Django 1.8 is officially supported from Django untill april this year. I think that is a good idea to make it compatible with Django 1.8.

One approach can be:
```
try:
    models.SlugField(allow_unicode=True)
except TypeError:
    slug_extra_kwargs = {}
else:
    slug_extra_kwargs = {'allow_unicode': True}

...
slug = models.SlugField(_("slug"), max_length=255, db_index=True, unique=False, **slug_extra_kwargs)
```
Or something like that

jedie:
I stumbled across it by accident. Just a few notes:

What's about the JavaScript stuff that will replace non ASCII characters in frontend?

What's about `SLUG_REGEXP = `'[0-9A-Za-z-_.//]+'` in: https://github.com/divio/django-cms/blob/develop/cms/constants.py#L29 that would not match on non ASCII, isn't it? see also #4847

vThaian:
Develop don't support 1.8 so we don't need to handle it, but what with JS? Did you check this maybe?

FinalAngel:
closing this for now, please reopen with adjustments if needed :)

Thank you very much for the initial effort and comments!

cms/models/titlemodels.py
petersiucom: Hi All,
I tried to modified it in my project to allows unicode slugs.
cms/models/titlemodels.py
slug = models.SlugField(_("slug"), max_length=255, db_index=True, unique=False, allow_unicode=True)

it doesn't work, it complaint 'Slug must not be empty.'
django version:  1.11.26
Django-CMS version: 3.7 

any update for this issue?
thanks all

PR #6191, opened 2017-12-13
Fixed #6188 -- Only set published (or once published) languages into the public page

### LINKED ISSUE #6188: [3.5.x] Publishing a page sets all languages from draft into public
If a user has a page with two or more languages, and the user publishes one language, the cms will incorrectly set the `languages` field of the public version of the page to all languages from the draft version.

czpython:
Fixed #6188


stefanfoulis:
👍

PR #6224, opened 2018-01-10
Improved and simplified permissions documentation

evildmp:
[ci-skip]


FinalAngel:
@evildmp could you resolve the conflict so we can merge this into develop for 3.7?

PR #6237, opened 2018-01-15
Minor 3.5.rc1 bugfixes

vxsx:
- Removed "View on site" button when adding page


vxsx:
@czpython is also used in "Edit page" situation, not the worst thing to have?

czpython:
@vxsx we can do:

```
{% if not add and not popup %}
        <ul class="object-tools hide-in-modal">
            <li>
                {% url opts|admin_urlname:'preview_page' page.id language as admin_url %}
                <a href="{{ admin_url }}" target="_parent">{% trans "View on site" %}</a>
            </li>
        </ul>
    {% endif %}
```

vxsx:
Wasn't aware of "add" - fixed

PR #6239, opened 2018-01-16
Don't use multilingual URLs on the welcome template redirect

### LINKED ISSUE #5070: Welcome wizard redirect to language-prefixed home, even when i18n is disabled
While following [Martin's YouTube Demo](https://www.youtube.com/watch?v=NbsRVfLCE1U&list=PLbNxojMo86v42AOM47uhaEFHIJFUM4d6M&index=12) 
I answered NO to the I18N/L10N question, yet the installer still asked for a language choice.  I entered **'en'** and proceeded.  Then I:
1. `manage.py runserver 0:8000`
2. opened in browser
3. logged in 
The cms defaulted to URL `http://localhost:8000/en/?edit` which gives a **404 error**.

Changing the URL to `http://localhost:8000/?edit` actually opens home page editing view.

Please correct this behavior, for beginners it is very confusing.

edevil:
### Summary

Fixes #5070

Not everyone uses multilingual URLs so use the normal URLs.

### DIFF
--- a/cms/templates/cms/welcome.html
+++ b/cms/templates/cms/welcome.html
@@ -98,7 +98,7 @@
                 }
             });
         {% else %}
-            window.location.href = '{% url "admin:login" %}?next={{ next_url }}';
+            window.location.href = '{% url "admin:login" %}?next={{ next_url }}?{{ cms_edit_on }}';
         {% endif %}
     </script>
     {% endlanguage %}
--- a/cms/templates/cms/welcome.html
+++ b/cms/templates/cms/welcome.html
@@ -98,11 +98,7 @@
                 }
             });
         {% else %}
-            {% if LANGUAGE_CODE %}
-                window.location.href = '{% url "admin:login" %}?next=/{{ LANGUAGE_CODE }}/?{{ cms_edit_on }}';
-            {% else %}
-                window.location.href = '{% url "admin:login" %}?next=/?{{ cms_edit_on }}';
-            {% endif %}
+            window.location.href = '{% url "admin:login" %}?next=/?{{ cms_edit_on }}';
         {% endif %}
     </script>
     {% endlanguage %}
--- a/cms/templates/cms/welcome.html
+++ b/cms/templates/cms/welcome.html
@@ -98,7 +98,7 @@
                 }
             });
         {% else %}
-            window.location.href = '{% url "admin:login" %}?next=/?{{ cms_edit_on }}';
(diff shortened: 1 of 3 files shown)


czpython:
Hello @edevil,
Thanks for the patch.

We would still like to keep the correct url depending on the project's language settings.

Please add a `next_url` context variable to https://github.com/divio/django-cms/blob/develop/cms/page_rendering.py#L73 which reverses the cms root url.
`next_url = reverse('pages-root')`. Then this variable can be used as the `next` GET parameter in the template.

Doing this lets Django determine whether the url should have a language prefix or not.

edevil:
@czpython Wow, that was fast. It's great to see this commitment from maintainers. :)

I've made the changes you requested. How does Django determine if the URL should have the language or not?

czpython:
@edevil thanks!
When configuring the project, if you use the i18n patterns, django knows to prefix the current language to the url. Check out https://docs.djangoproject.com/en/2.0/topics/i18n/translation/#language-prefix-in-url-patterns.

cms/templates/cms/welcome.html
```diff
@@ -98,11 +98,7 @@ <h2 class="cms-hidden">{% trans "Installation Notes" %}</h2>
                 }
             });
         {% else %}
-            {% if LANGUAGE_CODE %}
-                window.location.href = '{% url "admin:login" %}?next=/{{ LANGUAGE_CODE }}/?{{ cms_edit_on }}';
-            {% else %}
-                window.location.href = '{% url "admin:login" %}?next=/?{{ cms_edit_on }}';
-            {% endif %}
+            window.location.href = '{% url "admin:login" %}?next={{ next_url }}';
```
czpython: missing the `?{{ cms_edit_on }} `

edevil: Sorry, done.

edevil:
Added the missing parameter back.

edevil:
@czpython Is anything else needed?

czpython:
@edevil One last thing, please add an entry to the CHANGELOG file under 3.5.x.

evildmp:
Thanks @edevil.

For the benefit of developers and maintainers, please

* [ ] Add a line to the CHANGELOG: ``Welcome page no longer uses multilingual URLs when not required.``

edevil:
@czpython @evildmp done

czpython:
Thanks!

PR #6256, opened 2018-02-01
Simplify code and use the tools provided by django-CMS itself

### LINKED ISSUE #6251: Nodes from CMSAttachMenu are ignored by cms.admin.pageadmin.get_published_pagelist
### Summary

Nodes added by a class inheriting from from `CMSAttachMenu` are ignored by `cms.admin.pageadmin.get_published_pagelist` which builds a results list, when querying  *Redirect* in the *Advanced Settings* of **Pages**.

### Expected behaviour

Since Nodes returned by `CMSAttachMenu.get_nodes()` are part of the CMS menu tree, I would expect them to be offered by the *Redirect* input field of *Advances Settings*.

### Actual behaviour

Only CMS pages are returned, but not nodes added by a menu extension.

By reading the code, it seems as if this feature was forgotten, since not attempt has been made to add these nodes to the result set.

### Environment

* Python version: all
* Django version: 1.11
* django CMS version: 3.4.5

jrief:
### Summary

Today I created a monkey patch to solve #6251. I then discovered, that by using the tools provided by django-CMS itself, the method `cms.admin.pageadmin.PageAdmin.get_published_pagelist()` can be even simplified compared to the current implementation. It furthermore is more standard compliant and solves a problem I have in django-SHOP.

Therefore I would like to share this code for review. Otherwise I must add a monkey patch to django-SHOP, which isn't something I'd really like to do.

@czpython 
Please don't close pull requests/issues which solve/describe real problems others have in their projects. In other OSS communities such decisions are made through voting and not by a single person.

Before adding unit tests, I'd like to have a review on this. I then will port it to version 3.5. 

Fixes #6251

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -1420,12 +1420,17 @@ class PageAdmin(PlaceholderAdminMixin, admin.ModelAdmin):
 
     def get_published_pagelist(self, *args, **kwargs):
         """
-         This view is used by the PageSmartLinkWidget as the user type to feed the autocomplete drop-down.
+        This view is used by the PageSmartLinkWidget as the user type to feed the autocomplete drop-down.
         """
         request = args[0]
         if not request.is_ajax():
             return HttpResponseForbidden()
         query_term = request.GET.get('q','').strip('/').lower()
+
+        # override request language if set
+        language_code = request.GET.get('language_code')
+        if language_code:
+            request.LANGUAGE_CODE = language_code
 
         # add menu items from the menu pool
         results = []
@@ -1434,7 +1439,7 @@ class PageAdmin(PlaceholderAdminMixin, admin.ModelAdmin):
             if query_term in path.lower() or query_term in node.title.lower():
                 results.append({
                     'path': path,
-                    'title': node.title,
+                    'title': node.get_menu_title(),
                     'redirect_url': node.get_absolute_url(),
                 })
                 if len(results) > 15:
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -8,7 +8,6 @@ import uuid
 
 import django
 from django.contrib.admin.helpers import AdminForm
-from django.conf import settings
 from django.conf.urls import url
 from django.contrib import admin, messages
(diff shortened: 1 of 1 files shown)


czpython:
Hello @jrief,

From me this is a no-go.

The menu system is purposely decoupled from the cms core code.
As I've stated before, any feature proposal should go over to the [django CMS developers](https://groups.google.com/forum/#!forum/django-cms-developers) google group where other devs can vote on it prior to creating the pull request.

By design the cms will display only pages in the redirect field.
Changing this would require much more work as there's difference between "published pages" and "published pages in the menu" which is the current state of this pr.

If you would like to get others involved, please re-open this on the developers group.

Also note that any new feature / change in behavior will land on 3.6 as 3.4 and 3.5 are maintenance only.

evildmp:
Please see http://docs.django-cms.org/en/latest/contributing/development-policies.html#proposal-and-discussion-of-significant-changes:

> Proposal and discussion of significant changes
New features and backward-incompatible changes should be proposed using the django CMS developers email list. Discussion should take place there before any pull requests are made.
>
> This is in the interests of openness and transparency, and to give the community a chance to participate in and understand the decisions taken by the project.

jrief:
Hi Daniele,

> Proposal and discussion of significant changes

This change is completely insignificant. The function which is affected is only used for one single purpose, which is to be an HTTP endpoint serving a list of URLs which may be used for redirection. It is rarely used and only available in the Advanced Section of the Page Admin.

Since Paulos claim "*By design the cms will display only pages in the redirect field*", isn't documented anywhere, and since such a restriction doesn't even make sense, from my point of view, this is a bug. Hence opening an issue for it is perfectly OK.

czpython:
This is an internal endpoint, as such it is not documented and/or open for usage outside of the core.

jrief:
@czpython by closing this pull request, one problem remains:
If the CMS contains many pages and the search query is empty or short, the queryset might become very long. Then all items are serialized and transferred to the client, who may experience a long delay. This should be shortened to say 15-25 items, since the Select2 widget for the redirect field usually does not offer much more space.

Issue #6267, opened 2018-02-02
cms.0017_pagetype... fails __fake__.MultipleObjectsReturned: get() returned more than one Page -- it returned 2!


LaurensBosscher:
### Summary

I'm trying to run the migrations on a site that has two Page trees; one for the Dutch version and one for the Belgium version. We started the project with only a Dutch version and at some point we made a copy (using the "cms copy" management command) of this so that the BE editors could edit the page tree. 

We use two page trees here because the .be version of our website has different content than the Dutch version, so using a translation didn't fit our requirements.

After updating to 3.5 we can't run the migrations.

### Expected behaviour

I expected the migrations to run :)

### Actual behaviour

The migrations fail at cms.0017_pagetype because of:

`
__fake__.MultipleObjectsReturned: get() returned more than one Page -- it returned 2!
`

This seems to be related to this line: https://github.com/divio/django-cms/blob/fd5956bef4b74b0a8c144c2ecde74fdae67756a7/cms/migrations/0017_pagetype.py#L21 in the migrations.

When I run the query on our DB I see that we do have two of these objects:

```
>>> for i in Page.objects.filter(reverse_id='page_types', publisher_is_draft=True):
...   print(i.pk)
...   print(i.site)
... 
838
www.<site name>.nl
1515
www.<site name>.be
```

```
>>> for i in Page.objects.filter(reverse_id='page_types'):
...   print(i.site)
...   print(i.pk)
... 
www.<site name>.nl
838
www.<site name>.nl
952
www.<site name>.be
1515
www.<site name>.be
1797
```

I couldn't find any information on what page_types is supposed to do. Can we safely delete this? If so, it might be worth adding this to the release notes :) 

If not, it might be worth changing the migration to iterate over these objects instead of running a .get.

Thanks!

### Environment

* Python version: 3.6
* Django version: 1.11
* django CMS version: 3.5

LaurensBosscher:
Okay, so I've very poorly hacked together an updated migration that seems to work:

```
root@b05e7dca2651:/code# cat /usr/local/lib/python3.6/site-packages/cms/migrations/0017_pagetype.py 
# -*- coding: utf-8 -*-
from __future__ import unicode_literals

from django.db import migrations, models


def update_descendants(parent, **data):
    parent.children.update(**data)

    for child in parent.children.iterator():
        update_descendants(child, **data)


def migrate_to_page_types(apps, schema_editor):
    Page = apps.get_model('cms', 'Page')
[...]

PR #6269, opened 2018-02-03
Changed migration 0017_pagetype to work with multiple page_type_root …

LaurensBosscher:
### Summary

This is the fix that worked for me in https://github.com/divio/django-cms/issues/6267. Please do take a critical look at this, this fix seems to work but I do not feel knowledgeable enough of Django CMS to fully grasps the implications. This fix seems harmless but that doesn't mean much ;)

Fixes #

https://github.com/divio/django-cms/issues/6267

### DIFF
--- a/cms/migrations/0017_pagetype.py
+++ b/cms/migrations/0017_pagetype.py
@@ -15,31 +15,24 @@ def migrate_to_page_types(apps, schema_editor):
     Page = apps.get_model('cms', 'Page')
     db_alias = schema_editor.connection.alias
 
-    try:
-        page_types_root = Page.objects.using(db_alias).get(
-            publisher_is_draft=True,
-            reverse_id='page_types',
-        )
-    except Page.DoesNotExist:
-        return
+    for page_types_root in Page.objects.using(db_alias).filter(publisher_is_draft=True, reverse_id='page_types'):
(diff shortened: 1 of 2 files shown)


cms/migrations/0017_pagetype.py
```diff
@@ -15,31 +15,24 @@ def migrate_to_page_types(apps, schema_editor):
...
-    if page_types_root_public:
-        # very strange case.. technically page-types should never be published.
-        # but nothing actually prevents it, so update public pages (if any).
-        update_descendants(page_types_root_public, is_page_type=True)
-
-        # Remove reverse id from public page
-        page_types_root_public.reverse_id = ''
-        page_types_root_public.is_page_type = True
-        page_types_root_public.save(update_fields=['reverse_id', 'is_page_type'])
+    for page_types_root in Page.objects.using(db_alias).filter(publisher_is_draft=True, reverse_id='page_types'):
```
czpython:
```
Please put the lookup on one line like:

page_types = Page.objects.using(db_alias).filter(
    reverse_id='page_types',
    publisher_is_draft=True,
)
for page_types_root in page_types:
   ...
```

LaurensBosscher:
Thanks for the feedback. I've updated the pull-request!

PR #6281, opened 2018-02-15
Django 2.0 compatibility: first steps

russmo:
### Summary

Squashing some incompatibilities between Django 1.11 and 2.0 with some try/except shims. I have knocked out basic errors on django-cms, django-filer, and many plugins that raise exceptions when trying to run django-cms 3.5.0 on Django 2.0.2.  I sent separate pull requests to those repos.  Should be backwards compatible with Django versions <2.0.  My test site ran fine on Django 1.10 and 1.11 with these changes.  Many basic features worked ok on Django 2.0, but there is much work to be done still.  I just wanted to help get a jump start on this to speed the transition.

If django-cms 3.6.x will drop support for Django < 1.11, then a few of these shims might not be necessary.  Most deal with breaking differences between 1.11 and 2.0, though.  Maybe this should get pulled to czpython:future/3.6.x referenced in #6279 .

Fixes #

* Added on_delete=models.CASCADE on all model foreign key fields and in all older migrations.
* Shimmed module naming and location inconsistencies between Django<=1.11 and 2.0.
* Attempted some shims on some Django API changes in 2.0.

### Remaining Issues

* Order of form media javascript files is different on plugin admin forms rendered by Django 2.0.  This causes problems with jQuery not being loaded before plugins.  This needs to be addressed in django-cms, djangocms-??? plugins that use filer, and possibly django-filer itself.  For example, adding image plugins (or file, folder, etc) does not work, but adding text and link plugins works as expected.
* I'm sure many many more things that need to get resolved as part of a bigger push for Django 2.0 compatibility.

### DIFF
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
 
-__version__ = '3.5.0-django2'
+__version__ = '3.5.0'
 
 default_app_config = 'cms.apps.CMSConfig'
--- a/cms/__init__.py
+++ b/cms/__init__.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
 
-__version__ = '3.5.0'
+__version__ = '3.5.0-django2'
 
 default_app_config = 'cms.apps.CMSConfig'
--- a/cms/appresolver.py
+++ b/cms/appresolver.py
@@ -5,11 +5,10 @@ from importlib import import_module
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 try:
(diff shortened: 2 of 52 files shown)


vstoykov:
Probably adding a compatibility function somewhere for checking if the user is authenticated is better that `try:..except:` on all places.

Also probably some migration need to be fixed to represent the change `on_delete=models.CASCADE`.

czpython:
Thanks for the effort on this.
We've began a separate effort in https://github.com/divio/django-cms/pull/6402

russmo:
Thanks for moving this forward.  Sorry I’ve been MIA the last couple months.

PR #6294, opened 2018-03-02
Fixed a bug due to incorrect localization in toolbar JS

rfleschenberg:
Fixed https://github.com/divio/django-cms/issues/6296
When editing a page with a pk > 9999 and having `USE_THOUSAND_SEPARATOR = True` in my settings, `ValueError` is thrown. This change fixes the issue for me.

Note: I did not read through the code to check if there are more instances where values should be unlocalized. It seems like a good idea to do that. Maybe it would be even better to turn off localization in some templates and explicitly localize values instead.


rfleschenberg:
I think the build errors are unrelated to the change (empty job lob).

PR #6303, opened 2018-03-07
Fixed #6302 -- Copy the __qualname__ attribute on the cms_perms decorator

### LINKED ISSUE #6302: Views from CMSApps are not compatible with 'django.contrib.admindocs'
### Summary

The documentation of views from CMSApps is not correctly served by [django.contrib.admindocs](https://docs.djangoproject.com/en/2.0/ref/contrib/admin/admindocs/) if those CMSApps have `permissions = True`. To my understanding, this happens because the current implementation of [`cms.utils.decorators.cms_perms`](https://github.com/divio/django-cms/blob/develop/cms/utils/decorators.py#L11) does not copy the `__qualname__` attribute of the wrapped view function (or its class) to the wrapping function.

### Expected behaviour
Proper dotted paths to the actual view (class)
admindocs_expected

### Actual behaviour
Gibberish of dotted path to module and decorator `__qualname__`
admindcos_actual

### Environment

* Python version: 3.5.3
* Django version: 1.11.9
* django CMS version: 3.4.5

schwobaseggl:
#6302 Copy the `__qualname__` attribute from the wrapped view function to the wrapping function. This seems sensible as tools like django.contrib.admindocs require this attribute to be set correctly.

### Summary

Makes code documentation of CMSApp views available in admindocs (for Python 3.3+) which accesses the this attribute in order to import the view and retrieve its doc string. 

Copy the `__qualname__` attribute from the wrapped view function. Use the view function's class as a fallback source. This covers function based views, class based views (that are included via their `as_view` method) as well as view classes that implement `__call__`. The proposed changes should be fairly harmless as they do not tamper with any attribute that was used before by cms code.

### DIFF
--- a/cms/utils/decorators.py
+++ b/cms/utils/decorators.py
@@ -23,6 +23,11 @@ def cms_perms(func):
     elif hasattr(func, '__class__'):
         inner.__name__ = func.__class__.__name__
 
+    if hasattr(func, '__qualname__'):
+        inner.__qualname__ = func.__qualname__
+    elif hasattr(func, '__class__') and hasattr(func.__class__, '__qualname__'):
+        inner.__qualname__ = func.__class__.__qualname__
+
     if getattr(func, 'csrf_exempt', False):
         # view has set csrf_exempt flag
         # so pass it down to the decorator.


czpython:
Hello @schwobaseggl,
Thanks for your contribution.
Can you add an entry to the CHANGELOG under 3.4.6 (unreleased)?

schwobaseggl:
Hi @czpython,
Sorry for the late response. Of course, I write a changelog entry. Since 3.4.6 has been released, should I open a 3.4.7 (unreleased) section and add the entry there?

czpython:
@schwobaseggl 
Yes that would work. Thanks!

czpython:
Hello @schwobaseggl,
Could you reopen this pr against the develop branch?
I'll add the changelog entry on the backport.

PR #6314, opened 2018-03-20
Fixed a bug where non-staff user would be able to open empty structureboard

vxsx:
`content_mode_active` is only active when the user is staff: https://github.com/divio/django-cms/blob/52d31944afbda90dc5281b5a2ae7e99f2e2a931b/cms/toolbar/toolbar.py#L102

the js expects that whenever user is in edit mode CMS.settings.mode to be `edit` and not `structure` as it was before.

### DIFF
--- a/cms/tests/frontend/integration/nonadmin.js
+++ b/cms/tests/frontend/integration/nonadmin.js
@@ -1,0 +1,87 @@
+'use strict';
+
+var helpers = require('djangocms-casper-helpers');
+var globals = helpers.settings;
+var casperjs = require('casper');
+var xPath = casperjs.selectXPath;
+var cms = helpers(casperjs);
+
+casper.test.setUp(function(done) {
+    casper
+        .start()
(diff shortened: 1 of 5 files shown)


vxsx:
@czpython tests are up

cms/tests/frontend/integration/nonadmin.js
```diff
@@ -0,0 +1,87 @@
...
+var cms = helpers(casperjs);
+
+casper.test.setUp(function(done) {
+    casper
+        .start()
+        .then(cms.login())
+        .then(cms.addPage({ title: 'Home page' }))
+        .then(
+            cms.addPlugin({
+                type: 'TextPlugin',
```
czpython: please use a plugin from the cms tests instead.
like style, etc..

vxsx: thisisfine.jpg

PR #6380, opened 2018-05-20
Fixed #3734 -- Automated check that PlaceholderFields use Placeholder…

### LINKED ISSUE #3734: Automated check that PlaceholderFields use PlaceholderAdminMixin
Haven't looked into the feasibility of this but would like for the checks framework to detect all models with PlaceholderField, if any of those models are also registered in the admin that the admin class inherits from PlaceholderAdminMixin. It's an error to ever do otherwise, so would like this to be loud instead of silent.

vaquer:
Fixed #3734 -- Automated check that PlaceholderFields use PlaceholderAdminMixin

### Summary
ModelAdmin instances that are using PlaceholderField fields should be also a subclass of PlaceholderAdminMixin


cms/utils/check.py
```diff
@@ -343,6 +343,35 @@ def get_class(method_name, model):
...
+                section.success("%s is subclass of PlaceholderAdminMixin" % model_admin)
+            else:
+                section.error(output.colorize(
+                    "%s is subclass of PlaceholderAdminMixin" % model_admin,
+                    opts=['bold'],
+                    fg='red'
+                ))
+
+        if section.successful:
+            section.finish_success("PlaceHolderAdminMixin configuration okay")
```
czpython: Change `PlaceHolderAdminMixin`  to `PlaceholderField`

cms/utils/check.py
```diff
@@ -343,6 +343,35 @@ def get_class(method_name, model):
...
+        for model, model_admin in site._registry.items():
+            ph_fields = [field for field in model._meta.get_fields() if isinstance(field, PlaceholderField)]
+            if len(ph_fields) == 0:
+                continue
+
+            if isinstance(model_admin, PlaceholderAdminMixin):
+                section.success("%s is subclass of PlaceholderAdminMixin" % model_admin)
+            else:
+                section.error(output.colorize(
+                    "%s is subclass of PlaceholderAdminMixin" % model_admin,
```
czpython: ```%s does not subclass PlaceholderAdminMixin```

cms/utils/check.py
```diff
@@ -343,6 +343,35 @@ def get_class(method_name, model):
...
+def check_placeholder_mixin(output):
+    """
+    ModelAdmin instances that are using PlaceholderField fields
+    should be also a subclass of PlaceholderAdminMixin
+    """
+    from django.contrib.admin import site
+    from cms.models.fields import PlaceholderField
+    from cms.admin.placeholderadmin import PlaceholderAdminMixin
+
+    with output.section("PlaceHolderAdminMixin") as section:
```
czpython: Change `PlaceHolderAdminMixin` to `PlaceholderField`

cms/utils/check.py
```diff
@@ -343,6 +343,35 @@ def get_class(method_name, model):
                                    'https://django-cms.readthedocs.io/en/latest/extending_cms/extending_page_title.html#handling-relations.')  # noqa
 
 
+@define_check
+def check_placeholder_mixin(output):
```
czpython: Change to `check_placeholder_fields`

cms/utils/check.py
```diff
@@ -343,6 +343,35 @@ def get_class(method_name, model):
...
+    from cms.models.fields import PlaceholderField
+    from cms.admin.placeholderadmin import PlaceholderAdminMixin
+
+    with output.section("PlaceHolderAdminMixin") as section:
+        for model, model_admin in site._registry.items():
+            ph_fields = [field for field in model._meta.get_fields() if isinstance(field, PlaceholderField)]
+            if len(ph_fields) == 0:
+                continue
+
+            if isinstance(model_admin, PlaceholderAdminMixin):
```
czpython: I think is ok to remove this from output and instead only log if there's error.
so would result in:

```python
if not isinstance(model_admin, PlaceholderAdminMixin):
    ...
```

cms/tests/test_check.py
```diff
@@ -118,6 +119,19 @@ def test_non_numeric_site_id(self):
         with self.settings(SITE_ID='broken'):
             self.assertCheck(False, warnings=0, errors=1)
 
+    @isolate_apps("test_app")
+    def test_placeholder_mixin(self):
```
czpython: Rename to `test_placeholder_field`

cms/utils/check.py
```diff
@@ -343,6 +343,35 @@ def get_class(method_name, model):
...
+                section.success("%s is subclass of PlaceholderAdminMixin" % model_admin)
+            else:
+                section.error(output.colorize(
```
czpython: `section.error` already does colorization, can remove the call to `output.colorize`

vinitkumar:
@czpython Do you think this is merge ready?

cms/utils/check.py
```diff
@@ -339,6 +339,31 @@ def get_class(method_name, model):
...
+    should be also a subclass of PlaceholderAdminMixin
+    """
+    from django.contrib.admin import site
+    from cms.models.fields import PlaceholderField
+    from cms.admin.placeholderadmin import PlaceholderAdminMixin
+
+    with output.section("PlaceholderField") as section:
+        for model, model_admin in site._registry.items():
+            ph_fields = [field for field in model._meta.get_fields() if isinstance(field, PlaceholderField)]
+            if len(ph_fields) == 0:
```
czpython, line 355: you can do `if not ph_fields` instead

czpython (approved):
LGTM
Left one more small suggestion

goutnet:
I updated the branch, let's see if this one can merge

goutnet (approved):
Approving based on the comments from the core team

PR #6391, opened 2018-05-30
Fixed some spelling and build issues

evildmp:


evildmp:
@czpython This fixes a number of issues related to building the documentation and performing spelling checks, but I can't get rid of: ``/docs/reference/api_references.rst:99: WARNING: py:class reference target not found: datetime``.

I think it's an import issue - if the ``datetime`` class were available, it wouldn't give this error.

PR #6409, opened 2018-06-06
Updated apphooks (how-to and introduction)

evildmp:
[ci only docs]


docs/how_to/apphooks.rst
```diff
@@ -26,65 +26,78 @@ The file needs to contain a :class:`CMSApp <cms.app_base.CMSApp>` sub-class. For
...
 
-If you fail to this, then any templates in the application that invoke URLs using the form ``{% url 'myapp:index' %}``
-or views that call (for example) ``reverse('myapp:index')`` will throw a ``NoReverseMatch`` error.
+    class MyApphook(CMSApp):
+        # ...
+        def get_urls(self, page=None, language=None, **kwargs):
+            return [
+                url(r'^$', SomeListView.as_view()),
+                url(r'^(?P<slug>[\w-]+)/?$', SomeDetailView.as_view()),
+                ]
```
evildmp: @czpython Before merging, we need to explain how namespacing would be handled when returning URLs manually using this method. What do you suggest?

czpython: What do you mean? is handled same way

PR #6421, opened 2018-06-21
WIP: App registration

monikasulik:
### Summary
This is the new app registration system. It allows django apps that extend the cms to register their functionality with the cms and introduces a way to set configuration for other apps that use this additional functionality.
Atm this is a WIP. The description of this PR will also be updated with more info soon.

### Documentation checklist

    * [ ] general notes
    * [ ] bug-fixes
    * [ ] required upgrade steps
    * [ ] names of contributors


monikasulik:
@evildmp From what I understand you will be creating documentation for this. Please let me know if you need more than what I've put in the PR description to create this. Also, on the topic of docs there's a comment from Noel - it might be better to use a differently named app for the example so that nobody gets confused that these docs describe how to actually integrate with the actual versioning app.

monikasulik:
@czpython I ran the description in this PR by Matus and Dami who are planning on designing the generic (i.e. configurable) part of moderation imminently and gave me detailed feedback with that very much in mind. There's some stuff I probably need to add to the description of this PR to make a few things clearer based on that discussion (@evildmp I'll try to do that on Monday), but also there are two suggestions for changing how this would work:

1. Instead of `versioning_enabled = True`, `moderation_enabled = True` etc. just have `supported_apps = ['moderation', 'versioning']` in the config. Interestingly, even though Matus and Dami read through the docs separately and didn't consult with each other before speaking to me, they were both confused by the "enabled" config options and found the idea of having a supported apps config much clearer and more intuitive. Although I wasn't confused by the enabled options when I first saw them, I do agree that the `supported_apps` option would be a clearer interface.
1. Matus seemed to have the same confusion as I did when I first saw Paulo's prototype. To me initially it was very confusing that we didn't have two base classes - something like `CMSAppConfig` and `CMSAppExtension`. cms_config.py files could then define either one or two classes. Classes that inherit from CMSAppExtension would have to implement the configure_app method. Classes which inherit from CMSAppConfig would have to set `supported_apps` and any other app-specific settings. Dami was not as confused, but agreed that to him this would also be clearer.

IMO it would be good to implement both of those :) But obviously want to consult with @czpython 

Some other questions also arose:

- What about namespacing the settings? Should we do that so as to avoid setting name clash between different apps?
- What about if two apps conflict with each other? For example lets say we have an app called 'pink_admin' which makes the templates in the admin turn pink and an app called 'purple_admin' which makes the admin turn purple. If both are enabled then surely one will win over the other or something else odd will happen. Should we worry about that now?

CHANGELOG.txt
```diff
@@ -1,3 +1,8 @@
+=== 4.0.0 (unreleased) ===
```
czpython: I know the PR guidelines suggest to do this but for 4.0.x, let's leave the changelog as is.
Once we're closer to release, then we can use the pr descriptions to compose the changelog.
The reason for this is that otherwise there will be conflicts all the time with changelog file and will need constant rebasing, etc..

cms/app_base.py
```diff
@@ -93,3 +93,24 @@ def get_urls(self, page=None, language=None, **kwargs):
...
+        attribute like "<app_label>_enabled" as True on its cms app
+        config class.
+        So for example if app A with label "app_a" implements this
+        method and app B and app C define app_a_enabled = True on their
+        cms config classes, the method app A has defined will run twice,
+        once for app B and once for app C.
+
+        :param app: django app that has defined the feature as enabled
+        """
+        raise NotImplementedError()
```
czpython: when not passing arguments, raise `NotImplementedError`, vs `NotImplementedError()`

cms/utils/setup.py
```diff
@@ -2,6 +2,9 @@
 from django.core.exceptions import ImproperlyConfigured
 
 from cms.utils.compat.dj import is_installed as app_is_installed
+from cms.app_registration import (
+    autodiscover_cms_configs, get_cms_apps,
+    configure_cms_apps)
```
czpython: please change to:

```python
from cms.app_registration import (
    autodiscover_cms_configs,
    configure_cms_apps,
    get_cms_apps,
)
```

cms/test_utils/project/app_with_bad_cms_file/cms_config.py
```diff
@@ -0,0 +1 @@
+raise KeyError
```
czpython: Is there a reason for the KeyError? maybe use a `RuntimeError` instead.
Also some comments here would be good.

cms/app_registration.py
```diff
@@ -0,0 +1,85 @@
...
+    Returns django app configs of apps with a cms config
+    """
+    cms_apps = []
+
+    for app_config in apps.get_app_configs():
+        # The cms_app attr is added by the autodiscover_cms_configs
+        # function if a cms_config.py file with a suitable class is found.
+        if hasattr(app_config, 'cms_app'):
+            cms_apps.append(app_config)
+
```
czpython: remove newline

cms/app_registration.py
```diff
@@ -0,0 +1,85 @@
...
+                raise ImproperlyConfigured(
+                    "cms_config.py files must define exactly one "
+                    "class which inherits from CMSAppConfig")
+
+
+def get_cms_apps():
+    """
+    Returns django app configs of apps with a cms config
+    """
+    cms_apps = []
```
czpython: I think this can be a list comprehension

cms/app_registration.py
```diff
@@ -0,0 +1,85 @@
...
+                # all kinds of limitations as to what can be imported
+                # in django's apps.py and this could cause issues
+                app_config.cms_app = cms_app_classes[0]()
+            else:
+                raise ImproperlyConfigured(
+                    "cms_config.py files must define exactly one "
+                    "class which inherits from CMSAppConfig")
+
+
+def get_cms_apps():
```
czpython: decorate with `lru_cache(max_size=None)`

czpython:
@monikasulik 
My answers below:

1. I'm not a big fan of list based config, mostly because it makes it a bit more difficult to turn off one feature (moderation) in an installation. One of my ideas for the boolean approach was to add a setting which defaults to True for each app supported. Example:

```python
djangocms_moderation_enabled = settings.ENABLE_MODERATION_FOR_BLOG
```

We can achieve the same with list/tuple approach but IMHO is less readable.

2. After seeing the app in full (via this pr) and based on your initial feedback, I'm now convinced that separate classes will be cleaner. So 👍 from me.

3. The way I saw settings was using the app name as prefix, so something like djangocms_moderation_<option>, but I'm not sure is something we need to do automatically.
I'm seeing more as convention because its up to the app to use/validate its own setting.

4. This is already an existing issue with django apps, so I don't think is something we need to address. The general rule is, last loaded app wins.

mmoravcik:
@czpython, re point 1 in your answer.

I thought that the purpose of `foo_enabled` is to determine if the app supports `foo` integration, not that it is enabled for this integration? E.g. if I look at some package, and I don't see `foo_enabled`, I am not sure if the app supports `foo` or not. That was the idea behind `supported_apps = ['foo', 'bar']` and then we can even have the extra one `foo_enabled=settings.FOO_ENABLED`.

monikasulik:
@czpython I think I've acted on all comments now.

monikasulik:
@evildmp I have now improved the PR description. Let me know if you need any extra info to write the actual docs.

monikasulik:
@czpython From what I can see tests are failing because `lru_cache` does not exist in python 2.7. I'm not really sure what kind of conventions you follow for these sorts of things or how you would like me to handle it.

cms/tests/test_app_registration.py
```diff
@@ -0,0 +1,313 @@
...
+            app_list[0].cms_extension.__class__.__name__, 'CMSSomeFeatureConfig')
+        self.assertTrue(hasattr(app_list[1], 'cms_extension'))
+        self.assertEqual(
+            app_list[1].cms_extension.__class__.__name__, 'CMSExtension')
+        self.assertFalse(hasattr(app_list[2], 'cms_extension'))
+
+    @override_settings(INSTALLED_APPS=[
+        'cms.test_utils.project.app_with_bad_cms_file',
+    ])
+    def test_raises_exception_raised_in_cms_file(self):
```
czpython: rename to `test_exception_propagates_from_cms_file`

cms/app_registration.py
```diff
@@ -0,0 +1,137 @@
...
+    Helper function.
+
+    Returns the class inheriting from CMSAppConfig in the given module.
+    If no such class exists in the module, returns None.
+    If multiple classes inherit from CMSAppConfig, raises
+    ImproperlyConfigured exception.
+    """
+    cms_config_classes = []
+    # Find all classes that inherit from CMSAppConfig
+    for name, obj in inspect.getmembers(cms_module):
```
czpython: this logic should be a function because is used by the `_find_extension` above.

cms/app_registration.py
```diff
@@ -0,0 +1,137 @@
...
+def configure_cms_apps(apps_with_features):
+    """
+    Check installed apps for apps that are configured to use cms addons
+    and run code to register them with their config
+    """
+    for app_with_feature in apps_with_features:
+        enabled_property = "{}_enabled".format(app_with_feature.label)
+        configure_app = app_with_feature.cms_extension.configure_app
+
+        for app_config in apps.get_app_configs():
```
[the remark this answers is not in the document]
monikasulik: In the spirit of YAGNI, wouldn't it be better to add that when we actually need it rather than before? It's a simple thing to add later if it is needed.

cms/app_registration.py
```diff
@@ -0,0 +1,137 @@
...
+            if not hasattr(app_config, 'cms_config'):
+                # Not a cms app with a config, so ignore
+                continue
+            if getattr(app_config.cms_config, enabled_property, False):
+                # Feature enabled for this app so configure
+                configure_app(app_config)
```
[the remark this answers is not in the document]
monikasulik: Ok, that will mean that I'll also need to override the `__init__` to take the django app config as in your original design. People might need things like app label, verbose_name etc. for things.

cms/app_registration.py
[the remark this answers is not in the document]
czpython: We would use it here, and will be good because then we can compute it once at runtime

cms/app_registration.py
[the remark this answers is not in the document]
czpython: Yup

cms/tests/test_app_registration.py
```diff
@@ -0,0 +1,313 @@
...
+from cms.utils import setup
+from cms.test_utils.testcases import CMSTestCase
+
+
+class AutodiscoverTestCase(CMSTestCase):
+
+    @override_settings(INSTALLED_APPS=[
+        'cms.test_utils.project.app_with_cms_config',
+        'cms.test_utils.project.app_with_cms_feature_and_config',
+        'cms.test_utils.project.app_without_cms_file'
```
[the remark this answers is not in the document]
monikasulik: The reason it's there is to test that when cms_config.py isn't defined, cms_config and cms_extension do not get added. IMO either leave it as it is (where we test all three cases for cms_config in one test and all three for cms_extension in the other) OR separate these tests into 6 tests where each of these apps gets tested for both attributes OR separate these tests into 3 tests where each app gets tested for both cms_config and cms_extension in one test. Separating only the without_cms_file one into 1 or 2 extra tests would be more confusing IMO. So would rather you chose one of the 3 options above.

cms/app_registration.py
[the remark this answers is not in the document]
monikasulik: You want one with all cms apps (both extensions and configs) or one with just the cms apps that define a config? Cause actually I think in this case it makes more sense if it's just the apps that have a cms_config attribute.

cms/tests/test_app_registration.py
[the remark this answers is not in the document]
czpython: Separating into 3 tests seems like a good compromise.

cms/app_registration.py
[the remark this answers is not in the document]
czpython: Correct, just the ones with cms config

cms/app_registration.py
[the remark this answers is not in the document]
monikasulik: Now that I've just done this and looked at how the code is now, it has occurred to me that it would probably be neater to do `def configure_app(cms_config, app_config)` rather than override init and do `cms_config.app_config` :wink: Thoughts?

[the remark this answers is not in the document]
czpython: Don't like it because it implies that they're different while they're not, most cases will not need to use app_config.

monikasulik:
@czpython I think all your comments have been addressed now.
Still need to update the PR description though @evildmp because the changes have changed how people would interact with app registration.

cms/app_base.py
```diff
@@ -93,3 +94,32 @@ def get_urls(self, page=None, language=None, **kwargs):
...
+
+        This method will be run once for every app that defines an
+        attribute like "<app_label>_enabled" as True on its cms app
+        config class.
+        So for example if app A with label "app_a" implements this
+        method and app B and app C define app_a_enabled = True on their
+        cms config classes, the method app A has defined will run twice,
+        once for app B and once for app C.
+
+        :param app: django app that has defined the feature as enabled
```
czpython: param is now `cms_config`

cms/app_base.py
```diff
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from abc import ABC, abstractmethod
```
czpython: Have a look at https://stackoverflow.com/a/35673504/389453 for python2 and 3 compat

cms/test_utils/project/app_with_cms_feature/cms_config.py
```diff
@@ -0,0 +1,31 @@
+from cms.app_base import CMSAppExtension
+
+
+def some_function(*args, **kwargs):
```
czpython: this is a bit odd. what does this test?

cms/app_registration.py
```diff
@@ -0,0 +1,142 @@
...
+
+from django.apps import apps
+from django.utils.lru_cache import lru_cache
+from django.utils.module_loading import module_has_submodule
+from django.core.exceptions import ImproperlyConfigured
+
+from cms.app_base import CMSAppConfig, CMSAppExtension
+
+
+CMS_CONFIG_NAME = 'cms_config'
```
czpython: I think this can go into the `constants.py`

cms/app_registration.py
```diff
@@ -0,0 +1,142 @@
...
+def _find_subclasses(cms_module, klass):
+    """
+    Helper function.
+
+    Returns a list of classes in cms_module which inherit from klass.
+    """
+    classes = []
+    # Find all classes that inherit from klass
+    for name, obj in inspect.getmembers(cms_module):
+        is_cms_config = (
```
czpython: rename to is_subclass

cms/app_registration.py
```diff
@@ -0,0 +1,142 @@
...
+from django.utils.module_loading import module_has_submodule
+from django.core.exceptions import ImproperlyConfigured
+
+from cms.app_base import CMSAppConfig, CMSAppExtension
+
+
+CMS_CONFIG_NAME = 'cms_config'
+
+
+def _find_subclasses(cms_module, klass):
```
czpython: cms_module -> module

cms/app_registration.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Helper function.
+
+    Returns a list of classes in cms_module which inherit from klass.
```
czpython: cms_module -> module

cms/app_registration.py
```diff
@@ -0,0 +1,142 @@
...
+            extension = _find_extension(cms_module)
+            # We are adding these attributes here rather than in
+            # django's app config definition because there are
+            # all kinds of limitations as to what can be imported
+            # in django's apps.py and leaving it to devs to define this
+            # there could cause issues
+            if config:
+                app_config.cms_config = config(app_config)
+            if extension:
+                app_config.cms_extension = extension()
```
czpython: should we also pass app_config here for consistency?

cms/tests/test_app_registration.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def setUp(self):
+        # The result of get_cms_config_apps is cached. Clear this cache
+        # because installed apps change between tests and therefore
+        # unlike in a live environment, results of this function
+        # can change between tests
+        app_registration.get_cms_config_apps.cache_clear()
+
+    @patch.object(apps, 'get_app_configs')
+    def test_runs_configure_app_method_for_app_with_enabled_config(
```
czpython: one line is fine here

cms/tests/test_app_registration.py
```diff
@@ -0,0 +1,351 @@
...
+
+        app_registration.configure_cms_apps([feature_app])
+
+        # If an app has enabled a feature, the configure method
+        # for that feature should have run with that app as the arg
+        feature_app.cms_extension.configure_app.assert_called_once_with(
+            config_app.cms_config)
+
+    @patch.object(apps, 'get_app_configs')
+    def test_doesnt_run_configure_app_method_for_disabled_app(
```
czpython: one line

cms/tests/test_app_registration.py
```diff
@@ -0,0 +1,351 @@
...
+            feature_app, config_app_disabled1, config_app_disabled2]
+
+        app_registration.configure_cms_apps([feature_app])
+
+        # If an app has not enabled a feature, the configure method
+        # for that feature should not have been run for that app
+        self.assertFalse(feature_app.cms_extension.configure_app.called)
+
+    @patch.object(apps, 'get_app_configs')
+    def test_doesnt_raise_exception_if_not_cms_app(
```
czpython: one line

cms/tests/test_app_registration.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # An app that does not define a cms config should just be
+        # ignored and not cause any exceptions
+        try:
+            app_registration.configure_cms_apps([feature_app])
+        except AttributeError:
+            self.fail("Exception raised when cms app config not defined")
+
+    @patch.object(apps, 'get_app_configs')
+    def test_runs_configure_app_method_for_correct_apps_when_multiple_apps(
```
czpython: here is ok

cms/test_utils/project/app_with_cms_feature/cms_config.py
monikasulik: As the comment in the next line says - that autodiscover doesn't pick this up :) Just wanted to add some random functions and classes in there, so that we know that autodiscover only ever picks up classes that inherit from CMSAppConfig and CMSAppExtension. Possibly now a bit redundant because the separation into two classes means that the code would have to check the subclass anyway, which wasn't true before.

cms/app_registration.py
monikasulik: IMO no because nobody should really be accessing `app_config.cms_extension` anyway, so there's no need for `app_config.cms_extension.django_app`. They are two different classes so don't think they need the init to be the same.

cms/test_utils/project/app_with_cms_feature/cms_config.py
czpython: Yeah I think is fine to remove this

cms/app_registration.py
czpython: Ok

PR #6429, opened 2018-06-27
Fixed #6428 -- Force page nodes to render based on the request page instead of request mode

### LINKED ISSUE #6428: Menu node not marked as selected
When requesting a page in edit mode with a staff user with no permissions to change the page, the menu will not mark the current page as selected.

This is because the menu incorrectly renders the draft menu while the request page object is public because the user had no permission to change the draft.

czpython:
Fixed #6428


tschale:
Why was this just closed?

PR #6436, opened 2018-07-05
Integrate wizards with app registration

monikasulik:
### Summary

Implement wizards via the new app registration system. Ensure backwards compatibility

This changes (although ensures backwards compatibility with) how wizards will be registered with the cms going forwards. From now on to register wizards one needs to do the following:

1. Create a cms_config.py file in the app (if one doesn't exist already)
1. Implement a class that inherits from `app_base.CMSAppConfig` (if it doesn't exist already) and make sure it has the following attributes:
```
class Config(CMSAppConfig):
    cms_enabled = True
    cms_wizards = [wizard1, wizard2]
```
where wizard1 and wizard2 are wizards as they are currently defined in cms_wizards.py

### Documentation checklist

    * [ ] general notes
    * [ ] bug-fixes
    * [ ] required upgrade steps
    * [ ] names of contributors

### DIFF
--- a/cms/cms_config.py
+++ b/cms/cms_config.py
@@ -1,0 +1,22 @@
+from cms.app_base import CMSAppExtension, CMSAppConfig
+from cms.cms_wizards import cms_page_wizard, cms_subpage_wizard
+
+
+class CMSCoreConfig(CMSAppConfig):
+    cms_enabled = True
+    cms_wizards = [cms_page_wizard, cms_subpage_wizard]
+
+
+class CMSCoreExtensions(CMSAppExtension):
+
+    def __init__(self):
+        self.wizards = {}  # this is instead of wizard_pool._entries
+
+    def configure_wizards(self, cms_config):
+        if not hasattr(cms_config, 'cms_wizards'):
+            return
+        for wizard in cms_config.cms_wizards:
+            self.wizards[wizard.id] = wizard
+
+    def configure_app(self, cms_config):
+        self.configure_wizards(cms_config)
--- a/cms/cms_config.py
+++ b/cms/cms_config.py
@@ -13,9 +13,13 @@ class CMSCoreConfig(CMSAppConfig):
 class CMSCoreExtensions(CMSAppExtension):
 
     def __init__(self):
-        self.wizards = {}  # this is instead of wizard_pool._entries
+        self.wizards = {}
 
     def configure_wizards(self, cms_config):
+        """
+        Adds all registered wizards from apps that define them to the
+        wizards dictionary on this class
+        """
         if not hasattr(cms_config, 'cms_wizards'):
             # The cms_wizards settings is optional. If it's not here
             # just move on.
@@ -27,8 +31,8 @@ class CMSCoreExtensions(CMSAppExtension):
                 raise ImproperlyConfigured(
                     "all wizards defined in cms_wizards must inherit "
                     "from cms.wizards.wizard_base.Wizard")
-        for wizard in cms_config.cms_wizards:
-            self.wizards[wizard.id] = wizard
+            else:
+                self.wizards[wizard.id] = wizard
 
     def configure_app(self, cms_config):
         self.configure_wizards(cms_config)
--- a/cms/cms_config.py
+++ b/cms/cms_config.py
@@ -1,3 +1,5 @@
+from collections import Iterable
+
 from django.core.exceptions import ImproperlyConfigured
 
 from cms.app_base import CMSAppExtension, CMSAppConfig
@@ -20,19 +22,27 @@ class CMSCoreExtensions(CMSAppExtension):
         Adds all registered wizards from apps that define them to the
         wizards dictionary on this class
         """
-        if not hasattr(cms_config, 'cms_wizards'):
-            # The cms_wizards settings is optional. If it's not here
-            # just move on.
-            return
-        if type(cms_config.cms_wizards) != list:
-            raise ImproperlyConfigured("cms_wizards must be a list")
+        if not isinstance(cms_config.cms_wizards, Iterable):
+            raise ImproperlyConfigured("cms_wizards must be iterable")
         for wizard in cms_config.cms_wizards:
             if not isinstance(wizard, Wizard):
                 raise ImproperlyConfigured(
-                    "all wizards defined in cms_wizards must inherit "
-                    "from cms.wizards.wizard_base.Wizard")
+                    "All wizards defined in cms_wizards must inherit "
+                    "from cms.wizards.wizard_base.Wizard"
+                )
+            elif wizard.id in self.wizards:
+                pass
+                # TODO: This is currently breaking various tests
+                # raise ImproperlyConfigured(
+                #    "Wizard for model {} has already been registered".format(
+                #        wizard.get_model()
+                #    )
+                #)
             else:
                 self.wizards[wizard.id] = wizard
 
     def configure_app(self, cms_config):
-        self.configure_wizards(cms_config)
+        # The cms_wizards settings is optional. If it's not here
+        # just move on.
+        if hasattr(cms_config, 'cms_wizards'):
+            self.configure_wizards(cms_config)
--- a/cms/cms_config.py
+++ b/cms/cms_config.py
@@ -1,3 +1,4 @@
+from logging import getLogger
 from collections import Iterable
 
 from django.core.exceptions import ImproperlyConfigured
@@ -5,6 +6,9 @@ from django.core.exceptions import ImproperlyConfigured
 from cms.app_base import CMSAppExtension, CMSAppConfig
 from cms.cms_wizards import cms_page_wizard, cms_subpage_wizard
 from cms.wizards.wizard_base import Wizard
+
+
+logger = getLogger(__name__)
 
 
 class CMSCoreConfig(CMSAppConfig):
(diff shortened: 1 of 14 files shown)


cms/wizards/wizard_pool.py
```diff
@@ -14,82 +15,63 @@ def entry_choices(user, page):
...
 class WizardPool(object):
-    _entries = {}
-    _discovered = False
-
-    def __init__(self):
-        self._reset()
-
-    # PRIVATE METHODS -----------------
-
-    def _discover(self):
```
monikasulik: After thinking about this later, I'm getting the impression I may have messed up backwards compatibility with this. Because although the method `wizard_pool.register` still works, there is now no code that actually looks for cms_wizards.py
@czpython Where would you like the code that searches for cms_wizards.py files to live now?

cms/tests/test_cms_config.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def test_raises_exception_if_doesnt_inherit_from_wizard_class(self):
+        """
+        If one or more of the wizards defined in the wizard setting
+        do not inherit from the Wizard class, raise an exception
+        """
+        extensions = CMSCoreExtensions()
+        wizard = Mock(id=3, spec=object)
+        cms_config = Mock(
+            cms_enabled=True, cms_wizards=[wizard])
```
czpython: single line

cms/tests/test_cms_config.py
```diff
@@ -0,0 +1,97 @@
...
+        with self.assertRaises(ImproperlyConfigured):
+            extensions.configure_wizards(cms_config)
+
+    def test_raises_exception_if_not_list(self):
+        """
+        If the wizard setting isn't a list, raise an exception.
+        """
+        extensions = CMSCoreExtensions()
+        wizard = Mock(id=6, spec=Wizard)
+        cms_config = Mock(
```
czpython: single line

cms/tests/test_cms_config.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        extensions = CMSCoreExtensions()
+        wizard = Mock(id=3, spec=object)
+        cms_config = Mock(
+            cms_enabled=True, cms_wizards=[wizard])
+
+        with self.assertRaises(ImproperlyConfigured):
+            extensions.configure_wizards(cms_config)
+
+    def test_raises_exception_if_not_list(self):
```
czpython: if not iterable

cms/tests/test_wizards.py
```diff
@@ -154,74 +163,109 @@ def test_endpoint_auth_required(self):
...
-        wizard_pool._reset()
-        self.assertFalse(wizard_pool._discovered)
-        self.assertEqual(len(wizard_pool._entries), 0)
-        wizard_pool._discover()
-        self.assertTrue(wizard_pool._discovered)
+    def tearDown(self):
+        # Clean up in case anything has been removed or added to the
+        # registered wizards, so other tests don't have problems
+        extension = apps.get_app_config('cms').cms_extension
+        expected_wizards = {
```
czpython: this means we would have to keep this dict updated for every new wizard in core or tests.
I think we can just trigger a reload of the apps no?

cms/tests/test_wizards.py
```diff
@@ -154,74 +163,109 @@ def test_endpoint_auth_required(self):
...
+            extension.wizards = expected_wizards
 
-    def test_register_unregister_isregistered(self):
-        wizard_pool._clear()
-        self.assertEqual(len(wizard_pool._entries), 0)
-        wizard_pool.register(self.page_wizard)
-        # Now, try to register the same thing
-        with self.assertRaises(AlreadyRegisteredException):
-            wizard_pool.register(self.page_wizard)
+    def test_discovered_returns_true(self):
```
czpython: can be removed

cms/tests/test_wizards.py
```diff
@@ -154,74 +163,109 @@ def test_endpoint_auth_required(self):
...
-        self.assertSequencesEqual(entries, wizards)
-
-        wizard_pool._clear()
-        wizard_pool.register(self.user_settings_wizard)
         wizard_pool.register(self.page_wizard)
-        wizards = [self.page_wizard, self.user_settings_wizard]
-        wizards = sorted(wizards, key=lambda e: getattr(e, 'weight'))
-        entries = wizard_pool.get_entries()
-        self.assertSequencesEqual(entries, wizards)
+
```
czpython: remove newline

cms/tests/test_wizards.py
```diff
@@ -154,74 +163,109 @@ def test_endpoint_auth_required(self):
...
-        wizard_pool._clear()
-        wizard_pool.register(self.page_wizard)
-        entry = wizard_pool.get_entry(self.page_wizard)
-        self.assertEqual(entry, self.page_wizard)
+    def test_unregister_registered_wizard(self):
+        """
+        Test for backwards compatibility of the unregister method.
+        Removes a wizard from the wizards dict.
+        """
+        was_unregistered = wizard_pool.unregister(cms_page_wizard)
```
czpython: remove newline

cms/tests/test_wizards.py
```diff
@@ -507,3 +551,91 @@ def test_create_page_with_existing_slug(self):
...
+
+class TestEntryChoices(CMSTestCase):
+
+    def test_generates_choices_in_weighted_order(self):
+        """
+        The entry_choices function returns the wizards ordered by weight
+        """
+        user = self.get_superuser()
+        page = create_page('home', 'nav_playground.html', 'en', published=True)
+
```
czpython: remove newline

cms/tests/test_wizards.py
```diff
@@ -507,3 +551,91 @@ def test_create_page_with_existing_slug(self):
...
+        wizard_choices = [option for option in entry_choices(user, page)]
+
```
czpython: remove newline

cms/tests/test_wizards.py
```diff
@@ -507,3 +551,91 @@ def test_create_page_with_existing_slug(self):
...
+    def test_doesnt_generate_choice_if_user_doesnt_have_permission(self):
+        """
+        The entry_choices function only returns the wizards that the
+        user has permissions for
+        """
+        user = self.get_superuser()
+        page = create_page('home', 'nav_playground.html', 'en', published=True)
+
+        wizard_choices = [option for option in entry_choices(user, page)]
+
```
czpython: remove newline

cms/tests/test_wizards.py
```diff
@@ -507,3 +551,91 @@ def test_create_page_with_existing_slug(self):
...
+        Mock(return_value=False)
+    )
+    def test_doesnt_generate_choice_if_user_doesnt_have_permission(self):
+        """
+        The entry_choices function only returns the wizards that the
+        user has permissions for
+        """
+        user = self.get_superuser()
+        page = create_page('home', 'nav_playground.html', 'en', published=True)
+
```
czpython: remove newline

cms/cms_config.py
```diff
@@ -0,0 +1,38 @@
...
+            # just move on.
+            return
+        if type(cms_config.cms_wizards) != list:
+            raise ImproperlyConfigured("cms_wizards must be a list")
+        for wizard in cms_config.cms_wizards:
+            if not isinstance(wizard, Wizard):
+                raise ImproperlyConfigured(
+                    "all wizards defined in cms_wizards must inherit "
+                    "from cms.wizards.wizard_base.Wizard")
+            else:
```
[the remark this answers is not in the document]
monikasulik: So I did this locally and am now struggling with random tests failing. What seems to be happening is:

1. The test setup is loading cms and sampleapp as expected and their wizards are getting registered
1. The individual tests use the `override_settings` decorator with INSTALLED_APPS and include the cms app on the revised list, which in turn triggers another autodiscover thing because the cms's ready method is getting called twice. And therefore raises an error.

I can think of some ways of fixing this, but all these ways seem really hacky and will be really annoying and unobvious for future test writers. Do you have an idea for a nice way of doing this?
Alternatively, because we're using a dict, nothing bad is actually going to happen if a wizard is added to the dict twice (it'll just override the previous wizard), so we could just change this to a `logger.warning`? Really our code can cope with double registration of wizards, it's just that the developer has done something fishy if that's what's happening, so we should warn them.

cms/tests/test_wizards.py
monikasulik: There could be side effects to that (such as getting an exception for registering a wizard twice, but there could be more stuff like that once more things are integrated with app registration). But I think there are other ways I can be more generic here.

cms/wizards/wizard_pool.py
monikasulik: Ok, this is bizarre. Just wrote a test with an app that doesn't have cms_config.py but only cms_wizards.py and it seems to have registered the wizard from cms_wizards.py without me adding this into the init! Will check what's up there Monday morning, but possibly all that is needed is that test?

czpython: Likely there is an import somewhere of cms_wizards.py which should be removed

cms/cms_config.py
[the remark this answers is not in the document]
czpython: Is the app config instance setup on the first load, persisting through to the second load on the tests that use `override_settings`?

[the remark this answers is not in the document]
monikasulik: @czpython I believe so, yes. I seem to be struggling on a similar issue with the integration test on backwards compatibility (needed for the work on retaining the autodiscover of cms_wizards) as well. I've just pushed that.

cms/wizards/helpers.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    Returns a list of (wizard.id, wizard) tuples (for all registered
+    wizards) ordered by weight
+    """
+    wizards = apps.get_app_config('cms').cms_extension.wizards
+    return [value for (key, value) in sorted(
+            wizards.items(), key=lambda e: getattr(e[1], 'weight'))]
+
+
+def get_entry(entry_key):
```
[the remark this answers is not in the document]
monikasulik: Done

cms/wizards/wizard_pool.py
monikasulik: So, I can't use the `__init__` in CMSCoreExtensions due to the order in which things happen (it means attempting to add wizards to a CMSCoreExtensions instance that doesn't exist yet). However, I have added this elsewhere. Hope this will work for you.

PR #6437, opened 2018-07-05
Refactored plugin tree

czpython:


cms/models/pluginmodel.py
```diff
@@ -21,7 +19,18 @@
 from cms.utils.conf import get_cms_setting
 from cms.utils.urlutils import admin_reverse
 
-from treebeard.mp_tree import MP_Node
+
```
kinkerl: raw sql? what dialects are supported?

cms/models/pluginmodel.py
```diff
@@ -134,6 +148,21 @@ def __repr__(self):
...
 
+    @classmethod
+    def _get_database_connection(cls, action):
+        return {
+            'read': connections[router.db_for_read(cls)],
+            'write': connections[router.db_for_write(cls)]
+        }[action]
+
+    @classmethod
+    def get_database_vendor(cls, action):
```
kinkerl: unused?

cms/models/pluginmodel.py
czpython: We'll support Postgres for now.
Next pr will add sqlite and mysql support

cms/models/pluginmodel.py
czpython: Added for the next pr

czpython:
@kinkerl 
The plugin architecture will only have a single field for tree related stuff which is the position field.
We don't have an internal documentation for this at the moment.

There will be no replacement for fix-tree because positions are absolute.

cms/admin/placeholderadmin.py
```diff
@@ -856,107 +764,63 @@ def _paste_placeholder(self, request, plugin, target_language,
...
-            'placeholder': target_placeholder,
-        }
-
-        source_language = plugin.language
-        source_placeholder = plugin.placeholder
-        source_tree_order = source_placeholder.get_plugin_tree_order(
-            language=source_language,
-            parent_id=plugin.parent_id,
-        )
+        if target_placeholder and not self.has_move_plugin_permission(request, plugin, source_placeholder):
```
chaosk: Shouldn't the last argument passed here be `target_placeholder`?

cms/admin/placeholderadmin.py
```diff
@@ -856,107 +764,63 @@ def _paste_placeholder(self, request, plugin, target_language,
...
             operation=operations.MOVE_PLUGIN,
             plugin=plugin,
-            source_language=source_language,
+            source_language=plugin.language,
             source_placeholder=source_placeholder,
             source_parent_id=plugin.parent_id,
-            source_order=source_tree_order,
-            target_language=target_language,
-            target_placeholder=target_placeholder,
+            target_language=language,
```
chaosk: Why the discrepancy between `source_language` and this? `language` point to the same thing

PR #6438, opened 2018-07-09
Removed unpublish button from toolbar

Aiky30:
Removed the unpublish features from the toolbar. No tests failed. The publish and unpublish buttons are still present on the page drop down in the toolbar.


Aiky30:
@czpython I have updated and removed some of the unused functions. :-)

PR #6439, opened 2018-07-09
Bumped the version number up to release 4.0.0 to avoid confusion with…

Aiky30:
There has been a few cases where developers have been working on the wrong version of the CMS, especially when testing the new app integration features. Please let me know if I need to change anything.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,3 @@
-=== 4.0.0 (unreleased) ===
-
-
 === 3.6.0 (unreleased) ===
 
 * Removed the ``cms moderator`` command.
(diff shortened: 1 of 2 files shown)


Aiky30:
@czpython I've made the changes although there seems to be a FE test failing that isn't in the FE list of accepted.

Aiky30:
The gulp build file has the following which would generate a build with the version hardcoded so I'm assuming there is a FE test looking for the version:

glpfile.js:
var CMS_VERSION = fs.readFileSync('cms/__init__.py', { encoding: 'utf-8' }).match(/__version__ = '(.*?)'/)[1];

Aiky30:
I've tried running the FE build and testing myself and it explodes. Haha. What's the CI setup, does it generate the FE build for each commit? Would make sense I guess.

czpython:
Closing this for now

PR #6445, opened 2018-07-12
Remove publish / unpublish buttons from page changelist

Aiky30:
- Remove publish button from changelist dropdown
- Remove unpublish button from changelist dropdown
- Remove any backend test that checks for these two buttons
- Leave the preview button
- New pages should show a green circle for the language that way just created
- New pages should show a white circle for the languages that are not yet configured


Aiky30:
@czpython Please review this PR and let me know if you want me to change anything. One thing I would like to raise is that I've used the old states for the colours, they are embedded into the sass styles and possibly js hooks. Let me know if they should also be changed. Ta.

PR #6446, opened 2018-07-12
Remove logic which publishes the first page page as soon as it is created

Aiky30:
- Remove logic in forms which publishes the first page of a site
- Remove tests that check for this


cms/test_utils/testcases.py
```diff
@@ -265,6 +265,7 @@ def get_new_page_data(self, parent_id=''):
             'title': 'test page %d' % self.counter,
             'slug': 'test-page-%d' % self.counter,
             'parent_node': parent_id,
+            'publish': True,  # Temp added as most tests require the first page to be published
```
czpython: this is not a field on the form, is not used.

PR #6452, opened 2018-07-23
Fil 190: Adding Generic Foreign Key to Placeholder model

anirbanlahiri-fidelity:


cms/models/placeholdermodel.py
```diff
@@ -38,6 +40,10 @@ class Placeholder(models.Model):
     cache_placeholder = True
     is_static = False
     is_editable = True
+    content_type = models.ForeignKey(ContentType, blank=True, null=True,
```
czpython: put each arg on its own line

czpython:
@anirbanlahiri-fidelity Make sure to rebase this with latest release/4.0.x branch

cms/migrations/0026_remove_page_placeholders.py
```diff
@@ -8,7 +8,8 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('cms', '0024_title_placeholders_data_migration'),
+        ('cms', '0023_auto_20180719_1638'),
```
czpython: remove 0023 from here

cms/models/placeholdermodel.py
```diff
@@ -35,6 +37,12 @@ class Placeholder(models.Model):
     """
     slot = models.CharField(_("slot"), max_length=255, db_index=True, editable=False)
     default_width = models.PositiveSmallIntegerField(_("width"), null=True, editable=False)
+    content_type = models.ForeignKey(ContentType,
```
czpython:
```
content_type = models.ForeignKey(
    ContentType,
    blank=True,
    null=True,
    on_delete=models.SET_NULL,
)
```

cms/migrations/0023_auto_20180719_1638.py
```diff
@@ -0,0 +1,27 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contenttypes', '0002_remove_content_type_name'),
+        ('cms', '0021_auto_20180507_1432'),
```
czpython: should depend on 0022

cms/migrations/0023_auto_20180719_1638.py
```diff
@@ -0,0 +1,26 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('cms', '0022_auto_20180620_1551'),
```
czpython: contenttypes should be added as dependency here

anirbanlahiri-fidelity: Krzysztof asked me to remove it and have a purely linear dependency flow.  I will put it back in if you say.

chaosk: I only talked about cms migrations depending on each other in linear fashion. This migration uses contenttypes, so it also needed to be in the dependency list.

anirbanlahiri-fidelity: Sorry, I misunderstood you then, since 0010_migrate_use_structure.py had a dependency on contenttypes.

PR #6454, opened 2018-07-23
Remove revert to live feature

Aiky30:
### Summary

Remove revert to live from toolbar
Remove tests for this feature
Remove revert_to_live method on Page model

### Documentation checklist

    * [ ] general notes
    * [ ] bug-fixes
    * [ ] required upgrade steps
    * [ ] names of contributors


cms/admin/pageadmin.py
```diff
@@ -718,19 +717,6 @@ def has_publish_permission(self, request, obj=None):
         site = self.get_site(request)
         return page_permissions.user_can_publish_page(request.user, page=obj, site=site)
 
-    def has_revert_to_live_permission(self, request, language, obj=None):
-        if not obj:
-            return False
-
-        site = self.get_site(request)
-        has_perm = page_permissions.user_can_revert_page_to_live(
```
czpython: when doing these kinds of removals, make sure to also remove the function being called (if not used anywhere else). So in this case remove `user_can_revert_page_to_live` and if there's a permission attached to it, also remove that.

Aiky30: @czpython I have done as you requested. In this instance the permission is also used elsewhere in the file so I believe that the permission should stay. Let me know if I've misunderstood. Ta.

Aiky30: Just spotted and removed a redundant new line. Will need to let it build and pass. It was a pass before this change so I don't see any issues with it going ahead.

czpython: Which permission are you referring to?

Aiky30: @czpython you mentioned "if there's a permission attached to it, also remove that.". There was a decorator with a permission label which is used in other parts. I figured you were referring to that.

Aiky30: -@auth_permission_required('change_page')

PR #6467, opened 2018-08-01
Fix wrong field name

schwobaseggl:
`extended_page` is not a field of a `PageExtension`. This should be `extended_object`.


vThaian:
Thanks, @schwobaseggl for your contribution 👍

PR #6482, opened 2018-08-20
Preview, Structure and Edit endpoints

Aiky30:
### Summary

Create dedicated Preview, Structure and Edit endpoints.

### Documentation checklist

    * [ ] general notes
    * [ ] bug-fixes
    * [ ] required upgrade steps
    * [ ] names of contributors


cms/admin/placeholderadmin.py
```diff
@@ -46,6 +46,7 @@
     has_reached_plugin_limit,
 )
 from cms.utils.urlutils import admin_reverse
+from cms.views import render_object_edit, render_object_structure, render_object_preview
```
czpython: add newline under this import

cms/middleware/toolbar.py
```diff
@@ -22,6 +23,14 @@ class ToolbarMiddleware(MiddlewareMixin):
     Middleware to set up CMS Toolbar.
     """
 
+    def is_edit_mode(self, request):
+        try:
+            match = resolve(request.path_info)
```
czpython: why not request.resolver_match

cms/toolbar/utils.py
```diff
@@ -116,3 +118,24 @@ def get_toolbar_from_request(request):
     from .toolbar import EmptyToolbar
 
     return getattr(request, 'toolbar', EmptyToolbar(request))
+
+
+def get_object_edit_url(obj):
+    content_type = ContentType.objects.get_for_model(obj)
+
+    with force_language(get_current_language()):
```
czpython: don't hardcode current language, pass language as a parameter and if none fallback to django's get_current_language instead of the cms one

czpython: This applies to all three functions

cms/utils/conf.py
```diff
@@ -273,6 +279,7 @@ def get_unihandecode_host():
     'CMS_TOOLBAR_URL__EDIT_OFF': get_toolbar_url__edit_off,
```
czpython: remove all these edit off / build, etc..

cms/toolbar/toolbar.py
```diff
@@ -74,28 +76,31 @@ def structure_renderer(self):
 
     @cached_property
     def structure_mode_active(self):
-        structure = get_cms_setting('CMS_TOOLBAR_URL__BUILD')
-        return self.is_staff and structure in self.request.GET
+        try:
+            match = resolve(self.request.path_info)
```
czpython: use request resolver match

cms/toolbar/toolbar.py
```diff
@@ -74,28 +76,31 @@ def structure_renderer(self):
...
+            return False
+        return self.is_staff and match.url_name == 'cms_placeholder_render_object_structure'
 
     @cached_property
     def edit_mode_active(self):
         if not self.show_toolbar:
             return False
-        return self.structure_mode_active or self.content_mode_active
+        try:
+            match = resolve(self.request.path_info)
```
czpython: same as above

cms/toolbar/toolbar.py
```diff
@@ -74,28 +76,31 @@ def structure_renderer(self):
...
+            return False
+        return self.structure_mode_active or match.url_name == 'cms_placeholder_render_object_edit'
 
     @cached_property
     def content_mode_active(self):
         if self.structure_mode_active:
             # Structure mode always takes precedence
             return False
-        return self.is_staff and self.request.session.get('cms_edit', False)
+        return self.is_staff and not self.edit_mode_active
```
czpython: content mode is active if is not structure and the user is on edit mode.
content mode != live mode

cms/toolbar/toolbar.py
```diff
@@ -445,9 +459,9 @@ def get_render_context(self):
         context = {
             'cms_toolbar': self,
             'cms_renderer': renderer,
-            'cms_edit_on': self.edit_mode_url_on,
-            'cms_edit_off': self.edit_mode_url_off,
-            'cms_structure_on': self.structure_mode_url_on,
+            'cms_edit_on': self.get_object_edit_url(),
+            'cms_edit_off': self.get_object_preview_url(),
```
czpython: please rename this to cms_edit_url, cms_preview_url, cms_structure_url

cms/toolbar/toolbar.py
```diff
@@ -314,25 +318,35 @@ def get_object_pk(self):
             return self.obj.pk
         return ''
 
-    def get_object_public_url(self):
+    def get_object_preview_url(self):
         if self.obj:
+            obj = self.obj
+
+            if obj.__class__.__name__ == Page.__name__:
+                obj = self.obj.get_title_obj(language=self.request_language)
```
czpython: do this in the view instead (set the title object as toolbar obj)

cms/toolbar/toolbar.py
```diff
@@ -314,25 +318,35 @@ def get_object_pk(self):
...
-                    except:
-                        pass
+                return get_object_edit_url(obj)
+        return ''
+
+    def get_object_structure_url(self):
+        if self.obj:
+            obj = self.obj
+
+            if obj.__class__.__name__ == Page.__name__:
```
czpython: same as above

cms/toolbar/toolbar.py
```diff
@@ -314,25 +318,35 @@ def get_object_pk(self):
...
-                    pass
+                return get_object_preview_url(obj)
         return ''
 
-    def get_object_draft_url(self):
+    def get_object_edit_url(self):
         if self.obj:
+            obj = self.obj
+
+            if obj.__class__.__name__ == Page.__name__:
```
czpython: same as above

cms/models/titlemodels.py
```diff
@@ -285,6 +286,12 @@ def get_xframe_options(self):
         except IndexError:
             return None
 
+    def get_absolute_url(self):
+        with force_language(self.language):
```
czpython: add language as parameter and call self.page.get_absolute_url(language)

cms/utils/helpers.py
```diff
@@ -82,3 +82,35 @@ def get_timezone_name():
     # Hence this paranoid conversion to create a valid cache key.
     tz_name = force_text(get_current_timezone_name(), errors='ignore')
     return tz_name.encode('ascii', 'ignore').decode('ascii').replace(' ', '_')
+
+
+def is_model_editable(model_class):
```
czpython: `is_editable_model`

cms/utils/helpers.py
```diff
@@ -82,3 +82,35 @@ def get_timezone_name():
...
+    # Check whether model_class has an admin class
+    # and whether its inherited from FrontendEditableAdminMixin
+    try:
+        admin_class = admin.site._registry[model_class]
+    except KeyError:
+        return False
+
+    if admin_class and issubclass(admin_class.__class__, FrontendEditableAdminMixin):
+        return True
+    return False
```
czpython: `return isinstance(admin_class, FrontendEditableAdminMixin)`

cms/views.py
```diff
@@ -159,15 +163,6 @@ def details(request, slug):
     # permission checks
     if page.login_required and not request.user.is_authenticated:
         return redirect_to_login(urlquote(request.get_full_path()), settings.LOGIN_URL)
-
-    if hasattr(request, 'toolbar'):
```
czpython: this is important

czpython: should be the title as mentioned on the toolbar.py comments

cms/views.py
```diff
@@ -188,3 +183,59 @@ def login(request):
...
+        content_type_obj = content_type.get_object_for_this_type(pk=object_id)
+    except ObjectDoesNotExist:
+        raise Http404
+
+    context = {
+        'object': content_type_obj,
+        'cms_toolbar': request.toolbar,
+    }
+
+    if hasattr(request, 'toolbar'):
```
czpython: call `get_toolbar_from_request`

cms/views.py
```diff
@@ -188,3 +183,59 @@ def login(request):
...
+    if not is_model_editable(content_type.model_class()):
+        raise Http404
+
+    try:
+        content_type_obj = content_type.get_object_for_this_type(pk=object_id)
+    except ObjectDoesNotExist:
+        raise Http404
+
+    if not hasattr(content_type_obj, 'get_absolute_url'):
+        raise NotImplementedError('get_absolute_url not implemented for object')
```
czpython: doesn't help end user.
return BadRequest()

cms/views.py
```diff
@@ -188,3 +183,59 @@ def login(request):
...
+
+    abs_url = content_type_obj.get_absolute_url()
+    match = resolve(abs_url)
+
+    if hasattr(request, 'toolbar'):
```
czpython: call `get_toolbar_from_request`

cms/views.py
```diff
@@ -188,3 +183,59 @@ def login(request):
...
+
+def render_object_preview(request, content_type_id, object_id):
+    content_type = ContentType.objects.get_for_id(content_type_id)
+
+    try:
+        content_type_obj = content_type.get_object_for_this_type(pk=object_id)
+    except ObjectDoesNotExist:
+        raise Http404
+
+    abs_url = content_type_obj.get_absolute_url()
```
czpython: this should also check for get_absolute_url

cms/views.py
```diff
@@ -188,3 +183,59 @@ def login(request):
...
+    match = resolve(abs_url)
+
+    if hasattr(request, 'toolbar'):
```
czpython: call get_toolbar_from_request

cms/middleware/toolbar.py
malindap92: during this time request.resolver_match is None. the same implementation is done in is_cms_request in the ToolbarMiddleware.

cms/toolbar/toolbar.py
malindap92: when going through the changes again, ``content_mode_active`` is not used anymore in ``edit_mode_active``. is it fine to delete this completely or is there a special usecase for this?

cms/middleware/toolbar.py
Aiky30: In the Middleware the resolver_match is always null in the Middleware. This is per Djangos docs:

https://docs.djangoproject.com/en/2.1/ref/request-response/#django.http.HttpRequest.resolver_match

cms/toolbar/toolbar.py
Aiky30: The resolver match is None here most of the time when set as a cached property. Removing the cached property allows resolver match to be used. When set as a cached property the only method of getting the information that resolver match would have is using this method.

Aiky30: This is most likely because Djangos cache is set by the Middleware.

czpython: Middleware should not trigger this property to be called.
This is why its a lazy object now

Aiky30: The failures occur (50+) when the RequestFactory is used. I believe that this is due to it accessing the toolbar Middleware directly to mock generating a response. The request factory is used directly in some tests and via the get_page_request method in the test cases:

https://github.com/divio/django-cms/pull/6482/files#diff-2a8a0d85dd243bf778b381658688d623R452

cms/toolbar/toolbar.py
monikasulik: There might be external apps using it (such as versioning :wink: ). If you're providing a different way to establish whether we're in edit, preview or structure mode then make sure that knowledge gets to us on versioning. And also probably you'll be wanting to keep all the old attributes for backwards compatibility (perhaps with deprecation warnings). But I guess that's for Paulo to decide.

PR #6487, opened 2018-08-22
Make on_delete configurable (#6083)

Julz089:
Fixes [issue](https://github.com/divio/django-cms/issues/6083#issuecomment-411738890)


filwaitman:
@Julz089 thanks for the PR!

The code looks good. For completion sake: would you be able to add some tests for it?

Julz089:
@filwaitman added some simple testcase
2 travis builds are failing, but it looks like a travis issue can you please take a look

filwaitman:
Hey @Julz089! Thanks for the tests. I've made a few changes, please take a look:

- Added another 3 tests:
  - test PageField.on_delete defaults to CASCADE
  - test PlaceholderField.on_delete can be set
  - test PlaceholderField.on_delete defaults to CASCADE
- Removed that migration file (as it is a test model)
- Moved the new test models to a test app
- Moved all those new tests to a new `test_fields.py` file.

I don't have push access to the repo as well, so I can't contribute directly to your PR. 

Could you please check [this patch file](https://gist.github.com/filwaitman/230f2527680a98e15f27648e9538ad9d), apply it locally (against your current PR branch) and let me know if you agree with my changes?

Julz089:
@filwaitman thanks for the refactor. The changes all make sense and should prevent future on_delete issues.

Julz089:
@filwaitman sorry I didn't think about it and pushed your changes to see the travis result.
Should I reset my branch and allow you access?

filwaitman:
Don't worry @Julz089, that's all good! No need to reset your branch.
Overall it looks good to me. @czpython please take a look when you have some time.

PR #6496, opened 2018-08-30
Placeholder source field implementation

malindap92:
### Summary

This PR consist of:
- Data migration to populate Placeholder source field for any content type which has a relationship to Placeholder.
- Implementation of usage of {% placeholder %} template tag in non-CMS objects.
- Implementation of generic relation field for Placeholder.

### Documentation checklist

    * [ ] general notes
    * [ ] bug-fixes
    * [ ] required upgrade steps
    * [ ] names of contributors


cms/models/fields.py
```diff
@@ -92,3 +94,15 @@ def formfield(self, **kwargs):
...
+
+
+class PlaceholderRelationField(GenericRelation):
+
+    def __init__(self, **kwargs):
+        kwargs.pop('object_id_field', None)
+        kwargs.pop('content_type_field', None)
+
+        super(PlaceholderRelationField, self).__init__(
+            Placeholder, object_id_field='object_id', content_type_field='content_type',
```
czpython: one arg on each line

cms/models/fields.py
```diff
@@ -92,3 +94,15 @@ def formfield(self, **kwargs):
...
         defaults.update(kwargs)
         return super(PageField, self).formfield(**defaults)
+
+
+class PlaceholderRelationField(GenericRelation):
+
+    def __init__(self, **kwargs):
+        kwargs.pop('object_id_field', None)
+        kwargs.pop('content_type_field', None)
+
```
czpython: remove newline

cms/migrations/0033_placeholder_source_data_migration.py
```diff
@@ -0,0 +1,64 @@
...
+
+from django.db import migrations
+
+
+def forwards(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    CMSPlugin = apps.get_model('cms', 'CMSPlugin')
+    Placeholder = apps.get_model('cms', 'Placeholder')
+    filtered_content_types = (
```
czpython: why not just go through all apps directly? instead of db query

cms/migrations/0033_placeholder_source_data_migration.py
```diff
@@ -0,0 +1,64 @@
...
+    for ct in filtered_content_types:
+        model_class = apps.get_model(ct.app_label, ct.model)
+
+        if issubclass(model_class, (CMSPlugin, Placeholder)):
+            # We ignore all models inherited from CMSPlugin or Placeholder.
+            continue
+
+        # Get all fields which has a relationship to Placeholder.
+        pl_related_fields = [
+            f.name for f in model_class._meta.get_fields()
```
czpython: collect fields instead of names only to be used below

cms/migrations/0033_placeholder_source_data_migration.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # Model has related field/s with Placeholder.
+        # Now get all the objects for the model and populate source field.
+        cur_ct_obj = ContentType.objects.get_for_model(model_class)
+
+        for obj in model_class.objects.using(db_alias):
+            for pl_field_name in pl_related_fields:
+                pl_field = getattr(obj, pl_field_name)
+
+                if pl_field.__class__.__name__ == 'ManyRelatedManager':
```
czpython: use the collected fields above to avoid doing this check.
there's some methods on the field instances like get_queryset() which you might be able to leverage for an update call or an attribute to check for m2m

cms/utils/placeholder.py
```diff
@@ -337,3 +337,32 @@ def _find_topmost_template(extend_node):
...
+
+    existing = OrderedDict()
+    declared_placeholders = get_declared_placeholders_for_obj(obj)
+    placeholders = [pl.slot for pl in declared_placeholders]
+
+    for placeholder in Placeholder.objects.get_for_obj(obj):
+        if placeholder.slot in placeholders:
+            existing[placeholder.slot] = placeholder
+
+    for placeholder in placeholders:
```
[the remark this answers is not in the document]
czpython: Technically this can still happen in the second loop.
Don't think we should worry about this scenario.

cms/utils/placeholder.py
```diff
@@ -337,3 +337,34 @@ def _find_topmost_template(extend_node):
...
+                source=obj,
+            )
+    return existing
+
+
+def get_declared_placeholders_for_obj(obj):
+    from cms.models import PageContent
+
+    if isinstance(obj, PageContent):
+        obj = obj.page
```
czpython: move get_template to PageContent and remove this condition check

cms/forms/utils.py
```diff
@@ -26,6 +26,8 @@ def get_sites():
 
 
 def get_page_choices_for_site(site, language):
+    from cms.models import PageContent
```
czpython: why?

cms/migrations/0033_placeholder_source_data_migration.py
```diff
@@ -0,0 +1,57 @@
...
+                for pl_field in pl_related_fields:
+                    obj_placeholder_field = getattr(obj, pl_field.name)
+
+                    if pl_field.many_to_many:
+                        obj_placeholder_field.update(
+                            content_type_id=cur_ct_obj.pk,
+                            object_id=obj.pk,
+                        )
+                    else:
+                        obj_placeholder_field.content_type_id=cur_ct_obj.pk,
```
czpython: missing space in assignment

cms/migrations/0033_placeholder_source_data_migration.py
```diff
@@ -0,0 +1,57 @@
...
+                        obj_placeholder_field.object_id=obj.pk
```
czpython: same as above

cms/forms/utils.py
malindap92: I did this because when I imported `PlaceholderRelationField` from `cms.models.fields` in titlemodels.py, cms/forms/utils.py throws an ImportError on PageContent. That's why it was moved inside the function.

cms/utils/placeholder.py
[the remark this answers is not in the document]
malindap92: I think we will need to keep both loops as it is (unless I'm missing something):
- The first loop will check for existing placeholders for the object. This loop will not execute if there are no placeholders (at time of PageContent creation and first time render of non-CMS objects)
- The second loops take care of the declared placeholders in the template and creates if a new placeholder is found.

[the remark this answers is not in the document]
czpython: You're right.
I misread this.

PR #6501, opened 2018-08-31
Fixed a bug with deleting a model from changelist inside modal

### LINKED ISSUE #6495: Can't delete model instances from admin list view that opens from the toolbar
### Summary

I can't delete model instances from the list view that opens from the toolbar. In the "normal" admin list view it does work. From the toolbar opened detailview of the model instance it also works.

### Expected behaviour

I'm expecting to be able to delete model instances from the toolbar list view.

### Actual behaviour

When confirming the deletion, the modal goes blank. The following error appears in the Javascript console:

```
Uncaught TypeError: e.match is not a function
    at HTMLAnchorElement.<anonymous> (bundle.toolbar.min.js:1)
    at HTMLAnchorElement.dispatch (bundle.toolbar.min.js:1)
    at HTMLAnchorElement.g.handle (bundle.toolbar.min.js:1)
```

### Environment

* Python version: 3.6
* Django version: 1.11
* django CMS version: 3.5.2

This is happending both in the Firefox and Chrome (latest at time of writing).

vxsx:
### Summary

Fixes #6495


creimers:
@vxsx any idea on when this might get merged?

PR #6502, opened 2018-08-31
Boost performance for page deletion

filwaitman:


filwaitman:
@chaosk can you take a look at it and merge if everything is ok?

I discussed this with @czpython and he thinks this approach is ok (in fact we're making something similar on cms 3.6).

PR #6528, opened 2018-09-20
Dedicated edit preview buttons

malindap92:
### Summary

This PR includes the implementation of splitting the template logic of rendering of edit/preview buttons to dedicated edit/preview buttons using CMSToolbar items.

### Documentation checklist

    * [ ] general notes
    * [ ] bug-fixes
    * [ ] required upgrade steps
    * [ ] names of contributors


cms/toolbar/toolbar.py
```diff
@@ -444,6 +444,7 @@ def post_template_populate(self):
         if not self.request.user.is_staff:
             return
         self._call_toolbar('post_template_populate')
+        self._render_object_editable_buttons()
```
czpython: this should not be done here.
Instead do it in the post_template_populate of the PlaceholderToolbar.
Placeholder toolbar and editable buttons are a cms specific feature that should not be in the generic toolbar class

cms/tests/test_cache.py
```diff
@@ -461,7 +461,7 @@ def test_no_page_cache_on_toolbar_edit(self):
 
             # Make an initial edit endpoint request
             with self.login_user_context(superuser):
-                with self.assertNumQueries(FuzzyInt(1, 35)):
+                with self.assertNumQueries(FuzzyInt(1, 41)):
```
czpython: that's a whole lot more queries.
why the increase?

cms/cms_toolbars.py
```diff
@@ -86,6 +88,107 @@ def add_wizard_button(self):
...
+                # User has explicitly requested structure mode.
+                # and the object (page, blog, etc..) allows for the non-legacy structure mode.
+                renderer = toolbar.structure_renderer
+            else:
+                renderer = toolbar.get_content_renderer()
+
+            self.placeholders = renderer.get_rendered_placeholders()
+
+    def add_edit_button(self):
+        if self.page and not user_can_change_page(self.request.user, page=self.page):
```
czpython: do the check in the caller instead.

cms/cms_toolbars.py
```diff
@@ -86,6 +88,107 @@ def add_wizard_button(self):
...
+        item.add_button(
+            _('Edit'),
+            url=url,
+            disabled=False,
+            extra_classes=['cms-btn', 'cms-btn-action', 'cms-btn-switch-edit'],
+        )
+        self.toolbar.add_item(item)
+
+    def add_preview_button(self):
+        if self.page and not user_can_change_page(self.request.user, page=self.page):
```
czpython: do the check in the caller instead or create a hidden method like _can_add_preview_button and then call the add_preview_button if true. Basically decouple the check from the action.

cms/cms_toolbars.py
```diff
@@ -86,6 +88,107 @@ def add_wizard_button(self):
...
+        item.add_button(
+            _('Preview'),
+            url=url,
+            disabled=False,
+            extra_classes=['cms-btn', 'cms-btn-switch-save'],
+        )
+        self.toolbar.add_item(item)
+
+    def add_structure_mode(self):
+        if (
```
czpython: Same as above. do the check in the caller instead.

cms/cms_toolbars.py
```diff
@@ -86,6 +88,107 @@ def add_wizard_button(self):
...
+                renderer = toolbar.get_content_renderer()
+
+            self.placeholders = renderer.get_rendered_placeholders()
+
+    def add_edit_button(self):
+        if self.page and not user_can_change_page(self.request.user, page=self.page):
+            return
+
+        item = ButtonList(side=self.toolbar.RIGHT)
+        url = get_object_edit_url(self.toolbar.obj, language=self.toolbar.request_language)
```
czpython: declare this above item

cms/cms_toolbars.py
```diff
@@ -86,6 +88,107 @@ def add_wizard_button(self):
...
+            extra_classes=['cms-btn', 'cms-btn-action', 'cms-btn-switch-edit'],
+        )
+        self.toolbar.add_item(item)
+
+    def add_preview_button(self):
+        if self.page and not user_can_change_page(self.request.user, page=self.page):
+            return
+
+        item = ButtonList(side=self.toolbar.RIGHT)
+        url = get_object_preview_url(self.toolbar.obj, language=self.toolbar.request_language)
```
czpython: declare this above item

cms/cms_toolbars.py
```diff
@@ -86,6 +88,107 @@ def add_wizard_button(self):
...
+        self.toolbar.add_item(item)
+
+    def add_structure_mode(self):
+        if (
+            self.page and not self.page.application_urls and
+            not user_can_change_page(self.request.user, page=self.page)
+        ):
+            return
+
+        if not any(ph for ph in self.placeholders if ph.has_change_permission(self.request.user)):
```
czpython: Same as above

cms/cms_toolbars.py
```diff
@@ -86,6 +88,107 @@ def add_wizard_button(self):
...
+            return
+
+        self.add_structure_mode_item()
+
+    def add_structure_mode_item(self, extra_classes=('cms-toolbar-item-cms-mode-switcher',)):
+        structure_active = self.toolbar.structure_mode_active
+        edit_mode_active = (not structure_active and self.toolbar.edit_mode_active)
+
+        with force_language(self.current_lang):
+            if self.request.user.has_perm('cms.use_structure'):
```
czpython: do this check in the caller. why call method and enter context manager to do a noop

malindap92:
@czpython Changes have been made. Please review.

cms/cms_toolbars.py
```diff
@@ -325,31 +395,8 @@ def populate(self):
...
-        for sp in self.statics:
-            if sp.has_change_permission(self.request):
-                return self.add_draft_live_item()
-
-    def add_draft_live_item(self, template='cms/toolbar/items/live_draft.html', extra_context=None):
-        context = {'cms_toolbar': self.toolbar}
-        context.update(extra_context or {})
-        pos = len(self.toolbar.right_items)
-        self.toolbar.add_item(TemplateItem(template, extra_context=context, side=self.toolbar.RIGHT), position=pos)
+        if self.page and self.toolbar.edit_mode_active and not self.title:
```
czpython: is this intentional? shouldn't it be `and self.title` instead of `and not self.title`?

malindap92: @czpython I basically kept the existing code as it is but when I checked it doesn't do anything and I think its because of the `and not self.title`. Is this function necessary as the Page settings item is added under Page menu?

chaosk: I've checked the code, add_page_settings_button has the following line:
`url = '%s?language=%s' % (admin_reverse('cms_page_change', args=[self.title.pk]), self.toolbar.request_language)`

Which would fail, as `self.title` would be None/EmptyPageContent. If anything, shouldn't this line be:
`url = '%s?language=%s' % (admin_reverse('cms_page_advanced', args=[self.page.pk]), self.toolbar.request_language)`?

czpython:
@malindap92 @chaosk will re-check soon. No later than end of week.

cms/cms_toolbars.py
czpython: Unless I'm missing something (very possible), we don't need add_page_settings_button at all.
The page settings are added in line 562 and advanced settings in 568.
So remove this method and the call to it 🙈

chaosk: Just out of curiosity, is this a legacy bit that handled displaying page settings button for when you're viewing a page with no title for specified language?

czpython: Yes and no, looks like it was adapted for 4.0.x changes but is not actually needed

PR #6547, opened 2018-10-28
Fixed #6336 -- request.current_page should always respect draft/live state (#6354)

### LINKED ISSUE #6336: Menu doesn't set selected node on responses from apphook
When the cms sets `request.current_page`, the page can be draft or live depending on the edit `GET` parameter.

On apphoked pages, `request.current_page` correctly reflects draft/live but on views outside of the page path:

```
/apphook-root/
/apphook-root/outside-of-page-path/
```

The cms will always set `request.current_page` to the live page, regardless of draft/live state.
This inconsistency was introduced in https://github.com/divio/django-cms/pull/1892 to work around some limitations with apphooks prior to 3.x.

This causes issues on the menu in django CMS >= 3.4.x because it will compare the node's id with ``request.current_page.pk`` and if they match, it sets that node as selected, in the case of edit mode on a view outside of page path, this comparison will always fail because ``request.current_page.pk`` is the live page id while the navigation node has the draft page id.

czpython:


FinalAngel:
@czpython could you merge this against develop?

FinalAngel:
closing for now (not backporting)

PR #6548, opened 2018-10-28
Fixed #6002 --  Prevent users from moving homepage under another page

### LINKED ISSUE #6002: Prevent users from moving homepage under another page
Related to https://github.com/divio/django-cms/pull/5999
Also add "." to the error message on set-home endpoint

czpython:


FinalAngel:
not backporting

PR #6553, opened 2018-11-06
WIP - Optionally disable the sideframe

Aiky30:
Added a setting for the Front end to use to disable the sideframe for a project. Requires front end input to actually bind on to disabling the sideframe.


chaosk:
Comment from Vadim:
>Pretty sure everything is covered [in the PR], apart from manual call to `sideframe.open`, which is now exposed and that should probably fail gracefully

cms/static/cms/js/modules/cms.sideframe.js
```diff
@@ -116,6 +116,13 @@ var Sideframe = new Class({
             throw new Error('The arguments passed to "open" were invalid.');
         }
 
+        // Fail gracefully when open is called when disabled
+
+
+        if (typeof CMS.settings.sideframe_enabled !== 'undefined' && CMS.settings.sideframe_enabled === false) {
```
vxsx: no need to check for typeof, since `===` would already do it (also empty lines before the comment)

haricot:
Hoping not to be off topic.

I will shouter the same functionality for the modal (to be able to deactivate it and to use the online edit function of ckeditor.) Only, it would take to do it conditionally, according to an option of the text plugin and not a global setting variable.

For the sideframe, would not it be interesting to be able to activate or deactivate according to the type url admin if it is possible ?

Aiky30:
> Hoping not to be off topic.
> 
> I will shouter the same functionality for the modal (to be able to deactivate it and to use the online edit function of ckeditor.) Only, it would take to do it conditionally, according to an option of the text plugin and not a global setting variable.
> 
> For the sideframe, would not it be interesting to be able to activate or deactivate according to the type url admin if it is possible ?

@haricot I can't help make a decision on this, all i can advise is that you make a good case for the feature and open a PR with the changes, I'm sure the contribution will be appreciated.

PR #6556, opened 2018-11-08
Updated core team and version support details

vThaian:
Forwardport of #6555


czpython:
@vThaian 
You need to use cherry-pick to forward port.
Otherwise author is lost.

PR #6561, opened 2018-11-19
Bumped version to 3.5.3

vThaian:


jrief:
..., unfortunately #6457 didn't make it into this release. It would use the SVG admin icons, since the PNG ones have been removed from Django.

PR #6577, opened 2018-12-12
Merged release/3.6.x into develop

evildmp:


czpython:
@evildmp 
This is quite dangerous.
We had decided not to forward port entire branches. What is the reason for it in this case?

PR #6584, opened 2018-12-19
Fixed #6585 -- Added missing clear cache for menu when page is created

### LINKED ISSUE #6585: Page is missing in menu after created
### Summary

The page is missing in the menu after created. Is working correctly after publish.

This is the problem with clearing cache for a menu after creating a new page.

### Expected behaviour

Link to the page should be visible after created.

### Actual behaviour

Link to the new page after created is missing in menu.

### Environment

* Python version: 3.5
* Django version: 1.11
* django CMS version: 3.6.0rc1

vThaian:
Based on issue #6585

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -827,8 +827,6 @@ class Page(models.Model):
 
         if created:
             self.created_by = self.changed_by
-
-        self.clear_cache(menu=True)
         super(Page, self).save(**kwargs)
 
     def save_base(self, *args, **kwargs):
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -827,6 +827,8 @@ class Page(models.Model):
 
         if created:
             self.created_by = self.changed_by
+
+        self.clear_cache(menu=True)
         super(Page, self).save(**kwargs)
 
     def save_base(self, *args, **kwargs):
--- a/cms/tests/test_menu.py
+++ b/cms/tests/test_menu.py
@@ -298,6 +298,9 @@ class FixturesMenuTests(MenusFixture, BaseMenuTest):
         )
 
     def test_show_new_draft_page_in_menu(self):
+        """
+        Test checks if the menu cache is cleaned after create a new draft page.
+        """
         with self.login_user_context(self.get_superuser()):
             page_data_1 = self.get_new_page_data()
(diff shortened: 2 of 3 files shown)


cms/models/pagemodel.py
```diff
@@ -827,6 +827,8 @@ def save(self, **kwargs):
 
         if created:
             self.created_by = self.changed_by
+
+        self.clear_cache(menu=True)
```
czpython: Please create gh tickets for all bugfixes on released versions (3.6 but not 4.0 in this case).
Because is an rc release regression, you don't need a changelog entry.
Regarding the fix, add it to `AddPageForm` instead.

cms/tests/test_menu.py
```diff
@@ -297,6 +297,29 @@ def test_basic_cms_menu(self):
...
+        renderer = menu_pool.get_renderer(request)
+        renderer.draft_mode_active = True
+        renderer.get_nodes()
+        self.assertEqual(CacheKey.objects.first().key, 'cms_3.6.0rc1_menu_nodes_en_1:draft')
+
+        with self.login_user_context(self.get_superuser()):
+            page_data = self.get_new_page_data()
+            response = self.client.post(URL_CMS_PAGE_ADD, page_data)
+            self.assertRedirects(response, URL_CMS_PAGE)
+            page = Title.objects.drafts().get(slug=page_data['slug']).page
```
czpython: are you sure page is not the same one created above? check that page_data['slug'] is not the same.

cms/tests/test_menu.py
```diff
@@ -297,6 +297,29 @@ def test_basic_cms_menu(self):
...
+    def test_show_new_draft_page_in_menu(self):
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.post(URL_CMS_PAGE_ADD, self.get_new_page_data())
+            self.assertRedirects(response, URL_CMS_PAGE)
+
+        request = self.get_request('/')
+        renderer = menu_pool.get_renderer(request)
+        renderer.draft_mode_active = True
+        renderer.get_nodes()
+        self.assertEqual(CacheKey.objects.first().key, 'cms_3.6.0rc1_menu_nodes_en_1:draft')
```
czpython: not going to work once we release new versions because of the hardcoded version number.
have a look at other menu tests, I know we have checked in the past whether we're being cached.

PR #6609, opened 2019-02-06
Added example of `get_plugin_instance` to the docs on nested plugins.

marksweb:
### Summary
Addition to the docs of `get_plugin_instance` alongside nested plugins content.


czpython:
Hello @marksweb,
thanks for the contribution.

`get_plugin_instance` is an internal API which can lead to serious performance issues depending on where its used so I suggest a warning in the docs. Also we were in the process of replacing it with `get_bound_plugin`.

If we do end up documenting this then I suggest to use that one instead.
It will still suffer from performance issues but at least the behavior will be consistent.

marksweb:
Thanks @czpython. I have seen `get_bound_plugin` but not used it before.

It's also been a while since I've needed `get_plugin_instance` but have modified a parent/child plugin setup to hold attributes on the parent that allow for filtering of choices on the child, so I needed that parent instance yesterday. I ended up asking on IRC about getting the attrs of the parent so when someone reminded me of `get_plugin_instance` I thought the proper way to do that needs to be documented.

Reviewing the source, it looks like I'd be better referencing `get_bound_plugin` in the docs because it's already used by `get_plugin_instance` & more often than not you'd not want that parent class to be returned as well.

Is there a timeline for dropping `get_plugin_instance`?

FinalAngel:
@marksweb thanks for the contribution, I'll adapt the requested changes in a new PR so we can get this merged 👍

PR #6614, opened 2019-02-15
Page has no attribute site_id [Docs]

pawelmarkowski:
There is a bug in the docs that concerns missing attribute. To get information about site_id in ExtensionToolbar you have to use page.node.site_id


vThaian:
Thanks, @pawelmarkowski. We forgot to reflect that change in our documentation.

PR #6699, opened 2019-07-03
Fix plugin js loading

### LINKED ISSUE #6698: Plugin javascript code, not executed upon adding it
### Summary
While writing integration tests for [my django-CMS plugin](https://github.com/s-weigand/djangocms-equation), which requires the execution of javascript code to be rendered properly, I found that the javascript code isn't executed when the plugin is added for the first time.
Instead the page needs to be published or refreshed, for the DOM to to use the scripts.

### Expected behaviour
The javascript code loads and executes, when the plugin is added to the page and it gets rendered properly.

### Actual behaviour
The html code of the updated page is requested via ajax and the innerHtml of the body is replaced with the updated innerHtml of the new body ([see](https://github.com/divio/django-cms/blob/8964d77b876ddb9221a12c64d1c4f5b5575b2049/cms/static/cms/js/modules/cms.structureboard.js#L1293)). This method leads for the scripts not to be executed (tested it with a minimal example in a static html file).
[djangocms-googlemap](https://github.com/divio/djangocms-googlemap) has the same problem.

### Environment

* Python version: All (Javascript issue)
* Django version: All (Javascript issue)
* django CMS version: All (Javascript issue)

s-weigand:
### Summary
This PR adds a feature that allows Plugin creators to inject their javascript code, when adding the plugin.
This is mostly important for plugins which require the execution of javascript code to be rendered properly. (i.e. [djangocms-googlemap](https://github.com/divio/djangocms-googlemap)).

It uses the class `'cms-execute-js-to-render'`  to determine if a script is needed for the plugin to be rendered.
If the javascript code is protected from prematurely executing by the EventListener for the event 'load' and/or 'DOMContentLoaded', the following classes can be added to the script tag:

| Classname                                   | Corresponding javascript code                            |
|---------------------------------------------|----------------------------------------------------------|
| cms-trigger-event-document-DOMContentLoaded | ``document.dispatchEvent(new Event('DOMContentLoaded')`` |
| cms-trigger-event-window-DOMContentLoaded   | ``window.dispatchEvent(new Event('DOMContentLoaded')``   |
| cms-trigger-event-window-load               | ``window.dispatchEvent(new Event('load')``               |

Also the scripts are only executed if they weren't present on the page before adding the plugin.

Fixes #6698 
Closes #6698 

### Documentation checklist

    * [ ] general notes
    * [ ] bug-fixes
    * [ ] required upgrade steps
    * [ ] names of contributors
* [x] I have updated other documentation
* [x] I have added my name to the AUTHORS file


FinalAngel:
@s-weigand thank you for the contribution! I've discussed this intensely with @vxsx and we both came to the conclusion to rather leverage the current system by listening to the existing ``cms-content-refresh`` event.

A good example can be found here:
https://github.com/divio/djangocms-googlemap/blob/15076ed92921ff378bdffd0826954e0f6f793105/djangocms_googlemap/static/djangocms_googlemap/js/djangocms.googlemap.js#L323-L327

s-weigand:
The problem with `cms-content-refresh` is, that it won't be executed until you refresh the page, since the script containing the event listener code won't be executed.
This behavior is due to the way [_replaceBodyWithHTML](https://github.com/divio/django-cms/blob/7dc241e086d6d13cab69390ebba2eb2635c7a5f7/cms/static/cms/js/modules/cms.structureboard.js#L1479) is implemented.
You can reproduce it by adding  [djangocms-googlemap](https://github.com/divio/djangocms-googlemap) map to a blank page.
The Map won't render until you refresh the page, since the listener for  `cms-content-refresh` never gets executed.

s-weigand:
@FinalAngel @vxsx Maybe I have something configured horribly wrong.
Could you be so kind and tell me if this example works at your end.

1. Create a new page
2. Add a [djangocms-googlemap](https://github.com/divio/djangocms-googlemap)

My question now now is:
Does it render with out a page refresh? 

For me it doesn't, same as my own plugin which also requires javascript code to render.
After I added one item of a Plugin, everything works fine and items render/rerender uppon add/change. My Problem is just with the initial adding of an plugin which needs js code to render.

Thanks in advance for testing.

vxsx:
First of all, sorry about delays.

You're right, I can reproduce the faulty behaviour on a clean project. The reason I couldn't reproduce it originally is because I've checked a production project I had that had custom google map js that was bundled via webpack that was always there and there was one more script which we included like so:

```
        <script defer src="{% static 'js/dist/cms.bundle.js' %}"></script>
{% endif %}
```

which roughly contained following (with a bunch of other stuff):

```
CMS.$(window).on('cms-content-refresh', () => {
    // SVGs for some reason don't like DOM refreshing, probably related to creation of elements
    // with incorrect namespace, but no time to look into it now ¯\_(ツ)_/¯
    CMS.$('svg').each((i, el) => {
        CMS.$(el).replaceWith(
            CMS.$(el)
                .clone()
                .wrap('<div></div>')
                .parent()
                .html()
        );
    });

    $(window).trigger('scroll');

    initWaypoints();
    initBlogCarousel();
    initGoogleMaps();
    lazyloadInstance.update();
});
```
all of that was handling the custom reloads and only loading when it was available.

I do understand that it's just one way of doing things and it's not enough out of the box.
I'll check the PR in more detail

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1295,16 +1298,108 @@ class StructureBoard {
...
-
-        Helpers._getWindow().dispatchEvent(new Event('load'));
         $(Helpers._getWindow()).trigger('cms-content-refresh');
 
         this._loadedContent = true;
     }
 
+    /**
+     * Checks if new scripts with the class 'cms-execute-js-to-render' exist
+     * and if the were present befor. If they weren't present before the will be downloaded
```
vxsx: and if the**y** were present befor**e**. If they weren't present before **-** the**y** will be downloaded

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1295,16 +1298,108 @@ class StructureBoard {
...
+     * and executed. If ther script also has the class 'cms-trigger-load-events' the
```
vxsx: ther -> the

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1295,16 +1298,108 @@ class StructureBoard {
...
+     * 'load' and 'DOMContentLoaded' events will be triggered
+     *
+     * @param {JQuery} new_scripts  JQuery selector of the scripts for the new body content
```
vxsx: `@param {jQuery}`

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1295,16 +1298,108 @@ class StructureBoard {
...
+        }
+    }
+
+    /**
+     * Downloads and executes all scripts given by `scriptSrcList` and returns a
+     * `Promise` when all are done. This is to prevent multiple triggering of
+     * load events by `triggerLoadEventsByClass` in `addJsScriptsNeededForRender`.
+     * Original solution from https://stackoverflow.com/a/11803418
+     *
+     * @param {array} scriptSrcList   Array of script sources
```
vxsx: `@param {String[]}`

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1295,16 +1298,108 @@ class StructureBoard {
...
+        }));
+
+        return $.when.apply($, _arr);
+    }
+
+    /**
+     * Triggers events if specific classes were in any of the scripts added by
+     * the method addJsScriptsNeededForRender
+     *
+     * @param {object} classListCollection  array of all classes the script tags had
```
vxsx: `@param {String[]}`

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1295,16 +1298,108 @@ class StructureBoard {
...
+        let _arr = $.map(scriptSrcList, function(scr) {
+            return $.getScript(scr);
+        });
+
+        // eslint-disable-next-line new-cap
+        _arr.push($.Deferred(function(deferred) {
+            $(deferred.resolve);
+        }));
+
+        return $.when.apply($, _arr);
```
vxsx: maybe this whole method could be boiled down `return Promise.all(scriptSrcList.map(s => $.getScript(s))`?

it's ok to use es6 features since the source is transpiled anyway

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1295,16 +1298,108 @@ class StructureBoard {
...
+     * the method addJsScriptsNeededForRender
+     *
+     * @param {object} classListCollection  array of all classes the script tags had
+     */
+    triggerLoadEventsByClass(classListCollection) {
+        if (classListCollection.indexOf('cms-trigger-event-document-DOMContentLoaded') > -1) {
+            Helpers._getWindow().document.dispatchEvent(new Event('DOMContentLoaded'));
+        }
+        if (classListCollection.indexOf('cms-trigger-event-window-DOMContentLoaded') > -1) {
+            Helpers._getWindow().dispatchEvent(new Event('DOMContentLoaded'));
```
vxsx: personally i really doubt this will be of much use and would prefer to avoid it, since listening to this on window is quite rare

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -59,6 +59,7 @@ class StructureBoard {
         this.state = false;
         this.dragging = false;
         this.latestAction = [];
+        this.trigger_window_load_event = true;
```
vxsx: this doesn't seem to be used

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1270,6 +1271,8 @@ class StructureBoard {
         }
         var fixedContentMarkup = contentMarkup;
         var newDoc = new DOMParser().parseFromString(fixedContentMarkup, 'text/html');
+        let new_scripts = $(newDoc).find('script');
```
vxsx: this is a nitpick, but in cms js we generally try to avoid snake case in variable names

s-weigand: Sorry for that, I'm a python guy, so snake case is my natural style, but I will change it 😃

cms/static/cms/js/modules/cms.structureboard.js
s-weigand: Must have been in there at some point and I forgot to remove it. After nearly half a year this looks like an artifact to me as well. I will remove it and see what happens.

cms/static/cms/js/modules/cms.structureboard.js
s-weigand: Since the events can be turned on and off by the choice of the plugin developer, I would leave it in. Just in case someone needs it in an edge case, where i.e. a CDN included code triggers on this event. If no one uses it the only downside is some unused lines of code, but if someone needs it it a game changer.

cms/static/cms/js/modules/cms.structureboard.js
s-weigand: I will try it out

s-weigand: But I don't think so, since this isn't a 'proper Promise' but a [jQuery implementation](https://api.jquery.com/jQuery.getScript/) with different methods.

s-weigand: It works, I will cherry pick the commit from my test branch, after all CI tests succeeded and push it here.

cms/static/cms/js/modules/cms.structureboard.js
s-weigand: You were right, it was an artifact.

s-weigand:
@vxsx if you are fine with https://github.com/divio/django-cms/pull/6699#discussion_r355014581, this should be ready to merge.

closes #6698

s-weigand:
@vxsx The failing CI test is due to a [racing condition](https://travis-ci.org/divio/django-cms/jobs/621992847#L2421) (as so often in integration tests 😒 ).
If you rerun that job on travis, it should pass.

cms/static/cms/js/modules/cms.structureboard.js
vxsx: fair enough 👍

vxsx:
LGTM

s-weigand:
@vxsx @FinalAngel I guess the `status: marked for rejection` label, which is now outdated, is idicating that this shouldn't be merged.

s-weigand:
@vxsx @FinalAngel any update on this? I just rebased and rebundled the 2nd time.
Also, IMHO the bundling should be done by the CI/CD and not on the developer side.

FinalAngel:
Thank you very much @s-weigand! Merging it now.

s-weigand:
As for the now failing travis tests on develop. `sphinxcontrib-spelling` dropped python 3.5 support with version 7.0.0](https://sphinxcontrib-spelling.readthedocs.io/en/latest/history.html#id2).
So you need to change:
https://github.com/django-cms/django-cms/blob/4d8aded242af5244d79e23d9726ea4443660cb86/test_requirements/requirements_base.txt#L16

to
```
sphinxcontrib-spelling<7.0.0
```

PR #6160, opened 2017-12-01
scriptname fixed in page resolver

PleiadiBot:
### Summary

I've noticed that if you run django (django-cms) under a subfolder, the Page Tollbar item for an app-hook page disappeared. 

Fixes #

Fixes page_resolver in order to take the scriptname (env SCRIPT_NAME) in account.

Issue #6521, opened 2018-09-17
page not show when CMS is not installed at the root of the server's URLs


rayer4u:
### Summary

When cms has a page  of "article", **aldryn_newsblog** plugin.  serve django with gunicorn use SCRIPT_NAME
```
export SCRIPT_NAME=/myprefix/aaa
```
url is "http://localhost:8000/myprefix/aaa/article/", in request head
```
PATH_INFO :'/article/'
```
When not use SCRIPT_NAME envirement valuable.
url is "http://localhost:8000/article/", in request head
```
PATH_INFO :'article'
```

### Expected behaviour
Show correct.

### Actual behaviour
```
AttributeError at /article/
'NoneType' object has no attribute 'exclude_featured'
```

### Trace
CurrentPageMiddleware -> get_page_from_request
```
    if clean_path:
        pages_root = reverse("pages-root")

        if path.startswith(pages_root):
            path = path[len(pages_root):]

        # strip any final slash
        if path.endswith("/"):
            path = path[:-1]
```
path is not deal well, only strip final slash, not first slash. '/article' match nothing. when  add code below
```
        if path.startswith("/"):
            path = path[1:]
```
my problem is gone. I don't know this fix is correct for any "some_path" situation?

### Environment

* Python version: 3.6
* Django version: 1.11
* django CMS version:3.5.2

bigbobah:
I have recently faced also with this part of code (CurrentPageMiddleware) but related to another bug. 

My scenario is next:
- Home page has an app hook with urls attached.
- User requests one of urls under this apphook in different languages.

For this case, **expected behavior** is Home page detected as *current* inside middleware, because this page is "owner" of app hook.

While **actual behavior** is Home page detected as *current* only for default language, and None was returned for all other languages. The reason is also that incorrectly implemented truncation of language prefix without first slash.

Issue #5804, opened 2016-12-12
Serving project from subfolder


PleiadiBot:
My site is currently running with: nginx, uwsgi, django (with djangocms).

I need to move everything under a subfolder /abc

nginx

location /abc {
                uwsgi_pass rak;
                include /etc/nginx/uwsgi_params;
                uwsgi_modifier1 30;
                uwsgi_param SCRIPT_NAME /abc;
        }

The site works well but when I try to load any cms page I get **404** or the "Create page" wizard.

All the other apps are correctly served under the /abc/ subfolder even when attached to the cms via apphook.
The url reverse works good and the admin site is ok.

Am I missing something?

czpython:
Hello @PleiadiBot  , please make sure to submit questions within the appropriate channels on either the [User Mailing List](https://groups.google.com/forum/#!forum/django-cms), IRC (freenode#django-cms) or [Stackoverflow](http://stackoverflow.com/questions/tagged/django-cms).

You can find additional information on our [support](https://www.django-cms.org/support) page.

In this case, I believe your question is best in the [Users Mailing List](https://groups.google.com/forum/#!forum/django-cms).

Issue #365, opened 2010-03-03
Multilingual urls break when django is not on domain root


hukka:
Steps to reproduce:
Configure Django project to work on a subfolder of a domain (eg. http://example.com/project/) with mod_wsgi. This should set the SCRIPT_NAME to "/project" so that reverse urls work correctly without prepending the subfolder to every line in url-conf.

Expected outcome:
Logging in at http://example.com/project/admin/ works.

Actual outcome:
User is redirected to http://example.com/en/project/admin/, which does not work.

PR #6726, opened 2019-09-01
 Add test under wsgi script_name

haricot:
### Summary

Fixes  #6160

possible #6521 #5804 #365 #6763

test if function `get_page_from_request` take context about script_name in wsgi request.

### General checklist

* [x] I have updated the CHANGELOG.txt
* [x] I have updated the documentation and/or amended the upgrade section
      if necessary


haricot:
At this stage the test does not pass, the function  `get_page_from_request  `does not take the prefix. The test work.

FinalAngel:
@haricot thank you for the contribution. Though this will most likely not land in 3,7 anymore.

@yakky what's your feedback on this? :)

cms/utils/page.py
```diff
@@ -152,8 +153,10 @@ def get_page_from_request(request, use_path=None, clean_path=None):
 
     draft = use_draft(request)
     preview = 'preview' in request.GET
-    path = request.path_info if use_path is None else use_path
-
+    if request.environ.get('SCRIPT_NAME'):
```
vstoykov: Instead of manually looking for `SCRIPT_NAME` i think that the easy way of achieving the same thing is by using `path` instead of `path_info`.
https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpRequest.path

haricot: @vstoykov thank you, it's much better.

haricot:
PR updated on dev branch.
I can rewrite this PR to have more complete commit descriptions if needed.

cms/utils/page.py
```diff
@@ -150,7 +150,7 @@ def get_page_from_request(request, use_path=None, clean_path=None):
 
     draft = use_draft(request)
     preview = 'preview' in request.GET
-    path = request.path_info if use_path is None else use_path
+    path = request.path if use_path is None else use_path
```
yakky, line 153: we currently use `path_info` everywhere, so I guess more clarifications why this is needed here would be needed

yakky (changes requested):
@haricot could you add a comment on the code how the change works and why it's needed?

If we have a problem with running when `SCRIPT_NAME` is not empty, I think a better solution would be to set it everytime, to improve coverage for the use case

cms/utils/page.py
haricot: When using djangocms with a defined script_name and using an apphook this one does not return the correct url because of Path_info, the prefix is ​​missing.

cms/test_utils/testcases.py
```diff
@@ -412,11 +415,17 @@ def get_request(self, path=None, language=None, post_data=None, enforce_csrf_che
                 language = settings.LANGUAGES[0][0]
             else:
                 language = settings.LANGUAGE_CODE
-
         if post_data:
             request = factory.post(path, post_data)
         else:
             request = factory.get(path)
+        if script_name:
```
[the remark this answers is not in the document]
haricot: Thanks for your advice, I modified this and I realize that the problem with this PR is not from "get_page_from_request" but rather from apphook logic when used with a defined SCRIPT_NAME.
The test passes without problem. Sorry for the inconvenience.

haricot:
The test passed, so the problem was not there, but rather the logic of the specific apphook if SCRIPT_NAME is used, i think. You will tell me if you are interested in this test, I will do this PR clean.

PR #6732, opened 2019-09-17
Store page status code into the cache

zbohm:

Some extensions can have error pages directly in cms administration. Such pages are cached. For this purpose, it is also necessary to save the status code of the corresponding error, otherwise HTTP 200 is set.


FinalAngel:
@zbohm thanks for the contribution, please reopen this PR against develop with the conflicts resolved :)

PR #6733, opened 2019-09-17
Add COLLATE "C" into table field cms_cmsplugin.path

zbohm:
### Fixes

For databases with a collate other than "C", the column `path` in table `cms_cmsplugin` must be set, otherwise the system will not work properly.


FinalAngel:
@zbohm thanks for the contribution, please reopen this PR against develop with the conflicts resolved :)

PR #6743, opened 2019-10-07
Improve apphooks documentation

adrien-delhorme:
Remove a reference to AppHook class `urls` attribute in the documentation.


FinalAngel:
Thanks for the contribution @adrien-delhorme :)

PR #6744, opened 2019-10-08
Fix documentation bug regarding CMSPluginBase

seppeljordan:
### Summary

The documentation advertises that you can set the module attribute of
custom plugins to None.  If the user attempts this django they won't be able
to start up when django.contrib.admin.site.urls is included in urls.py
(tested with django 1.11.25 and django-cms 3.7.0).

I strongly recommend backporting this to the 3.7.x release branch.

      if necessary


FinalAngel:
Thank you for the contribution @seppeljordan :)

PR #6748, opened 2019-10-14
Fix #6746: Pack page-ids in chunks to prevent query error on SQLite

### LINKED ISSUE #6746: On SQLite: Too many SQL variables if page contains more than 1000 descendants
### Summary

This error presumably only occurs on SQLite.
If a page contains more than 1000 descendants one can not set page permissions.

### Expected behaviour

In settings.py set `CMS_PERMISSION = True`.
Create a CMS page with more than 1000 descendants.
On the page's toolbar, select `Page > Permissions`
Set any king of permission in `Page Permission`

Login as a (staff / non-superuser) user with permissions on that page.
This causes an error in SQLite.

### Actual behaviour

SQLite complains with `Too many SQL variables`.
This is because there is a filter on model `Page` with too many values.

### Environment

* Python version: 3.6
* Django version: 2.1
* django CMS version: 3.6

jrief:
### Summary

Fixes #6746

This PR is ready for review.
If accepted, I will add some lines to the CHANGELOG.

### General checklist

      if necessary

### DIFF
--- a/cms/utils/permissions.py
+++ b/cms/utils/permissions.py
@@ -188,32 +188,38 @@ def get_page_actions_for_user(user, site):
     )
     nodes = [page.node for page in pages]
     pages_by_id = {}
+    page_permissions_chunks = []
 
-    for page in pages:
-        if page.node.is_root():
-            page.node._set_hierarchy(nodes)
-        page.node.__dict__['item'] = page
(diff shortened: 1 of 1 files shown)


goutnet:
@jrief  can you rebase this onto develop?

PR #6749, opened 2019-10-21
Fixes/sqlite3 too many sql variables

### LINKED ISSUE #6746: On SQLite: Too many SQL variables if page contains more than 1000 descendants

jrief:
### Summary

Fixes #6746
Ported from #6748 for django-CMS 3.7

Fixes the bug.

### General checklist

* [x] I have updated the CHANGELOG.txt
* [x] I have created backports if necessary
      if necessary

### DIFF
--- a/cms/utils/permissions.py
+++ b/cms/utils/permissions.py
@@ -188,32 +188,38 @@ def get_page_actions_for_user(user, site):
     )
     nodes = [page.node for page in pages]
     pages_by_id = {}
+    page_permissions_chunks = []
 
-    for page in pages:
-        if page.node.is_root():
-            page.node._set_hierarchy(nodes)
-        page.node.__dict__['item'] = page
-        pages_by_id[page.pk] = page
+    for offset in range(0, pages.count(), 500):
+        pages_ids_chunk = []
+        for page in pages[offset:offset + 500]:
+            if page.node.is_root():
+                page.node._set_hierarchy(nodes)
+            page.node.__dict__['item'] = page
+            pages_by_id[page.pk] = page
+            pages_ids_chunk.append(page.pk)
 
-    page_permissions = (
-        PagePermission
-        .objects
-        .with_user(user)
-        .filter(page__in=pages_by_id)
-    )
+        page_permissions_chunks.append(
+            PagePermission
+            .objects
+            .with_user(user)
+            .filter(page_id__in=pages_ids_chunk)
+        )
 
-    for perm in page_permissions.iterator():
-        # set internal fk cache to our page with loaded ancestors and descendants
-        if DJANGO_1_11:
-            perm._page_cache = pages_by_id[perm.page_id]
-        else:
-            # for django >= 2.0
-            PagePermission.page.field.set_cached_value(perm, pages_by_id[perm.page_id])
+    for page_permissions in page_permissions_chunks:
+        for perm in page_permissions.iterator():
+            # set internal fk cache to our page with loaded ancestors and descendants
+            if DJANGO_1_11:
+                perm._page_cache = pages_by_id[perm.page_id]
+            else:
+                # for django >= 2.0
+                PagePermission.page.field.set_cached_value(perm, pages_by_id[perm.page_id])
 
-        page_ids = frozenset(perm.get_page_ids())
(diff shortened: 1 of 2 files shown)


jrief:
Additional explanation:
In function `get_page_actions_for_user(...)`, a long list of PKs is generated. This list then is used to generate a filter query checking which page is inside that list.

For SQLite, the Django ORM creates SQL using a `?` as placeholder. In SQLite this number of placeholders is limited to 999, hence not more than 999 pages can be descendants of another page. By building chunks of 500, this problem can be avoided.

cms/utils/permissions.py
```diff
@@ -188,32 +188,38 @@ def get_page_actions_for_user(user, site):
     )
     nodes = [page.node for page in pages]
     pages_by_id = {}
+    page_permissions_chunks = []
+
+    for offset in range(0, pages.count(), 500):
```
FinalAngel: not a fan of having a magical number in here. As it only affects SQLite we should also probably target only SQLite with such an exception. I'm sadly not deep into the permission system by itself. WDYT @yakky or @czpython ?

jrief:
@FinalAngel , @yakky 
This limit is hard-coded into SQLite itself using `SQLITE_MAX_VARIABLE_NUMBER`. Changing it requires the recompilation of the SQLite code. Unfortunately I haven't found any interface to fetch that number using some sort of SQLite dialect. 500 is certainly a safe lower limit which helps to prevent these sorts of errors.

Proposal:

a) I make this number `MAX_INT` for non-SQLite databases and 500 for SQLite. Making this configurable, in my opinion doesn't make much sense, since rarely anybody knows that limit.
b) Through the `Page`-model, I check on which database it is running, in order to answer a).

cms/utils/permissions.py
yakky: I will make this number more explicit by setting a constant (something like PAGE_PERMISSION_CHUNK_SIZE)
We can then have different numbers for different database backends and make it very large (to the point of being infinite) for DBMS other than sqlite

czpython:
I honestly don't think this is the way to go, for being a limitation of sqlite only, worst-case scenario I would add a condition that avoids this issue only for sqlite.
A more involved solution would be to refactor this into subqueries instead of manual ids.

jrief:
@yakky puh, writing a test for this scenario would indeed require to add a lot of pages to the tree. I'm not sure if this is worthwhile.
> something like PAGE_PERMISSION_CHUNK_SIZE

This is perfectly fine, but that number shall *not* be configurable through the owners `settings.py`

@czpython 
>  I would add a condition that avoids this issue only for sqlite.

That's simple. We can use a variable instead of `500` and change that to `MAX_INT` for non-sqlite database.

> A more involved solution would be to refactor this into subqueries instead of manual ids.

I looked at this, but since the queryset which could be used for the subquery, is modified by calling some Django permission functions, I don't believe it's feasible. 

Shall I modify that pull request and use a meaningful variable name instead of 500 and apply that only to SQLite?

yakky:
@jrief 
agree `PAGE_PERMISSION_CHUNK_SIZE` is just an internal constant, not a setting.
As you wrote is set to something like:

```
if driver == "sqlite3":
   PAGE_PERMISSION_CHUNK_SIZE = 500
else:
   PAGE_PERMISSION_CHUNK_SIZE = MAX_INT
```

Subqueries would be great, and the `with_user` call can be probably rewritten, but I don't think it's worth the effort, unless we see some global improvement

goutnet:
@jrief would this be useful in develop for 3.9.0 ?

PR #6753, opened 2019-10-26
Added a virtual environment to remove the hassel and test the project…

subodhk01:
… directy and added a local databse with superuser (Username- user && Password- password)

### Summary

Fixes #

### General checklist

      if necessary


FinalAngel:
Thank you very much for the contribution @subodhk01. I sadly have to refuse this PR as testing should be triggered through ``python manage.py test`` by following the [documentation](http://docs.django-cms.org/en/latest/contributing/testing.html#running-tests) guidelines.

We do not want a pre-defined database committed to the project itself that holds user data (even if it's just the superuser). The tests should create that automatically for you. For example have a look at the [get_superuser test helper](https://github.com/divio/django-cms/blob/develop/cms/test_utils/testcases.py#L212).

PR #6757, opened 2019-11-07
Add COLLATE "C" into database table column cms_cmsplugin.path.

zbohm:
### Fixes

(Repoen PR #6733 - branch rebased to `develop`.)

For databases with a collate other than "C", the column path in table cms_cmsplugin must be set, otherwise the system will not work properly.


FinalAngel:
thank you for reopening the PR to develop @zbohm 👍 

Would you also be able to add an entry to the CHANGELOG once https://github.com/divio/django-cms/pull/6755 gets merged?

zbohm:
Changelog updated. Rebased to divio/develop.

goutnet:
@czpython @Aiky30 is this migration still needed on the current develop?

If so, let me know and let's move on merge.

zbohm:
The collate setting is necessary when using a database that has a default collate that has a different alphabetical order. Module [treebeard](https://github.com/django-treebeard/django-treebeard) finds position of the last plugin node by alphabetical order. For example function [get_root_nodes](https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/mp_tree.py#L904-L906) ask for the last node by query `SELECT * FROM cms_cmsplugin ORDER BY path`. If alphabetical order is different than it is expectected, the last node is not found properly and this state leads to IntegrityError.

```
CREATE TEMPORARY TABLE test_path (
    path  char(10)
);
INSERT INTO test_path (path) VALUES ('0001'), ('0002'), ('0003'), ('0004'), ('0005'), ('0006'), ('0007'), ('0008'), ('0009'), ('000A');

SELECT * FROM test_path ORDER BY path COLLATE "C";
    path    
------------
 0001      
 0002      
 0003      
 0004      
 0005      
 0006      
 0007      
 0008      
 0009      
 000A      
(10 rows)

SELECT * FROM test_path ORDER BY path COLLATE "cs_CZ";
    path    
------------
 000A      
 0001      
 0002      
 0003      
 0004      
 0005      
 0006      
 0007      
 0008      
 0009      
(10 rows)

```

Last item was incorectly found as a `0009`. Next node `000A` rises error:

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "cms_cmsplugin_path_4917bb44_uniq"
DETAIL:  Key (path)=(000A) already exists.
```

goutnet:
I am guessing it would be beneficial to have this rebased onto `release/3.9.x` for inclusion in `3.9.0` and tested? wat do you tihnk?

goutnet:
@vinitkumar  ^

vinitkumar:
@goutnet I think, it should be safe to merge to release/3.9.x because I have also seen this issue happening in wild with our own apps. However, @zbohm, could you please remove the translations file related changes? We pull translations from transifixs, so we won't need these changes here. They are also polluting the PR and hiding the real diff.

goutnet (changes requested):
* Please rebase on the `release/3.9.x` branch, (which really is develop plus some translations.
* remove all translations from this PR
* also remove the modifications on the CHANGELOG (this is now automated).

After that we can merge :)

goutnet:
I am changing the target on the rebase

goutnet:
I changed this to 3.9.x,

please remove translations and unrelated merges

zbohm:
I removed a modification from the CHANGELOG.

There are no changes in translations in the branch. It was only added only one file with migrations -  `0023_cms_cmsplugin_path_collage.py`.

Florian Delizy already merged the branch together with `develop` and `release/3.9.x`. So I cannot rebase without breaking these joins. Should I leave it as it is? Or I should rebase to `release/3.9.x`?

zbohm:
I added branch `add-collate-c-into-path` because this branch is in conflict with assertions of number of migrations with 4.x. The branch `add-collate-c-into-path` is based on `3.9.x` and it includes only one commit with only one file - the migration. You can choose this: https://github.com/django-cms/django-cms/compare/release/3.9.x...zbohm:add-collate-c-into-path?expand=1

goutnet:
Ah…

looking at this, the problem is we can't really add a migration at this point since this would break the migration process from 3.x to 4.x.

Is there any way we could takle this issue without adding a migration?

goutnet:
@Aiky30 what would you suggest to preserve the next migration process from 3.x to 4.x ?

Aiky30:
@goutnet I see that we have 2 options here, one is a migration merge in v4, the other is to first rollback this migration when installing v4. 

I personally prefer the first option as it keeps the codebases consistent, however path is no longer a field on the plugin model so I am unsure how current db's on v4 would handle this migration addition. I have never tried a merge with so many differing migrations after the conflict so may need to try this out.

Worst case we can always reverse this migration. Issues occur if more migrations come through as any additional migrations would also need to be reversed leaving us in a tricky situation.

Issue #6776, opened 2019-11-28
Cascading user deletion


robinvd:
## Description
After adding cms, Currentusermiddleware, and CMS_PERMISSION=1 to settings. Deleting a user will cascading delete all users this user ever created.

This is without ever mentioning cms outside of settings.py

See the POC django app (with relevant test) https://github.com/robinvd/djangocms-user-cascade

## Steps to reproduce

1. add cms, Currentusermiddleware, and CMS_PERMISSION=1 to settings.
2. create a user while logged in
3. delete the current user

in code:
```python
self.user = get_user_model().objects.create_user('a', 'b', 'pass')
self.client.login(username='a', password='pass')

response = self.client.post('/create_user/')

print(get_user_model().objects.count()) # prints 2
self.user.delete()
print(get_user_model().objects.count()) # prints 0
```

## Expected behaviour
The created user (in step 2) still exists

## Actual behaviour
The created user (in step 2) is cascaded by the delete of the first

## Additional information
cms version: 3.7.1
python version: 3.7.5
django version: 2.2.7

OMH-G:
I have an suggestion that if we store the current user in an instance before login and delete that current user after login .
 
self.user = get_user_model().objects.create_user('a', 'b', 'pass')
self.store=get_user_model().objects.get(username='a',password='pass')#Adding line
self.client.login(username='a', password='pass')

response = self.client.post('/create_user/')

print(get_user_model().objects.count()) # prints 2
self.store.delete()#Changes 
print(get_user_model().objects.count()) # prints 0

OMH-G:
I have an suggestion that if we store the current user in an instance before login and delete that current user after login .

self.user = get_user_model().objects.create_user('a', 'b', 'pass')
self.store=get_user_model().objects.get(username='a',password='pass')#Adding line
self.client.login(username='a', password='pass')

response = self.client.post('/create_user/')

print(get_user_model().objects.count()) # prints 2
self.store.delete()#Changes
print(get_user_model().objects.count()) # prints 1

PR #6778, opened 2019-12-03
made PageUser non cascading

robinvd:
## Description

A PageUser gets created for every user that is created, with a created_by to the original user.
On deleting the original user all PageUsers would be deleted by cascade.
And thus also the user connected to there PageUsers as well.

This sets created_by to on_delete=models.SET_NULL stopping this problem

## Related resources

* #6776

## Checklist

I can make a test if that is required
* [] I have added or modified the tests when changing logic


FinalAngel:

Issue #6754, opened 2019-10-29
[bug] Toolbar does not detect static plugin changes - and thus publication is not possible


petrklus:
### Summary

A custom application is using static placeholders to populate its sidebar contents. The changes to the sidebar are not detected, yet editing works as expected and toggling between "view published" and "edit" modes work.

### Expected behaviour

The publish button should appear when static placeholders are altered.

### Actual behaviour

The publish button never appears in the toolbar. As a workaround, we had created a CMS page with a template that features all static placeholders one after another - there the changes are detected correctly (even if only static placeholders are altered).

### Environment

* Python version: 3.6.8
* Django version: 1.11
* django CMS version: 3.7 (running on Divio)

marksweb:
I feel like this got fixed.

PR #6781, opened 2019-12-09
Static placeholder publication for non-CMS pages

petrklus:
The `PageToolbar.has_publish_permission` function always returns `False` when not on CMS pages. This is incorrect as static placeholders can be placed outside of CMS pages. This fix allows for proper static placeholder publication checking as (probably) originally intended.

## Description

Fixes an issue when static placeholders could not be published when placed outside of CMS pages.

## Related resources
* #6754 

## Checklist

### DIFF
--- a/cms/cms_toolbars.py
+++ b/cms/cms_toolbars.py
@@ -291,7 +291,7 @@ class PageToolbar(CMSToolbar):
         else:
             publish_permission = False
 
-        if publish_permission and self.statics:
+        if self.statics:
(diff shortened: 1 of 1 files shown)


yakky:
@petrklus I would rather fix this by checking the general `publish_page` permission in the `else` block
This will ensure that only users with general `publish_page` permission can publish the static placeholder.
Alternatively we can add a permission to publish static placeholders, but this will only be possible in a future feature release of django CMS

PR #6790, opened 2020-01-08
Adding support for Django 2.2 LTS to django-cms 4.0

jonathan-s:
## Description
Cherrypicks the work from develop that makes this compatible with django 2.2. The tests are currently failing because PyYaml does not support python 3.4. Furthermore python 3.4 is end of life as of March 2019 ( last year ). So I would suggest that python 3.4 is also removed from travis. Life would be simpler and everyone happier. 

@FinalAngel Could you please have a look at this too?

## Checklist


CHANGELOG.txt
```diff
@@ -1,4 +1,10 @@
-=== 3.6.0 (unreleased) ===
+
```
Aiky30: How did we get the version 3.7? Non eof the changes from 3.5 to 3.7 exist here so we shouldn't mention 3.7 until that time comes to back port changes.

docs/upgrade/3.7.rst
```diff
@@ -0,0 +1,84 @@
+.. _upgrade-to-3.7.0:
+
+###################
```
Aiky30: As mentioned above we shouldn't have anything mentioning 3.6 or 3.7 yet

Issue #5448, opened 2016-06-22
Use logging for apphook reload


czpython:
There are some places in the apphooks reload module that use print statements instead of logging.
https://github.com/divio/django-cms/blob/develop/cms/utils/apphook_reload.py#L96
These should use logging instead.

Related to https://github.com/divio/django-cms/issues/5084

jsma:
Was just about to open a ticket for this. I agree it should use logging as well, but I'd also like to improve *what* is logged so it's as useful as possible. For instance, it's very unclear why I see this kind of output or what I'm supposed to do with this information:

    New revision!!!! RELOAD!
       074c568c-1ce0-49fe-afe2-1027ddcaff0b (<class 'str'>)
    -> None (<class 'NoneType'>)

Or:

    reverse('my_test_app_view'): Reverse for 'my_test_app_view' with arguments '()' and keyword arguments '{}' not found. 0 pattern(s) tried: []

czpython:
In the hopes someone can tackle this, here's the explanation:

When an apphook is attached to a page, if the project has the `ApphookReloadMiddleware` middleware installed, the cms will trigger a reload of the configured url conf module.  This allows the content manager to continue to use the page with the apphook urls already applied.

Without the `ApphookReloadMiddleware` middleware installed, content managers would not be able to see the attached app until a sys admin restarts the wsgi process.

Example, user attaches the blog apphook to a page called blog.

If `ApphookReloadMiddleware` middleware is installed, content manager can start adding and editing articles, also the article detail pages would work fine like `/en/blog/django-cms-part-1/`

If `ApphookReloadMiddleware` middleware is **not** installed, content manager can probably add and edit articles via the admin (depending on the app) but if the content manager visits an article detail page (`/en/blog/django-cms-part-1/`) he'll get a 404 because the url conf module was loaded on memory when the wsgi process first started and so does not know about the article urls.

To manage this, there's a global revision and local revision records.
The global revision is stored in the database and the local one in thread.

The first message:

    New revision!!!! RELOAD!
       074c568c-1ce0-49fe-afe2-1027ddcaff0b (<class 'str'>)
    -> None (<class 'NoneType'>)

Tells that a new local revision is being created because a change has been detected and so url conf will be reloaded.

After the url conf module is reloaded, it's necessary to trigger a `reverse()` call in order to make sure django loads all included urls.

The second message:

[...]

PR #6791, opened 2020-01-15
Remove print statements from apphook_reload.py

stuaxo:
Remove print statements from apphook_reload.py

Debug logging in the codebase via print statements is OK locally, but should never have been committed to the codebase.

If someone wants these, they should be converted to use a logger at DEBUG level that is not enabled by default.

This is a fix for https://github.com/divio/django-cms/issues/5448

### DIFF
--- a/cms/utils/apphook_reload.py
+++ b/cms/utils/apphook_reload.py
@@ -28,15 +28,7 @@ def ensure_urlconf_is_up_to_date():
     if not local_revision:
         set_local_revision(global_revision)
     elif global_revision != local_revision:
-        if settings.DEBUG:
-            print("   New revision!!!! RELOAD!\n"
-                  "      {0} ({1})\n"
-                  "   -> {2} ({3})".format(
-                      global_revision, type(global_revision),
-                      local_revision, type(local_revision),))
-        debug_check_url('my_test_app_view')
         reload_urlconf(new_revision=global_revision)
-        debug_check_url('my_test_app_view')
 
 
 def get_local_revision(default=None):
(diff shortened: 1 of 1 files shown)


stuaxo:
Converting to log statements is probably a good call.

Like most other people I use ye old print debugging from time to time, but never commit that stuff.

It raised a red flag to do it, and probably similar for others.

yakky:
> Converting to log statements is probably a good call.

:tada: are you willing to do in this PR?

> 
> Like most other people I use ye old print debugging from time to time, but never commit that stuff.
> 
> It raised a red flag to do it, and probably similar for others.

Somehow it managed to survive quite a few refactorings of the area :joy:

stuaxo:
I'll see if I can fit it in.    I use Django CMS for work, but my contract only has days left to run.
I do, do free time coding, but it's doesn't usually cross over with the stuff I do for work.

So.. . maybe.

yakky:
> I'll see if I can fit it in. I use Django CMS for work, but my contract only has days left to run.
> I do, do free time coding, but it's doesn't usually cross over with the stuff I do for work.
> 
> So.. . maybe.

No worries, I can do this myself, if you cannot

stuaxo:
Thanks :) it'll may be a while until I come back to Django CMS.

FinalAngel:
@yakky any updates on this? :)

Aiky30:
@FinalAngel @yakky I think we maybe create an issue for this and close this PR as it would probably be more effort to bring this back alive than to just create a new PR from an issue. Let me know your thoughts and we can action.

stuaxo:
I'm happy for this to be closed, I'm unlikely to be using django cms for the next 12 months at least.

yakky:
@Aiky30 see #6917 , I don't think we need a changelog entry for this

stuaxo:
Thanks for sorting this, it was bugging me when I worked with dcms :)

PR #6793, opened 2020-01-16
Fixed issue where a Page object returns the wrong title object after publishing

arjandepooter:
## Description

Recently we ran into an issue in a job which builds a pagetree from remote data. If there are unpublished changes in existing pages it will publish those changes. When that happens the urls of the children of the published page are not in the correct translation but take the path of the title object with the first fallback language from `CMS_LANGUAGES` (see added unittest which reproduces this issue). After some debugging I discovered that because of the removal of the published language from the internal title cache combined with the order of handling `fallback` and `force_reload` in `Page._get_title_cache` it returns the fallback while it should reload the cache from database: `force_reload` is handled after `fallback`. Because the title cache is always filled at once for all languages, I think single language instances shouldn't be deleted (which is what happens now in the `publish` function) so the flow from `_get_title_cache` can stay as it is.

## Checklist

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -990,7 +990,7 @@ class Page(models.Model):
             self.mark_descendants_as_published(language)
 
         if language in self.title_cache:
-            del self.title_cache[language]
+            self._clear_internal_cache()
 
         # fire signal after publishing is done
         import cms.signals as cms_signals
--- a/cms/tests/test_page.py
+++ b/cms/tests/test_page.py
@@ -1146,6 +1146,30 @@ class PagesTestCase(TransactionCMSTestCase):
             self.assertContains(resp, public_text)
             self.assertNotContains(resp, draft_text)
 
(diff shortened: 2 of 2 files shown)


FinalAngel:

cms/models/pagemodel.py
```diff
@@ -990,7 +990,7 @@ def publish(self, language):
             self.mark_descendants_as_published(language)
 
         if language in self.title_cache:
-            del self.title_cache[language]
+            self._clear_internal_cache()
```
Aiky30, line 993: This is too heavy, you're effectively dropping the cached node tree and any prefetched objects. I would advise dropping only what is required, which appears to me to be clearing the title cache for the current and fallback language ("en" and "de" in the test). 

Just clear the title cache rather than clearing all caches!
"""
        self.title_cache = {}
"""

Aiky30 (changes requested):

PR #6795, opened 2020-01-16
Fixed issue where default fallbacks is not used when it's an empty list

arjandepooter:
## Description

When you want to define the default fallbacks in `CMS_LANGUAGES` as an empty list, this is ignored and it takes the default of all other available languages as fallbacks (see added unittest for reproduction of the bug). 

## Checklist


FinalAngel:

Aiky30 (changes requested):
I think this is ok but we need to be sure that developers are aware of this change as some projects may rely on this behaviour by accident where they have the full language list as fallbacks.

Needs a changelog entry to aid with the comment above.

goutnet (approved):
LGTM

PR #6796, opened 2020-01-23
Add Django 3 and Python 3.8 support

FinalAngel:
## Description

This pull request adds Django 3.0 and Python 3.8 support.

## Checklist


FinalAngel:
@yakky would you be able to have a look at this

cms/tests/test_i18n.py
```diff
@@ -364,9 +364,10 @@ def test_session_language(self):
         session = self.client.session
         session[LANGUAGE_SESSION_KEY] = 'fr'
         session.save()
-        response = self.client.get('/')
-        self.assertEqual(response.status_code, 302)
-        self.assertRedirects(response, '/fr/')
+        if not DJANGO_3_0:
```
FinalAngel, line 367: @yakky do you know any specific changes in Django 3 for language fullbacks, I cannot figure out what is going on there exactly.

cms/tests/test_static_analysis.py
```diff
@@ -10,5 +11,6 @@ class AboveStaticAnalysisCodeTest(TestCase):
     def test_pyflakes(self):
         import cms
         import menus
-        errors, message = pyflakes((cms, menus))
-        self.assertEqual(errors, 0, message)
+        if not DJANGO_3_0:
```
FinalAngel, line 14: need a new pyflakes release, as for tests only, just trivial

cms/tests/test_i18n.py
yakky: This is due to language not being stored in session anymore (see https://docs.djangoproject.com/en/3.0/releases/3.0/#id3 item 3)

I cannot find a better announcement regarding the change

I kept the logic unchanged on 2.2 and below to avoid changing existing behavior

cms/tests/test_static_analysis.py
yakky: only trivial changes needed here. I also bumped pyflakes for all django versions

Issue #164, opened 2009-09-23
Page editing breaks on monolingual sites


acdha:
The documentation says:

```
If your site is not multilingual you can leave out 'cms.middleware.MutilingualURLMiddleware'.
```

With the current master, if you actually do this the admin area will break when you attempt to update a page because cleaned_data won't have a key "language".

PR #6799, opened 2020-01-25
Dependency update of test_requirements

MLK97:
## Description

When trying to run tests you need to install the test requirements for django-2.2.
This also includes the pyenchant package, which at the current official version created an error, which is discussed [here](https://github.com/pyenchant/pyenchant/issues/164)
I updated the test_requirement.txt to the current pre-release and it is working now, though it should be changed as soon as the pre-release becomes an official release.

## Related resources

[ImportError: The 'enchant' C library was not found. Please install it via your OS package manager, or use a pre-built binary wheel from PyPI. #164](https://github.com/pyenchant/pyenchant/issues/164)

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -29,6 +29,7 @@ Changelog
 * Improved documentation related to nested plugins.
 * Updated installation tutorial.
 * Fixed a simple typo in the docstring for ``cms.utils.helpers.normalize_name``.
+* Changed pyenchant version in the base_requirement.txt to pre-release 3.0.0a1
 
 
 3.7.0 (2019-09-25)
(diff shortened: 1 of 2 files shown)


FinalAngel:
Thanks for the PR @MLK97 though it appears it's not resolved entirely yet. Also suffering the same in https://github.com/divio/django-cms/pull/6796.

Maybe @yakky has further insights here?

MLK97:
If I knew what the error code means I could look further into the problem, but I am honestly clueless what to do with the Travis-Ci output:
```bash
The command "if [ "$FRONTEND" != 1 ] && ([ "$TEST_DOCS" == 1 ] 
&& [ "$ONLY_DOCS" == 0 ] || [ "$ONLY_DOCS" != 0 ]); 
then coverage run --parallel-mode manage.py test; fi;" exited with 127.
```

yakky:
@MLK97 @FinalAngel this has been fixed in #6796 by pinning 3.0.1 which apparently solves the bug

PR #6803, opened 2020-02-03
EmptyPageContent should raise a 404

jonathan-s:
## Description

If a page does not contain any pagecontent it will return an `EmptyPageContent`. We should not try to render this `EmptyPageContent`, instead we should return a 404. This solves issues related to versioning down the line. If a page is in draft only, it will return an `EmptyPageContent`, and thus we don't want this page to be rendered at all. 

## Related resources
This PR depends on #6790, once that PR is merged this PR can be rebased against current master. I've based this PR on that branch to show that tests are passing. 

Please review the last commit to see the relevant changes. 

## Checklist

### DIFF
--- a/cms/tests/test_rendering.py
+++ b/cms/tests/test_rendering.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from django.core.cache import cache
+from django.http.response import Http404
 from django.test.utils import override_settings
 
 from sekizai.context import SekizaiContext
@@ -7,7 +8,7 @@ from sekizai.context import SekizaiContext
 from cms import plugin_rendering
 from cms.api import create_page, add_plugin
 from cms.cache.placeholder import get_placeholder_cache
-from cms.models import Page, Placeholder
+from cms.models import EmptyPageContent, Page, Placeholder
 from cms.plugin_rendering import PluginContext
 from cms.test_utils.project.placeholderapp.models import Example1
 from cms.test_utils.testcases import CMSTestCase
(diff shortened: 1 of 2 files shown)


jonathan-s:
@aiky30 Can you please review this.

cms/tests/test_rendering.py
```diff
@@ -52,7 +91,7 @@ class RenderingTestCase(CMSTestCase):
 
     def setUp(self):
         super(RenderingTestCase, self).setUp()
-        self.test_user = self._create_user("test", True, True)
+        self.test_user = self.get_superuser()
```
Aiky30: What is the reason that this user was changed?

cms/tests/test_rendering.py
```diff
@@ -40,6 +41,44 @@ def sample_plugin_context_processor(instance, placeholder, original_context):
...
+    CMS_TEMPLATES=[
+        (TEMPLATE_NAME, TEMPLATE_NAME),
+        ('extra_context.html', 'extra_context.html')
+    ],
+)
+class RenderingEmptyTestCase(CMSTestCase):
+    def setUp(self):
+        super().setUp()
+        self.test_data = {
+            'title': u'RenderingTestCase-title',
```
Aiky30: Does this need to be enforced as unicode, i'd avoid putting in any backwards compat if it's not needed. If it's absolutely need then fair enough.

cms/tests/test_rendering.py
jonathan-s: `_create_user("test", True, True)` is the same as `get_superuser()`. get_superuser is easier to understand so I changed that.

cms/tests/test_rendering.py
jonathan-s: It was a copy and paste. I'll remove `u`. It doesn't really have any effect in python 3 anyway.

cms/tests/test_rendering.py
Aiky30: So it does. Agreed better to use the util.

jonathan-s:
🎉

PR #6806, opened 2020-02-17
Provide a general get method that can be monkeypatched

jonathan-s:
## Description
Sometimes we need to monkeypatch the get method in djangocms-versioning to get drafts. Rather than monkeypatching whole blocks of methods we can provide a single method that needs to be monkeypatched. 

If new issues crops up related to djangocms-versioning this method would be used in that place. 

### DIFF
--- a/cms/utils/helpers.py
+++ b/cms/utils/helpers.py
@@ -111,3 +111,19 @@ def is_editable_model(model_class):
     except KeyError:
         return False
     return isinstance(admin_class, FrontendEditableAdminMixin)
+
+
+def get_admin_model_object_by_id(model_class, obj_id):
+    """
+    A method to fetch an object by object id.
+
+    3rd party applications may change the queryset by the use of monkey-patching.
+    djangocms-versioning is a prime example of this. By having this helper 3rd
+    party applications can monkeypatch this helper to ensure correct results
+    without monkeypatching entire functions of django-cms or through adding
+    their logic directly to django-cms.
+
+    This helper can be reused by applications that wish to get a model that
+    should be visible to the admin.
+    """
+    return model_class.objects.get(pk=obj_id)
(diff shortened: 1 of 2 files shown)


cms/utils/helpers.py
```diff
@@ -111,3 +111,12 @@ def is_editable_model(model_class):
     except KeyError:
         return False
     return isinstance(admin_class, FrontendEditableAdminMixin)
+
+
+def _get_admin_model_object(model_class, obj_id):
+    """
```
Aiky30: As this is a helper I don't think it should be private, Private helpers are utils. I'm also now thinking that we could use this helper or a slightly different helper to get the admin version in other places that are being patched currently i.e. latest version so there aren't patches etc.

I think we should maybe rewrite the name to state that it's getting an object from the id too. 'get_admin_model_object_by_id"

I'm also thinking that we should reword this docstring slightly to sell the use for it in the codebase but still explain that versioning will "hijack" it.

jonathan-s: > As this is a helper I don't think it should be private, Private helpers are utils.

I don't get that part. Private helpers are utils, what do you mean? 

> to get the admin version in other places that are being patched currently i.e. latest version so there aren't patches etc. 

Yeah, I'll include this where it is applicable once we have decided on an approach. 

> `get_admin_model_object_by_id` 

you are correct. That's a more descriptive name. 

> I'm also thinking that we should reword this docstring slightly to sell the use for it

Do you think you could expand on the selling part?

Aiky30: > I don't get that part. Private helpers are utils, what do you mean?

Anything private for the cms only is a utility, utilities are private helpers: utils.py. I don't think that this needs to be private though, 

> Do you think you could expand on the selling part?

By selling I mean that we should state what it is first, maybe what it can be used for and then why it's here. 

What it is:
A helper to return a queryset result for a given object id.

Why:
3rd party applications such as djangocms-versioning change the queryset by the use of monkey patching. By having a separate helper such as this the 3rd party application can amend this helper to ensure that the correct result is returned to the CMS without adding their logic directly to this package.

This helper can be reused by application that wishes to get a model that should be visible to the admin.

jonathan-s:
@Aiky30 This has now been tested manually and confirmed to work with the following PR -> https://github.com/divio/djangocms-versioning/pull/206

PR #6809, opened 2020-02-24
Update django.po

domlysi:
Everything is formal language but not this line.


FinalAngel:
@domlysi thank you very much for the PR, please adapt language strings on transifex directly:
https://www.transifex.com/divio/django-cms/

Issue #6433, opened 2018-07-04
No way to publish static placeholders from from the front-end


vasilty:
### Summary
Documentation says: 
> A static placeholder needs to be published to show up on live pages, and requires a name.

There is no way to publish static placeholders from the front-end after they have been changed. Maybe I couldn't find one, of course, but documentation also didn't point me to the correct direction.

### Environment

* Python version: 3.6.4
* Django version: 1.11
* django CMS version: 3.5.1

czpython:
Hello @vasilty,
Could you provide us with a screenshot of the frontend view for your static placeholder?
A "Publish" button should appear on any url that renders a template with a static placeholder.

vasilty:
@czpython I don't know where the "Publish" button should appear, so I'm not 100% sure I made the correct screenshots. I modified the static placeholder, but I don't see a "Publish button", I get the "View published" ("Veröffentlichung anzeigen"), which shows the page with the unmodified placeholder.

<img width="1279" alt="screen shot 2018-07-09 at 14 09 05" src="https://user-images.githubusercontent.com/19282383/42449659-d113e744-8381-11e8-9cb1-9fc2c712786a.png">
<img width="848" alt="screen shot 2018-07-09 at 14 07 50" src="https://user-images.githubusercontent.com/19282383/42449711-f94eccd8-8381-11e8-9c01-39aba7672932.png">

mihalikv:
Hi,
problem is that on non CMS pages `has_publish_permission` always return `False`. I don't think that it is alright, so i change it based on following logic:

- if CMS page - check `user_can_publish_page`
- elif check `has_publish_permission` on static placeholders
- else return False

mihalikv:
@mikesdnielsen currently we use this monkeypatch:


```
from cms.cms_toolbars import PageToolbar
from cms.utils import page_permissions


def monkeypatch_cms_has_publish_permission():
    def has_publish_permission(self):
        if self.page:
            publish_permission = page_permissions.user_can_publish_page(
                self.request.user,
                page=self.page,
                site=self.current_site
            )
        elif self.statics:
            publish_permission = all(sp.has_publish_permission(self.request) for sp in self.dirty_statics)
        else:
            publish_permission = False

        return publish_permission
    return has_publish_permission


PageToolbar.has_publish_permission = monkeypatch_cms_has_publish_permission()
```

nattyg93:
[...]

PR #6812, opened 2020-03-04
Fix a bug that breaks publishing of static placeholders (#6433)

victor-yunenko:
## Related resources

https://github.com/divio/django-cms/issues/6433

## Checklist

* [ ] I have added or modified the tests when changing logic - alas I don't have the time for it at the moment

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,10 @@
 Changelog
 =========
 
+3.7.2
+==================
+* Fixed a bug that didn't allow to publish changes related to static placeholders.
+
 
 3.7.1 (2019-11-26)
 ==================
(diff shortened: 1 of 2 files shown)


FinalAngel:
@victor-yunenko can you adapt this to latest develop so it could be merged? @Aiky30 would be glad for a review.

Aiky30 (changes requested):
@victor-yunenko To be sure that the fix that you have implemented is protected by any future changes we should always add tests, especially as you have identified a bug. The test should prove that there is a bug by failing without your fix and then passing with the fix.

goutnet:
@victor-yunenko what is the root cause of this issue, can we work on the proper fix instead?

Issue #6622, opened 2019-02-25
The slugs of child pages are built incorrectly and return a 404 error when a user adds a custom slug for another language of the parent page


victor-yunenko:
### Summary

The slugs of child pages are breaking when a user adds a custom slug for another language of the parent page, this leads to incorrect urls being displayed and the correct urls return a 404 error.

### Actual behaviour

1. Create an empty website with 2 languages - English and German.
2. Create a Page with the slug `parent-en` in English and publish it, leave the German slug empty.
3. Create a child Page of the first page with the slug `child-en` in English and `child-de` in German
4. Confirm that the created child page has the url `parent-en/child-en` and the respective German url, all is fine for now.
5. Open the parent Page and add a custom German slug, specifically `parent-de` and publish all the unpublished pages.
6. Now open the German version of the website, and see in the menu that the link to the child page is `parent-en/child-de` instead of the correct `parent-de/child-de` url. Confirm that the expected url `parent-de/child-de` actually returns a 404 error.

### Expected behaviour

On the 6th the url in the menu should be `parent-de/child-de` , and the url `parent-de/child-de` shouldn't return a 404 error.

### Environment

* the first one
    * Python version: 3.5.3
    * Django version: 1.11.20
    * django CMS version: 3.6.0

* the second one
    * Python version: 2.7.6
    * Django version: 1.9.5
    * django CMS version: 3.5.1

The issue doesn't appear to be present in django-cms 3.4.5 though.

christian-schneider:
I just experienced the same issue on a 3.5.3 installation. Glad I found this report, was quite worried for a moment ;)

Foobartender:
A prerequisite for reproducing this is that at least one of the parent's (i.e. parent-(en|de)) ancestors (including the root page) does not have a German translation. If all pages in the chain have a German translation, it works fine.

If a translation in the chain is unpublished, all children of the page will show a tooltip "unpublished parent" in the page tree view for the publish icon of the translation.

I also have a page with an CMSApp hook and a CMSAttachMenu only in English. The CMSAttachMenu will raise NoReverseMatch errors if another language is selected, presumably because the urlpatterns have not been loaded.

So, I'd guess the fallback mechanism isn't working properly (I have configured fallbacks in CMS_LANGUAGES).

Foobartender:
cms.models.pagemodel.mark_descendants_as_published() updates the paths of the published titles of its children:
[...]

PR #6813, opened 2020-03-04
A workaround for a bug that breaks page slugs in a multi-language configuration (#6622)

victor-yunenko:
## Description

This is a temporary fix for #6622 that simply calls `page._update_title_path(language)` method one more time when a user changes the page slug.

By the way as far as I remember that issue is also solvable by calling `page.save()` method one more time, but this workaround is better targeted at the source of the problem.

## Related resources

https://github.com/divio/django-cms/issues/6622

## Checklist

* [ ] I have added or modified the tests when changing logic - again alas I haven't had the time for it. In short this patch is in use on 3-4 projects in live at the moment, with django-cms from 3.5 to 3.7.1. I'm using it as part of [djangocms-helpers](https://pypi.org/project/djangocms-helpers/) package.

### DIFF
--- a/cms/signals/slug_update_fix.py
+++ b/cms/signals/slug_update_fix.py
@@ -1,0 +1,32 @@
+"""
+This is a temporary fix that should be replaced with an adequate bug fix to the Page logic.
+"""
+
+from django.dispatch import receiver
+from django.db.models.signals import post_save
+from fieldsignals import post_save_changed
+
+from cms.models import Title, Page
+
+
+@receiver(post_save, sender=Page)
+def fix_page_slugs(sender, instance, **kwargs):
+    _update_child_pages(instance)
+
+
+@receiver(post_save_changed, sender=Title, fields=['slug', 'path'])
+def fix_title_path(sender, instance, **kwargs):
+    _update_child_pages(instance.page)
+
+
+def _update_child_pages(page):
+    for language in page.get_languages():
+        _update_title_path(page, language)
+        for child_page in page.get_child_pages():
+            _update_title_path(child_page, language)
+
+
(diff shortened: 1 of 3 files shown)


FinalAngel:
@victor-yunenko I^m also not a fan of adding additional requirements. Are there alternatives available? /cc @Aiky30

Aiky30 (changes requested):
I'm not a fan of temporary work arounds, can we not just fix this in the model directly?

cms/signals/slug_update_fix.py
```diff
@@ -0,0 +1,32 @@
+"""
+This is a temporary fix that should be replaced with an adequate bug fix to the Page logic.
```
Aiky30, line 2: I'd opt to fix this properly rather than bring in a temporary fix that never actually gets fixed.

cms/signals/slug_update_fix.py
```diff
@@ -0,0 +1,32 @@
...
+from cms.models import Title, Page
+
+
+@receiver(post_save, sender=Page)
+def fix_page_slugs(sender, instance, **kwargs):
+    _update_child_pages(instance)
+
+
+@receiver(post_save_changed, sender=Title, fields=['slug', 'path'])
+def fix_title_path(sender, instance, **kwargs):
```
Aiky30, line 18: We do have signals already created, we should just bind to those. V4 reworks the signal in a better fashion.

Aiky30 (changes requested):

goutnet:
@Aiky30 can we look into a more proper fix in 3.9.x then? (since 4.x is  still a long way)

PR #6817, opened 2020-03-10
Add method that allows to monkeypatch how models are filtered

jonathan-s:
## Description
Instead of monkeypatching entire block of methods in djangocms-versioning we can monkeypatch single small methods to accomplish the same thing. 

This PR adds a method `filter_admin_model`. The intended use for this method is to be able to fetch objects that are in draft in djangocms-versioning, that package will monkeypatch that method to modify the behaviour so that draft objects can be found. 

## Checklist


jonathan-s:
Closing this as the approach won't give us the desirable outcome that we need. This coupled together with versioning renders the object in the wrong way.

PR #6825, opened 2020-03-25
Accessing a page 404 on draft language

jonathan-s:
## Description
When accessing a page in a specific language we can't use a fallback. If we do we'll show an english page even though we requested a french page.

We need to check that the page that we request is not EmptypageContent if it is the user should not see the page.

## Checklist

Since this logic only manifests itself when using djangocms-versioning the tests have been placed in that page. You can view the currently failing tests there. https://github.com/divio/djangocms-versioning/pull/217

### DIFF
--- a/cms/utils/__init__.py
+++ b/cms/utils/__init__.py
@@ -1,8 +1,13 @@
 # -*- coding: utf-8 -*-
 # TODO: this is just stuff from utils.py - should be splitted / moved
+from functools import lru_cache
+
 from django.conf import settings
 from django.core.files.storage import get_storage_class
 from django.utils.functional import LazyObject
+from django.urls import resolve
+from django.contrib.contenttypes.models import ContentType
+from django.urls.exceptions import Resolver404
 
 from cms.utils.conf import get_site_id  # nopyflakes
 from cms.utils.i18n import get_default_language
@@ -16,6 +21,34 @@ def get_current_site():
     return Site.objects.get_current()
 
 
+@lru_cache(512)
+def get_object_language(request):
+    """
+    We resolve the request and try to get content type if there is one.
+    If the content type has a language, it's the language we want to use.
+    """
+    try:
+        resolved = resolve(request.path)
+    except Resolver404:
+        return None
+
+    args = resolved.args
+    if len(args) != 2 or not all([a.isnumeric() for a in args]):
+        # if it is not a content type we can't get a language from it.
+        return None
+
+    content_type_id, object_id = args
+
+    try:
+        content_type = ContentType.objects.get_for_id(content_type_id)
+        content_type_obj = content_type.get_object_for_this_type(pk=object_id)
+        language = content_type_obj.language
+    except (ContentType.DoesNotExist, AttributeError):
+        return None
+
+    return language
+
+
 def get_language_from_request(request, current_page=None):
     """
     Return the most obvious language according the request
@@ -23,8 +56,14 @@ def get_language_from_request(request, current_page=None):
     language = None
     if hasattr(request, 'POST'):
         language = request.POST.get('language', None)
-    if hasattr(request, 'GET') and not language:
+    elif hasattr(request, 'GET'):
         language = request.GET.get('language', None)
+
+    if not language:
+        # if we still don't have language from get or post request
+        # we try to get it from the content object of the request.
+        language = get_object_language(request)
+
     site_id = current_page.node.site_id if current_page else None
     if language:
         language = get_language_code(language)
--- a/cms/views.py
+++ b/cms/views.py
@@ -164,7 +164,7 @@ def details(request, slug):
     if page.login_required and not request.user.is_authenticated:
         return redirect_to_login(urlquote(request.get_full_path()), settings.LOGIN_URL)
 
-    content = page.get_title_obj(language=request_language)
+    content = page.get_title_obj(language=request_language, fallback=False)
     # use the page object with populated cache
     content.page = page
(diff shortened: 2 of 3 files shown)


cms/utils/__init__.py
```diff
@@ -16,15 +21,42 @@ def get_current_site():
...
 
+@lru_cache(512)
+def get_object_language(request):
+    try:
+        resolved = resolve(request.path)
+    except Resolver404:
+        return None
+
+    args = resolved.args
+    if len(args) != 2 or not all([a.isnumeric() for a in args]):
```
Aiky30: Why !=2? Need to add comments to know what this is doing.

cms/utils/__init__.py
```diff
@@ -16,15 +21,42 @@ def get_current_site():
...
+
 def get_language_from_request(request, current_page=None):
     """
     Return the most obvious language according the request
     """
     language = None
     if hasattr(request, 'POST'):
         language = request.POST.get('language', None)
-    if hasattr(request, 'GET') and not language:
+    elif hasattr(request, 'GET'):
```
Aiky30: Makes sense, bad logic before.

cms/views.py
```diff
@@ -164,7 +164,7 @@ def details(request, slug):
     if page.login_required and not request.user.is_authenticated:
         return redirect_to_login(urlquote(request.get_full_path()), settings.LOGIN_URL)
 
-    content = page.get_title_obj(language=request_language)
+    content = page.get_title_obj(language=request_language, fallback=False)
```
Aiky30: I guess with this the fallback option is probably never actually ever used now. I get why you're doing this though.

cms/utils/__init__.py
```diff
@@ -16,15 +21,42 @@ def get_current_site():
...
         language = request.GET.get('language', None)
+
+    if not language:
```
Aiky30: Can you add a comment why we need to get the language here, maybe enforcing why you haven't used the fallback language. There arn't enough comments in the core. They don't have to be every line but defo anytime we need to do something out of the ordinary or not obvious why.

cms/utils/__init__.py
```diff
@@ -16,15 +21,42 @@ def get_current_site():
     return Site.objects.get_current()
 
 
+@lru_cache(512)
```
Aiky30: What is the justification for the value? Also lru_cache is not Python 2.7 compat is it?

cms/utils/__init__.py
```diff
@@ -16,15 +21,42 @@ def get_current_site():
...
+def get_object_language(request):
```
Aiky30: Maybe get_language_from_content_object, not that important.

Please add a doc type to help anyone using htis in the core, as it's a util it is only supposed to be used in the core checking that is as intended. Helpers are for external reuse to be kept backwards compatible where possible.

cms/views.py
jonathan-s: Tbh I don't fully understand why you'd want a fallback in the first place. Either you get the content you requested or you don't. If there was a "fallback" it should happen through a redirect in my opinion, not through secretly replacing the content.

cms/utils/__init__.py
jonathan-s: You are correct. Python 2.7 is end of life though. Is django-cms committed to maintain this for 2.7? For how long?

cms/views.py
jonathan-s: The fallback option might be used elsewhere implicitly. I haven't checked everywhere.

cms/utils/__init__.py
jonathan-s: Added a comment.

cms/utils/__init__.py
jonathan-s: in travis.yml they allow failures for python 2.7. So it looks like it is as good as deprecated.

Aiky30: I agree that it is as good as dead, issue is the project still supports it so our implementation should keep that support until a decision to not support is broadcast. We don't have that authority and It's changes like this that could jeopardise an agreement from Divio to support our changes. 2.7 is still in develop as supported: https://github.com/divio/django-cms/blob/develop/setup.py

cms/views.py
Aiky30: The fallback is for when you have a site in many languages there is a requirement that all of the pages didn't have to have a translation, it would serve a fallback. I'm not sure if this change would be accepted as this will change it's core behaviour:

"django CMS has a sophisticated multilingual capability. It is able to serve content in multiple languages, with fallbacks into other languages where translations have not been provided. It also has the facility for the user to set the preferred language and so on." 
http://docs.django-cms.org/en/latest/topics/multiple_languages.html

cms/utils/__init__.py
Aiky30: Our target is to merge into develop soon.

Aiky30:
@FinalAngel Can you please review this PR to get Divio's opinion on the various comments please?

FinalAngel:
@Aiky30 I'm probably the wrong one to review this change, maybe @victor-yunenko ?

Aiky30:
> @Aiky30 I'm probably the wrong one to review this change, maybe @victor-yunenko ?

@FinalAngel We have not yet proved that the fix is required, it will either be closed or merged by this time next week. We have a team of testers on the case now.

Aiky30:
The testers were unable to replicate this issue, it's possible that other changes / fixes have since resolved the issue addressed in this PR.

PR #6826, opened 2020-03-25
The language menu only shows active translations and points to preview url

jonathan-s:
## Description

The language menu should point to the preview url instead of the live url. The reason for this is that content may not be available for whatever reason. Ie if a page is not published in djangocms-versioning this page should return 404 for the live url. 

We further clarify that a translation is not available by disabling the item if we can't successfully resolve the path. 

## Checklist

Note: This currently does not work well with drafts/versioning, as `get_title_obj` can't find a draft.

### DIFF
--- a/cms/cms_toolbars.py
+++ b/cms/cms_toolbars.py
@@ -313,13 +313,16 @@ class BasicToolbar(CMSToolbar):
     def add_language_menu(self):
         if settings.USE_I18N and not self._language_menu:
             self._language_menu = self.toolbar.get_or_create_menu(LANGUAGE_MENU_IDENTIFIER, _('Language'), position=-1)
-            language_changer = getattr(self.request, '_language_changer', DefaultLanguageChanger(self.request))
+            page = self.request.current_page
             for code, name in get_language_tuple(self.current_site.pk):
                 try:
-                    url = language_changer(code)
-                except NoReverseMatch:
+                    content = page.get_title_obj(code, fallback=False)
+                    url = get_object_preview_url(content)
+                    disabled = False
+                except (NoReverseMatch, AttributeError):
                     url = DefaultLanguageChanger(self.request)(code)
-                self._language_menu.add_link_item(name, url=url, active=self.current_lang == code)
+                    disabled = True
+                self._language_menu.add_link_item(name, url=url, active=self.current_lang == code, disabled=disabled)
 
     def get_username(self, user=None, default=''):
         user = user or self.request.user
--- a/cms/tests/test_toolbar.py
+++ b/cms/tests/test_toolbar.py
@@ -463,6 +463,36 @@ class ToolbarTests(ToolbarTestBase):
         toolbar = CMSToolbar(request)
         self.assertTrue(toolbar.show_toolbar)
 
+    def test_change_language_menu_two_languages(self):
+        """
+        Only english language is active. German and English is not disabled.
+        The language link points to the preview url instead of live url.
+        """
+        page = create_page("english-page", "nav_playground.html", "en")
+        german_content = create_title("de", "german content", page)
+        english_content = page.get_title_obj('en')
+        staff = self.get_staff()
+        self.client.force_login(staff)
+
+        response = self.client.get(page.get_absolute_url())
+        menus = response.context['cms_toolbar'].menus
+        language_menu = menus['language-menu']
+        english = language_menu.items[0]
+        english_url = get_object_preview_url(english_content)
+        not_disabled_languages = [i for i in language_menu.items if i.disabled == False]
+
+        german = language_menu.items[1]
+        german_url = get_object_preview_url(german_content)
+
+        self.assertEqual(german.active, False)
+        self.assertEqual(german.disabled, False)
+        self.assertEqual(german.url, german_url)
+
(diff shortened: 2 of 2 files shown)


cms/cms_toolbars.py
```diff
@@ -313,13 +313,16 @@ def add_logout_button(self, parent):
     def add_language_menu(self):
         if settings.USE_I18N and not self._language_menu:
             self._language_menu = self.toolbar.get_or_create_menu(LANGUAGE_MENU_IDENTIFIER, _('Language'), position=-1)
-            language_changer = getattr(self.request, '_language_changer', DefaultLanguageChanger(self.request))
+            page = self.request.current_page
```
Aiky30: There seemed to be effort previously to not use page. I suspect that this can be used by other content types that use the toolbar, I'll need to check.

cms/cms_toolbars.py
```diff
@@ -313,13 +313,16 @@ def add_logout_button(self, parent):
...
             for code, name in get_language_tuple(self.current_site.pk):
                 try:
-                    url = language_changer(code)
-                except NoReverseMatch:
+                    content = page.get_title_obj(code, fallback=False)
+                    url = get_object_preview_url(content)
+                    disabled = False
```
Aiky30: Why has the disabled item been added? I do remember seeing the language menu disabling before, surely that's the active?

cms/cms_toolbars.py
Aiky30: I'm seeing _language_changer used in a lot of places, can I check that you're also testing the changes with the CMS in standalone and not with all of the additional addons installed? Most of the tests in this area are isolating the functionality so they'll still pass as they don't appear to be touching this add_language_menu method.

cms/cms_toolbars.py
jonathan-s: active and disabled are two different things. Active means that it's the item that is currently selected. Disabled means that you can't click on the item to begin with.

cms/cms_toolbars.py
jonathan-s: If I put in an `raise Exception` here it will fail in 97 different places. So `add_language_menu` is certainly being touched in quite a few different tests. 

> There seemed to be effort previously to not use page. I suspect that this can be used by other content types that use the toolbar, I'll need to check.

I don't really follow, to not use page? as in self.request.current_page?

If self.request.current_page happens to be None, because the url wasn't able to resolve. We fallback to using DefaultLanguageChanger which will give you the live url. So it would fail gracefully.

jonathan-s: @aiky30 To follow up on what will happen when no page exists such as for an alias placeholder. 

In the try ... except block will get an attribute error. So we catch the attribute error and use DefaultLanguageChanger instead. 

That will be called with the following code. 

https://github.com/divio/django-cms/blob/release/4.0.x/menus/utils.py#L151-L160

So it will give the correct url for the alias placeholder.

jonathan-s:
@FinalAngel Do you think you could review this as well.

cms/cms_toolbars.py
```diff
@@ -313,13 +313,16 @@ def add_logout_button(self, parent):
     def add_language_menu(self):
         if settings.USE_I18N and not self._language_menu:
             self._language_menu = self.toolbar.get_or_create_menu(LANGUAGE_MENU_IDENTIFIER, _('Language'), position=-1)
-            language_changer = getattr(self.request, '_language_changer', DefaultLanguageChanger(self.request))
+            page = self.request.current_page
             for code, name in get_language_tuple(self.current_site.pk):
                 try:
-                    url = language_changer(code)
-                except NoReverseMatch:
+                    content = page.get_title_obj(code, fallback=False)
```
jonathan-s: If we need to access drafts (which we do) we need to add `get_admin_title_obj` in here as described in this PR https://github.com/divio/django-cms/pull/6827

cms/cms_toolbars.py
Aiky30: Looking further into this this change is made to the BasicToolbar that won't always have a page hence the solution already in place here.

cms/cms_toolbars.py
```diff
@@ -313,13 +313,16 @@ def add_logout_button(self, parent):
...
+                    url = get_object_preview_url(content)
+                    disabled = False
+                except (NoReverseMatch, AttributeError):
                     url = DefaultLanguageChanger(self.request)(code)
-                self._language_menu.add_link_item(name, url=url, active=self.current_lang == code)
+                    disabled = True
```
jonathan-s, line 324: @Aiky30 I can't remember exactly, but when you click on the language and there is no content for that language yet. Should it create the page? If I remember it correctly the current behaviour is that it would fall back to different language, which is an odd behaviour.

cms/cms_toolbars.py
```diff
@@ -313,13 +313,16 @@ def add_logout_button(self, parent):
     def add_language_menu(self):
```
Aiky30, line 313: The test in this PR is still valid and the language link should still use the object_preview_url should be used for consistency. The following PR enables the features in Versioning: https://github.com/divio/djangocms-versioning/pull/223

Aiky30:
Covered by versioning and no longer required. Thank you.

PR #6827, opened 2020-03-27
Introduce get_admin_title_obj

jonathan-s:
The following method allow you to fetch pagecontent that is
safe for admin usage. This allows for modifying store objects
in the cache that would not show up for an end user.

It also allows for a clean way of mocking for external packages.

Tests for this functionality can be found here https://github.com/divio/djangocms-versioning/pull/217

This PR supersedes #6820


Aiky30:
This requires discussion and structured way of achieving this cms wide, especially for rethinking how versioning integrates with the cms in the future.

PR #6828, opened 2020-03-27
Display the correct url in change_language_menu

jonathan-s:
Depends on #6829 

The urls should not use `self.title.pk` it needs to use the pagecontent from their respective language to generate the correct urls. 

To access drafts we first need to merge #6827 and then modify the code to use `get_admin_title_obj`. 

### DIFF
--- a/cms/cms_toolbars.py
+++ b/cms/cms_toolbars.py
@@ -430,11 +430,11 @@ class PageToolbar(CMSToolbar):
                     add_plugins_menu.add_modal_item(name, url=url)
 
             if remove:
-                translation_delete_url = admin_reverse('cms_pagecontent_delete', args=(self.title.pk,))
-
                 remove_plugins_menu = language_menu.get_or_create_menu('{0}-del'.format(LANGUAGE_MENU_IDENTIFIER), _('Delete Translation'))
                 disabled = len(remove) == 1
                 for code, name in remove:
+                    pagecontent = self.page.get_title_obj(code)
+                    translation_delete_url = admin_reverse('cms_pagecontent_delete', args=(pagecontent.pk,))
                     url = add_url_parameters(translation_delete_url, language=code)
                     remove_plugins_menu.add_modal_item(name, url=url, disabled=disabled)
 
@@ -443,9 +443,9 @@ class PageToolbar(CMSToolbar):
                 title = _('from %s')
                 question = _('Are you sure you want to copy all plugins from %s?')
 
-                page_copy_url = admin_reverse('cms_pagecontent_copy_language', args=(self.title.pk,))
-
                 for code, name in copy:
+                    pagecontent = self.page.get_title_obj(code)
+                    page_copy_url = admin_reverse('cms_pagecontent_copy_language', args=(pagecontent.pk,))
                     copy_plugins_menu.add_ajax_item(
                         title % name, action=page_copy_url,
                         data={'source_language': code, 'target_language': self.current_lang},
--- a/cms/tests/test_toolbar.py
+++ b/cms/tests/test_toolbar.py
@@ -463,6 +463,44 @@ class ToolbarTests(ToolbarTestBase):
         toolbar = CMSToolbar(request)
         self.assertTrue(toolbar.show_toolbar)
 
+    def test_remove_and_copy_urls_are_correctly_associated_with_pagecontent(self):
+        """
+        The urls to copy and remove translations should be linked correctly.
+        """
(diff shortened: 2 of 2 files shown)


cms/cms_toolbars.py
```diff
@@ -430,11 +430,11 @@ def change_language_menu(self):
                     add_plugins_menu.add_modal_item(name, url=url)
 
             if remove:
-                translation_delete_url = admin_reverse('cms_pagecontent_delete', args=(self.title.pk,))
-
                 remove_plugins_menu = language_menu.get_or_create_menu('{0}-del'.format(LANGUAGE_MENU_IDENTIFIER), _('Delete Translation'))
                 disabled = len(remove) == 1
                 for code, name in remove:
+                    pagecontent = self.title.page.get_title_obj(code)
```
Aiky30: This is very strange, self.title is the title object

jonathan-s: Ah, we got `self.page` as well, so it would make more sense to use that.

Aiky30: Would this need to use the get_admin_title method to be able to get the drafts being as any delete or copy commands will be executed by an admin user only?

jonathan-s: I would say yes, it would indeed need to use `get_admin_title_obj` to be able to access drafts.

cms/cms_toolbars.py
```diff
@@ -430,11 +430,11 @@ def change_language_menu(self):
                     add_plugins_menu.add_modal_item(name, url=url)
 
             if remove:
-                translation_delete_url = admin_reverse('cms_pagecontent_delete', args=(self.title.pk,))
-
                 remove_plugins_menu = language_menu.get_or_create_menu('{0}-del'.format(LANGUAGE_MENU_IDENTIFIER), _('Delete Translation'))
                 disabled = len(remove) == 1
                 for code, name in remove:
+                    pagecontent = self.page.get_title_obj(code)
```
Aiky30, line 436: I've seen an issue with versioning where the edit button always points at the url prefixed /en/. This issue affects any actions selected in edit mode.

Steps to replicate:
got to page with multiple languages
select a language (takes you to the preview url)
select the edit button (redirects you to the edit url prefixed with /en/

Solution: 
Fix the edit redirect code in versioning

jonathan-s: If you make any new PR, send me a ping and I can at least read through it.

Aiky30: @jonathan-s This change belongs in Versioning, if this fix was merged here it would be overwritten by versioning: https://github.com/divio/djangocms-versioning/blob/master/djangocms_versioning/monkeypatch/cms_toolbars.py#L51

Am i Missing something or was this not tested in person but rather TDD standalone? Let me know and we can make the change in versioning to use the preview url for drafts.

jonathan-s: Hello there, so the idea was that we should be able to remove that entire block of code in versioning. Ie, one monkeypatch less than before.

Aiky30: @jonathan-s Hello mate. I think you're onto something here as the buttons would be generated wrong if versioning is not installed. I think we will still need to bring this change across to Versioning too in the short term as the remove feature is not yet required in versioning because versioning manages removing that feature for now. I think once we have an idea of how the remove should work if it should be present at all then we can look at removing the monkey patch in versioning all together.

jonathan-s: Yes, the one reason that it still needs to be in place is described here: https://github.com/divio/djangocms-versioning/pull/220

Once removal of urls is working correctly, you should be able to remove `change_language_menu` from versioning.

Aiky30: I'm confused why this is even monkey patched when you see a fix that allows the replacement of a toolbar rather than monkey patching it. I actually prefer that as it's a nicer way of redefining the methods rather than monkey patching. 

25th Oct: https://github.com/divio/djangocms-versioning/commit/97f4bc6b62a7844cf7657830295c00b6770cd65d

23rd Oct: https://github.com/divio/djangocms-versioning/commit/eecba2338b3f98f479d16645b16aa84cbc5b85e8

jonathan-s: I don't know why it was done the way it was. Either way, I hope you understand the reasoning of the code changes I've made.

Aiky30 (approved):

jonathan-s:
🎉

PR #6832, opened 2020-03-30
Add Oracle support to custom plugin queries.

jonathan-s:
This fixes two issues related to the database related to Oracle. 

### DIFF
--- a/cms/models/placeholdermodel.py
+++ b/cms/models/placeholdermodel.py
@@ -728,7 +728,7 @@ class Placeholder(models.Model):
             )
             sql = sql.format(connection.ops.quote_name(CMSPlugin._meta.db_table))
             cursor.execute(sql, [self.pk, language])
-        else:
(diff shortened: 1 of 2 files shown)


cms/models/placeholdermodel.py
```diff
@@ -729,6 +729,13 @@ def _recalculate_plugin_positions(self, language):
             sql = sql.format(connection.ops.quote_name(CMSPlugin._meta.db_table))
             cursor.execute(sql, [self.pk, language])
         else:
-            raise RuntimeError(
-                '{} is not supported by django-cms'.format(connection.vendor)
+            sql = (
```
Aiky30: The raise exception has now been removed for any other unsupported db's. This should be an else if that catches Oracle.

jonathan-s: Fixed.

PR #6847, opened 2020-04-29
Added ability to set placeholder global limit on children only

g3rb3n:
## Description
Added the ability to set placeholder children limit. 

With the current limits global setting, all plugins in the placeholder are counted. If one wants to restrict the placeholder to just one plugin, but with some plugins that allow for children this setting is useless. 

## Related resources
*https://groups.google.com/forum/#!topic/django-cms-developers/24UU7yc1A70

## Checklist


FinalAngel:

Aiky30:
I have seen requests for this change in the past, I think I've even seen some open issues. It would be good to know if this change satisfies previous use cases. I'll see what I can find to compare this change with requirements wise.

goutnet (approved):
LGTM

Issue #6448, opened 2018-07-14
Support prefix_default_language = False


tgckpg:
### Summary
When
`i18n_patterns( prefix_default_language = False )`

Page redirect settings do not work properly on the default language.

Problematic file:
```
Calls:
0 cms/views.py: def details( ... )
1 cms/views.py: if (is_language_prefix_patterns_used() ...
2 cms/utils/i18n.py: def is_language_prefix_patterns_used => True ( should be False on default language )
```

Reference fix:
`django/conf/urls/i18n.py: def is_language_prefix_patterns_used`

### Expected behaviour
On default language, page should be redirected without language prefix.

### Actual behaviour
On default language, page redirects with language prefix.

### Environment

* Python version: 3.6.0
* Django version: 1.11.7
* django CMS version: 3.4.6

nabinbhusal80:
I faced the same problem as well. The redirection does not work. If always sticks to the first language that is input and no changes take place after that.

tgckpg:
Since the devs didn't respond for a long time and the issue is rather critical IMO...

So here's the monkey patch I've written for anyone who come across this issue.

Place this under `urls.py` ( or some other better places ).
```python
from cms import views as cms_views
from django.urls import LocaleRegexURLResolver, get_resolver
def _patch_lang_pfx():
    for url_pattern in get_resolver(None).url_patterns:
        if isinstance(url_pattern, LocaleRegexURLResolver):
            return url_pattern.prefix_default_language
    return False
# Swap out the faulty method from cms.views
cms_views.is_language_prefix_patterns_used = _patch_lang_pfx
```

ivanfr90:
I'm suffering the same. 
Redirections always go to the path language without keep in mind if `prefix_default_language` is configured in url's definition.

jscontrust:
Same here. Would be great to have the `prefix_default_language` option implemented to django cms

tgckpg:
Thanks! Issue is fixed in 3.7.0. I am closing this now.

Zeerooth:
Unfortunately this still hasn't been fixed as of django-cms 3.7.1, django 2.2.11 and python 3.7. Redirects append the language prefix to urls even if prefix_default_language=False

Here are my url settings: 
```
urlpatterns += i18n_patterns(
    path('admin/', admin.site.urls),
    url(r'^', include('cms.urls')),
    url(r'^', include('filer.server.urls')),
    prefix_default_language=False,
)
```

[...]

PR #6851, opened 2020-05-05
prefix_default_language = False redirect behavior fixes

Zeerooth:
## Description

This patch aims to mitigate some issues with prefix_default_language set to False, mainly redirect errors:
* When page had a redirect set to it language prefix would ALWAYS get added at the beggining of the path
* When user browser language was set to a language that had a prefix added to it and a user tried to enter page in other language that was a default one while prefix_default_language = False CMS would redirect them to their preferred language therefore blocking them from ever entering that page

## Related resources

* https://github.com/divio/django-cms/issues/6448
* https://code.djangoproject.com/ticket/29425 (this behavior actually stems from Django itself but it was agreed that it is intended although I am sure that when use in CMS it leads to bad user experience and should be mitigated)

## Checklist


FinalAngel:

@Zeerooth would you be able to solve the merge conflicts from master and remove the release date in the changelog?

goutnet (approved):
LGTM

PR #6864, opened 2020-06-05
Remove duplicate `attr` attribute

utahguy:
## Description

The description for the `attr` attribute was listed twice, once with more detail and once with a partial description. Removed the less detailed one.

## Related resources
None

## Checklist

  * Don't think this is necessary for such a small change. But, if you'd like, happy to do so.
  * No logic changes.


FinalAngel:
Thank you very much @utahguy

FinalAngel (approved):

PR #6866, opened 2020-06-05
fix: Clear page permission cache on page create

### LINKED ISSUE #6865: Page permission cache does not contain newly created pages
## Description
When a staff user, with django permissions on the page content type and page permissions on the parent, saves a page and wants to change the page thereafter he is not allowed to do so.

The cause is in the permission cache, if I disable the caching by changing the following in cms/utils/page_permissions.py it works.
- page_ids = func(user, site, check_global=check_global) 
+ page_ids = func(user, site, check_global=check_global, use_cache=False)

Also a restart of the django server solves the problem. 

## Steps to reproduce
```
import logging

from cms import api
from cms.models import PagePermission
from django.contrib.sites.models import Site
from django.contrib.auth.models import Group, Permission
from django.contrib.contenttypes.models import ContentType
from django.test import Client, TestCase
from os import path


logger = logging.getLogger(__name__)
User = get_user_model()
language = "en"


class AddPageTestCase(TestCase):

    def add_permission(self, app_label, model, codename):
        content_type = ContentType.objects.get(
            app_label=app_label,
            model=model,
        )
        permission = {
            "codename": codename,
            "content_type": content_type,
        }
        permission = Permission.objects.get(**permission)
        self.group.permissions.add(permission)

    def setUp(self):
        template = settings.CMS_TEMPLATES[0][0]
        site = Site.objects.get(pk=1)
        page = api.create_page("Title", template, language, site=site)
        self.page = page

        group = Group.objects.create(name="editors")
        self.group = group

        user = User.objects.create_user("editor", password="password")
        user.is_staff = True
        user.groups.add(group)
        user.save()
        self.user = user

        self.add_permission("cms", "page", "add_page")
        self.add_permission("cms", "page", "change_page")
        self.add_permission("cms", "page", "delete_page")
        self.add_permission("cms", "page", "publish_page")
        self.add_permission("cms", "page", "view_page")

        PagePermission.objects.create(group=group, page=page)

        #usersites_api.create_user_site(user, site)

        settings.SITE_ID = 1

    def test_add_page(self):
        login = self.client.login(username=self.user.username, password='password')
        self.assertTrue(login)

        print('==========')
        print('Test 1')
        response = self.client.get('/{}/admin/'.format(language))
        self.assertEqual(response.status_code, 200)

        print('==========')
        print('Test 2')
        response = self.client.get("/{}/admin/cms/page/".format(language))
        self.assertEqual(response.status_code, 200)

        print('==========')
        print('Test 3')
        response = self.client.get("/{}/admin/cms/page/add/?parent_node={}".format(language, self.page.node.pk))
        self.assertEqual(response.status_code, 200)

        print('==========')
        print('Test 4')
        data = {
            "language": language,
            "source": "",
            "title": "test",
            "slug": "test",
            "menu_title": "",
            "page_title": "",
            "meta_description": "",
            "parent_node": self.page.node.pk,
            "_continue": "Opslaan en opnieuw bewerken"
        }
        response = self.client.post("/{}/admin/cms/page/add/?language={}&parent_node={}".format(language, language, self.page.node.pk), data)

        print('==========')
        print('Test 5')
        location = response.get('Location')
        print(location)
        response = self.client.get(location)
        self.assertEqual(response.status_code, 200)
```
## Expected behaviour
Newly created pages should be added to the permission cache.
Step 5 should pass.

## Actual behaviour
Step 5 gives a 403

## Additional information
python 3.6
Django 2.2.10
django-cms 3.7.1

g3rb3n:
## Description
Fixes #6865 
Just added one line to Page.save method
```
if created:
    clear_permission_cache()
```

## Related resources
* #6865 

## Checklist


FinalAngel:

marksweb (approved):

Issue #6026, opened 2017-07-19
Frontend Editing Delete Plugin does not work [branch develop]


SteinRobert:
### Summary

I was testing the new frontend editing behavior implmented in https://github.com/divio/django-cms/pull/5980 
It seems like deleting plugin does not work properly. They are deleted after reload.

### Expected behaviour

Plugins should disappear in structure and content mode after deleting.

#### Actual behaviour

Plugins are removed from the frontend after a reload. When the modal for deleting a plugin is opened an JavaScript error occurs: 

``Uncaught ReferenceError: django is not defined
    at cancel.js:9``

I am using a simple Django CMS setup with the TextPlugin.

#### Environment

* Python version: 2.7.10
* Django version: 1.10.7
* django CMS version: 3.4.3 (develop branch)

czpython:
Hello @SteinRobert,
Thanks for trying out the trunk version!

Currently develop is unstable as we're finalizing minor fixes on https://github.com/divio/django-cms/issues/6026. It would be great if you try out that branch instead.

Also, in order for the text plugin to work, you'll need the latest rc release, available in https://github.com/divio/djangocms-text-ckeditor.

PR #6871, opened 2020-06-19
add jQuery to local namespace in order to initialize cancel.js

jrief:
## Description

#6026 has been closed without being fixed. This pull request fixes that issue.

How to reproduce:
* Edit a CMS page and go into structure mode.
* Open Developer Tools in the browser.
* In Structure Mode, click onto **Delete** to delete a plugin.
* A popup window appears. The developer tools complain: `Uncaught ReferenceError: django is not defined`.
* Therefore the button "No, take me back" located inside the frame, doesn't work anymore.

This is because `cancel.js` attempts to access `django.jQuery` which is not yet added to the global JavaScript namespace.
And the reason for the latter is, that the Django View method
`cms.admin.placeholderadmin.PlaceholderAdminMixin.delete_plugin()`
does not define any media in its context.

## Checklist

* [x] I have opened this pull request against ``release/3.7.x``
* [o] I have added or modified the tests when changing logic


FinalAngel:

FinalAngel:
@jrief I'm unable to reproduce this issue using Django 2.2, 3.0 or 3.1. Is it only an issue for 1.11?

cms/admin/placeholderadmin.py
```diff
@@ -1105,6 +1106,7 @@ def delete_plugin(self, request, plugin_id):
             "protected": protected,
             "opts": opts,
             "app_label": opts.app_label,
+            "media": Media(js=['admin/js/vendor/jquery/jquery.min.js', 'admin/js/jquery.init.js'])
```
Aiky30, line 1109: 'admin/js/vendor/jquery/jquery.min.js' is redundant when 'admin/js/jquery.init.js' is present because jquery.init.js is already dependant on jquery.

Aiky30 (changes requested):

jrief:
I'm still able to reproduce this with the latest (3.7.4) version of Django-CMS:
* Add a plugin as a child of a container plugin.
* Open the developer tools of your browser.
* In structure mode click on the hamburger to the right of the plugin. After the context opens its popup, click on **Delete**.

The browser console shows this error:
Screenshot 2020-10-15 at 10 39 04

One of the reasons you might be unable to reproduce this, is because maybe you are using `djangocms-admin-style`.
I never use this package, because I prefer the default Django admin style. In my opinion, this however shouldn't be mandatory.

jrief:
@Aiky30 removing `'admin/js/vendor/jquery/jquery.min.js'` as you suggested isn't such a good idea. If I do that, I get this error in the browser console:

Screenshot 2020-10-15 at 13 51 05

cms/admin/placeholderadmin.py
jrief: Removing `'admin/js/vendor/jquery/jquery.min.js'` as you suggested isn't such a good idea. If I do that, I get this error in the browser console:

Screenshot 2020-10-15 at 13 51 05

Aiky30: Strange, the Django docs mention that you only need to add the init: https://docs.djangoproject.com/en/3.1/releases/2.2/#merging-of-form-media-assets

Aiky30 (approved):
This is potentially a fix that cam along as part of an asset loading change from Django: https://docs.djangoproject.com/en/3.1/releases/2.2/#merging-of-form-media-assets

cms/admin/placeholderadmin.py
jrief: Anyway, adding that `media`-attribute, shouldn't do any harm. Did you try yourself without `djangocms_admin_style` in `INSTALLED_APPS`?

goutnet:
@jrief  as this had been accepted some time ago, I'd like to merge it, but some questions:
* Is this a merge we should send onto `develop` ?
* is this still needed?

PR #6886, opened 2020-08-06
Django 3.1 support

FinalAngel:
* Added support for Django 3.1


FinalAngel:
@yakky and @Aiky30 would you be able to have a first look at the code? Overall it works so far, upgrading the other addons according to this.

yakky:
@FinalAngel it looks straighforward to me :+1:

cms/api.py
```diff
@@ -59,7 +57,7 @@ def _verify_apphook(apphook, namespace):
         apphook_name = apphook.__class__.__name__
     elif hasattr(apphook, '__module__') and issubclass(apphook, CMSApp):
         return apphook.__name__
-    elif isinstance(apphook, string_types):
+    elif isinstance(apphook, str):
```
Aiky30, line 60: I'm not sure if this is ok for Strings, I know you can't do this with list because Strings can be True when checking against a type of List.

cms/admin/pageadmin.py
```diff
@@ -9,7 +9,7 @@
 import django
 from django.contrib.admin.helpers import AdminForm
 from django.conf import settings
-from django.conf.urls import url
+from django.urls import re_path
```
Aiky30, line 12: I guess to ensure that the codebase worked as it did re_path is the same as url was, it is recommended to switch out to path but I think migrations like this should be staged and maybe pencil something in for a future change to use path vs re_path.

cms/tests/test_staticfiles.py
```diff
@@ -1,17 +0,0 @@
-from django.core import management
-from django.test import TestCase
-
-from cms.test_utils.util.context_managers import StdoutOverride
-from cms.test_utils.util.context_managers import TemporaryDirectory
-
-
-class StaticFilesTest(TestCase):
-
-    def test_collectstatic_with_cached_static_files_storage(self):
```
Aiky30, line 10: Why was this test removed?

setup.py
```diff
@@ -1,57 +1,58 @@
...
+#!/usr/bin/env python
 import os
-import cms
+
+from cms import __version__
+from setuptools import find_packages, setup
+
+
+REQUIREMENTS = [
+    'Django>=2.2,<4',
```
Aiky30, line 9: Should this be less than 4 or less than and equal to 3.2 as Django have committed to 3.0, 3.1 and 3.2, 4.0, 4.1, 4.2.

cms/admin/placeholderadmin.py
```diff
@@ -24,8 +26,6 @@
 from django.views.decorators.clickjacking import xframe_options_sameorigin
 from django.views.decorators.http import require_POST
 
-from six.moves.urllib.parse import parse_qsl, urlparse
-
 from six import get_unbound_function, get_method_function
```
Aiky30, line 29: From what I can see in the diff six looks to be installed just for these two imports, I assume logic that was for 2 that needs to still work in 3.

setup.py
FinalAngel: @yakky true but on the other hand restricting doesn't gain much either. That way people can try out django/master earlier without locally modifying the files.

yakky:
Sure, I'm entirely neutral on this
I'm fine with either solution
24 Aug 2020 09:38:39 Angelo Dini <notifications@github.com>:

cms/management/commands/subcommands/base.py
```diff
@@ -52,7 +52,7 @@ class SubcommandsCommand(BaseCommand):
     subcommand_dest = 'subcmd'
 
     def create_parser(self, prog_name, subcommand):
-        kwargs = {'cmd': self}
```
FinalAngel: @yakky I'm not sure if this is the right approach here, this is related to your comment on https://github.com/divio/django-cms/pull/6886#issuecomment-678237407

cms/tests/test_urlutils.py
```diff
@@ -19,11 +19,11 @@ def test_is_media_url(self):
...
             self.assertFalse(urlutils.is_media_request(request))
-        with self.settings(MEDIA_URL='http://testserver2.com/'):
+        with self.settings(MEDIA_URL='http://testserver.com/'):
             request = self.get_request('/')
             self.assertFalse(urlutils.is_media_request(request))
-        with self.settings(MEDIA_URL='http://testserver/media/'):
-            request = self.get_request('/media/')
-            self.assertTrue(urlutils.is_media_request(request))
+        with self.settings(MEDIA_URL='http://testserver.com/media/'):
+            # request = self.get_request('/media/')
```
FinalAngel, line 26: @yakky might also need help here to either rewrite the implementation or change the tests

cms/management/commands/subcommands/base.py
yakky: removing it looks fine and has no ill effect.

As it's been removed in 2.1, we can safely drop it. Other packages (django-constance, for example) did the same

cms/tests/test_urlutils.py
yakky: I pushed a fix in 0d2ad18c5

FinalAngel: awesome

FinalAngel:
@yakky wehee a passing build :D

FinalAngel:
@yakky or @jrief please approve if it can land :)

yakky (approved):

PR #6896, opened 2020-09-02
fix template label nested translation (#6687)

### LINKED ISSUE #6687: Advanced page settings not working
Advanced page settings raises `TypeError: __str__ returned non-string (type __proxy__)` exception.

Constant "Inherit the template of the nearest ancestor" (cms.utils.conf) is wrapped in ugettext_lazy 2 times. First in cms.utils.conf, then in cms.models.pagemodel (`template_choices` attribute). This raises exception when casting to str:

```python
from cms.models import Page
str(Page.template_choices[-1][1])
```

### Environment

* Python version: 3.6.5
* Django version: 2.1.8
* django CMS version: 3.6.0

jbazik:
## Description

Fixes a serious bug in (at least) 3.7.2 to dev that causes an exception in the admin when editing any page, when USE_I18N is set to False.  Fixes #6687. 

In utils/conf.get_templates, the inherit template label is stored as a gettext_lazy object:
```
        templates.append((constants.TEMPLATE_INHERITANCE_MAGIC, _('Inherit the template of the nearest ancestor')))
```
In models/pagemodel.py, all template label choices are stored as gettext_lazy objects:
```
    template_choices = [(x, y) for x, _(y) in get_cms_setting('TEMPLATES')]
```
This creates a nested lazy object, which results in an exception in django.template.base.render_value_in_context when it is coerced to a string:
```
            value = str(value)
```
This only happens when USE_I18N is False.  In that case, the function stored in the lazy object is django.utils.translation.gettext_noop, which just returns it's argument.  The str() constructor doesn't know what to do with the stored proxy object.

The docs suggest that CMS_TEMPLATES should contain translated labels, so the correct fix is to not apply translation when building the choices list in the Page model.

A test is included.  Note that it is not enough to `override_settings(USE_I18N=False)` when testing, since the translation mechanism in django is set up during initialization and is not affected by the override.  My test includes a mock to test against gettext_noop.

## Related resources

* Fixes #6687 

## Checklist

* [ x] I have opened this pull request against ``develop``
* [ x] I have updated the **CHANGELOG.rst**
* [x ] I have added or modified the tests when changing logic


FinalAngel:
any feedback on this @Aiky30 or @victor-yunenko

Aiky30:
I think that these changes fix when I18n is disabled but then break the scenario for the majority when it is enabled because the UI item will no longer be translated for users using a language other than English.

Aiky30 (changes requested):
I think that these changes fix when I18n is disabled but then break the scenario for the majority when it is enabled because the UI item will no longer be translated for users using a language other than English.

I think that building the template list should be different dependant on the I18N Setting. This then protects both use cases.

jbazik:
When should the labels be translated?  The docs indicate, by example (http://docs.django-cms.org/en/latest/reference/configuration.html#cms-templates), that they should be translated when configured.  In that case, my fix is correct for both cases.  If instead sites in all languages should configure the labels untranslated, then you are right.

There is an alternative fix: storing the TEMPLATE_INHERITANCE_MAGIC label untranslated.  That works only because nested gettext doesn't break with I18N turned on.  But it's the least disruptive.  If you'd prefer that, would you like me to change the docs, too, or leave them as-is?

John

Aiky30:
> There is an alternative fix: storing the TEMPLATE_INHERITANCE_MAGIC label untranslated. That works only because nested gettext doesn't break with I18N turned on. But it's the least disruptive. If you'd prefer that, would you like me to change the docs, too, or leave them as-is?

It was the template inheritance text that I thought would have then been broken but you're right that it's stored with translations active which we should keep: https://github.com/divio/django-cms/blob/c4b48fab0f84c237b436442039137ac1921e2093/cms/utils/conf.py#L168

I was thinking of a simple if else condition but you're right that it is covered already so no action needed, the test looks sane. I'm happy.

My apologies for the confusion.

Andrew

Aiky30:
@jbazik Can you bring the branch back up to date so that we can see the build pass. Once passed I will re-review and accept.

jbazik:
I added a changelog entry and rebased.

Aiky30 (approved):
I'm happy with this change.

Aiky30:
@FinalAngel I'm happy with the change.

9iaashalaby:
Pushed Merging

FinalAngel:
👍  will merge once the tests pass

marksweb:
@jbazik Looks like this might benefit from another merge of develop to pick up the differences in the coverage and coveralls requirements.

jbazik:
@marksweb done - thanks

marksweb:
@FinalAngel This is now passing after being updated

Aiky30 (changes requested):
Due to the nature of this change i.e. anyone who didn't follow the docs on translations can we please add a changelog entry, this will most likely make it into django-cms 3.9.x and could be considered a breaking change. My apologies, we have recently agreed that no changes should be coming in without a changelog entry, especially if there are potentially breaking changes. I'm sure you would appreciate this if the change affected your project.

I will personally take responsibility for getting the change merged ASAP with a changelog entry in place. 

Thank you.

jbazik:
@Aiky30 added changelog entry, thanks.

Aiky30 (approved):
Happy with the changes, just needs the tests to pass now.

jbazik:
Rebased against develop again.

marksweb (approved):

marksweb:
@victor-yunenko This PR requires a review from you to get wrapped up.

victor-yunenko (approved):

fsbraun:
@sparrowme Interestingly, it has. :-(

fsbraun:
@sparrowme Where exactly did you encounter the issue?

PR #6898, opened 2020-09-08
made PageUser non cascading

robinvd:
## Description

A PageUser gets created for every user that is created, with a created_by to the original user.
On deleting the original user all PageUsers would be deleted by cascade.
And thus also the user connected to there PageUsers as well.

This sets created_by to on_delete=models.SET_NULL stopping this problem

## Related resources

* #6776

## Checklist

I can make a test if that is required
* [] I have added or modified the tests when changing logic


robinvd:
@FinalAngel i saw there was an unrelated commit in the last PR, this is just the commit that changes the delete beheviour

FinalAngel:
thank you @robinvd, would you also be able to rebase develop and add a changelog entry? /cc @Aiky30 @victor-yunenko

robinvd:
@FinalAngel is this changelog entry discriptive enough?

FinalAngel:
thank you @robinvd, that seems fine. Let's wait for some feedback from @Aiky30

Aiky30:
Created by is a required field and the only breaking change to the api that v4 introduces. I don't think this should be set to null and there may be other logic in the current codebase that requires created by to be populated.

Aiky30 (changes requested):
I don't think that we should allow users to be deleted if they are attached to anything. This should be models.PROTECT!

We can add guidelines in the documentation that a user can have their permissions revoked but if a user has ever created anything we can't delete them due to the integrity of a projects history (especially versions).

marksweb:
Hi @robinvd, did you see the above feedback on this change?

I'd agree that setting to `models.PROTECT` instead of `SET_NULL` makes sense.

marksweb:
Closing this as I've raised the recommendation in #7206

PR #6819, opened 2020-03-13
Fix not rendering AliasPlugin in same static_placeholder with differrents languages  in edit mode

haricot:
## Description
I wanted to do a alias a navbar plugin in a same static_placeholder in translation context, but it did not appear in edit mode.

## Checklist

* [ ] I have opened this pull request against ``develop``
* [x] I have updated the **CHANGELOG.rst**
* [ ] I have added or modified the tests when changing logic

PR #6901, opened 2020-09-11
Add test_alias_plugins_no_recursion_across_page_translation and fix not rendering alias plugins in edit-mode

haricot:
## Description
Across page language translation:
In edit_mode, Alias_​plugins  and  Alias_placeholders are not rendered.
In published mode, Alias_​plugins are rendered, but with Alias_placeholders they are not.

## Related resources
 
* #6819

## Checklist

### DIFF
--- a/cms/tests/test_alias.py
+++ b/cms/tests/test_alias.py
@@ -109,10 +109,11 @@ class AliasTestCase(TransactionCMSTestCase):
         source_plugin = api.add_plugin(ph_2, 'StylePlugin', 'en', tag_type='div', class_name='info_2')
 
         api.create_title("fr", "page_1", page)
-        page.publish('fr')
 
         alias_placeholder = api.add_plugin(ph_1, 'AliasPlugin', 'fr', alias_placeholder=ph_1)
         alias_plugin = api.add_plugin(ph_2, 'AliasPlugin', 'fr', plugin=source_plugin)
+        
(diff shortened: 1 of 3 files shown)


cms/tests/test_alias.py
```diff
@@ -92,6 +92,44 @@ def test_alias_recursion_across_pages(self):
...
+        
+        page.publish('fr')
+
+        self.assertTrue(alias_placeholder.is_recursive())
+        self.assertTrue(alias_plugin.is_recursive())
+
+        with self.login_user_context(self.get_superuser()):
+            response = self.client.get(page.get_absolute_url(language='fr') + '?preview&edit_off')
+            self.assertEqual(response.status_code, 200)
+            self.assertContains(response, '<div class="info_2">', html=True)
```
haricot, line 124: At this point, in published rendering mode, the test with an alias plugin passes and after with an alias placeholder does not.

haricot:
Closed in favor of the djangocms-alias approach.
I realize that it is very interesting.
For example among other things, it would allow if i understood correctly to add a translation with only the fields that need to be different for the content of the associated site using `AliasContent`.
The advantage will be to be able to automatically synchronize the general modifications without taking care of the translations or without having to re-import everything. Via for example a menu item in the placeholder: 'aliased with the default language', this will generate and store the data of translations or content in the alias.

PR #6905, opened 2020-09-16
fix plural issue in django-cms

FinalAngel:
## Description

Fixes an issue in ``wizards/create.html`` where the error message did not use the plural form. This occurs when generating the translation files via `` django-admin.py makemessages -l en --no-location``.

Also adding missing changelog entries in preparation for the django CMS 3.8.0 RC1 release.

## Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,10 @@ Changelog
 * Introduced support for Django 3.1
 * Dropped support for Python 2.7 and Python 3.4
 * Dropped support for Django < 2.2
+* Removed ``djangocms-column`` from the manual installation instructions
+* Removed duplicate ``attr`` declaration from the documentation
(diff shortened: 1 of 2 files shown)


macolo (approved):

CHANGELOG.rst
```diff
@@ -9,6 +9,10 @@ Changelog
 * Introduced support for Django 3.1
 * Dropped support for Python 2.7 and Python 3.4
 * Dropped support for Django < 2.2
+* Removed ``djangocms-column`` from the manual installation instructions
```
Aiky30, line 12: @FinalAngel Should there be so many changelog entries as part of this change?

FinalAngel: @Aiky30 that's fine to add it here. We should ideally enforce new PRs to add changelog entries to avoid this issue

Aiky30 (approved):

PR #6906, opened 2020-09-16
Release django CMS RC1

FinalAngel:
* Introduced support for Django 3.1
* Removed ``djangocms-column`` from the manual installation instructions
* Removed duplicate ``attr`` declaration from the documentation
* Fixed a reference to a wrong variable in log messages in ``utils/conf.py``
* Fixed an issue in ``wizards/create.html`` where the error message did not use the plural form


Aiky30 (approved):
Looks OK to me.

PR #6912, opened 2020-09-21
Cache not invalidated when using a PlaceholderField outside the CMS

pierreben:
## Description

Since v3.6.0, when using a PlaceholderField outside the CMS, the cache is not invalidated when updating the placeholder content.
On [this commit](https://github.com/divio/django-cms/commit/99880979884de94b1f2784bb830ad5cf6d7bdea7#diff-5c9765ac0e61c2e87b3cea96392a6df2L1117), the value of the clear_cache attribute changed from True to False, and the [clear_cache method is never called in this case](https://github.com/divio/django-cms/blob/release/3.7.x/cms/models/placeholdermodel.py#L499).

## Related resources

* https://github.com/divio/django-cms/commit/99880979884de94b1f2784bb830ad5cf6d7bdea7#diff-5c9765ac0e61c2e87b3cea96392a6df2L1117
* https://github.com/divio/django-cms/blob/release/3.7.x/cms/models/placeholdermodel.py#L499
* https://github.com/nephila/djangocms-blog/issues/484

## Steps to reproduce

1. Create a custom model containing a PlaceholderField

```
class MyModel(models.Model):
    body_placeholder = PlaceholderField("Body")
```

2. Activate CACHE on the CMS

```
CMS_CACHE_DURATIONS = {
    "menus": 3600,
    "content": 3600,
    "permissions": 3600
}
```

3. Create a MyModel instance, a view and template to display it 

4. Edit the my_model_object placeholder field content within a template using the render_placeholder template tag:

`{% render_placeholder my_model_object.body_placeholder %}`

5. Open the MyModel instance view in a private session to compare the content.

6. Repeat steps 4 and 5 twice if the content doesn't differs.

## Expected behaviour

The contents are always the same between the two browser sessions.

## Actual behaviour

The contents differs as the cache is not invalidated when updating the body_placeholder content.

## Suggested correction

In the [mark_as_dirty method](https://github.com/divio/django-cms/blob/release/3.7.x/cms/models/placeholdermodel.py#L489), force clear_cache when the attached model is not a Page or a StaticPlaceholder. 

## Checklist


FinalAngel:
@pierreben please make sure to have a look at the checklist :)

yakky:
@pierreben I think the branch should be recreated including commit https://github.com/django-cms/django-cms/pull/6912/commits/c9a80df7bba861a7c7ee744f26b717b95bc1d4a3 only

pierreben:
I've didn't submit the correct branch.

I've created a new PR (https://github.com/django-cms/django-cms/pull/6956) with tests and only one change of 2 lines against develop.

PR #6918, opened 2020-10-08
Deprecate core Alias plugin

Aiky30:
## Description

The core Alias functionality is deprecated in favour of the package djangocms-alias. 

The changes in this PR:
- Remove the Alias from the UI (Page edit create alias links etc)
- Remove tests that would have now failed with no presence of the plugin
- Add deprecation notices where possible.

## Related resources

https://github.com/divio/djangocms-alias


Aiky30:
@FinalAngel I've added deprecation notices to this change, I'm not sure what the normal policy is for deprecation and version numbers, I've assumed the next major 4.1 would remove the code. Please advise.

FinalAngel:
@Aiky30 that depends a bit on the strategy. If we gonna release 3.8.0, the deprecation should probably be in 3.9 and be removed in 3.10. Depending on what happens with 4.x

Aiky30:
@FinalAngel What do you think is best for this, remove the deprecation notices and reengineer them once we have a plan?

FinalAngel (approved):

FinalAngel:
@Aiky30 I don't think removing them harms us for now. We want to get rid of Alias anyhow. We just need to come up with an alternative and migration path.

macolo:
All of the above makes 100% sense to me.

PR #6930, opened 2020-10-29
Patch defects

adam-murray:
## Description

* A number of unsanitised parameters being passed through code were highlighted by checkmarx when we scanned CMS4, this PR covers either sanitising those inputs or providing a test to prove they are non-issues


cms/tests/test_views.py
```diff
@@ -269,6 +272,34 @@ def test_incorrect_slug_for_language(self):
...
+        """
+            When sending a request the CMS uses get_page_from_request to return the appropriate page.
+            None should be returned
+        """
+        request = self.get_request("/")
+        request.path_info = "<script>alert('attack!')</script>"
+
+        response = get_page_from_request(request)
+
+
```
Aiky30, line 285:
```suggestion
```

cms/tests/test_views.py
```diff
@@ -269,6 +272,33 @@ def test_incorrect_slug_for_language(self):
...
+            None should be returned
+        """
+        request = self.get_request("/")
+        request.path_info = "<script>alert('attack!')</script>"
+
+        response = get_page_from_request(request)
+
+        # If this method is passed a parameter which is not a valid primary key
+        # for a page object nothing should be returned.
+
```
Aiky30, line 287:
```suggestion
```

Aiky30 (approved):

cms/tests/test_views.py
```diff
@@ -269,6 +272,32 @@ def test_incorrect_slug_for_language(self):
...
+        # If this method is passed a parameter which is not a valid primary key
+        # for a page object nothing should be returned.
+        self.assertEqual(response, None)
+
+    def test_malicious_content_login_request(self):
+        user = self.get_superuser()
+        request = self.get_request(
+            "/en/admin/login/?q=<script>alert('Attack')</script>",
+            post_data={"username": user.username, "password": user.username}
```
Aiky30, line 293:
```suggestion
            post_data={"username": username, "password": username}
```

@adam-murray This breaks the test suite, I'm not sure why the username field changes for just the Python 3.6 versions. To fix the test suite this needs to be:

```
username = getattr(user, get_user_model().USERNAME_FIELD)
```

Taken from: https://github.com/django-cms/django-cms/blob/f30f0204a8c3e498aa86de0aba35470521859873/cms/tests/test_no_i18n.py#L121

**With the suggestions above you'll still need to add the get_user_model() import at the top of the file:**
```
```

cms/tests/test_views.py
```diff
@@ -269,6 +272,32 @@ def test_incorrect_slug_for_language(self):
...
+        request.path_info = "<script>alert('attack!')</script>"
+
+        response = get_page_from_request(request)
+
+        # If this method is passed a parameter which is not a valid primary key
+        # for a page object nothing should be returned.
+        self.assertEqual(response, None)
+
+    def test_malicious_content_login_request(self):
+        user = self.get_superuser()
```
Aiky30, line 290:
```suggestion
        username = getattr(self.get_superuser(), get_user_model().USERNAME_FIELD)
```

Aiky30 (changes requested):
The two suggestions can be committed from the Github web front but they will still require the import to be manually added due to suggestion limitations.

Aiky30 (approved):
All good. Thanks.

yakky (approved):

FinalAngel (approved):

PR #6945, opened 2020-11-16
perf: Don't count users when CMS_RAW_ID_USERS=True

Pankrat:
## Description

When using CMS_RAW_ID_USERS=True on a Postgres database with many users, counting the users is slow and will always yield the same result.

Only count users when using an integer value as a threshold and reuse the same logic for both `PagePermissionInlineAdmin` and
`GlobalPagePermissionAdmin`.

### Background

When looking into performance issues for one of our websites, we found that Django CMS would regularly count all users in our Postgres database. Unfortunately, `SELECT COUNT(*) FROM huge_table` is slow in Postgres and easily takes 500ms or more if you have more than a million records in your table (exact numbers depend on the DB host of course).

I first looked into [improving the speed of the count query](https://www.cybertec-postgresql.com/en/postgresql-count-made-fast/), but this is not easy to achieve in a way that is compatible with different DB backends.

The easiest option was to not run this query at all when using the documented and widely used (AFAICT) `CMS_RAW_ID_USERS = True`.

The change should be backward compatible with the following exceptions:

* `CMS_RAW_ID_USERS = True` will use raw IDs, even if you have just one or zero users in the DB (since `2 > True` but not `1 >  True` in Python). Since the current behavior could be considered a (rather academic) bug, I don't think this is a deal-breaker.
* `CMS_RAW_ID_USERS = 0` will trigger a count and use raw IDs when there's one user or more. If that's a problem, I could check if `threshold` is "truthy" like in the original implementation.

## Checklist

### DIFF
--- a/cms/admin/permissionadmin.py
+++ b/cms/admin/permissionadmin.py
@@ -31,6 +31,29 @@ class TabularInline(admin.TabularInline):
     pass
 
 
+def users_exceed_threshold():
+    """
+    Check if the number of users exceed the configured threshold. Only bother
+    counting the users when using an integer threshold, otherwise return the
+    value of the setting to avoid a potentially expensive DB query.
+    """
+    threshold = get_cms_setting('RAW_ID_USERS')
+
+    # Don't bother counting the users when not using an integer threshold
+    if threshold is True or threshold is False:
+        return threshold
+
+    # Given a fresh django-cms install and a django settings with the
+    # CMS_RAW_ID_USERS = CMS_PERMISSION = True
+    # django throws an OperationalError when running
+    # ./manage migrate
+    # because auth_user doesn't exists yet
+    try:
+        return get_user_model().objects.count() > threshold
+    except OperationalError:
+        return False
+
+
 class PagePermissionInlineAdmin(TabularInline):
     model = PagePermission
     # use special form, so we can override of user and group field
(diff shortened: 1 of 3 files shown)


marksweb (approved):
Nicely done 👍

vinitkumar (changes requested):
Looks good. @Pankrat can you please rebase/merge with teh latest develop and also fix the flake8 errors if any?

Pankrat:
Hi @vinitkumar, it seems my earlier rebase ran into the [colors/faker issue on the CI system](https://github.com/django-cms/django-cms/runs/4766229157?check_suite_focus=true). I've just rebased again - hoping this now runs through.

cms/admin/permissionadmin.py
```diff
@@ -31,6 +31,29 @@ class TabularInline(admin.TabularInline):
...
+def users_exceed_threshold():
+    """
+    Check if the number of users exceed the configured threshold. Only bother
+    counting the users when using an integer threshold, otherwise return the
+    value of the setting to avoid a potentially expensive DB query.
+    """
+    threshold = get_cms_setting('RAW_ID_USERS')
+
+    # Don't bother counting the users when not using an integer threshold
+    if threshold is True or threshold is False:
```
marksweb, line 43: I feel like there should be some more explicit type checking done here if the type we're interested in is an `int`

marksweb (changes requested):

fsbraun:
Merged through #7414

PR #6602, opened 2019-01-25
prevents exception of page_attribute … as varname for invalid page_lo…

jrief:
### Summary

Fixes #5930

### Links to related discussion
https://github.com/divio/django-cms/issues/5930#issuecomment-457499815

### Proposed changes in this pull request
It adds the same behaviour to templatetag `{% page_attribute 'field_name' 'page_lookup' as varname %}` as is already implemented in templatetag `{% page_url 'page_lookup' as varname %}`

### Documentation checklist

* [x] I have updated CHANGELOG.txt if appropriate
* [ ] I have updated the release notes document if appropriate, with:
    * [ ] general notes
    * [ ] bug-fixes
    * [ ] improvements/new features
    * [ ] backwards-incompatible changes
    * [ ] required upgrade steps
    * [ ] names of contributors
* [ ] I have updated other documentation
* [x] I have added my name to the AUTHORS file
* [ ] This PR's documentation has been approved by Daniele Procida

Issue #5930, opened 2017-04-13
page_attribute "field-name" page_lookup as varname fails silently


jrief:
### Summary

when using the templatetag 

```
{% page_attribute "field-name" page_lookup as varname %}
```

using a ``page_lookup`` of a CMS page which does not exist, this templatetag fails silently, without rendering any part of the template. This makes is very hard to track down the reason.

### Expected behaviour

just as
```
{% page_url page_lookup as varname %}
```
can be used on a CMS page which does not exist, hence setting ``varname`` to ``None``, one would expect the same behaviour for the templatetag ``page_attribute``.

#### Environment

* Python version: any
* Django version: 1.10
* django CMS version: 3.4.2, HEAD of releases/3.4.x

czpython:
Hello @jrief,
I'm unable to reproduce this.
Can you provide me with a test project where I can do so?

czpython:
Closing as we're unable to reproduce this.

jrief:
This error occurred again, this time in version 3.5.3.
How to reproduce:

* Create two CMS pages A and B.
* On page A, in Advanced settings, set Page-ID to `id-a`.
* To the template used by page B add:
  ```
  {% page_url 'id-a' as a_url %}
  {% page_attribute 'menu_title' 'id-a' as a_title %}
  ```
* Publish both pages.
* Remove Page-ID from page A.
* Load Page B. If that doesn't raise a page-not-found exception, unpublish page B, then republish.

btw. there is an asymmetry in the code, while the class `PageUrl` implementing templatetag `page_url`, catches a page-not-found exception, the class `PageAttribute` does not catch this exception.

PR #6953, opened 2020-11-26
Fix 5930 page attribute as varname

jrief:
## Description

It adds the same behaviour to templatetag `{% page_attribute 'field_name' 'page_lookup' as varname %}` as is already implemented in templatetag `{% page_url 'page_lookup' as varname %}`. If `settings.DEBUG = True` and the underlying page is missing, the templatetag fails silently instead of raising a template exception.

This pull request superseeds #6602 and has been rebased against version 3.8.0.

## Related resources

* #5930
* #6602

## Checklist

* [x] I have opened this pull request against ``releases/3.8``

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,6 +3,12 @@ Changelog
 =========
 
 
+next
+====
+
+* Fixes templatetag ``page_attribute`` when extanding with ``as ...``.
+
+
 3.8.0 (2020-10-28)
 ==================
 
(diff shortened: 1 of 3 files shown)


yakky (approved):

goutnet:
Can we rebase this against develop maybe?

marksweb:
@jrief looks like this got lost in time. Is it still relevant? If so perhaps you could retarget at develop

PR #6958, opened 2020-12-08
Fix not checking slug uniqueness on page move

### LINKED ISSUE #6840: Slug uniqueness not checked when moving page
## Description

When page is moved, its slug uniqueness is not checked. Thus two pages with the same slug can be present on the same level.

## Steps to reproduce

1. Create a page at the root level and give it a slug "test". 
2. Publish the page
3. Using the Pages manager, copy the page and paste it as a child of the test page
4. Move the page back to the root - we now have two pages with the same slug (and both can be published)

## Expected behaviour

Slug should be checked and the page move should either not be allowed or slug should be automatically modified.

## Actual behaviour

Two pages with the same slug are allowed on the same level. Both can be published at the same time. The preview link in the Pages admin only points to the original (not to the copy).

## Additional information

Situation has been reproduced on the Divio platform using a clean project - https://cms-bug-demo-stage.us.aldryn.io/en/admin/cms/page/?language=en

yakky:
## Description

slug uniqueness must be checked during page move

Fix tap into `Page.update_title_path` to use existing `get_available_slug` function (used on page copy) to check for conflicting slug.

`update_title_path` is modified to accept the page to be moved as argument to exclude the page from the check. In this case, but published, unpublished and draft pages are excluded

## Related resources

Fix #6840

## Checklist


macolo:
Im rerunning the checks (new github actions)

marksweb:
Thanks @macolo

I was going to do that later to see if this is still good to review 👍

cms/utils/page.py
```diff
@@ -191,13 +191,15 @@ def get_all_pages_from_path(site, path, language):
     return pages.filter(title_set__language=language)
 
 
-def get_available_slug(site, path, language, suffix='copy', modified=False):
+def get_available_slug(site, path, language, suffix='copy', modified=False, current=None):
```
marksweb, line 194: I think my only comment on this change would be that `current` is ambiguous. I know it's part of the page utils, but `current_page` would perhaps make things more obvious.

victor-yunenko (approved):

PR #6961, opened 2020-12-14
✨(pagetree) search results are foldable

jbpenrath:
## Description
Now user can navigate into child nodes from search results page.

## Related resources

* https://www.django-cms.org/en/roadmap/ (Search results improvements in admin interface)
* https://django-cms-38x-drillable-search-stage.us.aldryn.io/ (admin:admin)

## Checklist
* [x] I have opened this pull request against develop
* [x] I have updated the CHANGELOG.rst

**Bounty program**
@NicolaiRidani


jbpenrath:
⬆ Update: I added a [test link](https://django-cms-38x-drillable-search-stage.us.aldryn.io/ (admin:admin)).

cms/utils/conf.py
```diff
@@ -82,6 +82,7 @@ def wrapper():
     'PAGE_WIZARD_CONTENT_PLUGIN': 'TextPlugin',
     'PAGE_WIZARD_CONTENT_PLUGIN_BODY': 'body',
     'PAGE_WIZARD_CONTENT_PLACEHOLDER': None,  # Use first placeholder it finds.
+    'PAGETREE_DESCENDANTS_LIMIT': float('inf'),
```
Aiky30, line 85: What is the reason for a float vs an integer?

cms/tests/test_admin.py
```diff
@@ -295,56 +295,6 @@ def test_change_template(self):
             )
             self.assertEqual(response.status_code, 200)
 
-    def test_changelist_items(self):
```
Aiky30, line 298: What is the justification for removing the two tests?

cms/tests/frontend/integration/pagetree.js
```diff
@@ -13,55 +13,55 @@ var xPath = casperjs.selectXPath;
 var createJSTreeXPathFromTree = cms.createJSTreeXPathFromTree;
 var getPasteHelpersXPath = cms.getPasteHelpersXPath;
 
-var closeWizard = function() {
-    return function() {
+var closeWizard = function () {
```
Aiky30, line 16: Formatting in this file should not have occurred.

cms/admin/pageadmin.py
```diff
@@ -1420,17 +1384,50 @@ def get_tree(self, request):
...
         open_nodes = list(map(int, request.GET.getlist('openNodes[]')))
+        node_page = None
+        root_pages = None
+
+        # Retrieve search parameters if exist
+        query = request.GET.get('q', None)
+        changelist_form = self.changelist_form(request.GET)
+
+        try:
+            changelist_form.full_clean()
```
Aiky30, line 1395: I'm not convinced that this logic should have been removed from the ChangeList view, it is directly associated with the Changelist view where as get_tree can be used in isolation i.e. it's not dependant on the Changelist view. With this change get_tree is now Changeview dependant when it shouldn't be.

cms/admin/pageadmin.py
```diff
@@ -816,10 +782,8 @@ def changelist_view(self, request, extra_context=None):
             'tree': {
                 'site': site,
                 'sites': self.get_sites_for_user(request.user),
-                'query': query,
+                'query': request.GET.get('q', ''),
                 'is_filtered': changelist_form.is_filtered(),
-                'items': pages,
-                'has_items': has_items,
```
Aiky30, line 822: Someone has gone to extra effort to keep the tree and the changelist separate. Paulo wouldn't have ran extra queries if he didn't need to.

Aiky30 (changes requested):
I don't think that we should be removing the reusability of the get_tree method by tightly coupling the endpoint to the change list.

cms/utils/conf.py
jbpenrath: It is just a way to set this limit to `Infinity` by default. I cannot use `int('int')` to achieve that.

cms/tests/frontend/integration/pagetree.js
jbpenrath: Automatic linting mistake. I undid this.

cms/utils/conf.py
jbpenrath: Did you prefer to set this settings to `None` by default ?

cms/tests/test_admin.py
jbpenrath: About `test_changelist_items`, we did not pass anymore `items` into context, so this test is no more relevant.
Then concerning `test_changelist_get_results`, this is the same, we did get node through a query into changelist.

cms/admin/pageadmin.py
jbpenrath: As `changelist_form` is a class attribute, I considered its existence intrinsic to `BasePageAdmin` which contains also `get_tree` method so from my point of view, we could always use `ChangeListForm` into `get_tree`.

I used it as a helper to get pages from query params pass to `get_tree` methods but I could also retrieve pages by executing manually queries without using a `ChangeListForm`. WDYT?

jbpenrath:
Hi @Aiky30, did you see my comments?

vinitkumar:
@jbpenrath Can you please fix the conflicts with this branch?

goutnet:
Also, let's try to not clean this PR so it does'nt include translations in there :) (those are automatically pulled in by release anyway)

goutnet (changes requested):
Please do not add modifications on the translation files.

jbpenrath:
@vinitkumar, @goutnet my branch is now up to date and I removed translations.

vinitkumar:
@Aiky30 You made some comments on this PR, can you please look and see if this is good enough or do you want some more changes?

goutnet:
@jbpenrath  please remove the modification to the CHANGELOG.rst those are now automated.

@Aiky30  what is your opinion on last comment, could you suggest a path toward resolution maybe?

Aiky30:
@vinitkumar @goutnet the changes to make get_tree tightly coupled to the ChangeListView cannot be brought into cms 4.0.x, if you wish to bring the feature into v3 that is totally fine by me. The reason it cannot be brought forward is because the chengelist can currently accept a new get_tree method which is currently done for djangocms-page-admin where the navigation is no longer page driven and is a separate application (exactly why menu is a seperate app in django-cms, it was a future direction of the built in menu too). This is the reason that I believe Paulo put in extra effort to keep the two things separate.

PR #6962, opened 2020-12-14
Use the Site ID to key the cache for page permissions

mattgoins:
## Description

This addresses a problem with how page permissions are cached. To reproduce:

* Create a django cms site with Redis cache
* Add a second Site to the cms
* Create an auth group, e.g. "Editors", and a non-super user with staff status, who is in the Editors group
* Create a Page on each site, and in the individual Page permissions (not the Global Permissions), grant some typical permissions to the Editors group, e.g. Add, Edit, Publish
* Log in as the non-super user from above
* Observe that on the page admin, on the main site, you can see the "add sub page" plus symbol next to the page
* Switch to the other site using the Site toggle button, and see that the plus sign for "add sub page" should be
      missing now, even though the individual permissions
     on this page match the other page.

Solution:
* Pass the Site to the function that checks permissions, and
* Key the cached permission in a way that includes the primary key of the site


cms/cache/permissions.py
```diff
@@ -55,8 +56,9 @@ def clear_user_permission_cache(user):
     Cleans permission cache for given user.
     """
     from django.core.cache import cache
-    for key in PERMISSION_KEYS:
-        cache.delete(get_cache_key(user, key), version=get_cache_permission_version())
+    for site in Site.objects.all():
```
marksweb, line 59: This could be a bit more efficient. Rather than grabbing each `Site` object, you could just grab their IDs because that's all that's actually getting used.

It'd become something like;

```python
def clear_user_permission_cache(user):
    """
    Cleans permission cache for given user.
    """
    from django.core.cache import cache
    for site_id in Site.objects.values_list('id', flat=True):
        for key in PERMISSION_KEYS:
            cache.delete(
                get_cache_key(user, "%s:%s" % (site_id, key)),
                version=get_cache_permission_version()
            )
```

PR #6965, opened 2021-01-19
Fix #6622 bad Title.path in ML sites if parent slug is created or mod…

### LINKED ISSUE #6622: The slugs of child pages are built incorrectly and return a 404 error when a user adds a custom slug for another language of the parent page

fp4code:
…ified

## Description

Fix #6622 Bad Title.path in Multilanguage sites if parent slug is created or modified

### First bug:

In 'en','fr' site
1) Create /en/en-1/ page
2) Create from this page a subpage with slug "en-1-1"
3) Create "fr" translation with slug "fr-1-1"
4) Create "fr" translation of parent /en/en-1/ with slug "fr-1"

As a result, the path of subpage "fr-1-1" is not updated. Here the paths:

/en/en-1/
/en/en-1/en-1-1/
/fr/fr-1/
/fr/en-1/fr-1-1/   <-- Bad Title.path "en-1/fr-1-1" instead of "fr-1/fr-1-1"

### Second bug:

Change "en" slug of /en/en-1/ page to "en-1b",
As a result the path of subpage "en-1-1" is changed to "fr-1/en-1-1" instead of "en-1b/en-1-1". Here the paths:

/en/en-1b/
/en/fr-1/en-1-1/   <-- Bad Title.path "fr-1/en-1-1" instead of "en-1b/fr-1-1"
/fr/fr-1/
/fr/en-1/fr-1-1/   <-- Bad Title.path "en-1/fr-1-1" instead of "fr-1/fr-1-1"

## Checklist

* [x] I have added or modified the tests when changing logic (no changed logic)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,11 +2,17 @@
 Changelog
 =========
 
+fp-6622 (2021-01-19)
+====================
+
+* Fixed 66622 bad Title.path in multilingual sites when parent slug is created or modified
+
 
 Unreleased
 ==================
 
 * Fixed builds on RTD
+
 
 3.8.0 (2020-10-28)
 ==================
--- a/cms/admin/forms.py
+++ b/cms/admin/forms.py
@@ -472,8 +472,9 @@ class ChangePageForm(BasePageForm):
 
         if update_count == 0:
             api.create_title(language=self._language, page=cms_page, **translation_data)
-        else:
-            cms_page._update_title_path_recursive(self._language)
+        # _update_title_path_recursive should be called if the new page is the parent
+        # of already created children in multilingual sites.
+        cms_page._update_title_path_recursive(self._language, slug=self.data['slug'])
         cms_page.clear_cache(menu=True)
         return cms_page
 
--- a/cms/management/commands/subcommands/titlepaths.py
+++ b/cms/management/commands/subcommands/titlepaths.py
@@ -1,0 +1,64 @@
+from cms.models import CMSPlugin, TreeNode, Title, Page
+
+from .base import SubcommandsCommand
+
+class CheckTitlePathsCommand(SubcommandsCommand):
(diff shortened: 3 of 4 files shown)


Aiky30 (changes requested):
This change will need a test that replicates and enforces the business requirements that have been identified. Without tests we cannot guarantee the fix in the future.

Once a test is implemented and approved we can look to get it released to unblock you ASAP.

fp4code:
Sorry for the mess, I don't know how to combine PR.
#6965 is the fix of #6622
#6967 add a test file with no fix to observe the issue #6622
#6968 the fix with test file included.

PR #6966, opened 2021-01-20
Allow migrations to be ran backwards

django-doctor:
currently django would raise exception if we tried to reverse the migrations because the data migration has no backwards. This avoids that problem [read more](https://django.doctor/advice/C8001)

I found this problem during auditing your codebase. [Read the full report](https://django.doctor/django-oscar/django-oscar)


Aiky30 (commented):
I'm not o sure on these changes. If someone tried to reverse the migrations and error will indeed be thrown, but they have also been prevented from "losing" data by reversing migrations that do not provide a way to move data backwards and forwards.

PR #6971, opened 2021-01-28
Changed description of django CMS

NicolaiRidani:
## Checklist

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -37,7 +37,7 @@ django CMS documentation
 Overview
 ********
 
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -37,7 +37,7 @@ django CMS documentation
 Overview
 ********
 
-django CMS is a modern web publishing platform built with `Django`_, the web application framework "for
+django CMS is a modern and beautiful web publishing platform built with `Django`_, the web application framework "for
```
evildmp, line 40: True, but we prefer to use "very beautiful" in our official style guide.

PR #6985, opened 2021-02-19
Fixes/flex page tree

jrief:
## Description

We are using django-CMS without djangocms-admin-styles. After upgrading to Django version 3.1 I experienced a broken layout when rendering the pages tree:

Screenshot 2021-02-19 at 12 58 31

From what I have seen, this is because Django itself switched to use `flex` in version 3.0.
With these little fixes, django-CMS renders as usual. That fix has been tested against Django-3.1 and Django-3.0.

## Related resources

after patching the pages tree page looks as we know it:

Screenshot 2021-02-19 at 12 58 53

## Checklist


marksweb (approved):

greengoaxe:
Hi @macolo,
may you merge it?
Thanks,
Francesco

vinitkumar (approved):

PR #6993, opened 2021-03-12
(WIP Don't merge) Implement Github Actions

vinitkumar:


Aiky30:
@vinitkumar I've been looking at how others have configured the workflows, I was initially not a fan of grouping jobs by the database but I now actually think it makes sense from a maintenance point of view as the DB change the least often.

I've updated the document with a Grouping Jobs section with the jobs:
- frontend
- docs
- database-postgres
- database-mysql
- database-sqlite

PR #7002, opened 2021-03-31
Fixes issue with template specific static placeholder configuration.

svenvandescheur:
## Description

When configuring `CMS_PLACEHOLDER_CONF`. A combination of the template name can be used to create a specific placeholder configuration for a specific page. This works as intended for regular placeholder and allow the CMS to, for instance create default placeholders when creating a new page.

This behaviour is also useful for static placeholders (for instance: a navigation bar which should be present on all pages but has slightly different (default) settings for a specific template). However, this does not seem to work.

## Related resources

* https://github.com/django-cms/django-cms/issues/7001

## Checklist


svenvandescheur:
I think I misunderstood some of the concepts of the static placeholder, I would think they'd be template specific, turns out they're not and fixing this issue will not allow for various templates to have various "static" placeholders. I'm closing the PR for this reason.

PR #7018, opened 2021-04-10
Add support for Django 3.2

vinitkumar:
## Description

Adds a new Django 3.2 requirements file and add it to the matrix.
It will most likely fail but will provide us a window to see what fails
and what needs to  be fixed to get it ready.

Authored-by: Vinit Kumar <vinit.kumar@socialschools.nl>
Signed-off-by: Vinit Kumar <vinit.kumar@socialschools.nl>

## Related resources

- Django 3.2 release notes: https://docs.djangoproject.com/en/3.2/releases/3.2/

## Checklist


marksweb (approved):
Looks good. Nice & simple

marksweb (approved):

cms/tests/test_po.py
```diff
@@ -33,10 +35,13 @@ def compile_messages():
         for dirpath, dirnames, filenames in os.walk(basedir):
             for f in filenames:
                 if f.endswith('.po'):
-                    fn = os.path.join(dirpath, f)
-                    if has_bom(fn):
-                        raise CommandError("The %s file has a BOM (Byte Order Mark). Django only supports .po files encoded in UTF-8 and without any BOM." % fn)
-                    pf = os.path.splitext(fn)[0]
+                    if DJANGO_2_2 or DJANGO_3_0 or DJANGO_3_1:
+                        pfn = os.path.join(dirpath, f)
+                    elif DJANGO_3_2:
```
pcolmant, line 40: With "elif DJANGO_3_2: instead of "else:", what happens when django 3.3, etc. is released? As I understand it, there is a matrix of (ptyhon /) django resleases intended for a specific version of Django CMS.  So I would have used "else:" and based on that matrix, adapt the code later for the next version of (python /) django.

vinitkumar: Now, when I think more deeply. I think you are right. Using an else, makes it easily future proof because if there is a change, we can just add a new condition.

vinitkumar: I will make these changes and commit later tonight.

vinitkumar: @pcolmat please check my latest commit.

marksweb:
@vinitkumar I've just been looking over old tickets & come across this;

https://github.com/django-cms/django-cms/blob/develop/cms/utils/compat/__init__.py

Perhaps nows a good time to put 3.2 into there now, before we forget.

vinitkumar:
> @vinitkumar I've just been looking over old tickets & come across this;
> 
> https://github.com/django-cms/django-cms/blob/develop/cms/utils/compat/__init__.py
> 
> Perhaps nows a good time to put 3.2 into there now, before we forget.

https://github.com/django-cms/django-cms/pull/7018/files#diff-7966586f25b4db1d9d9c45e65dfa7d27c2e19cb9e59fced89cd014dc62cef151R14

Already done. @marksweb

cms/cache/page.py
```diff
@@ -77,16 +78,28 @@ def set_page_cache(response):
...
-                _page_cache_key(request),
-                (
-                    response.content,
-                    response._headers,
-                    expires_datetime,
-                ),
-                ttl,
-                version=version
-            )
+            if not DJANGO_3_2:
```
Aiky30, line 81: As above, there's a lot of duplicated code for the header change. We should just use a new variable for the header I think. Something like, this also makes it much clearer for anyone cleaning out support logic in the future:

```
# support < django 3.2
if not DJANGO_3_2:
    response_headers = response._headers
# support >= django 3.2
else:
    response_headers = response.headers

cache.set(
    _page_cache_key(request),
    (
        response.content,
        response_headers,
        expires_datetime,
    ),
    ttl,
    version=version
)
```

cms/admin/pageadmin.py
```diff
@@ -416,9 +417,15 @@ def change_view(self, request, object_id, form_url='', extra_context=None):
 
         response = super().change_view(
             request, object_id, form_url=form_url, extra_context=extra_context)
-        if tab_language and response.status_code == 302 and response._headers['location'][1] == request.path_info:
-            location = response._headers['location']
-            response._headers['location'] = (location[0], "%s?language=%s" % (location[1], tab_language))
+        if not DJANGO_3_2:
```
Aiky30, line 420: There's a lot of duplicated code for the header change. We should just use a new variable for the header I think. Something like, this also makes it much clearer for anyone cleaning out support logic in the future:

```
# support < django 3.2
if not DJANGO_3_2:
    response_headers = "_headers"
# support >= django 3.2
else:
    response_headers = "headers"

if tab_language and response.status_code == 302 and response['response_headers']['location'][1] == request.path_info:
    location = response['response_headers']['location']
    response['response_headers']['location'] = (location[0], "%s?language=%s" % (location[1], tab_language))
```

I'm not sure if we created a reference to response.headers and then made a change to the value of our new variable would have changed the reference or created a new reference to the value. This may work as:

```
# support < django 3.2
if not DJANGO_3_2:
    response_headers = response._headers
# support >= django 3.2
else:
    response_headers = response.headers

if tab_language and response.status_code == 302 and response_headers['location'][1] == request.path_info:
    location = response['response_headers']['location']
    response_headers['location'] = (location[0], "%s?language=%s" % (location[1], tab_language))
```

The manual test for the above is that the following memory references are the same, if not the code above doesn't work and a setter would be required:

```
if not DJANGO_3_2:
    id(response_headers) == id(response._headers)
else:
    id(response_headers) == id(response.headers)
```

cms/tests/test_site.py
```diff
@@ -63,7 +64,12 @@ def test_site_preview(self):
             # simulate user clicks on preview icon
             response = self.client.get(admin_reverse('cms_page_preview_page', args=[page.pk, 'de']))
             self.assertEqual(response.status_code, 302)
-            self.assertEqual(response._headers['location'][1], 'http://sample2.com{}&language=de'.format(page_edit_url_on))
+            if DJANGO_2_2 or DJANGO_3_0 or DJANGO_3_1:
```
Aiky30, line 67: This is strange to do, normally you would have a separate test for the versions and use SkipIf(DJANGO_2_2) decorated on the function. I guess this will be ok to leave.

Aiky30 (commented):
Ok but could be cleaned up.

cms/cache/page.py
vinitkumar: Righto!

goutnet:
@NicolaiRidani as we discussed yesterday, do we have a chance to merge this for the coming 3.8.1? let me know.

goutnet:
@FinalAngel  @macolo  can you have a look?

Aiky30:
Hi @vinitkumar, Can you please fix the conflict in the changelog and I'll review asap. Ta.

vinitkumar:
> Hi @vinitkumar, Can you please fix the conflict in the changelog and I'll review asap. Ta.

It's done.

Issue #3378, opened 2014-08-13
The purpose of current_page in applications_page_check


mmarzantowicz:
What is the purpose of `current_page` argument in following function (https://github.com/divio/django-cms/blob/develop/cms/appresolver.py#L29)?

``` Python
def applications_page_check(request, current_page=None, path=None):
    """Tries to find if given path was resolved over application.
    Applications have higher priority than other cms pages.
    """
    if current_page:
        return current_page
```

Quick grep over source code shows that it's not used anywhere. What is more, if it is `not None` it returns itself and makes `applications_page_check` quite ... useless?

``` Bash
$ grep -nIR 'applications_page_check'
cms/middleware/page.py:6:    from cms.appresolver import applications_page_check
cms/middleware/page.py:14:            request._current_page_cache = applications_page_check(request)
cms/appresolver.py:28:def applications_page_check(request, current_page=None, path=None):
cms/tests/apphooks.py:10:from cms.appresolver import applications_page_check, clear_app_resolvers, get_app_patterns
cms/tests/apphooks.py:175:            attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash
cms/tests/apphooks.py:187:            attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash and language prefix
cms/tests/apphooks.py:234:                    attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash
cms/tests/apphooks.py:241:                    attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash
cms/tests/apphooks.py:260:            attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash
cms/tests/apphooks.py:277:            attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash
cms/tests/apphooks.py:289:            attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash
cms/tests/apphooks.py:355:            attached_to_page = applications_page_check(request, path=path[1:])  # strip leading slash
cms/views.py:153:        # since we always have a page at this point, applications_page_check is
cms/views.py:155:        # page = applications_page_check(request, page, slug)
```

Can I remove this argument or deprecate it with warning or there are plans to use it somewhere it the future?

markrofail:
Why is this still open? Wasn't it resolved by #7020

marksweb:
@markrofail Good spot, I'll close this, thanks!

PR #7020, opened 2021-04-14
Remove unused argument from `applications_page_check`

marksweb:
Addressing an issue from August 2014 here!!

## Description

Remove unused argument `current_page` from `applications_page_check`

* #3378

## Checklist


TiredFingers:
Signature changing always scary me. Using PyCharm I found 26 usages without any attemptions to use `current_page` argument. All checks passed. This change will not broke any. But I think we need somebody with expert level with django cms

marksweb:
> Signature changing always scary me. Using PyCharm I found 26 usages without any attemptions to use `current_page` argument. All checks passed. This change will not broke any. But I think we need somebody with expert level with django cms

The issue this fixes was agreed by the core team in 2014 at which point it was assumed to be an argument leftover from some old functionality. So I've actioned it on the grounds it had been accepted by the people that know this system the best.

TiredFingers (approved):

victor-yunenko (approved):

greengoaxe:
Hi @vinitkumar,
may you merge it?
Thanks,
greengoaxe

cms/appresolver.py
```diff
@@ -23,12 +23,10 @@ def clear_app_resolvers():
     APP_RESOLVERS = []
 
 
-def applications_page_check(request, current_page=None, path=None):
+def applications_page_check(request, path=None):
     """Tries to find if given path was resolved over application.
     Applications have higher priority than other cms pages.
     """
-    if current_page:
```
Aiky30, line 30: @marksweb How do you know that this is unused? This could be required for apphooks.

marksweb: @Aiky30 Based on the discussions on the associated issue (#3378) and a search of the project for use of this function.

The original issue is old, but has input from core members who agree that it's superfluous and likely stems from before the switch to treebeard.

vinitkumar:
@greengoaxe Yes, I will merge it once the test passes.

PR #7027, opened 2021-04-29
Fix github action job database-postgres

fp4code:
## Description

Fix github action database-postgres which was using sqlite3 (missing DATABASE_URL)

## Related resources

#7000

## Checklist

* [ ] I don't have updated the **CHANGELOG.rst** (fix an already written thing)


.github/workflows/test.yml
```diff
@@ -89,6 +92,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install pytest
         pip install -r test_requirements/${{ matrix.requirements-file }}
+        pip install -r test_requirements/mysql.txt
```
vinitkumar, line 95: Installing mysqlclient and psycopg2 doesn't in base is fine IMO.

.github/workflows/test.yml
```diff
@@ -43,11 +43,14 @@ jobs:
...
         pip install pytest
         pip install -r test_requirements/${{ matrix.requirements-file }}
+        pip install -r test_requirements/postgres.txt
         python setup.py install
 
     - name: Test with django test runner
       run: |
         python manage.py test
+      env:
+        DATABASE_URL: postgres://postgres:postgres@127.0.0.1/postgres
```
vinitkumar, line 53: This is a great catch! Thanks for spotting this.

.github/workflows/test.yml
vinitkumar: Can you please revert this changes & put mysqlclient and psycopg2 back in requirements_base.txt file?

vinitkumar (changes requested):
I like what you have done here. Please look at these comments and do the updates accordingly. Thanks for your contributions. 🎉

.github/workflows/test.yml
fp4code: We have to think about people that want to run `./manage.py test` on their computer, just using SQLite database, or using the postgreSQL database. If you include `mysqlclient==2.0.3` in `requirements_base.txt` file, the test will crash because it needs mySQL ou mariadDB to be installed on the computer:
    OSError: mysql_config not found
    mysql_config --version
    mariadb_config --version
    mysql_config --libs

vinitkumar: @fp4code Alright, let me think about it.

fp4code: Pleas think about a Django-CMS tester. He will tray to to a test (using just default sqlite).
`./manage.py test` Result if we do not split `mysqlclient` and `psycopg2` out or requirement_base.txt file, the test will crash. Very bad for Django-CMS fame!

fp4code:
Maybe we can use only one `databases.txt` file containing
`
mysqlclient==2.0.3
psycopg2==2.8.6
`

marksweb (approved):

Issue #7030, opened 2021-05-06
Permission signals don't work when user model is customized


monim67:
## Description

Permission signals don't work when the user model is customized by AUTH_USER_MODEL settings.

## Steps to reproduce

1. Start with a fresh django-cms installation.
2. Customize the default user model by setting AUTH_USER_MODEL to your own user model.
3. Migrate and start the development server.
4. Login to CMS and create a new user.

## Expected behaviour

A `post_save_user` signal should create a `PageUser` model instance for the user created.

## Actual behaviour

The `PageUser` model instance is not created.

PR #7034, opened 2021-05-13
Fix hard-coded primary key of PageUser

monim67:
## Description

This PR fixes the issue #7030 for cases where the user model is named other than User.

The `post_save_user` signal is using a hard-coded value of primary key `user_ptr_id` while creating `PageUser` instance. So if the user model is named anything other than User e.g Admin then the database column name becomes `admin_ptr_id` which raises an error while executing the signal.

This PR uses `pk` to assign primary key field.


vinitkumar (approved):
Looks good to me. Let's get this merged if all the tests pass 🚢  🚀

marksweb (approved):
Makes sense. My custom users are always `User` so I guess this never came up!

PR #7058, opened 2021-05-25
tests: adding test to ensure no migration can collide with 4.x upgrad…

goutnet:
## Description

Following conversation on slack, Adding a test here to ensure no new migration can be added to the application (as this would make problems on 4.x migration path).

@marksweb  @Aiky30  @vinitkumar

### DIFF
--- a/cms/tests/test_compatibility_with_4x_branch.py
+++ b/cms/tests/test_compatibility_with_4x_branch.py
@@ -1,0 +1,24 @@
+import cms
+import os
+from django.test import testcases
+
+class Compatibility4xTestCase(testcases.TestCase):
+
+
+    def test_ensure_no_migration_is_added(self):
+        """
+        to ensure the next version is compatible with the 4.x branch, we need to make sure
+        no new migration is added (otherwise, this will thnn conflicts with what is present in the 4.x
+        branch
+        """
+
+        migration = os.path.join("cms", "migrations")
+        MAX = 22
+
+        for root, _, files in os.walk(migration):
+            for name in files:
+                if name == "__init__.py":
+                    continue
+
+                mid = int(name.split("_")[0])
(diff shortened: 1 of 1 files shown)


vinitkumar (approved):

vinitkumar (changes requested):
Please check the tests, they are failing.

cms/tests/test_compatibility_with_4x_branch.py
```diff
@@ -0,0 +1,24 @@
+import cms
```
marksweb, line 1: This isn't being used according to the checks so can be removed 👌

marksweb (changes requested):

cms/tests/test_compatibility_with_4x_branch.py
goutnet: thanks! (and done) :)

goutnet:
tests still failing, let me check

Aiky30:
@marksweb @vinitkumar Seems the settings for the CI are configured to enforce the two tests: database-mysql, database-postgres. Are you doing something behind the scenes with this as it's not in the codebase yet or else it would have ran here.

goutnet:
Actually they just run all the tests in the different env. This is just one
test run in all scenarios :)

goutnet:
(closed&re-open to force travis to run again)

goutnet:
must be an issue in the config on github, let me check

vinitkumar:
@goutnet still comes up
Like I said earlier, do this from a fresh branch, some weird github actions caching issue is going on here. Also, the fastest way to get this resolved. It's just a diff from a single file, so shouldn't take much time.

goutnet:
not really sure what's going on on 3.6 there …

vinitkumar (approved):

PR #7065, opened 2021-05-28
fix: Placeholder copy orphaned plugin children

Aiky30:
## Description

Created logic to handle children that have been orphaned by badly positioned plugins. It's possible to position a plugin with a higher number than the parent position. This coupled with a mismatch of id's leaves us in a tricky situation. The id's also play a part. The fact is there are projects that now have this inconsistency so we need to handle the inconsistency.

We may also need to stop the UI / API allowing a plugin to be added as a child with a position that's higher than the parent, requires heavy lifting to remap the placeholders positions.

To recreate the issue:

Create 2 plugins
- Plugin 1, pk 1, position 1
- Plugin 2, pk 2, position 2

Then move Plugin 1 as a child of plugin 2:
- Plugin 2, pk 2, position 2
    - Plugin 1, pk 1, position 1 (Child)

The copy logic will now see plugin 1 first and try and allocate plugin 2 as a parent, it can't do that because plugin 2 hasn't been copied yet ......


goutnet:
@Aiky30  shouldn't this be sent to `develop-4` instead of `release/4.0.x` ?

Aiky30:
@goutnet this is a critical fix for the client using this branch. It will be released once verified as a fix from the client and we will probably want to port the change back up to the develop branch, maybe the release number needs to be pulled up to the develop-4 branch too.

goutnet (approved):
LGTM, following comments we should merge this here and cherry pick to `develop-4` too then.

PR #7078, opened 2021-06-18
Switch docs theme to furo

marksweb:
## Description

Following discussion in the documentation work group, this switches the docs theme to furo.


crydotsnake (changes requested):
We could change Divio AG too djangoCMS Association in the Copyright section

vinitkumar:
@marksweb There are tests failing as well due to missing dependencies. Please fix them too.

marksweb:
> @marksweb There are tests failing as well due to missing dependencies. Please fix them too.

That highlights a problem with those tests. The sphinx requirements are in the docs requirements, so the tests need reviewing that are failing as to what they're trying to do.

marksweb:
@vinitkumar So it turns out there's a unit test which tests the docs, so there were some of the docs requirements in the base test requirements.

No need for this duplication, so I've removed the docs requirements from the test requirements and added an install step for the docs requirements in the unit tests. Things are starting to pass now. I'll address @crydotsnake point next

docs/conf.py
```diff
@@ -64,7 +64,7 @@
 current_year = datetime.datetime.now().year
 # General information about the project.
 project = 'django cms'
-copyright = '2009-{}, Divio AG and contributors'.format(current_year)
+copyright = '2009-{}, djangoCMS Association and contributors'.format(current_year)
```
crydotsnake, line 67: Should we also change the year here? Because the djangoCMS Association exists since 2020 i think.

crydotsnake (changes requested):

docs/conf.py
FinalAngel: should be "django CMS Association" though

marksweb: @FinalAngel yeah good point, updated.

crydotsnake (approved):

vinitkumar (approved):

PR #7081, opened 2021-06-28
feat: add django3.2 in the framework identifier for setup.py

vinitkumar:
We support Django3.2, but we forgot to add it to the setup.py, If not
present, it will not show the support for django3.2 in PyPI.

Authored-by: Vinit Kumar <vinit.kumar@socialschools.nl>


marksweb (approved):

goutnet:
good point, and also a last minute change before release.

Thanks for the catch!

PR #7082, opened 2021-06-29
fix: missing tests of django3.2 for mysql and sqlite

vinitkumar:
- Add missing tests for django3.2. We somehow missed adding them to the CI when we added support for Django3.2.


goutnet (approved):
LGTM

PR #7070, opened 2021-06-03
Fixed browser freezing on cms toolbar markup refresh after plugin edit

nichoski:
## Related resources
This fixes the bug reported here:
https://github.com/django-cms/django-cms/issues/7069

To update the contents of the django-cms toolbar after a plugin update, instead of creating a diff between old content and new content and applying the diff, this fix simply replaces the old content with the new content. This is because on sites with more content, the diff creating process is incredibly slow. Please see issue 7069 for more info.

* [x] I have opened this pull request against ``develop``
* [ ] I have added or modified the tests when changing logic
* [ ] I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog
* [ ] I have read the [contribution guidelines ](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

PR #7086, opened 2021-07-01
Replacing toolbar contents and attributes intead of the whole toolbar…

nichoski:
## Description

When a page is in non legacy structure mode, clicking the toolbar switcher button in order to show the cms structure board (sidebar), reloads the contents of the toolbar before making the structure board (sidebar) visible. And since our recent change in PR #7070, when contents of the toolbar are reloaded, the old toolbar element is replaced with the newly fetched one. However, the cms structure board does not update its reference to the toolbar object in this situation, and it ends up referencing the old, deleted toolbar. This causes a problem when the structure board tries to calculate the css style 'right' attribute, as it needs to account for the width of the toolbar object, and it ends up showing the toolbar outside the viewport. We didn't catch this when preparing PR #7070 because all of the pages we tested with initially were in legacy structure mode. And updating the toolbar after a plugin edit did not bring this issue, because the structure board was updated too in those cases.

## Fix

Instead of replacing the old toolbar object with the new, we are now replacing only its contents and attributes. That way all the parts of the code that referenced the old toolbar object are still referencing the correct toolbar object.

## Checklist


marksweb:
Closing this as I believe it's been done via another PR.

PR #7089, opened 2021-07-05
Documentation for multi-site set-ups

MacLake:
## Description

As announced in the documentation channel on Slack, I have written some documentation on multi-site installations. and added it to the how to section. I’m not a native English speaker, so feel free to check the language and anythiing else that should be improved. Anyway, at least it’s a start.

Cheers,
Jens-Erik

## Related resources

* #...
* #...

## Checklist


marksweb:
This is a bit of an opinionated area of django where there's various ways of doing things.

If I get chance I'll consider expanding on this. I'm running a multisite django instance where we run 1 application on 1 stack, with 1 copy of the key files (wsgi, settings etc). Locally things are driven from the hosts file to differentiate between sites and the application is configured largely with environment variables. It's a very straightforward and low maintenance way to run things.

MacLake:
Sure, if we have another way to do multi-site instances, it would be good to know about it. So far we didn’t have any in the docs.  Well, strictly spoken, it’s a matter of Django and not Django CMS anyway, but still it would be good to have it in the docs here.

TiredFingers (approved):
Thanks for your work

PR #7095, opened 2021-07-16
fix: discrepancy around python 3.6 compatibility in release 3.9

### LINKED ISSUE #7091: [BUG] Django CMS 3.9 Isn't Supported on Python 3.6
## Description

According to the [Django/Python compatibility table](https://github.com/django-cms/django-cms/blame/3.9.0/docs/index.rst#L166), Django CMS 3.9 isn't supported on Python 3.6. This seems odd as it currently lists that every Django CMS from 3.4.5 through 4.0 support Python 3.6, _except_ Django CMS 3.9. Also the Python 3.6 classifier is still listed in the [setup.py](https://github.com/django-cms/django-cms/blob/3.9.0/setup.py#L27) for Django CMS 3.9. So either the documentation or the setup.py is wrong.

## Steps to reproduce

1. Go to https://github.com/django-cms/django-cms/blame/3.9.0/docs/index.rst#L166
2. Observe that Django CMS 3.9 is listed as not being compatible with Python 3.6.
3. Then go to https://github.com/django-cms/django-cms/blob/3.9.0/setup.py#L27, which still lists the Python 3.6 classifier is still listed.

## Expected behaviour

The documentation should be consistent with the setup.py classifiers.

## Actual behaviour

The documentation is not consistent with the setup.py classifiers.

Django CMS 3.9.0
Python 3.6

marksweb:
## Description

Fix #7091: Tidying up compatibility around python 3.6 in version 3.9 which was unreleased at the time of updates to the django/python compatibility table in the docs.

Support now aligned with django at python 3.6, 3.7, 3.8, 3.9.

## Related resources

* #7091 


marksweb:
I'm planning another PR against develop for similar changes. Unless we think it only needs the same then it can jusy be taken across directly.

goutnet:
I think we should just push everything into `develop`, as next release will  be pulling from develop (and maybe onto 3.9.0 or anything). at this point there is no real need to start from the release branch :)

marksweb:
@goutnet Yeah that's fair. I've merged develop into this branch and updated the base for the PR.

goutnet (approved):
LGTM

Issue #7102, opened 2021-08-01
Flake8 checks being ignored


marksweb:
Looking over a PR I've submitted I've noticed we're running `flake8`, but ignoring any errors. (using the `--exit-zero` flag)

This seems a bit pointless.

It's certainly something we should run, but clearly it needs better configuration. If the current state of the project is acceptable, the current failures should be added to the excluded list & the exit zero flag removed from the linting action.

Aiky30:
Agreed, I've not seen any of the projects actively ignoring flake8 errors. If there are errors / warnings that should be ignored they are normally added to the setup.cfg file.

marksweb:
@Aiky30 I'll go over things, make some sensible choices over ignoring things & changing code 👍

PR #7104, opened 2021-08-03
Stop ignoring flake8 errors [#7102]

marksweb:
## Description

This change makes the `flake8` check meaningful by actually letting it fail & cause the test suite to fail.

Until now the linting action has been running `flake8 --exit-zero` so it's reported errors if you looked at it's output, but it wouldn't actually return a failure.

For the majority of these changes, it's fairly simple but there's a couple of significant changes to point out.

Until now we've had a fairly old test which would call patch `pyflakes` and test that there are no errors or warnings. This seems pointless given that we're running `flake8` and pyflakes wraps it. So I've deleted this test and the patching module.

Also, the commit `Fix W605 (invalid escape sequence)`, updates regex usage, which I assume have been in place since there was python 2 support. For simplicities sake, I've made these strings raw.

I also wanted to get rid of all bare `except:` usage, however some instances of this isn't worth changing and where I had tried I came up against the [error from tests](https://github.com/marksweb/django-cms/runs/3231042231?check_suite_focus=true);

```
TypeError: catching classes that do not inherit from BaseException is not allowed
```

This was only in trying to catch errors like `AttributeError`, `NoReverseMatch` and `TypeError` itself. I've not seen this error before and couldn't track down where it was being raised, but these exceptions do inherit from `BaseException` so I must be missing something here. For now I've added explicit `# noqa: E722` to allow the bare except handlers.

## Related resources

* #7102 

## Checklist


vinitkumar:
@marksweb This is a good one mark. The flake8 --exit-zero was only supposed to be temporary, but it ran longer than it's due. Anyway, once this build passes, I will merge this PR and we will enforce the code quality checks using flake8.

cms/middleware/toolbar.py
```diff
@@ -40,7 +40,7 @@ def is_cms_request(self, request):
 
         try:
             match = resolve(request.path_info)
-        except:
+        except Resolver404:
```
Aiky30, line 43: This is fine but I think this needs to be listed somewhere as a possible breaking change as this handles every request and may have been swallowing other errors.

menus/templatetags/menu_tags.py
```diff
@@ -157,16 +157,14 @@ def get_context(self, context, from_level, to_level, extra_inactive,
...
-            context['children'] = children
-            context['template'] = template
-            context['from_level'] = from_level
-            context['to_level'] = to_level
-            context['extra_inactive'] = extra_inactive
-            context['extra_active'] = extra_active
-            context['namespace'] = namespace
-        except:
-            context = {"template": template}
+        context['children'] = children
```
Aiky30, line 160: I don't think this should be changed, it was clearly catching something in the past to have a try except block. Do you know what it was catching before and why it's no longer required?

Aiky30 (commented):
A couple of cautious comments to check you have explored them.

Everything else looks good.

cms/middleware/toolbar.py
marksweb: Yeah that's a fair point. One for the release notes & us to remember perhaps?

menus/templatetags/menu_tags.py
marksweb: I couldn't see how it could be catching anything as its not really doing anything.

marksweb: @Aiky30 I think it was written in 2010 to catch an error on `children`, but since then theres a default of `children = []`
This is where it started;
https://github.com/django-cms/django-cms/commit/bda49b26cd5f1216dfadb756ec3b693348295d3f#diff-2f98e9f55ac912e040227934eccb938c8a4d4b511f0f3222466753e17a23f713R80

Aiky30: Yup agreed, that looks really bad. Haha

Aiky30 (approved):
Happy with all of the changes.

vinitkumar (approved):

PR #7105, opened 2021-08-04
Django 3.0 support

Aiky30:


cms/api.py
```diff
@@ -37,6 +36,8 @@
 from cms.utils.plugins import copy_plugins_to_placeholder
 from menus.menu_pool import menu_pool
 
+from six import string_types
```
auvipy, line 39: I am not sure why python 2 support is being kept with six?

cms/models/permissionmodels.py
```diff
@@ -5,14 +5,16 @@
...
-from django.utils.encoding import force_text, python_2_unicode_compatible
+from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
 
 from cms.models import Page
 from cms.models.managers import (PagePermissionManager,
                                  GlobalPagePermissionManager)
 from cms.utils.compat import DJANGO_1_11
 
+from six import python_2_unicode_compatible
```
auvipy, line 16: django 1.11 is the last version to keep python 2 support isn;t it?

cms/utils/compat/__init__.py
```diff
@@ -12,3 +12,4 @@
 DJANGO_2_0 = LooseVersion(DJANGO_VERSION) < LooseVersion('2.1')
```
auvipy, line 12: django 2,0 and 2.1 is not supported anymore. not even 3.0

auvipy (changes requested):

cms/models/permissionmodels.py
marksweb: Regardless of which version it was, python 2 support shouldn't be maintained anymore 👍

Aiky30: This is a backport of the original change for django 3.0. This has been done to have a direct comparison to the original change. There is another PR for django 3.1 and another for 3.2 which is a few PRs combined.

cms/utils/compat/__init__.py
Aiky30: Same as above:

This is a backport of the original change for django 3.0. This has been done to have a direct comparison to the original change. There is another PR for django 3.1 and another for 3.2 which is a few PRs combined.

cms/api.py
Aiky30: Django 3.0 is also redundant 

This is a backport of the original change for django 3.0. This has been done to have a direct comparison to the original change. There is another PR for django 3.1 and another for 3.2 which is a few PRs combined.

Issue #7106, opened 2021-08-04
[BUG] - Language Middleware does not support Django LANGUAGE_COOKIE_* settings


jbpenrath:
## Description
Except `LANGUAGE_COOKIE_NAME`, all other `LANGUAGE_COOKIE` django settings are not take in account.

Furthermore, modern browsers will start soon to reject cookies which have same-site sets to None and are not secure. Currently it is impossible to respect this policy with Django CMS.

## Steps to reproduce

1. In your settings.py, set `LANGUAGE_COOKIE_SAMESITE='Lax'` or any other `LANGUAGE_COOKIE_` settings except `LANGUAGE_COOKIE_NAME`.
2. Get a cms page (from browser, clear your cookies)
3. Check the response header, you'll see that `Set-Cookie: django_language=en; expires=Thu, 04 Aug 2022 15:40:41 GMT; Max-Age=31536000; Path=/;` doest not contain `SameSite=Lax`

## Expected behaviour
Cookie related to language preference should take in account Django `LANGUAGE_COOKIE_*` settings.

## Actual behaviour
Only `LANGUAGE_COOKIE_NAME` setting is take in account.

## Screenshots
<img width="655" alt="Capture d’écran 2021-08-04 à 17 57 37" src="https://user-images.githubusercontent.com/9265241/128213998-64fb517e-8497-465b-9308-8aa8f1149d2f.png">

## Additional information (CMS/Python/Django versions)

Django 3.2.6
Django CMS 3.9.0
Python 3.7.9

jbpenrath:
Duplicate with #7092

Issue #7092, opened 2021-07-15
[BUG] LanguageCookieMiddleware Should Respect All LANGUAGE_COOKIE_* Settings


gopackgo90:
## Description

Today, the `django_cms.cms.middleware.language.LanguageCookieMiddleware` doesn't respect numerous Django settings such as `LANGUAGE_COOKIE_HTTPONLY`, `LANGUAGE_COOKIE_SAMESITE`, `LANGUAGE_COOKIE_SECURE`, etc.

## Steps to reproduce

1. Use `django_cms.cms.middleware.language.LanguageCookieMiddleware` in `MIDDLEWARE` in `settings.py`.
2. Set one of the unused LANGUAGE_COOKIE_* settings in `settings.py`, such as `LANGUAGE_COOKIE_HTTPONLY = True`.
3. Open up the Django CMS application in a new session.
4. Observe that the `LANGUAGE_COOKIE_NAME` is created without the HttpOnly flag being set to true.

## Expected behaviour

The LANGUAGE_COOKIE_* settings should be respected in `django_cms.cms.middleware.language.LanguageCookieMiddleware`.

## Actual behaviour

The LANGUAGE_COOKIE_* settings are not respected in `django_cms.cms.middleware.language.LanguageCookieMiddleware`.

## Additional information (CMS/Python/Django versions)

`django_cms.cms.middleware.language.LanguageCookieMiddleware` should use all of the LANGUAGE_COOKIE_* settings used in `django.views.i18n.set_language`, see https://github.com/django/django/blob/3.2.5/django/views/i18n.py#L61-L69.

Including `LANGUAGE_COOKIE_AGE` as the `max_age` argument instead of `expires`. Also note that the [current calculation erroneously](https://github.com/django-cms/django-cms/blob/3.9.0/cms/middleware/language.py#L19) says it's for 10 years, but really it's only 1 year.

PR #7107, opened 2021-08-04
fix(middleware): add support of Django language_cookie settings 

jbpenrath:
## Description

LanguageCookieMiddleware currently does not take in account LANGUAGE_COOKIE_* Django settings. So we are not able to customize language cookie properties except its name. 

Some modern browsers start to reject cookies which have same-site property defined to `None` and `Secure` defined to `False`. Currently we are not able to customize these properties and SameSite=None and Secure=False are the default value to the Django language cookie.

## Related resources

Issues #7106 #7092

## Checklist

_There are no tests related to this middleware._


marksweb:
Looks ok to me, but you could be explicit about it given all the `kwargs` now being passed & define `value=language`.

Not the end of the world if we leave it as is though.

jbpenrath:
> Looks ok to me, but you could be explicit about it given all the `kwargs` now being passed & define `value=language`.
> 
> Not the end of the world if we leave it as is though.

Done !

marksweb (approved):

vinitkumar (approved):

PR #7115, opened 2021-08-19
fix: do not replace & with &amp; in URLs

nichoski:
## Description

Do not convert & to &amp; in URLs. This is fixing the problem described in issue [7113](https://github.com/django-cms/django-cms/issues/7113). This is for branch release/3.7.x


vinitkumar:
@nichoski The fix looks legit, thought I am not convinced that it would work properly without bumping the version as well. To make sure it works fine, we need to bump the version and the generated files needs to build in a 3.7.5 directory to ensure there are no weird issues because of cached assets even on the new release.

vinitkumar (changes requested):
@nichoski The fix looks legit, thought I am not convinced that it would work properly without bumping the version as well. To make sure it works fine, we need to bump the version and the generated files needs to build in a 3.7.5 directory to ensure there are no weird issues because of cached assets even on the new release.

fsbraun:
Implemented by #7114

Issue #7110, opened 2021-08-09
Possibility to optimize downcast_plugins function


zkne:
Hello everyone I’ve noticed one potential performance improvement.

I have an issue rendering a plugin with one-to-many relations and related objects have relations to other objects. Rendering all data in the template is causing a lot of queries as every object is doing an additional query to fetch the related object.

I’ve noticed the when rendering plugin [downcast_plugins](https://github.com/django-cms/django-cms/blob/18e146495a64e88b35a2d7d0192c48c1887e2596/cms/utils/plugins.py#L259) function is used.
It internally uses [get_render_queryset](https://github.com/django-cms/django-cms/blob/e2c684a2710738de2d86e659038c59513e865e6a/cms/plugin_base.py#L170) classmethod of plugin.
Doing prefetch related on queryset is not possible as the _downcast_plugins_ uses queryset iterator on this [line](https://github.com/django-cms/django-cms/blob/18e146495a64e88b35a2d7d0192c48c1887e2596/cms/utils/plugins.py#L284).

Is there a need to use iterator here as the chunk size is 2000 and there shouldn't ever be more than 2000 plugins per page so it's not providing any memory optimization but strips the way to use _prefetch_related_ on queryset. Here is the reference: https://docs.djangoproject.com/en/3.2/ref/models/querysets/#iterator

My opinion is that there is no need for an iterator. Removing it will give the possibility to do a _prefetch_related_ on queryset for a plugin.

I'll be glad to contribute a PR for this issue.

jrief:
Thanks @zkne for finding this. Could you please make a benchmark and test the difference in rendering a page containing a lot of plugins, once as

```python
        for instance in plugin_qs.iterator():
```

and once as

```python
        for instance in plugin_qs.all():
```

would be interesting to see the difference.

zkne:
Hey, @jrief I haven't created a public benchmark as you requested. I've checked it for a private project I've been working on and performance gains are significant if there are a lot of plugins on the page that have a complex object structure.

If I find time I will create some POC project to show the potential performance gains. 🤞

PR #7121, opened 2021-09-29
Optimizing downcast_plugins function

zkne:
## Description

Changed the way how iteration is done for plugins queryset. Changed from `.iterator()` to `.all()` as it give us ability to prefetch related object in the plugins queryset.

## Related resources

#7110 

## Checklist

### DIFF
--- a/cms/utils/plugins.py
+++ b/cms/utils/plugins.py
@@ -281,7 +281,7 @@ def downcast_plugins(plugins,
 
         # put them in a map so we can replace the base CMSPlugins with their
         # downcasted versions
-        for instance in plugin_qs.iterator():
+        for instance in plugin_qs.all():
             placeholder = placeholders_by_id.get(instance.placeholder_id)
 
             if placeholder:


vinitkumar:
@zkne I am not so sure, if doing this will actually improve the performance.

This is from the docs:

```
Evaluates the QuerySet (by performing the query) and returns an iterator (see PEP 234) over the results. A QuerySet typically caches its results internally so that repeated evaluations do not result in additional queries. In contrast, iterator() will read results directly, without doing any caching at the QuerySet level (internally, the default iterator calls iterator() and caches the return value). For a QuerySet which returns a large number of objects that you only need to access once, this can result in better performance and a significant reduction in memory.

Note that using iterator() on a QuerySet which has already been evaluated will force it to evaluate again, repeating the query.

Also, use of iterator() causes previous prefetch_related() calls to be ignored since these two optimizations do not make sense together.

Depending on the database backend, query results will either be loaded all at once or streamed from the database using server-side cursors.
```

zkne:
@vinitkumar You pointed out the crucial difference between iterator() and all(). 
```
Also, use of iterator() causes previous prefetch_related() calls to be ignored since these two optimizations do not make sense together.
```
If there is complex object relation for the plugins you won't be able to do a prefetch_related() and then we end up with an N+1 problem. For example, a plugin that has a reference to some model and that model has M2M relations. 
Changing from iterator() to all() won't give an increase in performance out of the box. But when rendering if you prefetch all objects it will result in the performance gain (prefetching objects in get_render_queryset() method on the plugin).

I don't know if I made myself clear enough.

PR #7144, opened 2021-10-25
fix: Prefer titles matching request language

### LINKED ISSUE #3548: Homographs in slugs for different pages and different languages should not raise MultipleObjectsReturned
Your site has two sibling pages:

```
# a page about sticks
/it/bastone
/fr/canne
/en/cane

# a page about dogs
/it/cane
/fr/chien
/en/dog
```

At present, as soon as you try to publish (it will let you save a draft) the Italian translation, you'll get:

```
MultipleObjectsReturned at /it/cane/
get() returned more than one Page -- it returned 2!
```

Issues to be decided:
- [ ] should we allow the creation of the slug? We would need to stop ignoring the language prefix when processing URLs.
- [ ] if we do, how would we decide what serve when the user asked for `/fr/cane/` - should it serve up `chien` or `canne` (see #3547 on django CMS's page-guessing behaviour)
- [ ] if we don't, we should advise the user of a slug conflict, instead of raising the `MultipleObjectsReturned` error.

macro1:
## Description

Fixes #3548

Currently language is not taken into account when resolving paths through a Title to a page. This PR uses the solution proposed in #3548 to prefer using a Title with a language that matches the request. This will allow homographs between different languages to return different pages.

Notably as this change only sorts the order in which Title records resolve, it should not remove any existing functionality. It should only improve the quality of results when there is a possibility of conflicting.

## Related resources

Getting the language code for a request:
https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#how-django-discovers-language-preference

The `ExpressionWrapper()` I used for the query changes is a little strange, but we're stuck with it till Django 4.0:
https://code.djangoproject.com/ticket/27021

## Checklist


cms/utils/page.py
```diff
@@ -92,6 +93,17 @@ def get_page_from_path(site, path, preview=False, draft=False):
         titles = titles.filter(publisher_is_draft=False)
     else:
         titles = titles.filter(published=True, publisher_is_draft=False)
+    if language_code is not None:
```
marksweb, line 96: Could you add some comments through these changes to hopefully make it obvious to anyone reading through, what is happening?

We get quite a broad range of django knowledge from contributors, and I think it's fair to say `annotate` isn't basic functionality so it might just make sure everyone can follow what's happening here. 👍

marksweb (changes requested):

cms/utils/page.py
macro1: Added a brief description. I'm not sure if code comments are the best way to get a crash course in `annotate()` but hopefully the note and link to Django's documentation will be helpful.

marksweb (approved):

PR #7156, opened 2021-11-11
Update releases.json

andre-dasilva:
I found that in the [djangocms-admin-style](https://github.com/django-cms/djangocms-admin-style) package there is a update notification for the django-cms version

the package uses the url https://releases.django-cms.org/ to check if there is new version, and this url is a redirect to: https://raw.githubusercontent.com/divio/django-cms/releases/releases.json.

The following things should be changed:
1. The redirect should be changed to https://raw.githubusercontent.com/django-cms/django-cms/releases/releases.json (I would do it but i don't know where 😅)
2. The releases.json file is outdated and this PR should update this.

### DIFF
--- a/releases.json
+++ b/releases.json
@@ -1,36 +1,72 @@
 {
     "latest": {
-        "version": "3.8.0",
-        "url": "http://docs.django-cms.org/en/latest/upgrade/3.8.0.html"
+        "version": "3.9.0",
+        "url": "https://docs.django-cms.org/en/latest/upgrade/3.9.0.html"
     },
     "patches": [
         {
+            "version": "3.8.0",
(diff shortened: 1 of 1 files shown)


marksweb:
@NicolaiRidani The redirect mentioned here needs to be updated, likely by someone with DNS access to the domain as this is using a subdomain to redirect.

NicolaiRidani:
Hi @FinalAngel is this something you can change?

andre-dasilva:
@NicolaiRidani the redirect is an issue. but i also included some changes for the releases.json file. thats why it is a PR.

NicolaiRidani:
> @NicolaiRidani the redirect is an issue. but i also included some changes for the releases.json file. thats why it is a PR.

Ok thx

PR #7188, opened 2022-01-22
fix: Unlocalize page and node ids when rendering the page tree in the admin

### LINKED ISSUE #7175: [BUG] Admin page tree broken when on sites with more than 1000 pages and L10N turned on
## Description
On sites that have more than 1000 pages, when rendering [admin/cms/page/tree/menu.html](https://github.com/django-cms/django-cms/blob/develop/cms/templates/admin/cms/page/tree/menu.html#L20) all the page ids and node ids are rendered localized, with e.g. commas or apostrophes for thousands separators.

When we then expand a node, those node's `data-id` are picked up and stored in local storage under `cms_cookie` in their localized form:

image

On the next page reload, the content of `cms_cookie` is passed in the URL, still localized:

```
/en/admin/cms/page/get-tree/?language=en&openNodes%5B%5D=1&openNodes%5B%5D=2262&openNodes%5B%5D=2%2C262&nodeId=&site=1&_=1640334050634
```

Resulting in an `ValueError: invalid literal for int() with base 10: '2,262'` error on [cms/admin/pageadmin.py:1426](https://github.com/django-cms/django-cms/blob/f622008b471e664249b72db6dec528d4d8bd4255/cms/admin/pageadmin.py#L1430)

## Steps to reproduce

1. Have a site with more than 1'000 pages (or having a page with an id > 1000), one of these pages with an id > 1000 has a child page
2. In the admin page tree expand that page to reveal its child
3. Refresh the admin tree list
4. Error 500, page tree is broken.

## Expected behaviour

Page tree should refresh and display normally.

## Actual behaviour

Mayhem.

image

Django-cms 3.9.0
Any supported version of Django
Any supported version of Python

Four possible courses of action, anyone of these fixes the problem, but you might want to apply them all:

- Apply `|unescape` from `l10n` in `admin/cms/page/tree/menu.html` whenever we render page or node ids. 
- In the JavaScript code storing the values in localstorage, sanitize the identifiers being stored
- In the JavaScript code reading the values from localstorage and passing them into the URL, sanitize the values again
- In `cms/admin/pageadmin.py` before blindly mapping to `int`, make sure we received actual ints, sanitize them first.

I'd gladly submit a PR for the first and/or fourth fix in the list above, but please advise on which one(s) of the options above you'd like to use to fix this.

mbi:
## Description

This fixes #7175 by:

1. unlocalizing page and node ids when rendering admin/cms/page/tree/menu.html
2. sanitizing page and node ids received from `request.GET` in cms/admin/pageadmin.py to make sure only unlocalized integers are passed on to `int`. 

See #7175 for a detailed description of the problem being addressed. 

<strike>NOTE: the new test takes approximately 10 seconds (on my old laptop) because it has to create 1000 pages to trigger the problem. If you see a more convenient way of triggering the problem, please let me know!</strike>

## Checklist

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -1,5 +1,6 @@
 import copy
 import json
+import re
 import sys
 import uuid
 from collections import namedtuple
@@ -1401,8 +1402,12 @@ class BasePageAdmin(PlaceholderAdminMixin, admin.ModelAdmin):
         """
         site = self.get_site(request)
         pages = self.get_queryset(request)
-        node_id = request.GET.get('nodeId')
-        open_nodes = list(map(int, request.GET.getlist('openNodes[]')))
+        node_id = re.sub(r'[^\d]', '', request.GET.get('nodeId', '')) or None
+        open_nodes = list(map(
+            int,
+            [re.sub(r'[^\d]', '', node) for node in
+                request.GET.getlist('openNodes[]')]
+            ))
 
         if node_id:
             page = get_object_or_404(pages, node_id=int(node_id))
--- a/cms/templates/admin/cms/page/tree/base.html
+++ b/cms/templates/admin/cms/page/tree/base.html
@@ -107,7 +107,7 @@
                                     <span class="label">{% trans "Sites" %}</span>
                                 </li>
                                 {% for site in tree.sites %}
-                                    <li{% if site.pk|unlocalize == tree.site.pk|unlocalize %} class="active"{% endif %}>
+                                    <li{% if site.pk == tree.site.pk %} class="active"{% endif %}>
                                         <a href="#{{ site.pk|unlocalize }}" class="js-cms-pagetree-site-trigger" data-id="{{ site.pk|unlocalize }}">{{ site.name }}</a>
                                     </li>
                                 {% endfor %}
--- a/cms/templates/admin/cms/page/tree/base.html
+++ b/cms/templates/admin/cms/page/tree/base.html
@@ -1,5 +1,5 @@
 {% extends "admin/change_list.html" %}
-{% load i18n admin_list static admin_urls cms_admin cms_js_tags cms_static cms_tags %}
+{% load i18n l10n admin_list static admin_urls cms_admin cms_js_tags cms_static cms_tags %}
 
 {# TODO might not need that #}
 {% block title %}{% trans "List of pages" %}{% endblock %}
@@ -107,8 +107,8 @@
                                     <span class="label">{% trans "Sites" %}</span>
                                 </li>
                                 {% for site in tree.sites %}
-                                    <li{% if site.pk == tree.site.pk %} class="active"{% endif %}>
-                                        <a href="#{{ site.pk }}" class="js-cms-pagetree-site-trigger" data-id="{{ site.pk }}">{{ site.name }}</a>
+                                    <li{% if site.pk|unlocalize == tree.site.pk|unlocalize %} class="active"{% endif %}>
+                                        <a href="#{{ site.pk|unlocalize }}" class="js-cms-pagetree-site-trigger" data-id="{{ site.pk|unlocalize }}">{{ site.name }}</a>
                                     </li>
                                 {% endfor %}
                                 {% if has_recover_permission %}
@@ -127,7 +127,7 @@
                         <form method="post" class="js-cms-pagetree-site-form cms-hidden">
                             <select name="site">
                                 {% for site in tree.sites %}
-                                    <option value="{{ site.pk }}">{{ site.name }}</option>
+                                    <option value="{{ site.pk|unlocalize }}">{{ site.name }}</option>
                                 {% endfor %}
                             </select>
                             {% csrf_token %}
@@ -186,7 +186,7 @@
                             "permission": {{ CMS_PERMISSION|bool }},
                             "debug": {{ DEBUG|bool }},
                             "filtered": {% if tree.is_filtered %}true{% else %}false{% endif %},
-                            "site": {{ tree.site.pk }},
+                            "site": {{ tree.site.pk|unlocalize }},
                             "hasAddRootPermission": {{ has_add_permission|yesno:"true,false" }},
                             "lang": {
                                 "code": "{{ preview_language|lower }}",
@@ -250,7 +250,7 @@
                         "permission": {{ CMS_PERMISSION|bool }},
                         "debug": {{ DEBUG|bool }},
                         "filtered": {% if tree.is_filtered %}true{% else %}false{% endif %},
-                        "site": {{ tree.site.pk }},
+                        "site": {{ tree.site.pk|unlocalize }},
                         "hasAddRootPermission": {{ has_add_permission|yesno:"true,false" }},
                         "lang": {
                             "code": "{{ preview_language|lower }}",
--- a/cms/templates/admin/cms/page/tree/menu.html
+++ b/cms/templates/admin/cms/page/tree/menu.html
@@ -1,4 +1,4 @@
-{% load i18n cms_admin admin_urls %}
+{% load i18n cms_admin admin_urls l10n %}
 
 {# INFO: columns are defined in base.html options #}
 {% spaceless %}
@@ -17,8 +17,8 @@
     {% endblock %}
     "
     {% if is_popup %}onclick="opener.dismissRelatedLookupPopup(window, {{ page.id }}); return false;"{% endif %}
-    data-id="{{ page.pk }}"
-    data-node-id="{{ node.pk }}"
+    data-id="{{ page.pk|unlocalize }}"
+    data-node-id="{{ node.pk|unlocalize }}"
     data-slug="{{ page.get_slug }}"
     data-is-home="{{ page.is_home|yesno:"true,false" }}"
     data-move-permission="{{ has_move_page_permission|yesno:"true,false" }}"
@@ -198,7 +198,7 @@
                     <div{% if has_add_page_permission %} class="cms-hover-tooltip cms-hover-tooltip-left cms-hover-tooltip-delay"
                         data-cms-tooltip="{% autoescape on %}{% trans 'New sub page' %}{% endautoescape %}"{% endif %}>
                         {% if has_add_page_permission %}
-                            <a href="{% url opts|admin_urlname:'add' %}?parent_node={{ node.pk }}"
+                            <a href="{% url opts|admin_urlname:'add' %}?parent_node={{ node.pk|unlocalize }}"
                                 class="js-cms-pagetree-add-page cms-btn cms-btn-default cms-icon cms-icon-plus">
                         {% else %}
                             <span class="cms-btn cms-btn-default cms-btn-disabled cms-icon cms-icon-plus">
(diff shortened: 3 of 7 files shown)


vinitkumar:
@mbi Thanks for the fix, Marco. Need to look into this fix with a bit more care.

Also, please don't mix the fixes with the formatting and import changes, it hides the real fix and makes reviewing the code harder.

mbi:
> Also, please don't mix the fixes with the formatting and import changes, it hides the real fix and makes reviewing the code harder.

Apologies about that, my editor's isort settings apparently didn't pick up the project settings.

mbi:
@marksweb this PR is a few months old, but the original bug (#7175), even if it was closed by the stalebot, is still very relevant. Would you mind taking a second look? :pray:

marksweb:
@mbi yeah sure.

Stalebot is in use to help us keep on top of issues.

When it updates things, we can see the important issues because people like yourself let us know. We had hundreds of open issues and this is helping to manage them.

marksweb:
So looking at this new test, I don't think we can really cope with test workflows needing an hour+ to run. Quick fixes won't be possible anymore.

Obviously the test has merit, but we need a way to isolate it and then we can mark it as optional for checks to pass.

@fsbraun would be good to get your thoughts on this PR change & any potential solutions to what I'm thinking here.

fsbraun:
@mbi @marksweb I have two thoughts:

1. Can we **set the auto increment** for the id in the database to 1000, e.g. by issuing an sql statement in the test? (e.g., `ALTER TABLE cms_page AUTO_INCREMENT=1000;`)
2. Is it possible for a **linter** to search through all template files to look for `{{ something.pk }}` and `{{ something.id }}` not followed by `| unlocalize` or `| safe`? We'd be able to skip the test and we would be able to transfer this to all packages.

Finally, I am wondering if `| safe` is better than `| unlocalize` since it does not require the `l10n` template tags to be loaded. Or is that considered a hack? (It clearly does the trick.)

cms/admin/pageadmin.py
```diff
@@ -1401,8 +1402,12 @@ def get_tree(self, request):
         """
         site = self.get_site(request)
         pages = self.get_queryset(request)
-        node_id = request.GET.get('nodeId')
-        open_nodes = list(map(int, request.GET.getlist('openNodes[]')))
+        node_id = re.sub(r'[^\d]', '', request.GET.get('nodeId', '')) or None
```
fsbraun, line 1405: Sorry, I am totally not a regex person. What does this do and why is it important?

cms/templates/admin/cms/page/tree/menu.html
```diff
@@ -17,8 +17,8 @@
     {% endblock %}
     "
     {% if is_popup %}onclick="opener.dismissRelatedLookupPopup(window, {{ page.id }}); return false;"{% endif %}
-    data-id="{{ page.pk }}"
-    data-node-id="{{ node.pk }}"
+    data-id="{{ page.pk|unlocalize }}"
```
fsbraun, line 20: I would love to get other peoples opinion about `unlocalize` vs. `safe`

cms/tests/test_page_admin.py
```diff
@@ -1912,6 +1912,54 @@ def test_clear_placeholder_marks_page_as_dirty(self):
             self.assertEqual(page.reload().get_publisher_state("en"), PUBLISHER_STATE_DIRTY)
 
 
+    @override_settings(USE_THOUSAND_SEPARATOR=True, USE_L10N=True)
+    def test_page_tree_render_localized_page_ids(self):
+        admin_user = self.get_superuser()
+        root = create_page("home", "nav_playground.html", "fr",
```
fsbraun, line 1918: Potentially something like

    Page.objects.raw('ALTER TABLE cms_page AUTO_INCREMENT=1001;')

In stead of loop. I have not tried this, though.

cms/admin/pageadmin.py
mbi: The (possibly naive) idea here is to remove all non-digit characters from the passed data. So this would remove all the possible values for [THOUSAND_SEPARATOR](https://docs.djangoproject.com/en/4.1/ref/settings/#std-setting-THOUSAND_SEPARATOR) that Django could pass in the request and just leave the raw integer value.

cms/templates/admin/cms/page/tree/menu.html
mbi: AFAICS they have a completely different scope and the output of either filter is different: `unlocalize` will force Django not to localize the value, i.e. by not applying THOUSAND_SEPARATOR in this case, while `safe` instructs Django that it doesn't need to html-escape the passed variable.  So filtering through `safe` in this case would *not* resolve the issue because at best we'd get  `1´000` instead of `1&apos;000` in the back-end, both being unparseable when passed directly to the ORM.

cms/tests/test_page_admin.py
mbi: I will test this. My first impression is that there isn't a db-agnostic SQL syntax to fix the auto-icrement, but I might be wrong.

cms/templates/admin/cms/page/tree/menu.html
fsbraun: Absolutely (that's why it may be called a hack). The process, however, is that first the number is marked as "safe" and with this marker later **not converted** (neither by escaping html nor by number formatting). The `1&apos;000` situation should not happen. However, I am happy to go with the `unlocalized` version as it has more meaning.

cms/tests/test_page_admin.py
fsbraun: https://github.com/django-cms/django-cms/pull/7188/commits/666a88a8449741388668da99287784cf05ced6b2 looks like a practical solution to me. 👍. 

I am also looking at https://djlint.com/docs as a template linter. It would allow to add a rule like

```
- rule:
    name: D034
    message: .pk or .id used without unlocalize
    flags: re.DOTALL|re.I
    patterns:
      - {{\s*[a-zA-Z0-9_.]*.pk\s*}}
      - {{\s*[a-zA-Z0-9_.]*.id\s*}}
```

to find all unlocalized references to database ids. In total I would feel better to add the linter to all our projects to avoid this issue wherever it happens. Then IMHO we could skip the test altogether. Will try with djangocms-frontend. What do you think @marksweb?

fsbraun: Here's my try for djangocms-frontend (https://github.com/django-cms/djangocms-frontend/pull/58). Immediatly found ~5 bugs in the templates!!

fsbraun: I have now run a linter through the cms code base. There are more breaking templates if page numbers exceed 1000. 
The following files were detected:

```
cms/templates/admin/cms/page/tree/actions_dropdown.html
cms/templates/admin/cms/page/tree/base.html
cms/templates/admin/cms/page/tree/menu.html (already part of this PR)
cms/templates/cms/toolbar/dragitem.html
cms/test_utils/project/templates/integration/plugins/no_custom_model.html 
cms/test_utils/project/templates/pluginapp/plugins/no_custom_model.html
cms/test_utils/project/templates/plugins/no_custom_model.html
```

Can we extend the PR to cover at least the non-testing templates (first 4)? The solution would be to `{% load l10n %}` and append `|unlocalize` to the use of `.pk`.

fsbraun: Here's the full lint output:

```
Linting 254/254 files ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 00:00    


cms/templates/admin/cms/page/tree/actions_dropdown.html
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
D034 8:25 .pk or .id should only be used with "|unlocalize" {{ page.pk }}
D034 9:30 .pk or .id should only be used with "|unlocalize" {{ node.pk }}
D034 21:25 .pk or .id should only be used with "|unlocalize" {{ page.pk }}
D034 22:30 .pk or .id should only be used with "|unlocalize" {{ node.pk }}
D034 32:33 .pk or .id should only be used with "|unlocalize" {{ page.pk }}
D034 33:30 .pk or .id should only be used with "|unlocalize" {{ node.pk }}

cms/templates/admin/cms/page/tree/base.html
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
D034 111:50 .pk or .id should only be used with "|unlocalize" {{ site.pk }}
D034 111:111 .pk or .id should only be used with "|unlocalize" {{ site.pk }}
D034 130:51 .pk or .id should only be used with "|unlocalize" {{ site.pk }}
D034 198:36 .pk or .id should only be used with "|unlocalize" {{ tree.site.pk }}
D034 262:32 .pk or .id should only be used with "|unlocalize" {{ tree.site.pk }}

cms/templates/admin/cms/page/tree/menu.html
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
D034 19:71 .pk or .id should only be used with "|unlocalize" {{ page.id }}
D034 20:13 .pk or .id should only be used with "|unlocalize" {{ page.pk }}
D034 21:18 .pk or .id should only be used with "|unlocalize" {{ node.pk }}
D034 150:84 .pk or .id should only be used with "|unlocalize" {{ node.pk }}
D034 164:37 .pk or .id should only be used with "|unlocalize" {{ node.pk }}
D034 164:61 .pk or .id should only be used with "|unlocalize" {{ page.pk }}
D034 186:58 .pk or .id should only be used with "|unlocalize" {{ node.pk }}
D034 186:82 .pk or .id should only be used with "|unlocalize" {{ page.pk }}

cms/templates/cms/toolbar/dragitem.html
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
D034 50:76 .pk or .id should only be used with "|unlocalize" {{ plugin.pk }}

cms/test_utils/project/templates/integration/plugins/no_custom_model.html
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
D034 1:3 .pk or .id should only be used with "|unlocalize" {{ instance.pk }}

cms/test_utils/project/templates/pluginapp/plugins/no_custom_model.html
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
D034 1:3 .pk or .id should only be used with "|unlocalize" {{ instance.pk }}

cms/test_utils/project/templates/plugins/no_custom_model.html
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
D034 1:3 .pk or .id should only be used with "|unlocalize" {{ instance.pk }}

Linted 254 files, found 23 errors.

```

mbi: @fsbraun done (in f99ad42423db24482d74f2f39412ee455fa51cb9). I didn't find the last three templates you mentioned in `develop`, though!

fsbraun: @mbi Excellent work! Thank you so much! I believe you've just fixed an extremely difficult to find bug. 🙌

fsbraun (approved):
I really appreciate this PR! Excellent work. Will take the learnings to the tech committee.

cms/templates/admin/cms/page/tree/menu.html
```diff
@@ -76,7 +76,7 @@
         <div class="cms-tree-col">
             <div class="cms-tree-item cms-tree-item-preview
                 {# INFO: highlight active icon when in sidebar mode #}
-                {% if page.id|slugify == request.GET.page_id|slugify %} cms-tree-preview-active{% endif %}">
```
fsbraun, line 79: One more thing: I guess the if statement does not need the `unlocalize`.

cms/templates/admin/cms/page/tree/base.html
```diff
@@ -107,8 +107,8 @@ <h1>
                                     <span class="label">{% trans "Sites" %}</span>
                                 </li>
                                 {% for site in tree.sites %}
-                                    <li{% if site.pk == tree.site.pk %} class="active"{% endif %}>
-                                        <a href="#{{ site.pk }}" class="js-cms-pagetree-site-trigger" data-id="{{ site.pk }}">{{ site.name }}</a>
+                                    <li{% if site.pk|unlocalize == tree.site.pk|unlocalize %} class="active"{% endif %}>
```
fsbraun, line 110: The if statement works without the `unlocalize`. Can you remove it here?

fsbraun (changes requested):
Sorry, two minor changes have come up. Thanks so much @mbi !

cms/templates/admin/cms/page/tree/menu.html
mbi: Done

cms/templates/admin/cms/page/tree/base.html
mbi: ... and done.

fsbraun (approved):

marksweb:
@mbi Sorry, one last thing before I merge - this needs a changelog entry.

PR #7191, opened 2022-01-23
Feature/drop eol python django

marksweb:
## Description

Drop python 3.6 and django 3.0 as they're insecure and no longer supported so we shouldn't release supporting software.

Typically this would trigger a major version change, but I think our situation is a bit unique in that sense. 

## Checklist


marksweb:
Checks here did pass but the result didn't make it back here...

https://github.com/django-cms/django-cms/actions/runs/1737062710

vinitkumar:
@marksweb Indeed and if you notice it is all django3.0 which is removed now, so this unpassed tests are actually something cached in GitHub's view.

PR #7200, opened 2022-01-23
Enabled isort workflow

marksweb:
## Description

The isort test/check was disabled & has been since the move to github actions.

This reenables it and has ran isort so there shouldn't be any errors like there was with this check earlier.

## Related resources

https://github.com/django-cms/django-cms/runs/4913405002?check_suite_focus=true

## Checklist


vinitkumar:
@marksweb Same issue here. All tests pass, but the view here showing those cached ones.

PR #7208, opened 2022-01-26
Upgrade Gulp and Nodejs

vinitkumar:
This upgrade some core dependencies that were blocking the app to be built during release.

- Node.js (16.x)
- Gulp (4.x)


marksweb (approved):
This is incredible progress @vinitkumar!

What's next for this?

I think an `.nvmrc` file might be useful in the root so that if someone is using nvm they can activate the correct version.
https://github.com/nvm-sh/nvm#nvmrc

marksweb:
@vinitkumar The "expected" test is hanging for the same reason as the django 3.0 tests.

And we can't control the name of that so we just need to disable the branch protection for that check when we change versions like this.

vinitkumar:
@marksweb 

> This is incredible progress @vinitkumar!
> 
> What's next for this?
> 
> I think an `.nvmrc` file might be useful in the root so that if someone is using nvm they can activate the correct version. https://github.com/nvm-sh/nvm#nvmrc

I think the next step should be upgrade karma since it is not longer officially supported. However, it is an high effort low return task (for now), but we should do it soon. We can also look into upgrading webpack itself. This again is something we can do in near future and is not an urgent issue to tackle now.

marksweb (approved):

PR #7221, opened 2022-02-08
feat: Re-enable showing the toolbar to anonymous users

Aiky30:
Allowing anonymous users added as part of the following commit which was removed from v4, I suspect as part of a cleanup of the toolbar show / hide logic: https://github.com/django-cms/django-cms/commit/9aa11aebbf3f887fdc292fe31165023c0c247ad1

Provides the ability to enable the toolbar when not logged in. The use case for this is, you’re on a web page, you want to login but the only way currently is to type the login url directly or “/admin”, the side affect is that when you are redirected back when logging in you are redirected to the “/admin” url rather than the page you were on. 

Related documentation: https://hackmd.io/5Sj6X5XhTJOmZgNj8e_KCw?view#Configuration

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,6 +1,9 @@
 
 === 4.0.0 (unreleased) ===
-
+* feat: Re-enable showing the toolbar to anonymous users
+* Backport django 3.2 support
+* Backport django 3.1 support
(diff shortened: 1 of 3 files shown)


marksweb (approved):
Looks good 👍

Issue #7216, opened 2022-02-05
[BUG] djangoCMS 3.10.0RC1 structure tree doesn't open


crydotsnake:
## Description
The structure tree in djangoCMS 3.10.0RC1 doesn't open anymore.

## Steps to reproduce

- Install djangoCMS 3.10.0rc1
- Login to the djangoCMS backend and try to open the structure tree

![djangocms3 10 0rc1](https://user-images.githubusercontent.com/39345336/152661867-b021fd9c-99e5-4f5f-8f37-487b69a3c8c0.gif)

## Expected behaviour
The structure tree should open and list all the plugins.

## Actual behaviour
The structure tree doesn't open. (see the GIF above)

## Additional information (CMS/Python/Django versions)
djangoCMS 3.10.RC1
Python 3.9.7
Django: 2.2

fsbraun:
If I try, the structure tree appears if I reduce the window size beyond some breakpoint. CSS issue?

goutnet:
@vinitkumar @marksweb  this might be related to the new build system?  can you help check?

Does this replicates also on `develop` ?

crydotsnake:
@fsbraun and I can confirm that the toolbar becomes visible when you minimize the browser window to a certain breakpoint. So it might be a CSS issue here.

![djangocms-toolbar-resize-window](https://user-images.githubusercontent.com/39345336/152699381-872e78f8-7cc5-42a7-89d0-685ef69bc9a6.gif)

fsbraun:
@crydotsnake @goutnet A little inspection shows that on a wide screen the class `.cms-structure` gets annotated by an inline style `right: -1031px;` (the particular number seems to vary but is large negative) which makes the structure view disappear to the left:

    <div class="cms-structure cms-structure-condensed" style="right: -1031px; display: block;">
        ...
    </div>

In 3.9. this reads `right: -15px` which seems fine.

I could image that there is a JS routine at work where the inline style is calculated which for whatever reason has gone stale.

MacLake:
The breakpoint is 2^10. If the viewport width is 1024 px or less, the structure board becomes visible and takes the whole width instead of just a part.

Btw.: Since 3.5.3 I’ve been encountering a strange bug when opening the structure board, most likely it’s not related, but who knows?
https://github.com/django-cms/django-cms/issues/6600

fsbraun:
[...]

PR #7223, opened 2022-02-09
Fix:		toolbar bug 3.10.rc1

fsbraun:
## Description

Fixes issue #7216

## Related resources

Implements https://github.com/django-cms/django-cms/issues/7216#issuecomment-1032279809

## Checklist


cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -736,7 +736,7 @@ var Toolbar = new Class({
     _refreshMarkup: function(newToolbar) {
         const switcher = this.ui.toolbarSwitcher.detach();
 
-        $(this.ui.toolbar[0]).replaceWith(newToolbar[0]);
+        $(this.ui.toolbar).html(newToolbar.children());
```
vinitkumar, line 739: I like this fix, it would awesome if you add some comments about what was wrong and how does your fix resolves that? Easier to review and refer in future.

vinitkumar (changes requested):

cms/static/cms/js/modules/cms.toolbar.js
fsbraun: Thx! Goes back to the analysis I wrote in the issue comments:

There's been a change to `cms.toolbar.js`. Also affected is line 739:

https://github.com/django-cms/django-cms/blob/04e843337d18afe64478d09f00051ed13a49ff55/cms/static/cms/js/modules/cms.toolbar.js#L736-L740

Hypthesis: The DOM is changed without updating the reference in `this.ui.toolbar` leaving the latter pointing to something that is not displayed and hence has the width 0? Also, as far as I understand, `this.ui.toolbar` here belongs to the `Toolbar` class. Besides this shortcut also the similar shortcut in the `StructureBoard` class in `cms.structureboard.js` might have to be updated. 

It seems that before the commit the DOM **was patched and not replaced** potentially leaving the pointers in tact.

Fix implemented for line 739 in `cms.toolbar.js`:

    this.ui.toolbar.html(newToolbar.children());

This leaves the DOM element unchanged, to which the two `this.ui.toolbar` point.

PR #7235, opened 2022-02-12
TASK: Add missing watchFiles task

crydotsnake:
## Description

watchFiles was defined as a function, but has not been added as a gulp task yet. So you couldn't use the gulp watchFiles command before.

## Related resources
Bildschirmfoto 2022-02-12 um 15 22 24

So now you can use the `gulp watchFiles` command. This will track your work, and if something is changed in the js or sass files, it will tell you it. It also shows you if there any errors.


crydotsnake:
I just saw it was already there: `gulp.task('watch', gulp.parallel(watchFiles));` so i will close this PR...

PR #7238, opened 2022-02-14
New feature: Dark mode for toolbar, page tree, structure tree, modals, login and wizard

fsbraun:
## Description

Since Django 3.2 the admin interface respects "dark mode", i.e. if selected by the user, the admin shows on dark background instead if light background. Dark mode is often used by IT people. Currently, django CMS does not respect a selected dark mode leading to a admin pages partly on dark background, partly on bright background giving the impression of a broken UI design.

This PR introduces changes to the css so than django CMS can respect dark mode. This affects the toolbar, page tree, structure tree and modals. It does not affect the admin site nor the frontend editing forms.

Changes:

* Replaced the $gray sass variables with css custom properties `--dca-gray` etc. (dca for django CMS association)
* Added different color codes for dark mode for those variables
* Replaced brighten and darken sass functions with css alternatives (mostly transform: `brightness()` since functions do not work on custom properties.
* Removed a few `!important` statements to allow other apps to properly adjust their admin forms to dark mode.
* Replaced a few left-over absolute black and white properties with $back  and $white to become flexible.
* Reflected @marksweb’s suggestion to have the django CMS color adjusted in dark mode to be better for the eye

## Related resources

There is a similar PR for `djangocms-admin-style`. With both accepted django CMS will give a nice dark mode design: [PR 443](https://github.com/django-cms/djangocms-admin-style/pull/443)

## Checklist

No tests are affected. 


marksweb (changes requested):
@fsbraun your previous commit for the toolbar bug contained modifications to the 3.9 static files which has already been released so we don't want those files changed.

I've made changes to the develop branch which removed those changes while maintaining your fix.

So we need these changes isolating from the changes to the 3.9 bundle.

I'd suggest you create a new branch from upstream develop and cherry pick the changes needed for dark mode. Then delete your develop branch locally and on your form origin. Then you can create a new develop branch from upstream and raise the PR from the cherry picked branch.

fsbraun:
Technically, does it help if I revert the static files in this PR to the version of Florian's commit when releasing 3.9 (https://github.com/django-cms/django-cms/commit/abf8e80ac6da3cb8261b515a2219bfc6bffa574a) or remove them (`.map`)?

marksweb:
> Thx, @marksweb ! Just for future understanding: Those static files get crated when testing but I am supposed to **not** commit those changes to the static files? The commits should only contain the changes in the source folders (`js`, `sass`, etc.)?

Yeah that's right. Sometimes we need to generate them to develop something, but they are generated by the release process so we don't modify them in the main repo.

As far as reverting them goes, I tried that a bunch of different times with the original toolbar fix but nothing seemed to work. I ended up having to reset & force push.

PR #7241, opened 2022-02-15
ci: Some tests misusing assertTrue for comparisons fix

### LINKED ISSUE #7240: Some tests misusing assertTrue for comparisons
`assertTrue` is not for comparing arguments, should use `assertEqual` for that.

The developer's intent of the test was to compare argument 1 with argument 2, which is not happening. Really what is happening is the test is passing because first argument is truthy. The correct method to use is assertEqual. [more details](https://codereview.doctor/features/python/best-practice/avoid-misusing-unittest-assert-true)

https://github.com/django-cms/django-cms/blob/04e843337d18afe64478d09f00051ed13a49ff55/cms/tests/test_alias.py#L404
https://github.com/django-cms/django-cms/blob/04e843337d18afe64478d09f00051ed13a49ff55/cms/tests/test_api.py#L202
https://github.com/django-cms/django-cms/blob/04e843337d18afe64478d09f00051ed13a49ff55/cms/tests/test_publisher.py#L731
https://github.com/django-cms/django-cms/blob/04e843337d18afe64478d09f00051ed13a49ff55/cms/tests/test_toolbar.py#L647

I found this issue automatically, see other issues [here](https://codereview.doctor/django-cms/django-cms)

code-review-doctor:
Fixes #7240


creyD (approved):
Looks good so far.

creyD:
@code-review-doctor do you know why the tests fail?

marksweb:
Thanks @fsbraun 👏

PR #7243, opened 2022-02-17
[docs] Removed "version added" notation

marksweb:
## Description

The most recent "version added" notation was from 3.3, release in 2016 so I don't think any of them are necessary anymore. Should someone go back through the versions of the docs, they'll still see these, but now at 3.9, I think everything highlighted is now the norm for most, if not all, developers.

## Checklist


TiredFingers (approved):
Let's try, why not. If somebody will feels pain about it, we can always back things back

crydotsnake (approved):

PR #7268, opened 2022-03-15
Feat/django4 support

vinitkumar:
Add Django 4.0 Support


marksweb (approved):
I've got this running with django 4. So happy enough.

We just need to update corresponding apps to work with django 4 now. Primarily text-ckeditor because the page creation wizard depends on that.

```
  File "/Users/mwalker/apps/django-cms-3/.env/lib/python3.9/site-packages/djangocms_text_ckeditor/models.py", line 2, in <module>
    from django.utils.encoding import force_text
ImportError: cannot import name 'force_text' from 'django.utils.encoding'
```

And there's a couple of other packages that I had installed that also need work for django 4;

```
  File "/Users/mwalker/apps/django-cms-3/.env/lib/python3.9/site-packages/djangocms_transfer/datastructures.py", line 5, in <module>
    from django.utils.encoding import force_text
ImportError: cannot import name 'force_text' from 'django.utils.encoding'
```

```
  File "/Users/mwalker/apps/django-cms-3/.env/lib/python3.9/site-packages/djangocms_attributes_field/fields.py", line 10, in <module>
    from django.utils.translation import ugettext as _
ImportError: cannot import name 'ugettext' from 'django.utils.translation' 
```

fsbraun:
@marksweb I think @jrief has replaced `force_text` by `force_str` in the development branch.

PR #7271, opened 2022-03-22
fix(security): Hide sensible information (#7270)

### LINKED ISSUE #7270: cms/toolbar/toolbar_javascript.html shows python and django version
When we are logged in as CMS users and have access to the cms toolbar, the "toolbar_javascript.html" template source code shows version information of Django, Python, etc. I think this could lead to security issues. A good solution would be to hide this data for production when settings.DEBUG=False

Screenshot from 2022-03-22 12-51-58

django-cms==3.9.0

DmytroLitvinov:
## Description

Hide sensible information about python/django/django cms versions which project uses.

Closes #7270 

## Related resources

* #7270

## Checklist

* [X] I have read the [contribution guidelines ](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

### DIFF
--- a/cms/templates/cms/toolbar/toolbar_javascript.html
+++ b/cms/templates/cms/toolbar/toolbar_javascript.html
@@ -24,7 +24,7 @@ CMS.config = {
         'toolbar': '{% language cms_toolbar.request_language %}{% cms_admin_url "cms_usersettings_get_toolbar" %}{% endlanguage %}'
     },
     'lang': {
-        {% if user.is_authenticated %}
+        {% if debug and user.is_authenticated %}
         'debug': '{% filter escapejs %}{% blocktrans %}Development version using django CMS {{ cms_version }}, Django {{ django_version }}, Python {{ python_version }}{% endblocktrans %}{% endfilter %}',
         {% endif %}
         'cancel': '{% filter escapejs %}{% trans "Cancel" %}{% endfilter %}',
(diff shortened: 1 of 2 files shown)


DmytroLitvinov:
Hi @marksweb ,
Pushed changes :) (confused JS syntax with Django template syntax :sweat_smile: )
It is ready to be reviewed.

cms/templates/cms/toolbar/toolbar_javascript.html
```diff
@@ -24,7 +24,7 @@
         'toolbar': '{% language cms_toolbar.request_language %}{% cms_admin_url "cms_usersettings_get_toolbar" %}{% endlanguage %}'
     },
     'lang': {
-        {% if user.is_authenticated %}
+        {% if debug and user.is_authenticated %}
```
marksweb, line 27: This assumes that `debug` is available in the context.

On my projects it is, because I make sure context processors provide it. But I don't think we can guarantee that, so I think we need something like the following in `BaseToolbar.render_with_structure()`

```

    def render_with_structure(self, context, nodelist):
        self.populate()

        context.update(self.get_render_context())

        with force_language(self.toolbar_language):
            # needed to populate the context with sekizai content
            if 'debug' not in context:
                context['debug'] = settings.DEBUG
            render_to_string('cms/toolbar/toolbar_javascript.html', flatten_context(context))

marksweb (changes requested):

cms/templates/cms/toolbar/toolbar_javascript.html
DmytroLitvinov: Thank you for feedback. I provided that changes to `cms/toolbar/toolbar.py` file.

marksweb (approved):

Issue #7222, opened 2022-02-08
[BUG] Copying a plugin hides the "Publish page changes" and breaks "View published" button


petrklus:
## Description

On a page with pending changes, using the "Copy" plugin function breaks the "View published" button and hides the "Publish changes" button.

## Steps to reproduce

1. Load up a CMS page with unpublished changes
2. Open the structure mode
3. Select "Copy" for any of the plugins
4. Toolbar updates with the buttons as per the issue above

## Expected behaviour

For the "View published" button to continue working and the publication button to not disappear.

## Actual behaviour

"View published" button breaks (wrong URL) and publication button disappears.

## Screenshots

[Screen recording](https://youtu.be/zInXtFXr5pc)

## Additional information (CMS/Python/Django versions)

Django 3.2.12.0, Django CMS 3.9.0.5

fsbraun:
A similar but different issue is found in django CMS v4.0.0rc11. There a the toolbar shows "Edit" despite having the structure board open:
<img width="1157" alt="image" src="https://user-images.githubusercontent.com/16904477/159787350-adeb2802-99de-4fdc-975b-eb3a4024c220.png">

fsbraun:
Fixed in version 3.10

Issue #7113, opened 2021-08-18
[BUG] Publish button disappears after plugin editing


nichoski:
## Description
The publish button blinks after editing plugins on some projects, and completely disappears on others.

## Steps to reproduce
Use django newer than 2.2.19 and any django-cms version newer than September 2014. Create a text plugin. Edit the text plugin. In most cases you will notice the publish button blink. In rare occasions (in some project not so rare, almost always) the publish button disappears and does not return until you refresh the page.

## Expected behaviour
Editing plugins should not make the publish button disappear or blink.

## Technical Description
After editing a plugin, django-cms' frontend makes two requests to the server. One to get the toolbar, and one to get the current page's contents. Using the response from both of them, the toolbar's contents are being replaced. The toolbar contents are replaces twice, I don't know why. The request that fetches the toolbar alone is broken and lacks the publish button in its response. The one that fetches the current page's contents is ok. These requests are executed asynchronously. Almost always, the request that fetches the toolbar finishes first, and the request that fetches the entire page's contents finishes second. This usually hides the problem with the missing publish button from the request that fetches the toolbar. You can notice the publish button blinking after editing a text plugin for example. But on some projects they finish in reversed order, and the page ends up missing the publish button. 

The request that fetches the toolbar contents needs to be fixed. I have a working solution, but I want to avoid breaking other things, so I need your input. I will first explain below what and why is happening.

[...]

PR #7272, opened 2022-03-23
Bugfix v4: Structure mode toggle button disappearing from toolbar

fsbraun:
## Description

When editing with the structure board open any change in the page, e.g. by moving a plugin to a new position, will make the toggle button in the toolbar disappear as @marksweb  reported here: https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1645542305165849

**Please note the empty upper right corner.**

image

## Related resources

* The disappearing of the button happens in two steps: 

  1. The toolbar is reloaded after each editing step (I am not sure this is necessary, btw.) 

  2. If the page content has changed, the page is reloaded using xhr. The reload includes the toolbar.
  
* The first step failed to send the current placeholder as get parameters since the url was encoded twice. As a result the backend returns the toolbar with just the create button and without edit button. This new toolbar is rendered in the browser window.

* When the page reload delivers the correct toolbar, it is used to **patch** the current toolbar. This leads fails to restore the toggle button since there is no button to patch. 

The fix removes the call to `Helpers.updateUrlWithPath` which encodes the already encoded url a second time replacing the get parameter separator `&` by `&amp;`.

**NOTE:** `Helpers.updateUrlWithPath` can also change the url path according to some rules. This is not the case in the call affected (imho).

The double xhr call to the backend is unnecessary and the first call can be dropped. This PR does that by rearranging the code. It leads to better responsiveness of the frontend editing experience.

**This PR should be ported to v3 also.** I expect it to solve issue #7222 .

## Checklist


fsbraun:
**Additional fix in `settingsadmin.py`:**

* If the only the toolbar is loaded by the js, the toolbar settings need to be set for the calling page (supplied in the `cms_page` get parameter). The toolbar class accounts for it and expects that the request object passed on initialization in that case contains a property `resolver_match`. 

* This property was not set in `settingsadmin.py` leading to the toolbar not recognizing it already is in editing mode.

* As a result a copy operation (only operation that leads to this situation) would generate a wrong toolbar clicking upon which would lead not to edit but to structure mode and potentially to a dead endpoint (get instead of post).

* Finally, I have removed three lines for python 2 support.

cms/admin/settingsadmin.py
```diff
@@ -97,6 +97,10 @@ def get_toolbar(self, request):
         request = copy.copy(request)
         request.GET = data
         request.current_page = current_page
+        try:
+            request.resolver_match = resolve(origin_url.path)
+        except Resolver404:
```
Aiky30, line 102: What is the scenario for the 404 being thrown?

There should be a unit test that protects the need for this, it will also expose the scenario where the 404 occurs.

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1070,12 +1068,17 @@ class StructureBoard {
         const placeholderIds = getPlaceholderIds(CMS._plugins).map(id => `placeholders[]=${id}`).join('&');
 
         return $.ajax({
-            url: Helpers.updateUrlWithPath(
+            url: (
```
Aiky30, line 1071: What were you thoughts when removing: `Helpers.updateUrlWithPath`?

cms/admin/settingsadmin.py
fsbraun: Someone could type an url into the browser with a typo in the get parameter `cms_path`.  This will cause the url resolver to fail, say `http://localhost:8000/admin/cms/usersettings/cms-toolbar/?placeholders[]=3&placeholders[]=13&placeholders[]=6&obj_id=7&obj_type=cms.pagecontent&cms_path=%2Fadmin%2Fcms%2Fplaceholder%2Fobject%2F10%2Feat_it%2F7%2F`. The `eat_it` part will not resolve.

Also, code symmetry with https://github.com/django-cms/django-cms/blob/914558d283c197b4035ae7e1a084860f486c9429/cms/toolbar/toolbar.py#L39-L47

Note that line 40/41 prepared for exactly the behavior proposed here.

I think I can add a test for this.

cms/static/cms/js/modules/cms.structureboard.js
fsbraun: `Helpers.updateUrlWithPath` is the root caues of the bug. It replaces query separator`&` by `&amp`s. Since it is used in many places within django CMS, I did not change it. I cannot anticipate the side effects. 

In any case: `Helpers.updateUrlWithPath('/local_url?x=1&y=2')` returns `'/local_url?x=1&amp;y=2&amp;cms_path=%2Fadmin%2Fcms%2Fplaceholder%2Fobject%2F10%2F'` which is not a meanigful path to get the toolbar for. The fix in the same situation creates a meaningful url: `'/local_url?x=1&y=2&cms_path=%2Fadmin%2Fcms%2Fplaceholder%2Fobject%2F10%2F'`

Frankly, I still struggle to comprehend the implementation of `Helpers.makeUrl` which is called by `Helpers.updateUrlWithPath`: https://github.com/django-cms/django-cms/blob/914558d283c197b4035ae7e1a084860f486c9429/cms/static/cms/js/modules/cms.base.js#L291-L308

fsbraun: An alternative way would be to drop the `urljs` dependence and use the browsers' `URLSearchParams` interface. This would allow to have `Helpers.updateUrlWithPath` to be present. It should build correct urls. But it also requires a small change in `cms.pagetree.js`.  Overall this is probably cleaner but a be it small change deep in the code.

If you think that is better, I'll give it a go!

fsbraun: After more pondering I come to the conclusion to essentially remove lines 303-307 in `makeURL`:  It seems to never have worked right with more than one query parameter. Overall, this deems to be the most systematic solution. 

I still have to make the discrimination test for `Resolver404` work.

fsbraun:
@Aiky30 The resolver test ist implemented and works. 

I have refactored the patch to change the underlying `Helpers.makeUrl`. I now believe this has never worked correctly. Also the test were testing for **incorrect** behavior? I have not found a single browser that accepts `&amp;` as a search parameter delimiter in js. (It seems ok for html.)

Test action fails when testing the documents build (which incidentally works in the separate automation). The reason seems to be sphinx 1.8 importing incorrectly from jinja2. Certainly nothing to do with this PR. I suggest to remove the fix to Sphinx==1.8.x in the requirements and backport the changes https://github.com/django-cms/django-cms/commit/f3285b659eb5094c3a1e318e8c520aae63555c9a to the documentation (`highlight` etc.)

fsbraun:
@Aiky30 What are your thoughts on getting this merged?

Aiky30:
@fsbraun I haven't forgotten about your changes here. The makeUrl function replacing & with &amp; appears to have been a quick and dirty workaround here: https://github.com/django-cms/django-cms/issues/3404. I guess you may already know this. 

Once I understand the reason for it's existence I'll be able to approve your rework. My fear right now is any regression, especially if this is used in other parts of the CMS / external packages.

Aiky30:
A frontend test that replicates this bug: https://github.com/django-cms/django-cms/issues/3434

Then we can see with these changes if that issue has been reintroduced.

fsbraun:
I understand your concerns. The "quick and dirty workaround" from 2014 does not work in the first place, it does break _every_ url with more than one parameter as of Django 2.2.19. See nichoski's discussion in https://github.com/django-cms/django-cms/issues/7113. In Django < 2.2.19 things only work _accidentally_.

The underlying jQuery bug has been fixed. I did not research the exact version but tested it with CMS' jQuery verision. 

To mitigate your concerns please be aware that this has also been corrected and in the field with v3.10 (PR #7114). 

Finally, also "Add page type" works in 3.10 with the proposed fix. In v4 page types are gone, hence nothing to test there.

Aiky30:
> I understand your concerns. The "quick and dirty workaround" from 2014 does not work in the first place, it does break _every_ url with more than one parameter as of Django 2.2.19. See nichoski's discussion in #7113. In Django < 2.2.19 things only work _accidentally_.
> 
> The underlying jQuery bug has been fixed. I did not research the exact version but tested it with CMS' jQuery verision.
> 
> To mitigate your concerns please be aware that this has also been corrected and in the field with v3.10 (PR #7114).
> 
> Finally, also "Add page type" works in 3.10 with the proposed fix. In v4 page types are gone, hence nothing to test there.

@fsbraun I could see the change was made to django-cms 3.10 and I guess as you mentioned the issue was in an older release of django. PageType is still in v4 but it just hasn't been configured in to use the edit, preview and structure endpoints.

Aiky30:
After seeing the various tickets listing bugs and fixes I'm happy to allow this reverse of the fix.

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -1074,7 +1072,7 @@ class StructureBoard {
                 `${CMS.config.request.toolbar}?` +
                     placeholderIds +
                     '&' +
-                    `obj_id=${CMS.config.request.pk}&` +
+                     `obj_id=${CMS.config.request.pk}&` +
```
Aiky30, line 1075:
```suggestion
                    `obj_id=${CMS.config.request.pk}&` +
```
Some un related formatting changes that should be reversed.

cms/static/cms/js/modules/cms.toolbar.js
```diff
@@ -725,7 +725,6 @@ var Toolbar = new Class({
         const diff = dd.diff(this.ui.toolbar[0], newToolbar[0]);
 
         dd.apply(this.ui.toolbar[0], diff);
```
Aiky30, line 727:
```suggestion
        dd.apply(this.ui.toolbar[0], diff);

```

Unrelated formatting changes added.

cms/static/cms/js/modules/cms.base.js
```diff
@@ -299,12 +299,7 @@ export const Helpers = {
         });
 
         return newUrl
-            .toString()
-            .split('#')
-            .map((part, i) => {
-                return i === 0 ? part.replace(/&/g, '&amp;') : part;
-            })
-            .join('#');
+            .toString();
```
Aiky30, line 302: The function documentation needs to be updated to suit the code, it no longer does as this describes:

    /**
     * Modifies the url with new params and sanitises
     * the ampersand within the url for #3404.
     *
     * @method makeURL
     * @param {String} url original url
     * @param {Array[]} [params] array of [`param`, `value`] arrays to update the url
     * @returns {String}
     */

Aiky30: Possibly:

/**
 * Modifies the url with new params and sanitises the url.
 *
 * @method makeURL
 * @param {String} url original url
 * @param {Array[]} [params] array of [`param`, `value`] arrays to update the url
 * @returns {String}
 */

cms/tests/test_toolbar.py
```diff
@@ -200,6 +200,18 @@ def test_toolbar_request_endpoint_validation(self):
...
+            response = self.client.get(
+                endpoint,
+                data={
+                    'obj_id': page_content.pk,
+                    'obj_type': 'cms.pagecontent',
+                    'cms_path': get_object_edit_url(page_content)+"q"  # Invalid
+                },
+            )
+            self.assertEqual(response.status_code, 200)
+            self.assertNotContains(response, "Clipboard")  # No clipboard exposed to invalid cms_path
```
Aiky30, line 214:
```suggestion
            # No clipboard exposed to invalid cms_path
            self.assertNotContains(response, "Clipboard")
```

It's not great having comments appended at the end of code that is not say settings, they are easily missed and having them on their own line helps to keep it consistent with the rest of the codebase.

Aiky30 (changes requested):
Minor formatting changes required, the only one I can't use GH suggestions for I have proposed a comment.

cms/static/cms/js/modules/cms.base.js
fsbraun: True! I've also added that any passed `&amp;` are converted to `&`

fsbraun:
Thanks very much for taking the time. The plan sounds good!! Fabian

PR #7289, opened 2022-03-31
Fixes #7288 by also catching AttributeError, when the current toolbar…

### LINKED ISSUE #7288: [BUG] CMS Toolbar wrongly assumes the current object has get_draft_url()
## Description

To add edit link items to the CMS toolbar for the currently displayed item, we're assigning the current detail object to the toolbar, the usual way:

```python 
    if hasattr(request, "toolbar"):
        request.toolbar.set_object(article)
```

Since django-cms 3.10, this now loads to an error:

```
AttributeError: 'Article' object has no attribute 'get_draft_url'
```

The issue is triggered [here](https://github.com/django-cms/django-cms/blob/develop/cms/toolbar/toolbar.py#L329), essentially `get_object_draft_url` assumes that the current toolbar object has a `get_draft_url()`method, and only catches `NoReverseMatch` but fails to handle the catch `AttributeError` as well, when no such method is defined.

## Steps to reproduce

1. Assign an object to the current toolbar with `request.toolbar.set_object(current_object)`

## Expected behaviour

No error raised

## Actual behaviour

```
AttributeError: 'MyModel' object has no attribute 'get_draft_url'
```

This is new in DjangoCMS 3.10, because prior to #7104 `get_draft_url` would be a bare `catch` and would thus also include `AttributeError`. Since 3.10 includes this change, this only happens in 3.10

mbi:
… object doesn't define get_draft_url()

Fixes https://github.com/django-cms/django-cms/issues/7288

## Description

See description in #7288: `get_object_draft_url` assumes that the current toolbar object has a `get_draft_url()` method, and only catches `NoReverseMatch` but fails to handle  `AttributeError` as well, when no such method is defined.

## Related resources

Discussion on Slack: https://django-cmsworkspace.slack.com/archives/C01EBCEK49E/p1648718708538749

## Checklist

* [ ] I have opened this pull request against ``develop`` (Opened against release/3.10.x as requested)

### DIFF
--- a/cms/toolbar/toolbar.py
+++ b/cms/toolbar/toolbar.py
@@ -326,7 +326,7 @@ class CMSToolbar(BaseToolbar):
             with force_language(self.request_language):
                 try:
                     return self.obj.get_draft_url()
-                except NoReverseMatch:
+                except (NoReverseMatch, AttributeError):
                     try:
                         return self.obj.get_absolute_url()
                     except NoReverseMatch:
--- a/cms/toolbar/toolbar.py
+++ b/cms/toolbar/toolbar.py
@@ -329,7 +329,7 @@ class CMSToolbar(BaseToolbar):
                 except (NoReverseMatch, AttributeError):
                     try:
                         return self.obj.get_absolute_url()
-                    except NoReverseMatch:
+                    except (NoReverseMatch, AttributeError):
                         pass
         return ''
 


cms/toolbar/toolbar.py
```diff
@@ -326,7 +326,7 @@ def get_object_draft_url(self):
             with force_language(self.request_language):
                 try:
                     return self.obj.get_draft_url()
-                except NoReverseMatch:
+                except (NoReverseMatch, AttributeError):
                     try:
                         return self.obj.get_absolute_url()
                     except NoReverseMatch:
```
marksweb, line 332: I think it's worth catching the `AttributeError` here also while we're making this change. Models might not have a `get_absolute_url` either.

marksweb (changes requested):

cms/toolbar/toolbar.py
mbi: I was wondering about... will do.

marksweb (approved):

PR #7292, opened 2022-04-01
ci: Add codespell to pre-commit and setup.cfg

cclauss:
## Description

Add [codespell](https://pypi.org/project/codespell) to pre-commit and setup.cfg and fix typos along the way.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,7 +3,7 @@ universal=1
 
 [codespell]
 ignore-words-list = cant,dne,statics
-skip = skip = *.js,*.po,./node_modules/*,./docs/env/*,./docs/build/*,./.env/*,./.venv/*,./build/*,./django_cms.egg-info/*
+skip = *.js,*.po,./node_modules/*,./docs/env/*,./docs/build/*,./.env/*,./.venv/*,./build/*,./django_cms.egg-info/*
 
 [flake8]
 ignore=E122,E123,E124,E125,E126,E127,E128,E129,E131,E201,E202,E203,E221,E225,E226,E231,E241,E251,E261,E262,E265,E271,E301,E302,E303,E501,E502,E701,E713,E714,E731,F401,F403
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,10 @@
 [bdist_wheel]
 universal=1
 
+[codespell]
+ignore-words-list = cant,dne,statics
+skip = cms/forms/widgets.py,*.js,*.po
+
 [flake8]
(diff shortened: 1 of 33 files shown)


cms/forms/widgets.py
```diff
@@ -78,14 +78,14 @@ def _build_widgets(self):
 
     def _build_script(self, name, value, attrs={}):
         return r"""<script type="text/javascript">
-                var CMS = window.CMS || {};
+                var CMS = window.CMS || {{}};
```
cclauss, line 81: ???  Is this change OK?

marksweb: @cclauss I don't think so. I think that's being interpreted as a python string when infact it's used by javascript where I don't think that's valid. Though I'm not sure how to get pyupgrade to skip it.

When I implemented the pre-commit config I didn't run it on all files, because that tends to make people nervous when a big change comes in. 

I'll take a look over the rest of these changes.

cclauss: After further review… Actually `pyupgrade` did the right thing and this syntax works as expected.

vinitkumar: @cclauss @marksweb This can't possibly be correct. This is an invalid syntax and all browsers crash on this. 

Please revert back all the changes that turns `{}` to `{{}}``

Screenshot 2022-04-02 at 3 29 23 PM

vinitkumar: Also, I feel pyupgrade shouldn't touch JS strings.

vinitkumar (changes requested):

cclauss:
Perhaps the JavaScript should go in .js files that are read by the Python code.

marksweb:
@cclauss yes I couldn't agree more. I was thinking the same thing when I was looking at this again last night.

Or even html files that contain the current string with placeholders for the necessary context variables. The widget can then render the template to a string so that it's still returning the same thing.

cclauss:
OK.  I have reverted all changes to `cms/forms/widgets.py` and requested (in `setup.cfg`) that codespell ignore that file for now.  We can attack `cms/forms/widgets.py` in a separate PR.

> html files that contain the current string with placeholders for the necessary context variables

I am unclear how to implement this so a short example would help.

setup.cfg
```diff
@@ -1,6 +1,10 @@
 [bdist_wheel]
 universal=1
 
+[codespell]
+ignore-words-list = cant,dne,statics
+skip = cms/forms/widgets.py,*.js,*.po
```
cclauss, line 6: `cms/forms/widgets.py,` can be removed after #7300 lands.

vinitkumar: @cclauss #7300 is merged now. Please go ahead and make this change.

marksweb (approved):
Alright, I've been over this now I think - took a while.

I think I'm pretty happy with where it is. Just not sure how we stop pre-commit changing those js strings again.

PR #7297, opened 2022-04-05
Add cache ttl extension point

### LINKED ISSUE #7296: [BUG] Can't change the cache ttl per page using business logic
## Description

Can't change the cache ttl per page using business logic.
On [Richie](https://github.com/openfun/richie), CMS for Open Education built on top of Django CMS, we need to limit the cache ttl of each page given a business rule. 
Example: after a time has expired the page presentation should change. So it is more efficient to limit the cache ttl for that page during its generation, so on the next request the page would be expired and the DjangoCMS would generate a new one.

It is related with this issue openfun/richie#1663 on Richie.

## Steps to reproduce

It is not easy to reproduce, to make it work it is need some timebased logic on the template logic.
In resume, we have a course page with image, description and videos, each course has multiple course runs (editions) each run has multiple dates, like when it is going to start and end. When its start passes, the page change its presentation.

## Expected behaviour

To exist an extension point that DjangoCMS calls during `set_page_cache`, where it could be added business logic that limit the ttl for that page/request.
So a new value could be added that would be used by the `min` other ttl's:
https://github.com/django-cms/django-cms/blob/develop/cms/cache/page.py#L68-L73

## Actual behaviour

Using a page cache for example of 1 day, the business logic based on date/time should change the presentation of the page, but the user still sees the old page content.
When we have found this issue, temporary we have republish the page to force the DjangoCMS to clear that page cache.

## Screenshots
No screenshots.

Any version.

* [X] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 

igobranco:
Adds the setting `CMS_CACHE_LIMIT_TTL_MODULE` that should have the
`limit_page_cache_ttl` function that would be called to limit the cache
ttl of a page using business logic.
Closes #7296

## Related resources

* #...
* #...

## Checklist


marksweb:
Hi @igobranco I like the sound of this 👌

Unfortunately I turned on the pre-commit bot earlier and were not ready for that. Its committing a run against all files in PRs so I've disabled it.

Could you please revert that pre-commit bots commit (or cherry pick your commit to a new branch and open a new PR) so that we can review without all the extra changes.

igobranco:
Ok @marksweb
Opened a replacement PR #7299

PR #7299, opened 2022-04-06
feat: add cache ttl extension point

### LINKED ISSUE #7296: [BUG] Can't change the cache ttl per page using business logic

igobranco:
## Description

Adds the setting `CMS_CACHE_LIMIT_TTL_MODULE` that should have the
`limit_page_cache_ttl` function that would be called to limit the cache
ttl of a page using business logic.
Closes #7296

## Related resources

* the issue #7296
* previous PR #7297

## Checklist

It will need a test so the interface of the extension point is preserved for the future.
If merged, this new feature would need to go to the docs.
The idea of this PR is to complement the issue.

### DIFF
--- a/cms/cache/page.py
+++ b/cms/cache/page.py
@@ -1,6 +1,7 @@
 
 import hashlib
 from datetime import timedelta
+from importlib import import_module
 
 from django.conf import settings
 from django.utils.cache import (
@@ -45,7 +46,7 @@ def set_page_cache(response):
 
     placeholders = toolbar.content_renderer.get_rendered_placeholders()
     # Checks if there's a plugin using the legacy "cache = False"
-    placeholder_ttl_list = []
+    ttl_list = []
     vary_cache_on_set = set()
     for ph in placeholders:
         # get_cache_expiration() always returns:
@@ -53,22 +54,29 @@ def set_page_cache(response):
         ttl = ph.get_cache_expiration(request, timestamp)
         vary_cache_on = ph.get_vary_cache_on(request)
 
-        placeholder_ttl_list.append(ttl)
+        ttl_list.append(ttl)
         if ttl and vary_cache_on:
             # We're only interested in vary headers if they come from
             # a cache-able placeholder.
             vary_cache_on_set |= set(vary_cache_on)
 
-    if EXPIRE_NOW not in placeholder_ttl_list:
-        if placeholder_ttl_list:
-            min_placeholder_ttl = min(x for x in placeholder_ttl_list)
-        else:
-            # Should only happen when there are no placeholders at all
-            min_placeholder_ttl = MAX_EXPIRATION_TTL
-        ttl = min(
-            get_cms_setting('CACHE_DURATIONS')['content'],
-            min_placeholder_ttl
-        )
+    if EXPIRE_NOW not in ttl_list:
+        ttl_list.append(get_cms_setting('CACHE_DURATIONS')['content'])
+        ttl_list.append(MAX_EXPIRATION_TTL)
+
+        if hasattr(settings, 'CMS_LIMIT_TTL_CACHE_FUNCTION'):
+            extension_point = settings.CMS_LIMIT_TTL_CACHE_FUNCTION
+
+            module, func_name = extension_point.rsplit('.', 1)
+            module = import_module(module)
+            limit_ttl_cache_function = getattr(module, func_name)
+            limit_ttl = limit_ttl_cache_function(response)
+
+            # if the extension point returns an integer as ttl
+            if isinstance(limit_ttl, int):
+                ttl_list.append(limit_ttl)
+
+        ttl = min(ttl_list)
 
         if ttl > 0:
             # Adds expiration, etc. to headers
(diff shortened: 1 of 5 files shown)


marksweb:
Thank you @igobranco 👍

cms/cache/page.py
```diff
@@ -45,30 +45,37 @@ def set_page_cache(response):
...
-        ttl = min(
-            get_cms_setting('CACHE_DURATIONS')['content'],
-            min_placeholder_ttl
-        )
+    if EXPIRE_NOW not in ttl_list:
+        ttl_list.append(get_cms_setting('CACHE_DURATIONS')['content'])
+        ttl_list.append(MAX_EXPIRATION_TTL)
+
+        # If defined the `CMS_CACHE_LIMIT_TTL_MODULE` extension point
+        if hasattr(settings, 'CMS_CACHE_LIMIT_TTL_MODULE'):
```
marksweb, line 67: From what I can see, `CMS_CACHE_LIMIT_TTL_MODULE` is new in this PR.

Therefore we'll need some documentation to show what this is expected to be, and how to use it.

I think some detail in here; https://docs.django-cms.org/en/latest/how_to/caching.html?highlight=cache

And also something more brief in the advanced settings; https://docs.django-cms.org/en/latest/reference/configuration.html?highlight=settings#advanced-settings

marksweb (changes requested):

cms/cache/page.py
igobranco: I've included the requested documentation.

igobranco:
@marksweb I've renamed the setting to `CMS_CACHE_LIMIT_TTL_CLASS`; included the requested documentation and added new tests. I think with this changes it is ready to a new review.

docs/reference/configuration.rst
```diff
@@ -990,6 +990,17 @@ Default value of the ``cache`` attribute of plugins. Should plugins be cached by
     If you disable the plugin cache be sure to restart the server and clear the cache afterwards.
 
 
+..  setting:: CMS_CACHE_LIMIT_TTL_CLASS
+
+CMS_CACHE_LIMIT_TTL_CLASS
+================
```
marksweb, line 996: I think this line needs to match the length of the title to avoid a syntax error.

marksweb (changes requested):

docs/reference/configuration.rst
igobranco: Fixed.

cms/cache/page.py
```diff
@@ -45,30 +45,40 @@ def set_page_cache(response):
...
+
+        # If defined the `CMS_CACHE_LIMIT_TTL_CLASS` extension point
+        if hasattr(settings, 'CMS_CACHE_LIMIT_TTL_CLASS'):
+            # then loads the class
+            class_name = settings.CMS_CACHE_LIMIT_TTL_CLASS
+
+            import importlib
+            module_name, class_name = class_name.rsplit(".", 1)
+            LimitPageCacheTTL = getattr(importlib.import_module(module_name), class_name)
+            instance = LimitPageCacheTTL()
```
sampaccoud, line 74: Why not directly point the setting to a function?

marksweb: Yeah this might be a bit neater as a utility function and it might avoid the local import that way.

igobranco: @sampaccoud  and @marksweb  I've reviewed the PR. Change the extension point to reference a function instead of a class. I've also renamed the extension point has: `CMS_LIMIT_TTL_CACHE_FUNCTION`

igobranco:
@marksweb can you review the PR?

igobranco:
> One thing I can see though is it needs an entry in the unreleased section of the changelog

Added!

marksweb (approved):

PR #7302, opened 2022-04-15
Fix ``get_object`` call which was missing request

marksweb:
## Description

In `get_content_obj` the first call is to `get_object` is missing an argument.

```
    def edit_title_fields(self, request, page_id, language):
        page = self.get_object(page_id)
```

But `get_object` takes the request as the first argument and then the object ID as the second.

## Related resources

* Django source; https://github.com/django/django/blame/main/django/contrib/admin/options.py#L851

## Checklist

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -714,7 +714,7 @@ class PageAdmin(admin.ModelAdmin):
         return HttpResponse(json.dumps({"id": new_page.pk}), content_type='application/json')
 
     def edit_title_fields(self, request, page_id, language):
-        page = self.get_object(request, page_id)
+        page = self.get_object(request, object_id=page_id)
         translation = page.get_title_obj(language, fallback=False)
 
         if not self.has_change_permission(request, obj=page):
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -705,16 +705,16 @@ class PageAdmin(admin.ModelAdmin):
         page_languages = page.get_languages()
         site_languages = get_language_list(site_id=site.pk)
 
-        if not any(lang in  page_languages for lang in site_languages):
+        if not any(lang in page_languages for lang in site_languages):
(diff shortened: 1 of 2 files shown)


cms/admin/pageadmin.py
```diff
@@ -705,16 +705,16 @@ def copy_page(self, request, page_id):
...
-                                   "translated in any of the languages configured by the target site."))
+                                  "translated in any of the languages configured by the target site."))
             return jsonify_request(HttpResponseBadRequest(message))
 
         new_page = form.copy_page()
         return HttpResponse(json.dumps({"id": new_page.pk}), content_type='application/json')
 
     def edit_title_fields(self, request, page_id, language):
-        page = self.get_object(page_id)
+        page = self.get_object(request, page_id)
```
Aiky30, line 717:
```suggestion
        page = self.get_object(request, object_id=page_id)
```

The above suggestion will keep the fix consistent with all other use cases of get_object on this page.

Aiky30 (changes requested):
Minor amendment required and requires changelog entry.

cms/admin/pageadmin.py
marksweb: @Aiky30 Well spotted!

marksweb:
@Aiky30 Changelog updates are something I really want to get automated. The way github releases generate them is really good.

Aiky30:
> @Aiky30 Changelog updates are something I really want to get automated. The way github releases generate them is really good.

Completely agree, I think this was already implemented in cms 3 develop, need to put in the time to pull it out and into 4.

marksweb:
@Aiky30 Yeah v3 generates the changelog using this script executed from the build script;
https://github.com/django-cms/django-cms/blob/develop/scripts/make-changelog

I really want to break up the build script into the necessary parts to try to build releases via GH actions. As I mentioned last Friday, we've got releases ready to go but we can't run the script.

PR #7321, opened 2022-05-22
build: Release 3.11.0

creyD:
## Related resources

* #...
* #...

## Checklist


cms/__init__.py
```diff
@@ -1,6 +1,6 @@
 import django
 
-__version__ = '3.10.0'
+__version__ = 'test'
```
marksweb, line 3: This is my fault. I've broken the release script where this is bumped 🤦 

Fix for this here; https://github.com/django-cms/django-cms/pull/7322

marksweb (changes requested):
The static files haven't been given the updated version number. I'm going to assume this is because of the bug I caused to bumping the version number.

cms/__init__.py
creyD: Should I run the script again, or would it be enough if I just search and replaced the 3.10.0?

marksweb: @creyD Yeah, my fix is merged so I think we'll close this down, delete the release branch from here, you can do the same from your fork. Then run it again and we should see this version as `3.11.0` and the static files bundle in a folder matching the version.

PR #7325, opened 2022-05-23
Release/3.11.0

creyD:
## Related resources

* #...
* #...

## Checklist


marksweb (changes requested):
This (correctly) removes the 3.10 static files, but doesn't include the new 3.11 static files.

Not entirely sure what's gone wrong. Might be worth checking your console output.

At the point my machine can't run the script anymore (preparing the documentation) I have the new static files;

Screenshot 2022-05-23 at 22 00 24

Just had a thought. It might be done to your node version. We have an `.nvmrc` file in the root, so if you've got nvm installed you can just run `nvm use` before executing the release script.

If you don't have it installed have a look here; https://github.com/nvm-sh/nvm#installing-and-updating

PR #7359, opened 2022-07-11
feat: Add live-url url query parameter to PageContent cms Preview and Edit endpoints

adam-murray:
## Description

A new feature added to v4 Preview and Edit endpoints. Currently there’s no way of knowing what the current pages live url is like you used to be able to with 3.x using ?edit, this new feature (off by default) adds a query string parameter to the preview and edit endpoints.

Without the new feature enabled
Preview endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/preview/372/
Edit endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/edit/372/
Live url: /en/blog/

With the new feature enabled
Preview endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/preview/372/?live-url=/en/blog/
Edit endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/edit/372/?live-url=/en/blog/
Live url: /en/blog/

This PR introduces two settings, CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM_ENABLED and CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM. With the former set True, and a param (as a string) provided with the latter, the preview and edit endpoints for PageContent will append the live-url as a query string parameter.

## Checklist

Transcript of discussions regarding this change:

[Andrew Aikman](https://app.slack.com/team/U01A4M52Z7G)
  [4:46 PM](https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1657640789237889)
Hi all,
A new feature added to v4 Preview and Edit endpoints. Currently there’s no way of knowing what the current pages live url is like you used to be able to with 3.x using ?edit, this new feature (off by default) adds a query string parameter to the preview and edit endpoints.

**Without the new feature enabled**
Preview endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/preview/372/
Edit endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/edit/372/
Live url: /en/blog/

**With the new feature enabled**
Preview endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/preview/372/?live-url=/en/blog/
Edit endpoint: http://127.0.0.1:8000/en/admin/cms/placeholder/object/3/edit/372/?live-url=/en/blog/
Live url: /en/blog/

Please review the change here: https://github.com/django-cms/django-cms/pull/7359 (edited) 

[Mike](https://app.slack.com/team/U01EG6FDEA3)
  [7:23 AM](https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1657693428212949)
that looks like a useful function to me, just to confirm, the full URL is included in the query srting?

[Andrew Aikman](https://app.slack.com/team/U01A4M52Z7G)
  [9:48 AM](https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1657702083961919)
[@Mike](https://django-cmsworkspace.slack.com/team/U01EG6FDEA3) It’s the relative url of the page (returned by the pages get_absolute_url function), it won’t have the protocol or domain name in.
It’s for Page Urls only.

[Mike](https://app.slack.com/team/U01EG6FDEA3)
  [9:54 AM](https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1657702446175209)
ok, so this won’t automatically work in our new articles app?  we would need to code that ourselves?

[Andrew Aikman](https://app.slack.com/team/U01A4M52Z7G)
  [10:04 AM](https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1657703040041939)
We did look at making it configurable but with no example of how it would be used it was difficult to justify the extra code. If you have a working example that we can load in it would help to know.
I think for now it’s best to keep it simple with Page only and then when it’s clear how others such as yourself need it to work for other content types we can add a configuration mechanism behind it.

[Andrew Aikman](https://app.slack.com/team/U01A4M52Z7G)
  [10:13 AM](https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1657703615652569)
Does that help [@Mike](https://django-cmsworkspace.slack.com/team/U01EG6FDEA3) ?

[Mike](https://app.slack.com/team/U01EG6FDEA3)
  [10:14 AM](https://django-cmsworkspace.slack.com/archives/C01E6NMK540/p1657703654445339)
yes thanks, that’s really useful, I will add it to the schedule to add to articles
and then share with you


michaeljcollinsuk (changes requested):

cms/toolbar/utils.py
```diff
@@ -117,14 +119,57 @@ def get_toolbar_from_request(request):
     return getattr(request, 'toolbar', EmptyToolbar(request))
 
 
+def _get_cms_endpoint_settings():
```
Aiky30, line 122: This function is no longer required.

cms/toolbar/utils.py
```diff
@@ -144,4 +192,7 @@ def get_object_structure_url(obj, language=None):
         language = get_language()
 
     with force_language(language):
-        return admin_reverse('cms_placeholder_render_object_structure', args=[content_type.pk, obj.pk])
+        url = admin_reverse('cms_placeholder_render_object_structure', args=[content_type.pk, obj.pk])
```
Aiky30, line 195: I don't think the structure url should have this param, it's not visible to a user.

Aiky30 (changes requested):

cms/toolbar/utils.py
```diff
@@ -117,14 +119,40 @@ def get_toolbar_from_request(request):
...
+    querystring_param_name, querystring_param_content = modifier(obj, language)
+    return f"?{querystring_param_name}={querystring_param_content}"
+
+
+def get_querystring_modifier(obj, content_type, url, language=None):
+    extension = _get_cms_extension()
+
+    method = extension.cms_endpoint_modifiers.get(content_type, None)
+    if method:
+        # TODO: Add validation to the url, to ensure we don't already have querystring params
```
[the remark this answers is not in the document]
adam-murray: I went for the split solution, as the url lib one was going to require far more boilerplate code. If you could have a look and let me know what you think, would be much appreciated.

cms/utils/conf.py
```diff
@@ -82,6 +87,8 @@ def wrapper():
     'PAGE_WIZARD_CONTENT_PLUGIN_BODY': 'body',
     'PAGE_WIZARD_CONTENT_PLACEHOLDER': None,  # Use first placeholder it finds.
     'SIDEFRAME_ENABLED': True,
+    'CMS_ENDPOINT_QUERYSTRING_PARAM_ENABLED': True,
```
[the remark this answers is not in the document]
Aiky30: The CMS prefix is added by this file. Also reworded to match the use case.

Aiky30 (changes requested):
Still more to clean out I'm afraid.

cms/toolbar/utils.py
```diff
@@ -117,14 +119,42 @@ def get_toolbar_from_request(request):
...
     content_type = ContentType.objects.get_for_model(obj)
 
     if not language:
         language = get_language()
 
     with force_language(language):
-        return admin_reverse('cms_placeholder_render_object_edit', args=[content_type.pk, obj.pk])
+        url = admin_reverse('cms_placeholder_render_object_edit', args=[content_type.pk, obj.pk])
+    if getattr(settings, 'CMS_ENDPOINT_QUERYSTRING_PARAM_ENABLED', False):
+        url = get_querystring_modifier(obj, url, language)
```
michaeljcollinsuk, line 156: I may have missed some conversations about this, but is the intention for the solution to still allow a custom method to be passed in? Because as I understood, we only need to add the live url for a Page? In which case, the method injection could be removed, and instead we can handle getting the live url and appending it internally here? For example then your conditional could be more like:
```
    if getattr(settings, 'CMS_ADD_PAGE_LIVE_URL_PARAM', False) and isinstance(obj, Page):
        url = add_page_live_url_querystring_param(page=obj, url=url, language=language)

    return url
```
And then to use some of your existing changes but without the injection might be something like:
```
def add_page_live_url_querystring_param(page, url, language=None):
    live_url = page.get_absolute_url(language=language)
    # or could use a setting for the key?
    live_url_param = f"live-url={live_url}"
    urls = url.split('?')
    if len(urls) > 1:
        url += f'&{live_url_param}'
    else:
        url += f'?{live_url_param}'
    return url
```

cms/toolbar/utils.py
```diff
@@ -117,14 +119,42 @@ def get_toolbar_from_request(request):
...
 
 
+def get_querystring_modifier(obj, url, language=None):
+    """
+    Get the querystring modifier if one exists for the content type, and add it to the endpoint url
+    """
+    qsp_config = getattr(settings, 'CMS_ENDPOINT_QUERYSTRING_CONFIGURATION', None)
+    search_field = f"{obj._meta.app_label}.{obj._meta.model_name}"
+    if not qsp_config:
+        raise ImproperlyConfigured(
```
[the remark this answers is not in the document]
adam-murray: Added

cms/tests/test_toolbar.py
```diff
@@ -25,10 +27,14 @@
 from cms.test_utils.testcases import (CMSTestCase, URL_CMS_USERSETTINGS)
 from cms.test_utils.util.context_managers import UserLoginContext
 from cms.toolbar_pool import toolbar_pool
+from cms.toolbar import utils
 from cms.toolbar.items import (ToolbarAPIMixin, LinkItem, ItemSearchResult,
                                Break, SubMenu, AjaxItem)
 from cms.toolbar.toolbar import CMSToolbar
-from cms.toolbar.utils import get_object_edit_url, get_object_preview_url, get_object_structure_url
+from cms.toolbar.utils import (
+    add_endpoint_querystring_params, get_object_edit_url,
```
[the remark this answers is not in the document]
adam-murray: This one is a bit unclear in the view provided by GitHub, what is the actual change here?

cms/toolbar/utils.py
```diff
@@ -117,14 +120,41 @@ def get_toolbar_from_request(request):
...
+    """
+    url_param = getattr(settings, 'CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM', None)
+    if not url_param:
+        raise ImproperlyConfigured(
+            """
+            With 'CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM_ENABLED' set to True,
+            CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM must be provided
+            """
+        )
+    if not isinstance(obj, PageContent):
```
michaeljcollinsuk, line 136: Could you add a test for this - so when the setting is enabled, but the object is not `PageContent`, the url is returned unchanged?

cms/toolbar/utils.py
```diff
@@ -117,14 +120,41 @@ def get_toolbar_from_request(request):
     return getattr(request, 'toolbar', EmptyToolbar(request))
 
 
+def get_querystring_modifier(obj, url, language=None):
```
michaeljcollinsuk, line 123: Now that this has been simplified, I am not sure the name of this function is reflective of what it does - it doesnt really _get_ anything, it _adds_ the live url to the querystring. So to me something like `add_live_url_to_querystring` would be more descriptive.

cms/toolbar/utils.py
adam-murray: No problem!

cms/tests/test_toolbar.py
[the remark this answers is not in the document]
Aiky30: Doubled up suggestions some how.

cms/tests/test_toolbar.py
```diff
@@ -1865,6 +1874,119 @@ def test_edit_page(self):
...
+
+        self.assertEqual(edit_url, expected_edit_url)
+        self.assertEqual(preview_url, expected_preview_url)
+        self.assertEqual(edit_url.count("?"), 0)
+        self.assertEqual(preview_url.count("?"), 0)
+        self.assertEqual(edit_url.count("&"), 0)
+        self.assertEqual(preview_url.count("&"), 0)
+
+    @override_settings(CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM_ENABLED=True)
+    def test_add_live_url_querystring_param_handles_missing_enabled_setting(self):
```
Aiky30, line 1952: Not required, this was part of the older logic I think where the configuration was required. The param has a default, currently you can't use that default because it's enforced that it's set.

cms/toolbar/utils.py
```diff
@@ -117,14 +120,41 @@ def get_toolbar_from_request(request):
     return getattr(request, 'toolbar', EmptyToolbar(request))
 
 
+def add_live_url_querystring_param(obj, url, language=None):
+    """
+    Get CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM as the parameter, and the live url, if the object is
+    PageContent
+    """
+    url_param = getattr(settings, 'CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM', None)
+    if not url_param:
```
Aiky30, line 129: Not required, it has a default of "live-url" so we know it always exists. You probably had issues with the tests because you weren't using the config settings which get_cms_setting does.

cms/toolbar/utils.py
```diff
@@ -117,14 +120,41 @@ def get_toolbar_from_request(request):
     return getattr(request, 'toolbar', EmptyToolbar(request))
 
 
+def add_live_url_querystring_param(obj, url, language=None):
+    """
+    Get CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM as the parameter, and the live url, if the object is
+    PageContent
+    """
+    url_param = getattr(settings, 'CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM', None)
```
Aiky30, line 128:
```suggestion
    url_param = get_cms_setting('ENDPOINT_LIVE_URL_QUERYSTRING_PARAM')
```

This isn't how you get settings in the core.

cms/toolbar/utils.py
```diff
@@ -134,7 +164,10 @@ def get_object_preview_url(obj, language=None):
         language = get_language()
 
     with force_language(language):
-        return admin_reverse('cms_placeholder_render_object_preview', args=[content_type.pk, obj.pk])
+        url = admin_reverse('cms_placeholder_render_object_preview', args=[content_type.pk, obj.pk])
+    if getattr(settings, 'CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM_ENABLED', False):
```
Aiky30, line 168:
```suggestion
    get_cms_setting('ENDPOINT_LIVE_URL_QUERYSTRING_PARAM_ENABLED'):
```

Aiky30 (changes requested):
More changes after testing this out.

FYI, get_cms_setting prefixes with "CMS_" so you still need in in any docs and in a projects settings.py but not in the code.

cms/toolbar/utils.py
```diff
@@ -117,14 +119,34 @@ def get_toolbar_from_request(request):
     return getattr(request, 'toolbar', EmptyToolbar(request))
 
 
+def add_live_url_querystring_param(obj, url, language=None):
+    """
+    Get CMS_ENDPOINT_LIVE_URL_QUERYSTRING_PARAM as the parameter, and the live url, if the object is
+    PageContent
```
Aiky30, line 125:
```suggestion
```

Aiky30 (changes requested):
Minor changes left including adding preview endpoints to tests that have missed it

Aiky30 (approved):

michaeljcollinsuk (approved):

Bernardvdv (approved):

PR #7369, opened 2022-07-27
Change `TitleExtension` to `PageContentExtension`

marksweb:
## Description

**This is a major change.**

Currently we have a `TitleExtension` which hooks up `PageContent`. We've talked about how this should have been changed in the early days of v4, so I've done it now before we actually get to the point of official release to avoid it being something we wish we'd done but didn't.


cms/api.py
```diff
@@ -204,7 +204,7 @@ def create_title(language, title, page, menu_title=None, slug=None,
                  created_by='python-api', soft_root=False, in_navigation=False,
                  template=TEMPLATE_INHERITANCE_MAGIC,
                  limit_visibility_in_menu=constants.VISIBILITY_ALL,
-                 xframe_options=constants.X_FRAME_OPTIONS_INHERIT):
+                 xframe_options=constants.X_FRAME_OPTIONS_INHERIT):  # TODO: Deprecate this
```
Aiky30, line 207: It makes this much easier without changing the create_title api as you have done. We will need a plan of how we could replace this in the future, it will affect all internal tests, and if we have 2 at the same time the tests would need to cover both too. Very ugly. 

This was left alone due to the sheer amount of changes that would be required to every other package and every other packages test suite.

Aiky30 (commented):
@marksweb Having had the time to think about this:

I think this change will also need to come along at the same time as djangocms-versioning, this is because the djangocms-versioning monkey patches rely on title_cache and get_title_obj. 

It might be worth having a smaller PR just for TitleExtension to PageContentExtension renaming to allow that to come in quicker. As soon as you merge this you lose djangocms-versioning.

marksweb:
@Aiky30 thanks for taking a look. I've got a PR on versoning to bring support for this change. So the two can come into play together. It'll just need a plan for how best to do that.

fsbraun (approved):

marksweb:
@fsbraun So I updated the branch and resolved the conflicts, but something in that [commit](https://github.com/django-cms/django-cms/pull/7369/commits/975d07ef560bb9856e71766bc41084b2def83e91) has destroyed the test results 😬

fsbraun:
It seems that the merge sneaked in `title_cache` again: 
image

marksweb:
@fsbraun Oh well spotted! That wasn't part of the conflicts either so I missed it.

PR #7372, opened 2022-08-01
ci: Flake8 fixes

marksweb:
## Description

This fixes some flake8 issues that I've just found while trying to update `release/3.11.x`


cms/admin/forms.py
```diff
@@ -820,7 +820,7 @@ def __init__(self, *args, **kwargs):
     def get_root_nodes(self):
         # TODO: this needs to avoid using the pages accessor directly
         nodes = TreeNode.get_root_nodes()
-        return nodes.exclude(cms_pages__is_page_type=not(self.page.is_page_type))
+        return nodes.exclude(cms_pages__is_page_type=not self.page.is_page_type)
```
vinitkumar, line 823: To prevent confusion, can we assign `not self.page.is_page_type` to a variable and use that in the query?

vinitkumar (commented):
Please check my comment for cms/admin/forms.py

cms/admin/forms.py
marksweb: @vinitkumar Yeah I suppose so, I was just trying to keep changes to a minimum. But that makes sense.

vinitkumar (approved):

Issue #7033, opened 2021-05-12
[BUG] management command "cms check" doesn't work


MichaelAnckaert:
## Description

When running the command `python manage.py cms check` on the develop branch there is an issue with the argument parsing of the custom management command. 

I am aware that this happens on a development branch. This issue is mainly to raise awareness and coordinate any efforts to fix this issue. 

## Steps to reproduce

Run the command  `python manage.py cms check` and the following output is logged:

> 
> Traceback (most recent call last):
>   File "/home/michael/projects/django-cms-demo/myproject/manage.py", line 22, in <module>
>     main()
>   File "/home/michael/projects/django-cms-demo/myproject/manage.py", line 18, in main
>     execute_from_command_line(sys.argv)
>   File "/home/michael/projects/django-cms-demo/venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
>     utility.execute()
>   File "/home/michael/projects/django-cms-demo/venv/lib/python3.9/site-packages/django/core/management/__init__.py", line 413, in execute
>     self.fetch_command(subcommand).run_from_argv(self.argv)
>   File "/home/michael/projects/django-cms-demo/venv/lib/python3.9/site-packages/django/core/management/base.py", line 355, in run_from_argv
>     self.execute(*args, **cmd_options)
>   File "/home/michael/projects/django-cms-demo/venv/lib/python3.9/site-packages/django/core/management/base.py", line 391, in execute
>     if self.requires_system_checks and not options['skip_checks']:
> KeyError: 'skip_checks'
> 

## Expected behaviour

CMS checks are performed without error. 

## Actual behaviour

The parsing of the management command fails with an error on parsing the arguments passed to the management command.

## Additional information (CMS/Python/Django versions)

Regular management commands (for example `manage.py check`) work without issue. 

* Django CMS: develop branch
* Python 3.9
* Installed packages:

asgiref==3.3.4
Django==3.2.2
django-classy-tags==1.0.0
django-cms @ file:///home/michael/projects/django-cms
django-formtools==2.2
django-sekizai==2.0.0
django-treebeard==4.3.1
djangocms-admin-style==2.0.2
pytz==2021.1
six==1.16.0
sqlparse==0.4.1

goutnet:
we will need to check this on the coming release 3.9.0 and fix it before release then.

@vinitkumar  @marksweb  could you help me check?

marksweb:
[...]

PR #7387, opened 2022-09-12
fix: Fixes #7386: consider any version of Django >= 3.0 when checking weth…

### LINKED ISSUE #7386: [BUG] management command "cms check" doesn't work (again).
## Description

This appears to be exactly the same bug that was fixed in #7033: 

When running the subcommand, [this line](https://github.com/django-cms/django-cms/blob/develop/cms/management/commands/subcommands/base.py#L39) doesn't check for `DJANGO_4_0`, so the command fails with 

```
  File "/home/marco/Code/kavea/.venv/lib/python3.9/site-packages/django/core/management/base.py", line 453, in execute
    if self.requires_system_checks and not options["skip_checks"]:
KeyError: 'skip_checks'
``` 

## Steps to reproduce

1. Install django-cms==3.11 and Django==4.0
2. python manage.py cms check

mbi:
…er the BaseCommand requires a `skip-checks` argument.

## Description

`skip_checks` seems to be required in management commands for *any* version of Django >= 3.0. The previous implementation kept adding new `DJANGO_X_Y` variables in the test, but with each new supported Django version, this command was forgotten and broke. See also #7033

## Checklist

### DIFF
--- a/cms/management/commands/subcommands/base.py
+++ b/cms/management/commands/subcommands/base.py
@@ -4,7 +4,7 @@ from collections import OrderedDict
 from django.core.management.base import BaseCommand, CommandParser
 from django.core.management.color import color_style, no_style
 
-from cms.utils.compat import DJANGO_3_0, DJANGO_3_1, DJANGO_3_2
+from cms.utils.compat import DJANGO_VERSION
 
 
 def add_builtin_arguments(parser):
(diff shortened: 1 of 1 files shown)


cms/management/commands/subcommands/base.py
```diff
@@ -36,7 +36,7 @@ def add_builtin_arguments(parser):
         help="Don't colorize the command output.")
     parser.add_argument('--force-color', action='store_true', dest='force_color', default=False,
         help="Colorize the command output.")
-    if DJANGO_3_0 or DJANGO_3_1 or DJANGO_3_2:
+    if DJANGO_VERSION > '3.0':
```
fsbraun, line 39: Since `DJANGO_2_2` is the last version without the skip checks command supported can this be `if not DJANGO_2_2:`? I agree that it is a long time before `DJANGO_VERSION > '3.0'` might fail incorrectly but I guess it also makes it transparent that the if can be removed all together once Django 2.2 is history.

mbi: Good point, fixed.

fsbraun (approved):

Issue #7392, opened 2022-09-15
[BUG] CMS v4: Moving plugins between placeholders broken.


fsbraun:
## Description

Moving plugins **between** placeholders in certain situations fails and leaves the frontend js plugin tree out of sync with the database plugin tree.

## Steps to reproduce

1. Create empty page with at least two palceholders
2. Add plugin with 2 children in the first placeholder
3. Add plugin with 10 (>> 3) in the second placeholder4. 
4. Move plugins from second placeholder somewhere in the "middle" of first placeholder 

![move-bug](https://user-images.githubusercontent.com/16904477/190337036-f3c71fdc-0f7c-4fd7-a1fb-b247d1e98521.gif)

## Expected behaviour

Plugins move there

## Actual behaviour

* Server error 500 is created.

```
django.db.utils.IntegrityError: UNIQUE constraint failed: cms_cmsplugin.placeholder_id, cms_cmsplugin.language, cms_cmsplugin.position
[15/Sep/2022 06:54:48] "POST /admin/cms/placeholder/move-plugin/?cms_path=%2Fadmin%2Fcms%2Fplaceholder%2Fobject%2F10%2Fstructure%2F2%2F HTTP/1.1" 500 163851
```

* Server error is briefly shown in an unreadable format then disapperas
* Plugin tree in database remains unchanged
* Plugin tree in js frontend changes and is out of sync with database
* Further manipulation of plugin tree might lead to inconsistent data structures, even trees spanning more than one placeholder

## Additional information

Since the **clipboard** is a separate placeholder the bug could potentially also appear when using the clipboard (copy & paste).

## Additional information (CMS/Python/Django versions)

djangocms-frontend 0.9.4 & 1.0
django-cms 4.0 develop-4
django 3.2.15
sqlite

## Do you want to help fix this issue?

* [x] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 
* [ ] No, I only want to report the issue.

fsbraun:
Is there a link to the Sentry report?

marksweb:
@fsbraun no, private sentry instance so I can't get those public links for it.

I can pull info out of it though.

```
UniqueViolation: duplicate key value violates unique constraint "cms_cmsplugin_placeholder_id_language_position_bad60ebe_uniq"
DETAIL:  Key (placeholder_id, language, "position")=(54, en, 27) already exists.

[...]

Issue #7391, opened 2022-09-15
[BUG] CMS v4: Inconsistent behaviour when re-ordering plugins


marksweb:
## Description

I've got a page with 1 placeholder. Using `djangocms-frontend` to build content, first in a container for big displays, then a container for smaller displays. Generally using copy/paste to begin setting up the 2nd container.

https://user-images.githubusercontent.com/1461191/190281318-380f926e-166c-4549-a4bd-41fb807c5f44.mov

Although this isn't only happening to plugins which have been pasted. I've also created a new spacing plugin and tried to move this around the page;

https://user-images.githubusercontent.com/1461191/190282224-acdda3df-a8f9-4edb-86d6-6b7bc61fee6f.mov

Also a text plugin. While the text plugin was moved amongst the two first level containers, it moved without issue. But once moved within a container, re-ordering didn't always result in the plugin being where I dragged it.

Focusing on the plugins at the bottom of the 2nd container;

| id      | position | parent | note
| ----------- | ----------- | ----------- | ----------- |
| 4823      | 186       | None | top level 2nd container
| 5081      | 260       | 4823 | 2nd level container
| 5085   | 263        | 5081 | First card
| 5096   | 264        | 5081 | Second card
| 5229   | 269        | 5081 | my-2 spacing
| 5108   | 270         | 5081 | py-4 spacing (1st)
| 5107   | 271          | 5081 | py-4 spacing (2nd)
| 5100   | 277          | 5081 | Third card

Here is a copy of the plugins table filtered to all plugins within the placeholder on this page;
[Playground-678.csv](https://github.com/django-cms/django-cms/files/9570705/Playground-678.csv)

## Additional information (CMS/Python/Django versions)

djangocms-frontend 0.9.4 & 1.0
django-cms 4.0 develop-4
django 3.2.15

## Do you want to help fix this issue?

* [x] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 
* [ ] No, I only want to report the issue.

marksweb:
Trying to recreate this scenario without using djangocms-frontend and I'm unable to with this plugin setup;

![Screenshot 2022-09-20 at 00 57 14](https://user-images.githubusercontent.com/1461191/191139213-d7180384-95f2-4340-83e6-e0d2cc2cceea.png)

fsbraun:
I just recreated it using Text and Icon plugins (somehow):
[...]

PR #7394, opened 2022-09-16
fix: CMS v4: Moving plugins between placeholders, plugin api

fsbraun:
## Description

This PR removes some bugs from handling the plugin tree within a placeholder:

* Moving plugins between placeholder now works reliably (fixing #7392). The position moving logic required for CTE to not fail because of violated uniqueness constraints [test added]
* adding a plugin as first child now works if the plugin tree contains more than 3 plugins already [test added]
* fix `.get_last_plugin_position` to give the position of the last child or grand-child. This caused the `cms.api.add_plugin` function to add a plugin before grand children of the last child [test added]
* fix `.get_last_plugin_position` to always give the largest position number and not the one retrieved from the database in last position (though they apparently regularly coincide) by adding `.order_py('position')`
* adding a `.check_tree` and `.fix_tree` method to the Placeholder model to at least be able to fix broken trees (#7392 )

## Related resources

It has been observed (though exact circumstances are unclear) that when the plugin tree in the js frontend and in the database get out of sync (e.g., by a bug like #7392) further editing commands can lead to unexpected results to the plugin tree, potentially even a plugin tree spanning placeholders (e.g., the clipboard). 

To this end the PR includes a `.check_tree` and `.fix_tree` methods which will check and fix 3 plugin tree properties

1. Plugin positions do not include gaps (should have been removed by `._recalculate_plugin_positions`)
2. Child plugins always have a higher position than their parent plugins
3. Plugins do not refer to parent plugins which are assigned to a different placeholder (attn: The fix breaks links to other placeholders by moving the plugin to the top-level (`parent = None`)

## Checklist


cms/models/placeholdermodel.py
```diff
@@ -739,3 +739,76 @@ def _recalculate_plugin_positions(self, language):
...
+                        messages.append(f"{language}: Children with positions lower than their parent's (id={parent}) position")
+
+        # Check 3: parents belonging to other placeholders
+        for plugin in self.cmsplugin_set.all():
+            if plugin.parent and plugin.parent.placeholder != self:
+                messages.append(f"{language}: Plugins claim to be children of parents in a different placeholder")
+
+        return messages
+
+    def fix_tree(self, language=None):
```
Aiky30, line 778: I fear that this will just end up like django-cms v3 where production container spin up always runs this command. It's not very efficient and people with versioning will want this to work very differently.

Aiky30 (commented):
I'm genuinely concerned by the amount of changes to the plugin tree here. 

Do we have a scenario as a unit test that currently fails without any changes?

I was under the impression that the issue was with shifting the plugins. Playing with the plugin tree is extremely risky.

My attention would have been here: https://github.com/django-cms/django-cms/blob/28d3fa54cb8795fc2276ada94e07f45fa2fe1def/cms/models/placeholdermodel.py#L611

cms/models/placeholdermodel.py
fsbraun: So far this is only a method for the Placeholder model. Not a management command. If added as a management command it would need to be only invoked on placeholders in draft models. If we fix the plugin trees we might never need it as a management command. Right now, @marksweb at least will need the fix for his site to work.

I suggest to add a management command for the develop-4. branch. We can decide to remove it before publishing v4. Clearly it should not be needed in a published version.

Aiky30:
@fsbraun Now you understand how the plugins are being ordered can you now see that the way the position is being created here in the front end could be creating them incorrectly? The cms always goes and gets the highest number: https://github.com/django-cms/djangocms-frontend/blob/f5691769b7d1d5c3e2a0cdd9d7162bd3f9bdbe41/djangocms_frontend/contrib/grid/cms_plugins.py#L125

You're making the core cms fit djangocms-frontend rather than make djangocms-frontend fit the cms. 

This could have unforeseen side affects that I would not want to be responsible for in production projects. 

I would advise having any fixing of trees done in a 3rd party package until it is seen to be stable, the same for any plugin tree changes. Just bringing them in could have devastating consequences, especially by modifying all placeholders in the system. I feel so strongly about this this change, that if it is just forced in then I will need to consider how much value I can add going forwards.

cms/models/placeholdermodel.py
Aiky30: The design principle of the core is to be as simple as possible, temporary logic should be a 3rd party package, not part of the core. It also allows changes to be optional and provides a much faster turn around. 

I agree Mark needs a fast solution, this can be a management command in the djangocms-frontend package because it is this package that is seeing these issues. The only time I saw the same issue is with the text editor logic I keep referring to. It had exactly the same side affects and it was because the cms wasn't used to get the position. Once the text editor got the position properly no position issues were seen again. 

```
        last_position = placeholder.get_last_plugin_position(language) or 0
        # A shift is only needed if the distance between the new plugin
        # and the last plugin is greater than 1 position.
        needs_shift = (position - last_position) < 1

        if needs_shift:
            # shift to the right to make space for the new ghost plugin
            placeholder._shift_plugin_positions(
                language,
                start=position,
                offset=last_position,
            )
```

Aiky30: The fix tre command you are creating probably only needs to be this method:
```
placeholder._shift_plugin_positions(
    language,
    start=position,
    offset=last_position,
)
```

Aiky30: @marksweb could you try this script locally and see if it reorders the plugins to get you out of your blocker? ^^

Aiky30: You can use version compare manually to see if it has reordered anything unintentionally. Compare is useless if all versions have been manipulated.

fsbraun:
@Aiky30 I understand your concerns about changes to the plugin tree at the very heart of v4. Obviously, you have by far the most massive experience with v4.

Nevertheless, all three new tests 

- `test_inter_placeholder_nested_plugin_move` https://github.com/django-cms/django-cms/blob/94c929e2c4ee932bcaf3921628c2de355fcd03cc/cms/tests/test_placeholder.py#L196
- `test_get_last_plugin_position` https://github.com/django-cms/django-cms/blob/94c929e2c4ee932bcaf3921628c2de355fcd03cc/cms/tests/test_placeholder.py#L230
- `test_get_last_plugin_position_order` https://github.com/django-cms/django-cms/blob/94c929e2c4ee932bcaf3921628c2de355fcd03cc/cms/tests/test_placeholder.py#L270

will fail/throw an exception without this PR. To me these address genuine bugs. Don't kill the messenger. 

This PR proposes no changes in the way the plugin tree works. It's just that you actually can now move plugins across placeholders by creating enough temporary space between plugin positions so that they can be moved to where they need to go. Currently, cms v4 in some cases only creates insufficient temporary space and moving the plugins to the new placeholder make them collide with an existing plugin there.

These bugs have nothing to do with djangocms-frontend and exist w/o it being installed. Here's a djangocms-frontend "free" example of #7392 - try it on any cms v4 installation and move **any** two+ plugins in front of a single plugin in a different placeholder:

move-bug-text

Just to be clear, I do not want to force anything on anybody. I just want to bring v4 forward. This is why I actively asked you to be involved and review in this PR. I hope you understand.

cms/models/placeholdermodel.py
fsbraun: For me it's also fine to have a fix-tree management command in a separate mini app or even script. Thanks for the code snippet to get the draft pages. I was thinking of let it run over all placeholders that are assigned to a ContentType object with "draft" status in versioning and, e.g., the clipboard.

marksweb: @Aiky30 Yeah need to give it a run. That earlier version that just went over all placeholders did resolve it, but obviously wasn't the correct approach so need to run this now.

marksweb: @Aiky30 I've tried to run your example and it didn't change what's happening.

Aiky30: Sorry guys, I saw what u thought was this just being merged without any approvals hence why I threw my toys out of the pram.

I do have genuine concerns at the scale if the changes here. 

If you're blocked then I guess you have to do something. 

@fsbraun  The tests will need to create multiple trees in different languages. 

Does there need to be the amount if changes made to the tree or can this be made less intrusive?

@marksweb it's a shame that the script I sent didn't work, I imagine it could with some tweaking but I'm away all weekend so not going to be much help on getting something fixed fast.

I would still lean towards making sure any fixing of previous trees is handled outside of the cms, this is very hacky and I'm pretty sure will crash on large projects.

Aiky30:
I do intend to pull this PR / branch down and check it over. I will at the very least have a lot of time this week to go over this, I'll also see Mark this week too so should have enough background to be of use with this change.

cms/templates/admin/cms/page/change_form.html
```diff
@@ -42,7 +42,7 @@
 
 {% block object-tools %}{% endblock %}
 
-<form {% if has_file_field %}enctype="multipart/form-data" {% endif %}action="?language={{ language }}{%if request.GET.parent_node %}&amp;parent_node={{ request.GET.parent_node }}{% endif %}{%if request.GET.source %}&amp;source={{ request.GET.source }}{% endif %}" method="post" id="page_form">
```
Aiky30, line 45: Wondering if these changes for &amp; are actually a separate bug / PR?

cms/models/placeholdermodel.py
```diff
@@ -498,7 +498,7 @@ def add_plugin(self, instance):
             self._shift_plugin_positions(
                 instance.language,
                 start=instance.position,
-                offset=last_position,
+                offset=last_position - instance.position + 2, # behind last_position plus one for the new
```
Aiky30, line 501: I can't see the need for this logic, because the "last" plugin should be the end, if we are adding a plugin we need to shift the tree behind where we are inserting all of the way to the end.

Aiky30 (changes requested):
The logic to moving the plugins looks sane. 

One comment on the add plugin offset that I don't understand. 

Asked for some documentation on the tests for the plugin tree to help anyone in the future, 

Is this change to the djangocms-text-ckeditor (https://github.com/django-cms/djangocms-text-ckeditor/pull/629/files) a side affect of the changes being made in this PR?

cms/templates/admin/cms/page/change_form.html
fsbraun: Right. This is not connected to the rest. It just came to mind that Django since 2.2.19 does not support semicolons as separators in urls any more (https://docs.djangoproject.com/en/4.1/releases/2.2.19/) Hence it does not understand `&amp;` as a separator. However, this seems not critical for the test. It passes w/o change.

cms/models/placeholdermodel.py
fsbraun: Like for the move between placeholders you have to both consider to be able to **move out the tail** (as you assume) but then also be able to **move the tail back in** after the new plugin has been added. For this you need one more spare position, since the gap has been reduced by the newly created plugin.

Moving the tail to the end amounts to an offset of `last_position - instance.position + 1`, then add `1` for the new plugin.

Consider your want to insert Plugin 0 at position 1 (positions in the first column):

```
1 Plugin 1
2 Plugin 2
```

First you move the tail out and put the new plugin into the database. Take as offset `last_position(2) - instance.position(1) + 2 = 3`:
```
1 Plugin 0
4 Plugin 1 
5 Plugin 2
```

Now you have the minimum position slots between `1` and `4` to move the tail back in (position 2 and up).

If you had just moved by `last_position(2)` the gap would not have been big enough to move back the tail.

fsbraun:
> The logic to moving the plugins looks sane.

Yes, it's good to have some sane logic there.

> One comment on the add plugin offset that I don't understand.
> Asked for some documentation on the tests for the plugin tree to help anyone in the future,

I've added an [elaborate explanation](https://github.com/django-cms/django-cms/pull/7394#discussion_r979382969) why the add plugin offset has to be what it is. The general issue for the offsets in both cases seems that not only they have to consider **moving the remaining part (tail) out far enough** to create space for new plugins and not overlap with the original plugin positions but also far enough out to after adding the plugins close the remaining gaps and **move the tail back in.**

I invested in some comments for the tests. I guess the new tests are above par with respect to comments and reflecting languages now. 😉

> Is this change to the djangocms-text-ckeditor (https://github.com/django-cms/djangocms-text-ckeditor/pull/629/files) a side affect of the changes being made in this PR?

At this point I have identified **two totally independent ways of getting the plugin tree into a inconsistent state**:
1. By moving a big number of plugins from one placeholder to another with a small number of plugins (**this PR**)
2. By generating a gap in the position field and moving plugins after the gap. Only one placeholder needed for this and this might be @marksweb case in #7391. The gap in the position field originates from **djangocms-text-ckeditor**. Hence this PR is not a side effect of the changes here.

Aiky30 (approved):

Issue #7389, opened 2022-09-14
[BUG] CMS v4: Placeholder burger menu "Empty all" does not work


fsbraun:
## Description

If the user selects "Empty all" from the placeholder burger menu in the structure board a modal window opens with the same edit view of the page (and its own structure board).

## Steps to reproduce

![image](https://user-images.githubusercontent.com/16904477/190152685-8261c4e1-2567-4e0e-93f6-9d178b03fe51.png)

## Expected behaviour

The expected behavior is to open a modal with a list of all plugins to be deleted and two options ("Cancel" and "Yes, I'm sure") 

## Screenshots

![image](https://user-images.githubusercontent.com/16904477/190151621-c18b3bb5-da0b-4973-ab11-4444cfb97772.png)

## Additional information (CMS/Python/Django versions)

django-cms@develop-4
django 3.2
Python 3.9

fsbraun:
Solved by #7403

PR #7403, opened 2022-09-26
fix: Empty all placeholder menu

fsbraun:
## Description

The empty placeholder menu currently does not work (#7389). The url is not rendered into the template.

This PR corrects the template and uses the `url` template tag to retrieve the admin url for "Empty all". 

empty-all

## Checklist

### DIFF
--- a/cms/tests/test_toolbar.py
+++ b/cms/tests/test_toolbar.py
@@ -437,28 +437,36 @@ class ToolbarTests(ToolbarTestBase):
         self.assertContains(response,
                             '<div class="cms-submenu-item cms-submenu-item-title"><span>Different Grouper</span>')
 
-    def test_extra_placeholder_menu_items(self):
+    def test_placeholder_menu_items(self):
         superuser = self.get_superuser()
         page = create_page("toolbar-page", "col_two.html", "en")
         page_content = self.get_page_title_obj(page)
         page_structure_url = get_object_structure_url(page_content)
+        placeholders = page_content.get_placeholders()
+        self.assertEqual(len(placeholders), 2)  # we get two placeholders
 
         with self.login_user_context(superuser):
             response = self.client.get(page_structure_url)
 
         self.assertEqual(response.status_code, 200)
-        self.assertContains(
-            response,
-            '<div class="cms-submenu-item"><a href="/some/url/" data-rel="ajax"'
+        expected_bits = (
+            # Copy all in placeholder menu
+            '<div class="cms-submenu-item"><a data-cms-icon="copy" data-rel="copy" href="#">Copy all</a></div>',
+            # Past in placeholder menu
+            '<div class="cms-submenu-item"><a data-cms-icon="paste" data-rel="paste" href="#">Paste</a></div>',
+            # Empty all placeholder menu (for both placeholders)
+            '<div class="cms-submenu-item"><a data-cms-icon="bin" data-rel="modal" href="' +
+            reverse("admin:cms_placeholder_clear_placeholder", args=(placeholders[0].id, )),
+            '<div class="cms-submenu-item"><a data-cms-icon="bin" data-rel="modal" href="' +
+            reverse("admin:cms_placeholder_clear_placeholder", args=(placeholders[1].id, )),
+            'data-name="sidebar column">Empty all</a></div>',
+            # Extra items in placeholder menu
+            '<div class="cms-submenu-item"><a href="/some/url/" data-rel="ajax"',
+            'data-on-success="REFRESH_PAGE" data-cms-icon="whatever" >Data item - not usable</a></div>',
+            '<div class="cms-submenu-item"><a href="/some/other/url/" data-rel="ajax_add"',
         )
-        self.assertContains(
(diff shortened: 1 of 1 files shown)


marksweb (approved):

Aiky30 (changes requested):
The change looks good, although I would question why no tests failed for this issue. Maybe it has no coverage. Here is an example of a test that checks the plugin menu, hopefully this will be super easy to implement: https://github.com/django-cms/django-cms/blob/28d3fa54cb8795fc2276ada94e07f45fa2fe1def/cms/tests/test_plugin_renderers.py#L32

I think when any bugs like this are found and no tests failed, we should implement a test to ensure it doesn't happen again. Paulo would regularly state that he had little confidence in the cms test suite, due to that we have tried to add confidence in the test suite early on in v4.

fsbraun:
I see your point. This was not detected since `{{ }}` expansions in templates fail silently. So what we would need to do is to find all those expansions and generate a test suite for those. Luckily `{% %}` template tags fail loudly and with this particular item I guess we have a good chance of being fine. 
I am actually thinking of a linter to identify those variable expansions. (There's a longstanding bug #7188 where page ids are put into an url in localized form rendering the page tree unusable in localizations with thousands separator - that's probably something to port to v4, too.)
Will look into other variable expansions at risk and see what tests make sense.

fsbraun:
@Aiky30  It turns out there was a test for the extended menu items. The new test also includes the standard items. (Copy, Paste, Empty all)  Other critical template elements (`{{ object.property }}`) can be found in these files:

```
(.venv.) $ find cms -name "*.html" -print -exec egrep 'href="{{\s*[a-zA-Z0-9_.]+\.[a-zA-Z0-9_]+\s*}}"' {} \;
cms/templates/cms/toolbar/toolbar_with_structure.html
    <p>{% trans "It is being redirected to:" %} <a href="{{ cms_toolbar.redirect_url }}">{{ cms_toolbar.redirect_url }}</a></p>
cms/templates/cms/toolbar/items/live_draft.html
    href="{{ cms_toolbar.get_object_preview_url }}">
    href="{{ cms_toolbar.get_object_edit_url }}">
cms/templates/cms/toolbar/dragitem_extra_menu.html
    <div class="cms-submenu-item"><a href="{{ item.url }}" data-rel="{{ item.action }}"{% if item.data %} data-post="{{ item.data }}"{% endif %}{% if question %} data-text="{{ question }}"{% endif %} data-on-success="REFRESH_PAGE" {% for key, attr in item.attributes.items %}data-{{ key }}="{{ attr }}" {% endfor %}>{{ item.name }}</a></div>
cms/templates/cms/toolbar/dragitem_menu.html
    <div class="cms-submenu-item"><a data-rel="add" href="{{ p.value }}">{{ p.name }}</a></div>
```

Afaics these are covered with tests.

Aiky30 (approved):
Awesome, and just like that, the world is a better place for all. :-)

fsbraun:
Wait: The first two are tested but do not detect missing links. I'll add two more tests.

fsbraun:
@Aiky30 Do you know if `cms/templates/toolbar/items/live_draft.html` is used at all in v4? I believe the edit, preview and publish buttons are generated programmatically in python.

Same with `cms/templates/cms/toolbar/toolbar_with_structure.html` (which is in use, obviously):  it seems that the `cms_toolbar.redirect_url` property is never used. Hence you never get the message "This page redirects to ...". I cannot go to edit a language that does not have a page content, correct?

Aiky30:
> @Aiky30 Do you know if `cms/templates/toolbar/items/live_draft.html` is used at all in v4? I believe the edit, preview and publish buttons are generated programmatically in python.
> 
> Same with `cms/templates/cms/toolbar/toolbar_with_structure.html` (which is in use, obviously): it seems that the `cms_toolbar.redirect_url` property is never used. Hence you never get the message "This page redirects to ...". I cannot go to edit a language that does not have a page content, correct? Hence the last part is not rendered (if not a custom toolbar causes it to be rendered):
> 
> ```
>     {% if cms_toolbar.redirect_url %}
>         <div class="cms-screenblock">
>             <div class="cms-screenblock-inner">
>                 <h1>{% trans "This page has no preview!" %}</h1>
>                 <p>{% trans "It is being redirected to:" %} <a href="{{ cms_toolbar.redirect_url }}">{{ cms_toolbar.redirect_url }}</a></p>
>             </div>
>         </div>
>     {% endif %}
> ```

You could possibly get this message if you have content in one language but you've not yet created content in another. 

cms/templates/toolbar/items/live_draft.html is possibly not used, although it could still be used and the name wasn't updated ...

fsbraun:
> You could possibly get this message if you have content in one language but you've not yet created content in another.

Is there an endpoint for that? It seems that the page content object needs to be created (and thereby the language) to edit it. If you can give me an edit endpoint for a non-existing language I can add a test for the message.

OK, it it could be used we better keep the `live_draft.html` template. There's not tests for it (and I see no way of testing it if it is not rendered at some point 😉).

Issue #6720, opened 2019-08-21
Is i18n get_current_language needed any more ?


stuaxo:
The [Django bug this references](https://code.djangoproject.com/ticket/9340) was shut 6 years ago, so maybe the function itself could be deprecated with a view to removing it ?

https://github.com/divio/django-cms/blob/develop/cms/utils/i18n.py#L75

```python
def get_current_language():
    """
    Returns the currently active language

    It's a replacement for Django's translation.get_language() to make sure the LANGUAGE_CODE will be found in LANGUAGES.
    Overcomes this issue: https://code.djangoproject.com/ticket/9340
    """
    language_code = translation.get_language()
    return get_language_code(language_code)

```

stuaxo:
Hi @Aiky30 know if anybody in DjangoCMS land can look at this and decide to fix it?

Any improvements to the docs probably help.

Aiky30:
Hi @stuaxo, Sure thing. I'll post it into the Slack channel and get it marked with a task todo.

stuaxo:
Cheers, trying not to be a drive by Dev too much these days :)

stuaxo:
Thanks for the fix @marksweb https://github.com/django-cms/django-cms/pull/7410

@marksweb @fsbraun it may worth adding the info to the docstring, but at least the info is there.

PR #7410, opened 2022-10-15
fix: Added deprecation warning to `get_current_language()`

marksweb:
## Description

This adds a deprecation warning to `get_current_language()` to address #6720.

After this I'll update all internal use of this function to use `translation.get_language()` from django itself (at least I think this is the correct course of action).

## Related resources

* #6720

## Checklist

### DIFF
--- a/cms/utils/i18n.py
+++ b/cms/utils/i18n.py
@@ -80,7 +80,8 @@ def get_current_language():
     """
     warnings.warn(
         "get_current_language() is deprecated; use django.utils.translation.get_language().",
-        DeprecationWarning
+        DeprecationWarning,
+        stacklevel=2
     )
     language_code = translation.get_language()
     return get_language_code(language_code)
--- a/cms/utils/i18n.py
+++ b/cms/utils/i18n.py
@@ -80,7 +80,7 @@ def get_current_language():
     """
     warnings.warn(
         "get_current_language() is deprecated; use django.utils.translation.get_language().",
-        warnings.DeprecationWarning
+        DeprecationWarning
     )
     language_code = translation.get_language()
     return get_language_code(language_code)
--- a/cms/utils/i18n.py
+++ b/cms/utils/i18n.py
@@ -1,3 +1,4 @@
+import warnings
 from contextlib import contextmanager
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


fsbraun:
We'd probably want to remove the use of `i18n.get_current_language()` from the codebase itself so that only third party packages will trigger the deprecation warning. Otherwise users might geht flooded by the warning despite them not using this function.

I've seen uses of `get_current_languages()`if a few files of the core:

    utils/i18n.py#L14
    forms/utils.py#L59,L81,L91
    models/pagemodel.py#L26,L#434

It also appears in a set of tests where I'd suggest to add the proposed `django.utils.translation.get_language()` code in addition to  `i18n.get_current_language()`.

marksweb:
@fsbraun Given that we don't bump major versions for something like this, I'd rather stick to a deprecation warning for now. It's following the way [django](https://github.com/django/django/blob/main/django/contrib/sitemaps/__init__.py#L133) removes things & gives notice.

If we just get rid of it, we might break third party packages suddenly & it'll be the users of those packages that find out first I imagine.

I have no issue with the project giving people warnings - by default I don't think you see them. But the point about warnings is that they are visible and people either update their code or raise issues on packages they depend on if warnings haven't been accounted for. Plus you can hide deprecation warnings - it's up to the project how they configure their logging.

fsbraun:
@marksweb Sorry, my comment was unclear.

 I was trying to refer to keeping the deprecation warning but replacing the calls to the deprecated function in the rest of the code base. This way users only see warnings if _they_ use the deprecated function and not as a result of using other bits of the CMS core that call the function internally. 

I guess that’s also what Django is doing. 
 
If I searched thoroughly only few places in the CMS codebase are affected.

marksweb:
@fsbraun Oh yeah, my plan is to do that in a different PR after we merge this, before we release either. So that the CMS itself doesn't use this function anymore.

fsbraun (approved):

cms/utils/i18n.py
```diff
@@ -77,6 +78,10 @@ def get_current_language():
     It's a replacement for Django's translation.get_language() to make sure the LANGUAGE_CODE will be found in LANGUAGES.
     Overcomes this issue: https://code.djangoproject.com/ticket/9340
     """
+    warnings.warn(
+        "get_current_language() is deprecated; use django.utils.translation.get_language().",
```
vinitkumar, line 82: Do you think it's a good idea to also add stacklevel to the warning.warn so that it shows users where they are using the deprecated code. 

https://www.youtube.com/watch?v=CtFdXBEwYfk

vinitkumar: Something like:

```python
warnings.warn(
    "get_current_language() is deprecated; use django.utils.translation.get_language().",
    DeprecationWarning, 
    stacklevel=2
)
```

vinitkumar (commented):
Suggested some improvements to the warnings module usage. Please check and see if you agree.

cms/utils/i18n.py
marksweb: Thanks @vinitkumar! That's really useful. Now I understand why you see so many different ways of raising deprecation warnings.

Issue #3157, opened 2014-05-10
graceful plugin exceptions


stefanfoulis:
Sometimes a plugin can cause an exception based on whatever is entered by the user, or because of other causes (e.g an api that is unreachable). Obviously this is a bug in the plugin, but it can happen and it can happen on production servers. Currently we let the exception bubble up and show a 500 page. In production this is a desaster, because with frontend editing, you can't even navigate the the page and remove the offending plugin anymore. The user is completely blocked concerning that page and has to get a developer to fix the bug before any more content can be entered on the page.

I propose that plugin rendering for each plugin has a try/except around it. To Visitors the offending plugin should just not be rendered (or an error message?). To content editors the plugin should be rendered in a way that is clear that something went wrong (perhaps with a red border and a message in the notification area at the top of the screen). In structure view the plugin should also be highlighted. In any case it should be possible to still edit or remove the plugin.
In any case the error should be logged with a traceback.

This behaviour should be controlled by a new setting that uses `DEBUG` as a default but can be overridden.

sidneijp:
Hi,

For those who can't wait for a release that fix it, we've develop a simple workaround for it. It basically dynamically decorates all registered plugins' `render` method to handle exceptions with a default fallback template (configurable). We recommend to call it just after `django.setup` in `project/wsgi.py`.

\<project\>/decorators.py
```
# coding: utf-8

from django.conf import settings
from functools import wraps
import logging
import traceback

logger = logging.getLogger(__name__)


def get_render_template(context, instance, placeholder):
    """Returns configured default template to show on plugins rendering exceptions"""
    return getattr(settings, 'CMS_PLUGIN_RENDER_ERROR_TEMPLATE', 'cms/render_plugin_error.html')


def plugin_exception_handler(func):
    """Decorates plugin's `render` method to set a default template as fallback
    for exceptions when rendering it. It prevents the whole page from break cause
    of a faulty plugin."""
    global get_render_template

    @wraps(func)
    def wrapper(self, *args, **kwargs):
        try:
            result = func(self, *args, **kwargs)
        except Exception, ex:
            logger.error(self.__class__)
            traceback.print_exc()
            logger.exception(ex)
            self.get_render_template = get_render_template
            result = {}
        return result
    return wrapper
```

\<project\>/utils.py
```
# coding: utf-8

def wrap_plugins_render_method(decorator):
    """Wraps all plugins' `render` method with `decorator`"""
    from cms.plugin_pool import plugin_pool
[...]

PR #7423, opened 2022-10-28
feat: graceful plugin exceptions

fsbraun:
## Description

Plugins raising an exception making the page uneditable. As a bit of sugar this PR catches exceptions that are raised

* within a plugins `render` method
* when renderings its `render_template`

Exceptions are logged (on error level) and developers can use Django's `LOGGING` setting to configure on how to process the logs.

Local development servers will typically show the traceback on the console.

Should an exception occur a placeholder with a short notice is rendered in edit mode. Double clicking on it will open the edit modal for the plugin. This way potential misconfigurations can be changed.

## Related resources

* #3157 

## Checklist


marksweb (approved):

cms/plugin_rendering.py
```diff
@@ -472,6 +493,19 @@ def render_plugin(self, instance, context, placeholder=None, editable=False):
             placeholder_cache.setdefault('plugins', []).append(instance)
         return mark_safe(content)
 
+    def render_exception(self, action, instance, context, placeholder, editable):
+        if editable:
+            exc, value, traceback = context['exc_info']
+            return f'<div style="display: inline-block; border: none; background: rgba(0, 0, 0, 0.5); ' \
```
vinitkumar, line 499: Hi @fsbraun. Do you think it is a good idea to write it like this inside `"""`?

```python
>>> a = 1
>>> b = 2
>>> sentence = f"""
... <div>
... {a}: {b}
... </div>
... """
>>> sentence
'\n<div>\n1: 2\n</div>\n'>>> a = 1
>>> b = 2
>>> sentence = f"""
... <div>
... {a}: {b}
... </div>
... """
>>> sentence
'\n<div>\n1: 2\n</div>\n'
```

This should just work, except that you don't have to escape each line using `\`

vinitkumar (commented):
@fsbraun Made a remark about alternative representation of string. Please let me know if you agree or disagree?

PR #7436, opened 2022-11-14
Add docstrings for autodoc

fsbraun:
## Related resources

This PR adds documentation docstring for sphinx' autodoc feature to pull from the source code.

* In `cms/models/pluginmodel.py/` it turns out that a defunct `copy_plugin` function was left over from versions 3.x
* This method is not covered by tests and in its original version must fail on each single call since it accesses treebeard properties of a plugin.
* This PR removes the `CMSPlugin.copy_plugin` function.
* For naming consistency, this PR renames the function `create_title` to `create_page_content` in `cms/api.py`. For ensure compatibility, a `create_title` function is retained. It issues a deprecation warning and calls `create_page_content`. **This will require a change in the monkey patch routine of djangocms-versioning**.  

## Checklist


marksweb (approved):
This looks good @fsbraun 

It doesn't need it, but it could also test that the old `create_title` raises the deprecation warning.

marksweb (approved):
Looks good @fsbraun 

The only thing I thought of was that there could be a test for ensuring `create_title` raises the warning.

PR #7439, opened 2022-11-14
ci: Introduce patching api to avoid side effects with patching

fsbraun:
## Description

Since django CMS v4 some packages patch the django CMS core to supply their features, esp. djangocms-versioning and djangocms-moderation.

To alert all contributors to this situation, this PR introduces rules for patches. **This is to ensure that future community contributors can avoid breaking packages that patch the django CMS core**:

* Patched methods, functions or properties require to be marked with the `@patch_hook` decorator. Example:
    ```
    from cms.utils.patching import patch_hook

    class MenuRenderer(object):
        ...
        @patch_hook
        @property
        def cache_key(self):
            ...
    ```
* Patching packages start using the `patch_cms` api. It essentially is equivalent to `setattr` but checks if the method, function or property has the `@patch_hook` decorator and otherwise raises an `ImproperlyConfigured` exception. The implementation is done so that patching remains fully backward compatible. Example:
    ```
    try:
         from cms.utils.patching import patch_cms
    except ModuleNotFoundError:
        patch_cms = setattr  # backwards compatibility

    patch_cms(MenuRenderer, "cache_key", property(menu_renderer_cache_key))
    ```

* Patching unmarked methods, functions, or properties becomes illegal. The intention is not at add more patch hooks than necessary.

## Related resources

* https://github.com/django-cms/djangocms-versioning/pull/298
* PR for djangocms-moderation is forthcoming

## Checklist

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -54,6 +54,7 @@ from cms.utils.i18n import (
     get_language_list, get_language_object, get_language_tuple,
     get_site_language_from_request,
 )
+from cms.utils.patching import patch_hook
 from cms.utils.plugins import copy_plugins_to_placeholder
 from cms.utils.urlutils import admin_reverse
 
@@ -832,6 +833,7 @@ class PageContentAdmin(admin.ModelAdmin):
             preserved_filters['language'] = lang
         return preserved_filters.urlencode()
 
+    @patch_hook
     def get_queryset(self, request):
         site = get_site(request)
         languages = get_language_list(site.pk)
@@ -1186,6 +1188,7 @@ class PageContentAdmin(admin.ModelAdmin):
 
         return HttpResponse(force_str(_("The template was successfully changed")))
 
+    @patch_hook
     @require_POST
     @transaction.atomic
     def copy_language(self, request, object_id):
--- a/cms/api.py
+++ b/cms/api.py
@@ -32,6 +32,7 @@ from cms.utils import get_current_site
 from cms.utils.conf import get_cms_setting
 from cms.utils.i18n import get_language_list
 from cms.utils.page import get_available_slug, get_clean_username
+from cms.utils.patching import patch_hook
 from cms.utils.permissions import _thread_locals
 from cms.utils.plugins import copy_plugins_to_placeholder
 from menus.menu_pool import menu_pool
@@ -196,6 +197,7 @@ def create_page(title, template, language, menu_title=None, slug=None,
     return page
 
 
+@patch_hook
 @transaction.atomic
 def create_title(language, title, page, menu_title=None, slug=None,
                  redirect=None, meta_description=None, parent=None,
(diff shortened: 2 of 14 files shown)


marksweb:
This sounds very sensible to me. I like the look of this 👌

Aiky30:
The idea of the patching was to first pull everything publishing related out of the core. Then where possible using the monkey patching as a guide to where needs to be updated in the core, add proper handles in the code that are "configurable" by using the app configuration created in the cms. Rather than thinking about how to make the core "patchable", it should be steered towards making it "configurable". This will need design and lots of thought to do right.

Aiky30:
An example of how djangocms-versioning uses the cms config to provide itself for any "versioned" package, very powerful and clever I'm sure that you'll agree: https://github.com/django-cms/djangocms-versioning/blob/master/djangocms_versioning/cms_config.py

Aiky30:
Here's a different type of configuration, moderation allows custom controls to be appended to the admin "Actions" column: https://github.com/django-cms/djangocms-moderation/blob/79aaae4558faf9c80292b493b36e35ff3cd28577/djangocms_moderation/cms_config.py#L28

fsbraun:
@Aiky30 @marksweb I totally agree with the objective of removing the monkey patches! 

I've had lots of thought on all the patching that djangocms-versioning does and in fact it seems only moderately complex to replace them (moderation patches versioning - I do not talk about that now). To me patches come in several categories:

- **Toolbar patches**: The alternative here is to allow third party apps to add mixins to `cms.toolbars.toolbars.CMSToolbar`. Just another `cms_config.py` setting.
- **Creating `Version` instances if a versioned model instance is created**. This can be automated by modifying the `PublishedContentManagerMixin`. It implies that other packages using versioning will no longer need to create the `Version` object manually. This is probably the biggest change necessary but it implies that the core can stay fully ignorant of whatever versioning package you want to use in the future.
- **Admin patches** can be moved to the already existing `VersioningCMSPageAdminMixin`
- **Extension patches** can be covered by allowing signals to react to changes in extension models.
- **Patches that revert the effect of the `PublishedContentManagerMixin`**: Typically they replace `PageContent.objects` by `PageContent._original_manager` to gain access. Those can be rewritten in the core to work both with  a custom manager or without.
- **Patches independent of versioning** (typically improving the core in some way): They can go into the core.

There is a patch to the Django ORM (removing the `unique_together` constraint for `PageContent`). This should stay.

The conclusion is: **We need to make a design decision on how to create the `Version` instances if a versioned object is created.** My suggestion is to move this to the `PublishedContentManagerMixin`. An important aspect is that the `Version` model needs a user instance which the versioned model might not have (as, e.g., djangocms-alias). I suggest the following new syntax for automatic creation of a `Version` model:

```
my_new_instance = MyModel.objects.with_user(request.user).create(**contents)
```

For backward compatibility the creation of the `Version` model may be left to the calling app if no `with_user()` is given (with a warning issued).

What do you think?

Issue #7400, opened 2022-09-21
[BUG] CMS V4: Copying and pasting an unpublished page renders the page tree unusable


fsbraun:
Copying and pasting an unpublished page renders the page tree and thereby the whole cms unusable. `./manage.py cms fix-tree` will not repair this tree.

## Steps to reproduce

Steps:
1.	Start from empty project with `./manage.py migrate`, `./manage.py createsuperuser`, and `./manage.py runserver`
2.	Go to `localhost:8000/admin` and log in.
3.	Create new page content (e.g., test)
4.	Copy the page from the burger menu
5.	Paste the page either from the global burger menu or from the page’s burger menu.

## Expected behaviour

The copied page should appear in the page tree.

## Actual behaviour

Page tree breaks with message “Error: Could not load node”. Switching languages does not help. The server reports an server error when serving `/admin/cms/pagecontent/get-tree/`.

The `cms_page` table contains two entries. The `cms_pagecontent` table contains one entry after the exception has been raised.

## Screenshots

![page-tree-bug](https://user-images.githubusercontent.com/16904477/191518386-20f9513e-bfd3-477a-9542-1f7cf45b2f4f.gif)

## Additional information (CMS/Python/Django versions)

* Django 3.2.12
* django-cms@develop-4
* djangocms-text-ckeditor@support/4.0.x
* djangocms-versioning@master

## Do you want to help fix this issue?

* [ ] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 
* [x] No, I only want to report the issue.

fsbraun:
@marksweb @Aiky30 I have been able to track down why this bug appears. **It appears that the page tree cannot render `EmptyPageContent`**. If - for what reason - there is no `PageContent` object this error prevents the page tree from being rendered.

Remedies:
1. Make the page tree not raise a server error if it contains an `EmptyPageContent` (this is an **[easy fix](https://github.com/fsbraun/django-cms/commit/e24eaabf3eb2bb2f9f1b80c01368a83e5ea2ed68)**)
2. Ensure that copying a page also copies the `PageContent` objects. The `Page` models copies `self.pagecontent_set.all()` which evaluates to an empty query set if `djangocms-versioning` is installed. **Does versioning change the way `pagecontent_set` needs to be evaluated?** (I would assume, the draft page contents should be copied.)
[...]

PR #7446, opened 2022-11-25
fix: Remove patching of PageContent by djangocms-versioning

fsbraun:
## Description

Patching easily introduces side-effects, especially if patches happen at diverse places in the code base. This PR removes the need for djangocms-versioning to patch the core.

## Additional interface through CMSConfig

The toolbar is at the heart of django CMS. This PR allows apps to register a mixin for the `cms.toolbar.toolbar.CMSToolbar` class. All it has to do is to set 

```
class MyCMSConfig(CMSAppConfig)
    cms_enabled = True
    cms_toolbar_mixin = MyMixinClass
```

djangocms-versioning (and other apps) can use this config instead of patching `CMSToolbar`. (Example for other use cases: Change the login process through the toolbar)

## Creating `Version` instances when a versioned model is created

So far, djangocms-versioning puts the burden of creating `Version` objects for versioned models on the app owning the versioned models. To keep the core agnostic of versioning capabilities. This PR (https://github.com/django-cms/djangocms-versioning/pull/300) lets djangocms-versioning automatically create a draft `Version` object as soon as a versioned object is created. This is achieved by modifying the versioned model's manager's `create` method. 

This removes the need to patch the creation of `PageContent` objects.

## Patches independent of versioning

Quite a few patches of djangocms-versioning are not connected to the versioning capability. They typically are improvement of the core and have been moved into the core with this PR.

## Ability to "see" unpublished `PageContent` objects

A number of patches by djangocms-versioning enable to core to see unpublished `PageContent` objects. An example are the Page wizards which need to be able to access newly created PageContent objects. This is done by using the Django manager `_base_manager` in these selected cases.

## Related resources

* https://github.com/django-cms/djangocms-versioning/pull/300
* #7439 becomes unnecessary with this patch
* #7400 will be solved.

## Checklist


cms/models/contentmodels.py
```diff
@@ -118,6 +118,13 @@ class PageContent(models.Model):
     )
 
     objects = PageContentManager()
+    admin_manager = PageContentAdminManager()
```
Aiky30, line 121: Could this new manager not inherit the existing `PageContentAdminManager` manager?

cms/utils/moderator.py
```diff
@@ -1,26 +0,0 @@
-
```
Aiky30, line 1: May need to document / deprecate the removal of these methods. One to document somewhere, potentially the Release document.

menus/menu_pool.py
```diff
@@ -104,7 +103,9 @@ def __init__(self, pool, request):
         self.request = request
         self.request_language = get_language_from_request(request, check_path=True)
         self.site = Site.objects.get_current(request)
-        self.draft_mode_active = use_draft(request)
+        toolbar = getattr(request, "toolbar", None)
+        # No toolbar - no draft mode
+        self.draft_mode_active = toolbar.edit_mode_active or toolbar.preview_mode_active if toolbar else False
```
Aiky30, line 108: The CMS has no concept of draft mode, this may be old logic that's never used (would have thrown an error if it did). Should be removed and if it is used anywhere replaced with bool(toolbar.edit_mode_active or toolbar.preview_mode_active) anywhere it is still referenced

. 
```suggestion
```

cms/api.py
```diff
@@ -321,7 +322,9 @@ def create_page_content(language, title, page, menu_title=None, slug=None,
         language=language,
     )
 
-    page_content = PageContent.objects.create(
+    # E.g., djangocms-versioning needs an User object to be passed when creating a versioned Object
+    user = getattr(_thread_locals, "user", "unknown user")
```
Aiky30, line 326: I'm not sure a default should be returned here, it needs to be an actual object not a string. If there's no user found it should error.

cms/api.py
```diff
@@ -533,7 +536,8 @@ def publish_page(page, user, language):
 
         For publishing functionality see `djangocms-versioning: <https://github.com/django-cms/djangocms-verisoning>`_
     """
-    warnings.warn('This API function has been removed', UserWarning)
+    warnings.warn('This API function has been removed. For publishing functionality use djangocms-versioning.',
```
Aiky30, line 539:
```suggestion
    warnings.warn('This API function has been removed. For publishing functionality use a package that adds publishing, such as: djangocms-versioning.',
```

Slight reword to state that djangocms-versioning is not the only package you can use.

cms/api.py
```diff
@@ -544,7 +548,8 @@ def publish_pages(include_unpublished=False, language=None, site=None):
...
```
Aiky30, line 551:
```suggestion
    warnings.warn('This API function has been removed. For publishing functionality use a package that adds publishing, such as: djangocms-versioning.',
```

cms/api.py
```diff
@@ -555,7 +560,8 @@ def get_page_draft(page):
 
         For draft functionality see `djangocms-versioning: <https://github.com/django-cms/djangocms-verisoning>`_
     """
-    warnings.warn('This API function has been removed', UserWarning)
+    warnings.warn('This API function has been removed. For publishing functionality use djangocms-versioning.',
```
Aiky30, line 563:
```suggestion
    warnings.warn('This API function has been removed. For publishing functionality use a package that adds publishing, such as: djangocms-versioning.',
```

cms/utils/helpers.py
```diff
@@ -109,19 +109,3 @@ def is_editable_model(model_class):
     except KeyError:
         return False
     return isinstance(admin_class, FrontendEditableAdminMixin)
-
-
-def get_admin_model_object_by_id(model_class, obj_id):
```
Aiky30, line 114: This was an early attempt at not just blindly patching the codebase. I prefer the new manager idea, much cleaner.

Aiky30 (commented):
Surprisingly a lot of old references to draft still in places. Well found. 

Some comments, 2 of any concern:
Page content creation user
Removal of force language, could be replaced / the original requirement sourced back to the wizard changes.

menus/menu_pool.py
fsbraun: This needs the toolbar which is not generally available in the menu. In practice the confusing thing is the legacy name `draft_mode_active`. In v4, the menu on the website is cached separately from the menu in edit or preview. I'd rather rename this.

cms/api.py
fsbraun: @Aiky30 So far, the requirement for a user object was only introduced by versioning. While I'd second the same requirement for the core it breaks the entire test suite: 573 out of 1025 tests do not provide a user object. Once versioning is active versioning will raise an error if there is no User object. Hence the default leads to the same behavior as with 4.0.x.

cms/utils/moderator.py
fsbraun: I have started a Deprecation PR for v3. Will add those into it.

cms/cms_toolbars.py
```diff
@@ -576,11 +570,6 @@ def add_page_menu(self):
             # first break
             current_page_menu.add_break(PAGE_MENU_FIRST_BREAK)
 
-            # page edit
-            with force_language(self.current_lang):
```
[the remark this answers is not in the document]
fsbraun: @Aiky30 I have deleted the `Page -> Edit` menu entry here. 

Largely, I did this because **with versioning it was a different action compared to the `Edit` button** (which potentially creates a new draft). So either versioning needs to replace the `Page -> Edit` menu entry (just like the language menu) to ensure both trigger the same action. Due to its limited usability (in the middle of a menu as opposed to the 1-click-button) I figured it is not worth the effort to replace it in versioning.

Frankly, I still am of this opinion, but had not put forward the reasons. In the end I am open to keeping it but I doubt it is of any use.

cms/models/contentmodels.py
fsbraun: Sorry, not sure what you mean here. I'd keep "user" and "admin" managers as separate as possible.

cms/cms_toolbars.py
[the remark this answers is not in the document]
fsbraun: I forgot one more argument: Since the toolbar now supports all content types they all should make similar use of menus etc.: All content-type specific actions go into a menu (Page settings, Alias settings etc.) Edit/Preview/Revert etc. are buttons in the toolbar. Alias, e.g., does **not** have an Edit option in its menu.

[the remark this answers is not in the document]
marksweb: I'm in favour of edit being part of the toolbar, not the menu.

That's basically the only way I ever go to edit a page. Same can be said for everyone I've worked with.

cms/api.py
Aiky30: Ok.

cms/models/contentmodels.py
Aiky30: It was more that the admin manager first inherits from the none admin manager, ensuring that any future changes to the standard manager are inherited. For sure the changes tot he admin manager should be separate. 

```
class PageContentAdminManager(PageContentManager):
    ...
```

cms/cms_toolbars.py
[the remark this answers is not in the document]
Aiky30: If another package (djangocms-versioning) in this case, needs to make changes, we can provide the functionality as default int he cms_config, and it can be overridden by other packages. We have to be careful that any existing functionality, although we may not personally use it, should still continue to work where possible.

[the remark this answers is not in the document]
fsbraun: I don't want to get hung up about this. Perhaps this is a different discussion. I reintroduced the `Page->Edit` menu entry.

cms/models/contentmodels.py
fsbraun: OK. Got it. Implemented it.

Aiky30 (approved):
Looks good. Read through it thoroughly once again. Nice to have a proper solution to the patching finally.

PR #7468, opened 2023-01-04
Update install.rst

bencipher:
Updated steps for installing and configuring sekizai

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/docs/how_to/install.rst
+++ b/docs/how_to/install.rst
@@ -210,16 +210,15 @@ After each of the steps below run ``cms check`` to verify that you have resolved
 Sekizai
 =======
 
-`Django Sekizai <https://github.com/ojii/django-sekizai>`_ is required by the CMS for static files management. You need
-to have::
+`Django Sekizai <https://github.com/ojii/django-sekizai>`_ is required by the CMS for static files management. For that, you need to install the package, add it to installed apps and to your context_processors by following steps::
+Install the package:
+     'pip install django-sekizai'
 
-     'sekizai'
-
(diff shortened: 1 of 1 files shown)


docs/how_to/install.rst
```diff
@@ -210,16 +210,15 @@ After each of the steps below run ``cms check`` to verify that you have resolved
 Sekizai
 =======
 
-`Django Sekizai <https://github.com/ojii/django-sekizai>`_ is required by the CMS for static files management. You need
-to have::
+`Django Sekizai <https://github.com/ojii/django-sekizai>`_ is required by the CMS for static files management. For that, you need to install the package, add it to installed apps and to your context_processors by following steps::
```
marksweb, line 213: Sekizai is one of the requirements of the cms so we shouldn't need to specify installing it separately.

See here: https://github.com/django-cms/django-cms/blob/develop/setup.py#L13

marksweb (commented):
I'm not sure this change is required. The package is installed as a dependency.

The syntax of these changes is also incorrect as you csn see in the test failures.

bencipher:
Yes you are correct about not needing to explicitly install the Selizai dependency. I think the doc was not formatted for clear interpretation at the configuration of sekizai. I could remove the PR or just remove the installation script for the Sekizai library, your call.

docs/how_to/install.rst
bencipher: Yes, I can remove the install line. I think for some reasons, it wasn’t installed for me and I had to manually install it and packaging library which was why I added it in the docs

PR #7488, opened 2023-01-25
fix: replace ' by ′ in fr translation − no more "page d\u0027accueil"!

corentinbettiol:
We can use this method (replace by a typographically valid char, but which does nothing in javascript) on other translations when there's a problem with  `escapejs` (ie using [those chars](https://github.com/django/django/blob/34b328814976a2e2f7907361a494202763649f3f/django/utils/html.py#L31-L42)) in order to fix those slight annoyances.

* [ ] I have added or modified the tests when changing logic (can't think of a way to retrieve all the translations that are displayed after a call to `escapejs`, so no test for now for this)

### DIFF
--- a/cms/locale/fr/LC_MESSAGES/django.po
+++ b/cms/locale/fr/LC_MESSAGES/django.po
@@ -1,7 +1,7 @@
 # SOME DESCRIPTIVE TITLE.
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the PACKAGE package.
-# 
+#
 # Translators:
 # adrien.delhorme, 2014
 # Adrien Delhorme, 2017,2019,2021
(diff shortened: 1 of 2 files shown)


fsbraun:
Thanks @corentinbettiol ! Since this are only a few changes, I can implement them using transifex. They will be pulled automatically with the next release.

fsbraun:
(Also I changed the django CMS 4 translations accordingly)

corentinbettiol:
Thanks!

vinitkumar (approved):

wiebren:
@corentinbettiol 
This seems to have introduced a bug in the `actions_dropdown.html` template:

https://github.com/django-cms/django-cms/commit/99c8843811929fe179b39961c32c8b2b5ce95a70#diff-d53f7c96ea020716cb9e33bd4a5aced4b41cbf0a463f7e07a278d4b6f3bd7237R108

`{{% trans "last change on" %}` should be `{% trans "last change on" %}`

corentinbettiol:
Oops, didnt saw that.

@fsbraun, what should I do to fix this thing? Add a new commit to this PR?

fsbraun:
The build process will have generated quite a few changes since the merge of this PR. I think it's best to create a micro PR to fix this! Can you do it, @corentinbettiol ?

corentinbettiol:
Yes, I will do that!

PR #7489, opened 2023-01-25
feat: add setting so redirect preserve params

igobranco:
## Description
Added a new setting that allow to configure globally if the django-cms
redirects preserving the query parameters.
`REDIRECT_PRESERVE_QUERY_PARAMS`.
This feature is usefull for example:
1. marketing campains extra parameters,
2. social networks extra parameters like `fbclick`,
3. custom developed parameters, after that page has been moved, the
older URLs for that page should preserve the functionality.

## Related resources
None

## Checklist

* [X] I have read the [contribution guidelines ](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

### DIFF
--- a/cms/tests/test_views.py
+++ b/cms/tests/test_views.py
@@ -152,6 +152,31 @@ class ViewTests(CMSTestCase):
             response = self.client.get(self.get_toolbar_disable_url(page_url))
             self.assertEqual(response.status_code, 302)
 
+    def test_redirect_not_preserving_query_parameters(self):
+        # test redirect checking that the query parameters aren't preserved
+        redirect = '/en/'
+        one = create_page("one", "nav_playground.html", "en", published=True,
+                          redirect=redirect)
+        url = one.get_absolute_url()
+        params = "?param_name=param_value"
+        request = self.get_request(url + params)
+        response = details(request, one.get_path())
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response['Location'], redirect)
+
+    @override_settings(CMS_REDIRECT_PRESERVE_QUERY_PARAMS=True)
+    def test_redirect_preserving_query_parameters(self):
+        # test redirect checking that query parameters are preserved
+        redirect = '/en/'
+        one = create_page("one", "nav_playground.html", "en", published=True,
+                          redirect=redirect)
+        url = one.get_absolute_url()
+        params = "?param_name=param_value"
+        request = self.get_request(url + params)
+        response = details(request, one.get_path())
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response['Location'], redirect + params)
+
     def test_login_required(self):
         self.create_homepage("page", "nav_playground.html", "en", published=True, login_required=True)
         plain_url = '/accounts/'
--- a/cms/views.py
+++ b/cms/views.py
@@ -162,6 +162,10 @@ def details(request, slug):
             toolbar.redirect_url = redirect_url
         elif redirect_url not in own_urls:
             # prevent redirect to self
+            if get_cms_setting('REDIRECT_PRESERVE_QUERY_PARAMS'):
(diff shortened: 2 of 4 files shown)


corentinbettiol:
Not against it, but I would like this to be an opt-out by default (can add a setting in order to enable it), since I think not *all* websites would want to preserve query parameters.

igobranco:
> Not against it, but I would like this to be an opt-out by default (can add a setting in order to enable it), since I think not _all_ websites would want to preserve query parameters.

Agree. 

I've changed the PR to add it has setting. By default it preserves the existing behavior.

Alternatively, I think this feature could be implemented adding the preserve query parameters as a boolean option on the Advanced Settings Form.

cms/views.py
```diff
@@ -162,6 +162,10 @@ def details(request, slug):
             toolbar.redirect_url = redirect_url
         elif redirect_url not in own_urls:
             # prevent redirect to self
+            if getattr(settings, 'REDIRECT_PRESERVE_QUERY_PARAMS', False):
```
marksweb, line 165: Could you change this to use `get_cms_setting` please.

It'll need the default adding here; https://github.com/django-cms/django-cms/blob/develop/cms/utils/conf.py#L43

And this will mean that the setting will be valid in the `CMS_` namespace. So the setting will be `CMS_REDIRECT_PRESERVE_QUERY_PARAMS`

cms/tests/test_views.py
```diff
@@ -152,6 +152,19 @@ def test_redirect_with_toolbar(self):
             response = self.client.get(self.get_toolbar_disable_url(page_url))
             self.assertEqual(response.status_code, 302)
 
+    @override_settings(REDIRECT_PRESERVE_QUERY_PARAMS=True)
```
marksweb, line 155: Is there also a test that ensures that params are removed on redirect?

marksweb (changes requested):

cms/views.py
igobranco: Done. 
Changed to use `get_cms_setting` and added on the `DEFAULTS`.

cms/tests/test_views.py
igobranco: Nop, it weren't. 

I've added 1 new test. So this PR has 2 new tests:
- test_redirect_not_preserving_query_parameters
- test_redirect_preserving_query_parameters

igobranco:
@marksweb I've also documented the new setting `CMS_REDIRECT_PRESERVE_QUERY_PARAMS` on `docs/reference/configuration.rst`.

fsbraun (approved):

vinitkumar (approved):

docs/reference/configuration.rst
```diff
@@ -1190,3 +1190,14 @@ when the "Content" field is filled in. There should be no need to change it,
...
+
+..  setting:: CMS_REDIRECT_PRESERVE_QUERY_PARAMS
+
+CMS_REDIRECT_PRESERVE_QUERY_PARAMS
+===================================
+
+default
+    ``False``
+
+This indicates the CMS that the redirects should preserve the query
```
marksweb, line 1202:
```suggestion
This indicates to the CMS that redirects should preserve the query
```

marksweb (changes requested):
Just a minor thing on the docs wording. Other than that - this is looking great now. Thanks!

docs/reference/configuration.rst
igobranco: Done. Thanks!

marksweb (approved):

igobranco:
I've rebased the branch, if someone with write privilegie could merge.
@marksweb @fsbraun @vinitkumar

PR #7493, opened 2023-01-31
fix: Copy page did not create versions for subpages

fsbraun:
## Description

This is an urgent fix when copying pages: 

* Children of the page copied did not have Version objects automatically created if djangocms-versioning is installed. 
   Installations can be fixed by running `./manage.py create_versions --username <some_username>` 
* Also, the `./manage.py cms copy` command now requires either a user id or a user name to create new page content models. They can be specified using the command line parameters `--username xy` or `--userid 23`.
* Finally, the "Duplicate page" now creates the correct Version objects automatically.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -449,7 +449,9 @@ class Page(models.Model):
         from cms.models import PageContent
         from cms.utils.page import get_available_slug
 
-        assert user is not None, "No anonymous page copying"
+        if user is None:
+            raise ValueError("Since django CMS 4 the page.copy method requires a user argument")
+
         if parent_node:
             new_node = parent_node.add_child(site=site)
             parent_page = parent_node.item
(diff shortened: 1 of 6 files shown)


cms/models/pagemodel.py
```diff
@@ -449,6 +449,9 @@ def copy(self, site, parent_node=None, language=None,
         from cms.models import PageContent
         from cms.utils.page import get_available_slug
 
+        if user is None:
```
marksweb, line 452: I feel like this would be better as 
```suggestion
        if not user:
```

Because there's not a type check so it could be improperly configured by doing the explicit `None` check.

For example;

```
>>> user = None
>>> if user is None:
...     print("Here")
... 
Here
>>> user = []
>>> if user is None:
...     print("Here")
... 
>>>
```

marksweb (changes requested):

marksweb (approved):

PR #7500, opened 2023-02-04
 feat: Improved admin_manager

fsbraun:
## Description

The `admin_manager` provided access to `current_content_iterator` which allowed to "filter" out all current versions of the model if some sort of versioning is installed by returning an iterator.

This method is now replaced by `current_content()` which is a filter that returns a queryset that only contains the current verison of the objects. Not only is it more convenient to have a qs returned, djangocms-versioning now can implement this much more efficiently.

## Related resources

* https://github.com/django-cms/djangocms-versioning/pull/318
* #...

## Checklist

### DIFF
--- a/cms/models/managers.py
+++ b/cms/models/managers.py
@@ -130,8 +130,8 @@ class PageContentAdminQuerySet(models.QuerySet):
         return self.filter(**kwargs)
     
     def latest(self):
-        """Returns the latest version (if a versioning package is installed). 
-        Without versioning every page content is the latest."""
+        """Returns the latest version (if a versioning package is installed) including discared 
+        or unpublished page content. Without versioning every page content is the latest."""
         return self
 
 
--- a/cms/models/managers.py
+++ b/cms/models/managers.py
@@ -125,13 +125,14 @@ class PageContentManager(WithUserMixin, models.Manager):
 
 class PageContentAdminQuerySet(models.QuerySet):
     def current_content(self, **kwargs):
-        """Returns the currently valid content that matches the filter given in kwargs.
-        (if a versioning package is installed). Without versioning every page us current."""
+        """If a versioning package is installed, this returns the currently valid content 
+        that matches the filter given in kwargs. Used to find content to be copied, e.g.. 
+        Without versioning every page is current."""
         return self.filter(**kwargs)
 
     def latest_content(self, **kwargs):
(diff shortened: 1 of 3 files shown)


cms/models/managers.py
```diff
@@ -124,8 +124,16 @@ def get_title(self, page, language, language_fallback=False):
 
 
 class PageContentAdminQuerySet(models.QuerySet):
-    def current_content_iterator(self, **kwargs):
-        return iter(self.filter(**kwargs))
+    def current_content(self, **kwargs):
```
marksweb, line 127: Because these both filter on arbitrary kwargs, it doesn't seem right to have two methods and to name them like this - you could have kwargs that lead to it filtering on something other that current/latest content.

fsbraun: The `kwargs` are not needed but provide a little syntactic sugar to allow for something like `PageContent.admin_manager.current_content(language="en")` wich would be identical to the longer `PageContent.admin_manager.filter(language="en").current_content()`. (Both give a qs of all "current" page contents in English - without versioning that's just all.)

Typically (but as you indicate correctly not necessarily) those kwargs would be one of the "extra grouping fields" known to versioning but unknown to the core. For page content we known it to be `language`. An alternative would be to test if `kwargs` contains anything else than a language argument and throw an exception. 

If we were to fully remove the `kwargs`, the API of djangocms-versioning would need an according change (since it adds non-trivial functionality to the method).

The kwargs-filtering is not used in the core code base but in a test which would need to be adjusted.

@marksweb , what's your take?

In any case I'll explain the existence of the methods in a docstring.

vinitkumar:
@fsbraun @marksweb What is blocking this PR to get merged?

vinitkumar (approved):

fsbraun:
@vinitkumar Thanks for the review! We're having a quick discussion of the signature of the filter https://github.com/django-cms/django-cms/pull/7500#discussion_r1105169217. Do you have any take on that?

cms/models/managers.py
marksweb: @fsbraun I'm not talking about removing the passing kwargs. That makes absolute sense.

It's the fact that two methods both return `self.filter(**kwargs)` that seems odd. Their names suggest that one should get the latest content and the other should get the current content. But without kwargs, they don't actually do either of those things.

PR #7504, opened 2023-02-14
ci: introduce ruff in place of flake8 for improved speed

vinitkumar:
Replace flake8 with ruff as it is mostly compatible but offer about 50x speedup. It also aims to replace many other tools like isort, flake8 plugins with one single tool.

Also, fixes some code quality checks suggested by ruff


cms/test_utils/testcases.py
```diff
@@ -433,23 +434,41 @@ def update(self, response):
...
+
     def failUnlessWarns(self, category, message, f, *args, **kwargs):
         warningsShown = []
-        cleanwarningsShown = []
         result = _collectWarnings(warningsShown.append, f, *args, **kwargs)
 
         if not warningsShown:
             self.fail("No warnings emitted")
-
-        for warning in warningsShown:
```
marksweb, line 444: How come this has been removed?

cms/utils/compat/forms.py
```diff
@@ -33,9 +33,9 @@
         else:
             from django.contrib.auth.forms import UserChangeForm
     except ImportError:
-        from django.contrib.auth.forms import UserChangeForm  # nopyflakes
-        from django.contrib.auth.forms import UserCreationForm  # nopyflakes
+        from django.contrib.auth.forms import UserChangeForm  # noqa
```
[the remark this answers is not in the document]
marksweb: If all of these are unused - can we remove them while they've first been flagged?

ruff.toml
```diff
@@ -0,0 +1,20 @@
+exclude = [
```
marksweb, line 1: Would rather we have a `pyproject.toml` to start bringing configs together in 1 place I think.

vinitkumar: Agreed.

ruff.toml
```diff
@@ -5,6 +5,7 @@ exclude = [
 ]
 ignore = [
```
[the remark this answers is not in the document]
vinitkumar: Done!

marksweb:
@vinitkumar I think it's worth adding the `I` job so we can remove isort as well.

cms/test_utils/testcases.py
```diff
@@ -433,23 +434,41 @@ def update(self, response):
         request._messages = MockStorage()
         return request
 
+    # TODO: Renable this.
```
fsbraun: I believe this is an outdated change. The tests should not be affected.

fsbraun (changes requested):
Probably the changes to the `testcases.py` file are not meant to be part of this PR. They seem outdated. Can you remove them?

Otherwise I feel that readability improves a lot. 👍

cms/api.py
```diff
@@ -437,7 +437,7 @@ def publish_pages(include_unpublished=False, language=None, site=None):
         qs = qs.filter(node__site=site)
 
     output_language = None
-    for i, page in enumerate(qs):
+    for _i, page in enumerate(qs):
```
fsbraun, line 440:
```suggestion
    for page in qs:
```

cms/toolbar/toolbar.py
```diff
@@ -191,7 +191,7 @@ def init_toolbar(self, request, request_path=None):
             self.clipboard = user_settings.clipboard
 
         if hasattr(self, 'toolbars'):
-            for key, toolbar in self.toolbars.items():
+            for key, _toolbar in self.toolbars.items():
```
fsbraun, line 194:
```suggestion
            for key in self.toolbars:
```

fsbraun (approved):
I am impressed how many clarifications ruff found! Two suggestions on how to simplify to code even more once ruff identified unused loop variables. LGTM. With respect to the CodeQL warnings I believe we have to acknowledge them in the UI since CodeQL does not respect `# pramga` comments etc.

cms/utils/compat/forms.py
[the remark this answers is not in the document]
vinitkumar: Removing them breaks it because they are imported somewhere else. I think we can address them in some other PR.

PR #7514, opened 2023-03-27
fix: typos in French translations

wasertech:
## Description

`avet` in not a verb in the french language, where as `avez` on the other hand...

Also since there is no page yet, I put page to singular.

## Related resources

* https://dictionnaire.lerobert.com/conjugaison/avoir


fsbraun (approved):

fsbraun:
@wasertech Thank you for pointing this out and correcting it. I have adjusted the translations on transifex, too. This is where translations are pulled upon release.

To this end I noticed that there are some untranslated French strings for django CMS 4. Let me know if you would like to have a look at them!

PR #7529, opened 2023-04-20
fix: remove superflous curly bracket left behind on PR 7488

corentinbettiol:
See [this comment](https://github.com/django-cms/django-cms/pull/7488\#issuecomment-1513517082) in PR #7488 for more infos.


corentinbettiol:
> I think it's best to create a micro PR to fix this!

@fsbraun: Here is the micro-pr!

fsbraun (approved):

PR #7533, opened 2023-04-21
Feat/integrate ruff linter

vinitkumar:
## Description

Add Ruff linting and fixes to CMS 4
- [x] lint cms
- [x] lint menus

## Related resources

* #...
* #...

## Checklist


cms/models/managers.py
```diff
@@ -278,7 +278,7 @@ def subordinate_to_user(self, user, site):
                                         A
                                     /      \
                                   user     B,E
-                                /     \      \
+                                /     \\      \
```
fsbraun, line 281: This change is probably unwanted... 

```suggestion
                                /     \\      \
                                /     \      \
```

cms/tests/test_static_analysis.py
```diff
@@ -1,14 +1,15 @@
-from unittest import TestCase
+# NOTE: we don't care about pyflakes or flake8 anymore
+# from unittest import TestCase
 
-from cms.test_utils.util.static_analysis import pyflakes
+# from cms.test_utils.util.static_analysis import pyflakes
```
fsbraun, line 4: Why would this test be commented out?

fsbraun (commented):
LGTM! Only real thing I am not sure about is why not to run pyflakes.

cms/tests/test_static_analysis.py
vinitkumar: 1) It was crashing because of the changes ruff did when auto formatting and fixing.
2) Using ruff to do all the linting instead of using pyflakes?

fsbraun: I'd added this to your PR, but I cannot change anything.

Shall we replace `cms/tests/test_static_analysis.py` by

```
import os
from unittest import TestCase

from ruff.__main__ import find_ruff_bin


class AboveStaticAnalysisCodeTest(TestCase):
    """
    Name is pretty lame, but ensure it's executed before every other test
    """
    def test_ruff(self):
        ruff = find_ruff_bin()
        self.assertEqual(os.spawnv(os.P_WAIT, ruff, ["ruff", "cms", "menus"]), 0)
```

and in `test_requirements/requirements_base.txt` the line with `pyflakes` by `ruff`?

Then the file `cms.test_utils.utils.static_analysis` could go away.

vinitkumar: Good idea.
I will just implement that.

fsbraun (approved):
GREAT! Let's get this merged!

PR #7549, opened 2023-05-08
feat: Update dark mode switch to be compatible with Django 4.2 admin dark mode

fsbraun:
## Description

Since version 4.2 Django's admin has its own dark mode toggle. While functionally similar, it uses a different naming convention. This PR adopts Django's naming convention

* The color scheme is defined by the `data-theme` attribute of the `html` tag (before: `data-color-scheme`)
* The toggle toggles between three states: light, dark, auto (before: two states - light and dark)
* The toggle button shows the current setting (sun: light, moon: dark, black-white-circle: auto) as opposed the setting that would be reached if the toggle is clicked

## Related resources

* #...
* #...

## Checklist


marksweb (approved):
LGTM 🎉

PR #7565, opened 2023-05-23
fix: create page wizard fails with Asian page titles/unicode slugs

fsbraun:
## Description

Django CMS supports unihandecode.js slug generation (if [configured](https://docs.django-cms.org/en/latest/reference/configuration.html?highlight=UNIHANDECODE#unicode-support-for-automated-slugs) in the settings). This PR resolves the following issues

* Unihandecode settings were ignored by the create page wizard
* Create page wizard crashed if unicode slugs were entered manually (#7564) (except for the root page)
* Misleading validation messages if non-ascii slugs were entered in the page settings

The root cause were inconsistent form validations and unihandecode being hard-wired into the page settings template while the wizard uses a custom SlugWidget:
* The SlugWidget moved into the AddPage/ChangePage forms
* unihandecode.js is loaded by the widget, not a form template: Thus it is available in both the page settings and create page wizard.
* Validation is handled by Django's SlugField validators and not inconsistently by different django CMS forms.

### Example: Create page wizard (zh language)
unicode-sligs

## Related resources

* #7564 
* #1605

## Checklist

### DIFF
--- a/cms/forms/wizards.py
+++ b/cms/forms/wizards.py
@@ -27,7 +27,7 @@ class SlugWidget(AdminSlugWidget):
     """Compatibility shim with deprecation warning:
     SlugWidget has moved to cms.admin.forms"""
     def __init__(self, *args, **kwargs):
-        warnings.warn("Import SlugWidget from cms.admin.forms. SlugWidget will be removed from cms.forms.wizzards",
+        warnings.warn("Import SlugWidget from cms.admin.forms. SlugWidget will be removed from cms.forms.wizards",
                       DeprecationWarning, stacklevel=2)
         super().__init__(*args, **kwargs)
 
--- a/cms/forms/wizards.py
+++ b/cms/forms/wizards.py
@@ -1,3 +1,5 @@
+import warnings
+
 from django import forms
 from django.core.exceptions import ValidationError
 from django.db import transaction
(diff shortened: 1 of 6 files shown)


cms/forms/wizards.py
```diff
@@ -21,16 +23,13 @@
...
-    class Media:
-        js = (
-            'admin/js/urlify.js',
-            static_with_version('cms/js/dist/bundle.forms.slugwidget.min.js'),
-        )
+class SlugWidget(AdminSlugWidget):
+    """Compatibility shim with deprecation warning:
+    SlugWidget has moved to cms.admin.forms"""
+    def __init__(self, *args, **kwargs):
+        warnings.warn("Import SlugWidget from cms.admin.forms. SlugWidget will be removed from cms.forms.wizzards",
```
vinitkumar, line 30: Typo on `cms.forms.wizzards`, should read `cms.forms.wizards`

vinitkumar (changes requested):

vinitkumar (changes requested):
Just found a typo and commented on it. Otherwise, everything else looks perfect according to me.

cms/forms/wizards.py
fsbraun: Changed! Thanks! (I tend to write wizard with double z 🤦‍♂️)

fsbraun:
@vinitkumar Typo corrected. Can you approve? Thanks so much!

vinitkumar (approved):

PR #7585, opened 2023-06-16
Replace PositiveSmallIntegerField against PositiveIntegerField

jrief:
## Description

In performing a migration towards version 4 I encountered a "value to big for small integer" exception.

This is because migration 0026 adds a temporary field for `title_id`. In my case some values do not fit in there.

### DIFF
--- a/cms/migrations/0026_title_placeholders.py
+++ b/cms/migrations/0026_title_placeholders.py
@@ -20,6 +20,6 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='placeholder',
             name='title_id',
-            field=models.PositiveSmallIntegerField(null=True),
+            field=models.PositiveIntegerField(null=True),
         ),
     ]


fsbraun:
Hi @jrief!

I fear the issue goes further than just changing the migration. The `position` field of the `CMSPlugin` model in both v3 and v4 is a `PositiveSmallIntegerField`. The restrictions caused by the `Small` part are different, however:

* On v3 it implies that a plugin can have at maximum 32767 children. The number of plugins in a placeholder is not limited.
* On v4 it implies that a placeholder can hold at maximum 32767 plugins (which obviously is a much stronger restriction). Considering how the CTE work, it even means, you in practice can work with "only" 16383 max plugins in a placeholder. (When inserting the CTE temporarily move plugins to positions behind the last current plugin position.) 

It seems you have more than 32767 plugins in at least one of your placeholders. **Could you please double-check?** 

I do not want to discuss the usability effects of having so-and-so many plugins in a placeholder, but generally feel that saving a few bytes in the DB should not be the reason for the restriction.

@marksweb @Aiky30 Does it make sense to change the `position` field to a regular `PositiveIntegerField`? To not block the migration, then all CMS v4 migrations should immediately also move from `PositiveSmallIntegerField` to `SmallIntegerField`.

fsbraun (approved):
Sorry, my previous comment was off-topic. Now, we're talking about a temporary field needed for migration 27 here to store the primary key. This must be at least `PositiveIntegerField`.

PR #7599, opened 2023-06-26
fix: update diff-dom and karma, run frontend tests on Chrome Headless

fsbraun:
## Description

This PR updates the karma JS testing environment to a current version. The PhantomJS headless browser is replaced by ChromeHeadless.

With this setup, `diff-dom` can be updated to 5.0.x, solving #7460 

## Related resources

* #7461 
* #7460
* #7583
* #7470 

## Checklist

* [x] I have opened this pull request against ``develop-4``
* [x] I have added or modified the tests when changing logic
* [x] I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog
* [x] I have read the [contribution guidelines ](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to find a “pr review buddy” who is going to review my pull request.

Issue #7460, opened 2022-12-20
[BUG] `diff-dom` freezing on CMS content refresh


petrklus:
## Description

We are experiencing issues with the CMS refresh. For certail plugin configurations and post-plugin modification, the browser tab first freezes, later to crash due to out of memory error. We have confirmed the same on Google Chrome / Mozilla Firefox.

## Steps to reproduce

1. Trigger content/structure board refresh
2. Tab crashes

## Expected behaviour

The page content to update.

## Actual behaviour

The diff between incoming document and actual document freezes up:

https://github.com/django-cms/django-cms/blob/develop/cms/static/cms/js/modules/cms.structureboard.js#L1291

Only tab restart fixes the issue.

## Screenshots

<img width="1211" alt="image" src="https://user-images.githubusercontent.com/1115667/208675313-8022ee9c-09f7-4d90-b1b3-809c68b65eb2.png">

## Additional information (CMS/Python/Django versions)

Django CMS 3.11.0 or 1, Django 3.2.

Tested on Mozilla Firefox and Google Chrome

## Do you want to help fix this issue?

We have looked into this further and the issue seems to be down to DiffDOM version 2.5.1. We have tried upgrading to the latest version 4.2.8, which works in debug mode, however, is not compatible with the current UglifyJS config.

* [x] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 
* [ ] No, I only want to report the issue.

Rogell:
This issue happened to me also and I can confirm that upgrading to the latest DiffDOM (version 5.0.4) fixes the problem.
I will not create a pull request as I don't have time to read all contribution requirements so if anyone wants to fix it here are my changes.

**Details:**
Branch: release/3.11.x
Python version: 3.11
NPM version: 6.14.4
System: Ubuntu 20.04.2 LTS on WSL2

**Changes:**
`package.json:52` -> `"diff-dom": "5.0.4",`
`cms/static/cms/js/modules/cms.structureboard.js:11` -> `import { DiffDOM } from 'diff-dom';`

petrklus:
@fsbraun would you know of anyone with the CMS editor code experience who can look into this?

marksweb:
@petrklus whats happening with the plugins that might give us a clue on how to reproduce this?

fsbraun:
[...]

PR #7600, opened 2023-06-27
fix: add updates same as PR: #7599

vinitkumar:
Code borrowed from:
https://github.com/django-cms/django-cms/pull/7599.

Also, based on on the ticket and work of https://github.com/django-cms/django-cms/issues/7460

## Description

Should fix the diff-dom issue in CMS 3.11.x series.

## Related resources

* (https://github.com/django-cms/django-cms/issues/7460)

## Checklist


package.json
```diff
@@ -3,25 +3,27 @@
   "private": true,
   "devDependencies": {
     "autoprefixer": "^6.3.6",
```
fsbraun, line 5: Let's try to also upgrade autoprefixer@^9

cms/static/cms/js/modules/cms.structureboard.js
```diff
@@ -8,7 +8,8 @@ import Plugin from './cms.plugins';
 import { getPlaceholderIds } from './cms.toolbar';
 import Clipboard from './cms.clipboard';
 import URI from 'urijs';
-import DiffDOM from 'diff-dom';
+import { DiffDOM } from 'diff-dom';
+import Promise from 'promise-polyfill';
```
fsbraun, line 12: Frankly, I am uncertain if polyfills should go into the `webpack.config.js` or into source files. What's the common practice?

vinitkumar: I don't like it either. Gulp lint gave me:

```/Users/vinitkumar/projects/python/django-cms/cms/tests/frontend/unit/cms.structureboard.test.js
  251:64  error  Promise.resolve() is not supported in IE Mobile 10, IE 10, bb 7  compat/compat
  344:54  error  Promise.resolve() is not supported in IE Mobile 10, IE 10, bb 7  compat/compat
  345:54  error  Promise.resolve() is not supported in IE Mobile 10, IE 10, bb 7  compat/compat
  421:62  error  Promise.resolve() is not supported in IE Mobile 10, IE 10, bb 7  compat/compat
  422:64  error  Promise.resolve() is not supported in IE Mobile 10, IE 10, bb 7  compat/compat
```
So either we don't care about these browsers, in that case, we just change the eslint config and get the lint to pass.

fsbraun:
@vinitkumar : Great work! I am glad, we can move this forward!
@petrklus: Can you confirm it solves your issue #7460 ? I'd appreciate if you could give it a try! Also, I seem to remember you needed to change some transpiling rules. I just want to make sure this was not for the bundle but for the tests with PhantomJS. This would explain why this PR does not need them.

cms/static/cms/js/modules/cms.structureboard.js
fsbraun: IE 10 is 10 years old and, according to [caniuse](https://caniuse.com/?search=promise), has a market share of 0.07%. 

@marksweb: I think we should drop those. Probably this requirement has hold us back from updating some of the packages.

vinitkumar: @fsbraun @marksweb In that case, I drop these and remove all browsers that don't support Promises and guarantee only modern ones will work.

marksweb (approved):

cms/static/cms/js/modules/cms.structureboard.js
marksweb: Yeah drop IE support. People should have moved on to edge by now

vinitkumar:
> 

Yes, sir

vinitkumar:
@fsbraun Just added a commit that updates the .nvmrc. Should be good enough to be merged now.

mihalikv:
@fsbraun Is it possible to build dist JS to develop branch with these new changes? I want to test it as pip git dependency but dist is not rebuilt. 
Thanks.

fsbraun:
@mihalikv The best way to test locally by installing the develop branch with `pip -e /path/to/develop/branch`. Then you can run `gulp bundle` on the develop branch to build the webpack js libs. 
If you need to install from github you'll have to create a fork, run `gulp bundle`, `gulp icons`, `gulp sass` and commit the assets to your branch. It is not possible to have the assets built in the main `develop` branch.

mihalikv:
@fsbraun thanks for the guidance, I built it in my own repo, deploy and it works! thx

fsbraun:
@mihalikv Thanks for the feedback! Great that it works! 🚀

DmytroLitvinov:
I can confirm that that PR fixes for us problem with freeze - more details [here](https://github.com/django-cms/django-cms/issues/7460#issuecomment-1711045725)

PR #7609, opened 2023-07-11
fix: Only return one Page in page_url templatetag

### LINKED ISSUE #7607: [BUG] Only return one Page in `page_url` templatetag
## Description

It's possible to use the `page_url` template tag with the id of a page and have multiple pages match. This results in an exception when rendering the page. I don't know how I ended up with a database of said state.

## Steps to reproduce

Can't reproduce but I didn't fiddle in the DB manually.

## Expected behaviour

Return only one page that matches the page's id and log a warning.

## Actual behaviour

The CMS throws an exception.

cms 3.11.3

I already created a patch that catches the exception and only returns one page. I guess this is only masking another bug or maybe corrupt database. I'm not sure this is the way to continue with this issue.

sepi:
In case there are multiple Pages matching, signal a warning message and return first Page found. Fixes #7607.

Date:      Tue Jul 11 15:01:31 2023 +0200

## Checklist

### DIFF
--- a/cms/templatetags/cms_tags.py
+++ b/cms/templatetags/cms_tags.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict, namedtuple
 from copy import copy
 from datetime import datetime
+import logging
 
 from classytags.arguments import Argument, MultiKeywordArgument, MultiValueArgument
 from classytags.core import Options, Tag
@@ -39,6 +40,7 @@ NULL = object()
 DeclaredPlaceholder = namedtuple('DeclaredPlaceholder', ['slot', 'inherit'])
 DeclaredStaticPlaceholder = namedtuple('DeclaredStaticPlaceholder', ['slot', 'site_bound'])
 
+logger = logging.getLogger(__name__)
 
 register = template.Library()
 
@@ -67,8 +69,15 @@ def _get_page_by_untyped_arg(page_lookup, request, site_id):
         raise TypeError('The page_lookup argument can be either a Dictionary, Integer, Page, or String.')
     site = Site.objects._get_site_by_id(site_id)
     try:
+        warning_message = _('The page_url tag unexpectedly found multiple pages matching the request: %(page_lookup)s') % {'page_lookup': page_lookup}
         if 'pk' in page_lookup:
-            page = Page.objects.select_related('node').get(**page_lookup)
+            try:
+                page = Page.objects.select_related('node').get(**page_lookup)
+            except Page.MultipleObjectsReturned:
+                messages.warning(request, warning_message)
+                logger.warning(warning_message)
+                page = Page.objects.select_related('node').filter(**page_lookup).first()
+
             if request and _use_draft(request):
                 if page.publisher_is_draft:
                     return page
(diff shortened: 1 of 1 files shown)


cms/templatetags/cms_tags.py
```diff
@@ -67,8 +69,15 @@ def _get_page_by_untyped_arg(page_lookup, request, site_id):
         raise TypeError('The page_lookup argument can be either a Dictionary, Integer, Page, or String.')
     site = Site.objects._get_site_by_id(site_id)
     try:
+        warning_message = _('The page_url tag unexpectedly found multiple pages matching the request: %(page_lookup)s') % {'page_lookup': page_lookup}
         if 'pk' in page_lookup:
-            page = Page.objects.select_related('node').get(**page_lookup)
+            try:
+                page = Page.objects.select_related('node').get(**page_lookup)
+            except Page.MultipleObjectsReturned:
```
fsbraun, line 76: I do not think this is a possible scenario: If the primary key `pk` is given, then exactly one (due to the uniqueness constraint on primary keys) or no `Page` will be found. The latter scenario is already taken care of in lines 99ff.
Therefore, I think this try except block is not necessary.

cms/templatetags/cms_tags.py
```diff
@@ -67,8 +69,15 @@ def _get_page_by_untyped_arg(page_lookup, request, site_id):
         raise TypeError('The page_lookup argument can be either a Dictionary, Integer, Page, or String.')
     site = Site.objects._get_site_by_id(site_id)
     try:
+        warning_message = _('The page_url tag unexpectedly found multiple pages matching the request: %(page_lookup)s') % {'page_lookup': page_lookup}
```
fsbraun, line 72: This should probably go into the `except:` clause in line 95ff.

cms/templatetags/cms_tags.py
```diff
@@ -81,7 +90,12 @@ def _get_page_by_untyped_arg(page_lookup, request, site_id):
                     return page
         else:
             pages = get_page_queryset(site, draft=_use_draft(request))
-            return pages.select_related('node').get(**page_lookup)
+            try:
+                return pages.select_related('node').get(**page_lookup)
+            except Page.MultipleObjectsReturned:
+                messages.warning(request, warning_message)
```
fsbraun, line 96: I find the `messages` warning inconsistent with the e-mail to admin setup for a non-existing page below.  A log entry in addition to the optional email might be a better solution.

cms/templatetags/cms_tags.py
```diff
@@ -81,7 +90,12 @@ def _get_page_by_untyped_arg(page_lookup, request, site_id):
...
+                logger.warning(warning_message)
+                return pages.select_related('node').filter(**page_lookup).first()
```
fsbraun, line 98: I am uncertain if returning the first page is a good solution. Something went wrong here, and I'd stay with `None`. Clearly, you cannot predict which of the multiple pages will be first.

fsbraun (changes requested):

fsbraun:
Hi @sepi ! Thank you very much for raising that issue and creating the PR! I appreciate it a lot!

I like your thinking and would like to make an additional suggestion:
There's already a try-except block catching the `Page.DoesNotExist` exception. Could this not work similarly. Something like:

```
    ...
    subject = None
    body = None
    try:
        ...
    except Page.DoesNotExist:
         subject = ...
         body = ...
         if settings.DEBUG:
             raise Page.DoesNotExist
    except Page.MultipleObjectsReturned:
         subject = ...
         body = ...
         if settings.DEBUG:
             raise Page.MultipleObjectsReturned
     if subject:  # error caught
         # Code to log
         # Code to send email if configured
```

Finally, I'd like to ask you to add a test to check if the exception is caught.

PR #7613, opened 2023-07-22
fix: Cross-talk between grouper admins due to common list initialization

fsbraun:
## Description

`GrouperModelAdmin` initialized a private list property to an empty list outside `__init__`. Since the list is a python object only a reference to a single list is stored in every `GrouperModelAdmin` instance leading to potential "cross-talk" between admins of different grouper models.

Fix better language tab layout of `GrouperModelAdmin` if used with plain Django admin.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/cms/admin/utils.py
+++ b/cms/admin/utils.py
@@ -240,13 +240,14 @@ class GrouperModelAdmin(ChangeListActionsMixin, ModelAdmin):
     EMPTY_CONTENT_VALUE = _("Empty content")
     LC_SORTED_FIELDS = (models.CharField,)
 
-    _content_obj_cache = {}
     _content_cache_request_hash = None
-    _content_qs_cache = {}
     _content_content_type = None
-    _content_subquery_fields = []
 
     def __init__(self, model, admin_site):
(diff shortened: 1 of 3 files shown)


cms/admin/utils.py
```diff
@@ -240,13 +240,14 @@ class Media:
...
     LC_SORTED_FIELDS = (models.CharField,)
 
-    _content_obj_cache = {}
     _content_cache_request_hash = None
-    _content_qs_cache = {}
     _content_content_type = None
-    _content_subquery_fields = []
 
     def __init__(self, model, admin_site):
+        self._content_subquery_fields = []
```
marksweb, line 247: Good catch. This is very important when creating class variables like these.

marksweb (approved):

cms/admin/utils.py
fsbraun: @marksweb Well, I created the issue myself, being too focused on other stuff (quite nice ORM calls etc.), and only noticed now... Glad we fixed this!

PR #7650, opened 2023-09-20
test: Run tests against django main branch

marksweb:
This introduces testing, allowed to fail, against the main branch of django so we have visibility of what is coming in future versions.


.github/workflows/test.yml
```diff
@@ -155,3 +155,135 @@ jobs:
...
+        django-version: [
+          'https://github.com/django/django/archive/main.tar.gz'
+        ]
+        os: [
+          ubuntu-20.04,
+        ]
+
+    services:
+      postgres:
+        image: postgres:12
```
fsbraun, line 210:
```suggestion
        image: postgres:13
```
Django main dropped postgres 12

.github/workflows/test.yml
```diff
@@ -155,3 +155,135 @@ jobs:
...
+        sudo apt install gettext gcc -y
+        python -m pip install --upgrade pip
+        pip install pytest ${{ matrix.django-version }}
+        pip install -r test_requirements/${{ matrix.requirements-file }}
+        pip install -r test_requirements/databases.txt
+        python setup.py install
+
+    - name: Test with django test runner
+      run: |
+        python manage.py test
```
fsbraun, line 239:
```suggestion
        python -c "from django import __version__ ; print(f'Django version {__version__}')"
        python manage.py test
``` 
For transparency let's print out which version the django main had...

.github/workflows/test.yml
```diff
@@ -155,3 +155,135 @@ jobs:
...
```
fsbraun, line 287:
```suggestion
        python -c "from django import __version__ ; print(f'Django version {__version__}')"
        python manage.py test
```
Same here.

fsbraun (changes requested):
Lines 184 and 185 need to be swapped. Otherwise, the requirements will uninstall the django main branch if there is a `<5` somewhere.

Like wise 232 and 233, and 280 and 281

fsbraun (approved):

.github/workflows/test.yml
```diff
@@ -155,3 +155,137 @@ jobs:
...
+        python -m pip install --upgrade pip
+        pip install -r test_requirements/${{ matrix.requirements-file }}
+        pip install pytest ${{ matrix.django-version }}
+        pip install -r test_requirements/databases.txt
+        python setup.py install
+
+    - name: Test with django test runner
+      run: |
+        python -c "from django import __version__ ; print(f'Django version {__version__}')"
+        python -c "from django import __version__ ; print(f'Django version {__version__}')"
```
fsbraun, line 240:
```suggestion
```

marksweb:
@fsbraun have we removed the Django 5 restriction from develop 4? Could do with updating my branches before merging to use it if we've dropped that limit.

fsbraun:
Just did. But also need to fix ruff... One minute.

PR #7651, opened 2023-09-21
chore: Add huntr.dev to SECURITY.md

huntr-helper:
As requested through the platform by @marksweb, this will point your security policy to [huntr.dev](https://huntr.dev/repos/django-cms/django-cms})


marksweb:
@fsbraun What do you think to adding this? It's good practice to have a security file.

fsbraun:
Sure. But I would not tie us to those guys. We should advertise our own email (from the blog): 

As ever, we remind our users and contributors that all security reports, patches and concerns be addressed only to our security team by email, at [security@django-cms.org](mailto:security@django-cms.org).

PR #7664, opened 2023-10-09
fix: preserve `view_class` in decorated views

Will-Hoey:
## Description

Add a check to preserve the `view_class` of a decorated function. Will allow retrieving a CBV name from a resolved URL

## Related resources

Open issue: https://github.com/django-cms/django-cms/issues/7653

## Checklist

### DIFF
--- a/cms/tests/test_apphooks.py
+++ b/cms/tests/test_apphooks.py
@@ -26,7 +26,7 @@ from cms.test_utils.project.placeholderapp.models import Example1
 from cms.test_utils.testcases import CMSTestCase
 from cms.tests.test_menu_utils import DumbPageLanguageUrl
 from cms.toolbar.toolbar import CMSToolbar
-from cms.utils.compat import DJANGO_2_2, DJANGO_3
+from cms.utils.compat import DJANGO_2_2, DJANGO_3, DJANGO_4_1
 from cms.utils.conf import get_cms_setting
 from cms.utils.urlutils import admin_reverse
 from menus.menu_pool import menu_pool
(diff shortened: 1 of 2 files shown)


cms/tests/test_apphooks.py
```diff
@@ -307,7 +307,7 @@ def test_apphook_permissions_preserves_view_name(self):
         view_names = (
             ('sample-settings', 'sample_view'),
             ('sample-class-view', 'ClassView'),
-            ('sample-class-based-view', 'view' if not (DJANGO_3 or DJANGO_2_2) else 'ClassBasedView' ),
+            ('sample-class-based-view', 'ClassBasedView' ),
```
marksweb, line 310: We still support older versions of django, so I feel like this still needs conditional logic.

cms/utils/decorators.py
```diff
@@ -20,6 +20,8 @@ def inner(request, *args, **kwargs):
         return func(request, *args, **kwargs)
     inner.__module__ = func.__module__
     inner.__doc__ = func.__doc__
+    if hasattr(func, "view_class"):
```
fsbraun, line 23: @Will-Hoey I'd prefer a conditional based on the Django version to keep the semantics and be able to remove the conditional completely once Django 3.2 is ancient history.

cms/tests/test_apphooks.py
fsbraun: @marksweb I think the PR is trying to achieve not having to have the Django version conditionals in custom code (or tests) but to always forward the view class name as presented by Django when using CMS decorators independently of the underlying Django version. @Will-Hoey Am I correct here?

cms/utils/decorators.py
fsbraun, line 23: @Will-Hoey I retract my earlier comment on this line. Sorry.

cms/tests/test_apphooks.py
Will-Hoey: @fsbraun That is correct yes!

fsbraun (approved):
@Will-Hoey Great work! Thank you!

PR #7683, opened 2023-10-26
Develop 4

dwintergruen:
## Description

I in the German version the permission seem to have sometimes(?) German names, so I had to add this.

Not changing the setup file let to a module cms not found error.

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -375,6 +375,9 @@ class PageAdmin(admin.ModelAdmin):
         if 'placeholder' in perms_needed:
             perms_needed.remove('placeholder')
 
+        if 'Seiten-Inhalt' in perms_needed:  # dwinter feel even worse in german version this is in german
+            perms_needed.remove('Seiten-Inhalt')
+
         if 'page content' in perms_needed:
             perms_needed.remove('page content')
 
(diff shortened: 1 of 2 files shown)


fsbraun:
@dwintergruen The django admin `get_deleted_objects` function returns verbose names (which can be localized). IMHO it would be the correct way to replace 
* `'page content'` -> `_("page content")`
* `'placeholder'` -> `_("placeholder")`.

cms/admin/pageadmin.py
```diff
@@ -375,6 +375,9 @@ def delete_view(self, request, object_id, extra_context=None):
         if 'placeholder' in perms_needed:
             perms_needed.remove('placeholder')
 
+        if 'Seiten-Inhalt' in perms_needed:  # dwinter feel even worse in german version this is in german
+            perms_needed.remove('Seiten-Inhalt')
+
```
fsbraun, line 380:
```suggestion
```

setup.py
```diff
@@ -47,7 +47,7 @@
 
 
 setup(
-    name='django-cms',
+    name='cms',
```
fsbraun, line 50:
```suggestion
    name='django-cms',
```

The django-cms package provides two apps: `cms` and `menus`.

fsbraun (changes requested):
I've created suggestions for the requested changes. Do you want to take a look at it?

fsbraun (approved):

PR #7686, opened 2023-10-27
feat: modernize django update CI for CMS 4

vinitkumar:
## Description

This PR uses django-upgrade to modernize our codebase to 3.2+, and also does some improvement to the CI.

## Related resources

* Django 5.0 https://docs.djangoproject.com/en/dev/releases/5.0/
* Drop support for Django 2.2, 3.1

## Checklist


fsbraun (approved):
LGTM! Let's drop Django 2.2 for v4.1. Mark agreed.

fsbraun (changes requested):
@vinitkumar  can you also update the `setup.py` file to require Django >= 3.2?

vinitkumar:
> @vinitkumar can you also update the `setup.py` file to require Django >= 3.2?

I just did the changes as you suggested.

fsbraun (approved):

cms/views.py
```diff
@@ -49,7 +49,7 @@
 
 if DJANGO_2_2:
```
marksweb, line 50: Is this a bad variable name where it means greater than 2.2?

I'm just thinking that while doing version support, we maybe can remove this check or name things better

cms/views.py
```diff
@@ -49,7 +49,7 @@
...
     from django.utils.http import (
-        is_safe_url as url_has_allowed_host_and_scheme,
+        url_has_allowed_host_and_scheme as url_has_allowed_host_and_scheme,
```
marksweb, line 52: If this check stays, we don't need the "as" because it's the same name.

setup.py
```diff
@@ -30,9 +30,6 @@
     'Programming Language :: Python :: 3.10',
     'Programming Language :: Python :: 3.11',
     'Framework :: Django',
-    'Framework :: Django :: 2.2',
-    'Framework :: Django :: 3.0',
-    'Framework :: Django :: 3.1',
     'Framework :: Django :: 3.2',
     'Framework :: Django :: 4.0',
     'Framework :: Django :: 4.1',
```
marksweb, line 35: Do we want to add 5 now?

cms/views.py
fsbraun: Nope, I think this is an automatic change. Only the else part should stay. `DJANGO_2_2` means version 2.2 or lower. It can go, since we drop Django 2.2 support.

cms/views.py
fsbraun: I think it needs to go. `DJANGO_2_2` means Django version 2.2 or lower.

cms/views.py
marksweb: That's what I was hoping!

setup.py
vinitkumar: I reverted the 5.0 addition, as we can wait for it until it turns stable to be added in the classifiers.

cms/views.py
vinitkumar: I cleaned up this code and all if conditions referencing DJANGO_2_2, DJANGO_3_0 and DJANGO_3_1

marksweb: @vinitkumar Have you pushed? I can't see any commits since Saturday.

vinitkumar: Omg! I am stupid. I have been pushing on my fork and not upstream and expecting you to see the changes. Sorry!

marksweb: Thanks!

vinitkumar:
@fsbraun  @marksweb Some tests are broken now. Will need to fix them for it to be merged.

marksweb:
@vinitkumar Yeah I was trying to workout why the language is failing 🤔

Issue #7692, opened 2023-10-30
[BUG] Language Switcher in toolbar and {% language_chooser %} dont work in v4.1rc3/4


benzkji:
## Description

Language Switcher in toolbar and {% language_chooser %} dont work in v4.1rc3/4, all languages provide the currently selected language in their links.

## Steps to reproduce

Demo repos:
https://github.com/benzkji/djangocms4test
Create a page, try the toolbars language switcher, or the on site rendered  {% language_chooser %}.

## Expected behaviour

Provide correct links. /de/ | /en/ | /fr/ | etc.

## Actual behaviour

All languages provide the currently selected language in their links.

## Screenshots

-

## Additional information (CMS/Python/Django versions)

cms 4.1rc3 and rc4

## Do you want to help fix this issue?

* [ ] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 
* [x] No, I only want to report the issue.

benzkji:
Update: Toolbar language switch seems to work when djangocms-versioning is installed. How bizarre.

benzkji:
I've done some research. Language chooser passes here:
- https://github.com/django-cms/django-cms/blob/develop-4/menus/utils.py#L144
- then https://github.com/django-cms/django-cms/blob/develop-4/menus/utils.py#L147 calls get_absolute_url() of the current PageContent (always returns the same url)
- but should probably just pass all the way down to https://github.com/django-cms/django-cms/blob/develop-4/menus/utils.py#L162. Or, pagecontent shoudl respect the current selected language, and not use just it's own?

I am not familiar with this code, but it's a basic functionality that's currently broken. Just wondering if nobody uses multiple languages?!

benzkji:
PageContent.get_absolute_url() can be passed a language. But many other `toolbar.object`'s get_absolute_url cannot accept a language parameter...?

benzkji:
3.X did it, but also had a slightly less complex mode. https://github.com/django-cms/django-cms/blob/develop/cms/templatetags/cms_tags.py#L230

benzkji:
@fsbraun it *MIGHT* be an easy fix ;-)

PR #7698, opened 2023-11-22
fix: Language chooser does not contain

fsbraun:
## Description

Fixes issue #7692 at least for content objects which do not have `language` as a grouping field or - if they have - allow for a `language` argument in their `.get_absolute_url()` method (such as, e.g., `PageContent`).

Overall, this still feels unsatisfying:
* The user is only directed to the public URLs, leaving preview or edit mode.
* Currently, there is no generic way for the django CMS core to identify content objects of a given language.
* Since the language menu uses the same `DefaultLanguageChanger` class, new content objects (such as alias or blog) have to provide their own language menu, creating unnecessary code repetition.

I make this PR a draft to resolve the remaining shortcomings. Options include
1. Add a complementary method `get_for_language(language)` to let a content model decide itself how to provide the appropriate language. Downside: Potentially inefficient since caching takes place in the grouper model (`Page`)
2. Add a complementary property to indicate how to find a content object's grouper object (would be `"page"` for `PageContent` since `page_content.page` gets the grouper object) to call its `get_content_object` method. Downside: A (potentially) unnecessary implicit convention.
3. Extend the CMS config for toolbar-enabled models to optionally provide the model's field to get to its grouper object. 

I am favouring the third option, but would appreciate any feedback! 

## Related resources

* #7692
* #...

## Checklist

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -731,7 +731,7 @@ class Page(models.Model):
             self.update_languages(page_languages)
 
     def update_languages(self, languages):
-        languages = ",".join(languages)
+        languages = ",".join(set(languages))
         # Update current instance
         self.languages = languages
         # Commit. It's important to not call save()
--- a/menus/utils.py
+++ b/menus/utils.py
@@ -1,5 +1,5 @@
 from django.conf import settings
-from django.urls import NoReverseMatch, resolve, reverse
+from django.urls import NoReverseMatch, resolve, Resolver404, reverse
 
 from cms.utils import get_current_site, get_language_from_request
 from cms.utils.i18n import (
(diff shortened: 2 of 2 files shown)


fsbraun:
@marksweb @Aiky30 @benzkji  Any thoughts?

benzkji:
With my limited knowledge: LTM

benzkji:
Sorry, did not read the whole post. 

- I would not imply that a model has a grouper. All of my apps use django-modeltranslation, so, no grouper there (for now ?)
- leaving edit mode is not very elegant, and could lead to user frustration
- I dont understand the "DefaultLanguageChanger", and how every app/model would need it's own?
- What are "toolbar-enabled models"?

As you see, more questions than answers ;-)

fsbraun:
To your points: 
* It has to work for both models with (PageContent) and without a grouper.  
* yes
* Supposedly it should be universal, but currently does not support edit or preview mode for frontend-editable models
* django CMS introduces frontend-editable models besides PageContent. Example: AliasContent. They can be previewed and edited in the frontend editor without having to have their own URL.

fsbraun:
I now implemented options three:

* Frontend-editable objects now have the option to register the name of the grouper field (e.g., `"page"` for `PageContent`, since `page_content.page` gives its grouper object)
* The language menu now has a cached way of identifying sibling objects of the toolbar object if the name of the grouper field is registered.
* Both language chooser and language menu now preserve preview or edit mode.

marksweb (approved):
I've been looking over this when I've found a time for a while. I haven't seen anything I don't like, but I've not ran it myself. Given I'm the bottleneck here I'll sign-off for the RC

Issue #7704, opened 2023-11-28
[BUG] ExtensionToolbar.get_title_extension_admin returns the wrong object/URL


jrief:
## Description

Method `cms.extensions.toolbar.ExtensionToolbar.get_page_extension_admin()` returns the wrong object and URL, if 
**djangocms-versioning** is configured. The object to be returned is of type `MyPageContentExtension` inheriting from `cms.extensions.models.PageContentExtension`. The URL to be returned points on the Django admin view to edit this object.

The problem arises because method `get_page_extension_admin` determines the `PageContentExtension` object [through the current page](https://github.com/django-cms/django-cms/blob/d031da33081f766395573d3e460f4c8675883b77/cms/extensions/toolbar.py#L44). In a versioned CMS, this however is the wrong approach, because instead it should return the currently edited version of that `PageContentExtension` object.

## Steps to reproduce

* Install and configure **djangocms-versioning**.
* Create a proprietary Django model inheriting from `PageContentExtension`.
* Register an editor for that proprietary model in the CMS toolbar.
* Edit the page and create multiple versions.
* Access the editor to edit the proprietary object from the CMS toolbar.

## Expected behaviour

The proprietary object should be that one associated with the *current* `PageContent` object.

## Actual behaviour

The proprietary object instead is associated with the *published* `PageContent` object.

## Workaround

I was able to create a workaround for this problem by overriding method `get_page_extension_admin` in my own class inheriting from `ExtensionToolbar`:

```python
from django.urls import NoReverseMatch, resolve, reverse

@toolbar_pool.register
class MyExtensionToolbar(ExtensionToolbar):
    model = MyPageContentExtension

    def get_title_extension_admin(self, language=None):
        urls = []
        resolver_match = resolve(self.request.path)
        if resolver_match.view_name != 'admin:cms_placeholder_render_object_edit':
            return super().get_title_extension_admin(language)
        page_content = PageContent.admin_manager.get(id=resolver_match.args[1])
        try:
            page_extension = self.model.objects.get(extended_object_id=page_content.pk)
        except self.model.DoesNotExist:
            page_extension = None
        model_name = self.model.__name__.lower()
        kwargs = {'app_label': self.model._meta.app_label, 'model_name': model_name}
        try:
            if page_extension:
                urls.append((page_content, admin_reverse(
                    '{app_label}_{model_name}_change'.format(**kwargs),
                    args=(page_extension.pk,)
                )))
            else:
[...]

PR #7708, opened 2023-11-29
fix: Extension toolbar

fsbraun:
## Description

This PR does fix the page content extension toolbar:
* Renames `get_title_extension_admin` to `get_page_content_extension_admin` (old name stays with a deprecation warning)
* Takes the current page content from the toolbar object if available (from page as a fallback). This works also, if going back to old versions when using djangocms-versioning.
* If used without the language argument, it picks the `latest_content` filter which - if versioning is installed - is the latest version (draft, published, or unpublished, archived in that order). I recommend to not use this variant, since it is not necessarily in sync with the currently shown toolbar object

@jrief can you test if this solves your issue #7704?

## Related resources

* #7704 
* #...

## Checklist

### DIFF
--- a/cms/extensions/toolbar.py
+++ b/cms/extensions/toolbar.py
@@ -1,3 +1,5 @@
+import warnings
+
 from django.urls import NoReverseMatch
 
 from cms.toolbar_base import CMSToolbar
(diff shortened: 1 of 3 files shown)


cms/extensions/toolbar.py
```diff
@@ -60,41 +70,61 @@ def get_page_extension_admin(self):
...
+        Return a tuple of the page content extension and the url; the extension is None
+        if no instance exists, the url is None is no admin is registered for the extension.
+        """
+        self._get_page()
+        page_content = page_content_obj or self.page_content
+        try:
+            pagecontent_extension = self.model.objects.get(extended_object_id=page_content.pk)
+        except self.model.DoesNotExist:
+            pagecontent_extension = None
+        try:
```
jrief, line 114: I would write:

```python
        kwargs = {'app_label': self.model._meta.app_label, 'model_name': self.model.__name__.lower()}
        try:
            if pagecontent_extension:
                urls.append((page_content, admin_reverse(
                    '{app_label}_{model_name}_change'.format(**kwargs),
                    args=(pagecontent_extension.pk,)
                )))
            else:
                urls.append((page_content, '{0}?extended_object={1}'.format(
                    admin_reverse('{app_label}_{model_name}_add'.format(**kwargs)),
                    page_content.pk,
                )))
        except NoReverseMatch:  # pragma: no cover
            pass
```

this reduces code duplication and uses Python `.format()`-method instead of Py-2's `%`.
From Zen of Python: "Readability counts".

jrief (changes requested):
After merging into my project, I can confirm that from a functional point of view, it now works as expected.

However I left two annotations on code and docs.

cms/extensions/toolbar.py
jrief: Sorry for the late edit. Rechanged that snippet.

jrief (changes requested):

fsbraun:
@jrief Docs currently live in a separate repo: Here are the accompanying changes to the docs: https://github.com/django-cms/docs/pull/36/files. I've replaced the `%` formatting by `.format`. Thanks for pointing this out. I gues that bit of code has not been touched for a while.

jrief (approved):
even better than `.format()`

PR #7739, opened 2024-01-05
fix: readable messages in dark-mode

### LINKED ISSUE #7738: [BUG] Dark mode makes the pop-up messages from the admin interface unreadable.
When the Django-CMS admin run in dark mode, the pop messages are unreadable. The text is dark on a dark background.  I also think they should not disappear so fast. In case of an error, you have not enough time to read it.

image

fsbraun:
## Description

Fixes #7738 for v3. 

The messages should always appear with white text on dark background.

## Related resources

* #7738 
* #...

## Checklist

### DIFF
--- a/cms/static/cms/sass/settings/_cms.scss
+++ b/cms/static/cms/sass/settings/_cms.scss
@@ -342,7 +342,7 @@ $dialog-shadow: $modal-shadow;
 // Toolbar messages + screenblock + login form
 $messages-width: 300px;
 $messages-padding: 6px 10px 8px;
-$messages-color: $white;
+$messages-color: white;
 $messages-font-size: 12px;
 $messages-line-height: 16px;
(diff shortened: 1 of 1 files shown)


cms/static/cms/sass/settings/_cms.scss
```diff
@@ -342,7 +342,7 @@ $dialog-shadow: $modal-shadow;
 // Toolbar messages + screenblock + login form
 $messages-width: 300px;
 $messages-padding: 6px 10px 8px;
-$messages-color: $white;
+$messages-color: white;
```
marksweb, line 345: Is this because the messages bar always has the same background colour?

fsbraun: Exactly: It is always shadow color and this is also dark in dark mode.

marksweb (approved):

Issue #7745, opened 2024-01-09
[BUG] Deleting a page sometimes impossible in languages other than English


jrief:
## Description

Deleting a page sometimes is impossible in languages other than English.

## Steps to reproduce

Setup Django-CMS with multiple languages.
Choose a language other than English.
Delete a page containing content, ie. filled placeholders.

* [x] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 
* [ ] No, I only want to report the issue.

fsbraun:
Fixed in #7745

PR #7747, opened 2024-01-09
fix: #7745 -- permission handling when deleting page

jrief:
discard specific verbose_name from perms_needed

## Description

This fixes #7746 , check issue for details.


fsbraun (approved):
Nice. I was not aware that `perms_needed` actually is a set. :-)

jrief:
> Nice. I was not aware that `perms_needed` actually is a set. :-)

and this is true for all supported Django versions (>3.0), hence using `.discard()` is safe.

PR #7748, opened 2024-01-09
Remove delete page from toolbar

jrief:
## Description

Currently the toolbar offers a menu item to delete the current page.
This is problematic, because the user first has to change into edit mode, before allowing him to delete that page.
Moreover, after deletion which page shall the browser render?

@fsbraun and I therefore discussed on Slack and concluded, that deleting a page only shall be possible through the page tree.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,3 +1,11 @@
+4.2 (in progress)
+=================
+
+Features:
+---------
+* Remove menu item "Delete page…" from toolbar.
+
+
 4.1.0 (2023-12-22)
 ==================
 
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -371,12 +371,10 @@ class PageAdmin(admin.ModelAdmin):
             **get_deleted_objects_additional_kwargs
         )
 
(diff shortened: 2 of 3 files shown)


fsbraun:
I would second the notion. Let me add my rationale:
* Together with djangocms-versioning pages can never be deleted unless they do not have a single page content object or a setting is changed.
* The impact of deleting a page is more transparent in the page tree: All languages and all child pages are affected. (I know all elements are shown in the confirmation dialogue - but that typically is extremely long and hard to grasp.) This is the asymmetry to the "New Page" menu entry.
* It should not depend on the edit mode or publication status of the PageContent.

The alternative would be:
* Only show the delete menu entry if deleting is actually possible in the setup (might be complex in relation with versioning)
* Allow deleting in all edit models: preview, edit and "view on site".
* Only grey out if the user lacks permissions

jrief:
**In the course of this, I would also remove the item "Duplicate this page".**

Reason is, that users have no way to specify the target and hence that duplicate would end up as a sibling of the target with a slug named "copy-of-…". By allowing the create duplicates from the page tree, this problem can be avoided because normally those duplicated pages are added to a different parent anyway.

PR #7753, opened 2024-01-10
fix: #7752 – change_form template uses the wrong id

### LINKED ISSUE #7752: [BUG] In PageContent admin first field isn't focused
## Description

The `<form>`-element in template `admin/cms/page/change_form.html` has hard coded `id="page_form"` pointing to the wrong model.

## Steps to reproduce

* Edit a `PageContent` object using Django's admin.
* Check the browser console, it states something like: 
Screenshot 2024-01-10 at 14 17 34

Reason is that the `<form>`-element in template `change_form.html` contains a hard coded `id="page_form"`, but the edited model is of type `page_content`.

This btw. prevent to set the focus on the first field.

Using the same pattern as the super template fixes this.

jrief:
## Description

Please check #7752 for details.

* [x] I have opened this pull request against `develop-4`


fsbraun (approved):
`change_form.js` is part of Django and needs the form id to be the model name + `"_form"`. @jrief Did you check the JS error disappears?

jrief:
> `change_form.js` is part of Django and needs the form id to be the model name + `"_form"`. @jrief Did you check the JS error disappears?

Yes it does. Btw. the (Django) super template does it exactly that way, so it must be the "right way".

Issue #7069, opened 2021-06-02
[BUG] Page becomes slow/unresponsive after plugin adding/editing/removing


nichoski:
## Description

On django-cms projects with a large number of sub-sites, editing/adding/removing/changing position on plugins freezes the browser for around a minute. On some devices the page crashes or becomes unresponsive indefinitely. We identified this to be caused by diff-dom library struggling with large amounts of html elements when trying to refresh the page's toolbar content. After a plugin is added or modified or removed, the page fetches new html data, then using diff-dom it creates a diff and applies it to the old toolbar html elements. However, the diff creation process takes unusually long when the toolbar contents contain a large amount of html elements. In our case, we have a django-cms project with 382 sites in it, and this process takes about 50 seconds. Increasing the number of sites increases the time exponentially.

We have a fix for this and will create a pull request shortly. If it's up to django-cms' standards, you can accept it, otherwise, feel free to suggest changes or add any of your own.

## Steps to reproduce

1. Create a django-cms project with 400 sub-sites. 
2. Create a page in any of the sites and add any plugin to it. We commonly tested this using the text plugin.
3. Confirm that after the plugin is added, the page stays greyed-out and unresponsive for around a minute. On some devices, the browser may crash.

## Expected behaviour

Refreshing the content after plugin addition should not take more than a few seconds.

## Actual behaviour

Adding a plugin on pages of projects with large amount of sites freezes the page for around a minute on some devices, or crashes the browser on others.

## Additional information (CMS/Python/Django versions)

django-cms version 3.7.4
django version 2.2.17

goutnet:
Hi @nichoski I see you are using release 3.7.4, do you think you could also try to upgrade to 3.9.0rc2 to test? (and then maybe send a PR for this version if you are working on one?)

nichoski:
@goutnet Unfortunately we don't have the option to upgrade the project's django-cms version to 3.9. Could this be backported in to a 3.7.x branch? Here is the pull request to develop:
https://github.com/django-cms/django-cms/pull/7070

PR #7787, opened 2024-01-28
fix: Port forward #7070 - faster DOM update after editing

fsbraun:
## Description

This PR ports forward the fix #7070 which improves editing workflow (solving #7069) by avoiding unresponsive pages.

The original fix in #7070 was amended by #7223 which is also included here.
 

## Related resources

* #7070
* #7223 
* #7069
## Checklist


cms/static/cms/js/modules/cms.changeform.js
```diff
@@ -2,7 +2,7 @@
 import $ from 'jquery';
 import addSlugHandlers from './slug';
 
-$(function() {
+$(function() {
```
fsbraun, line 5:
```suggestion
$(function() {
```

marksweb:
What was the diff providing here?

vinitkumar (approved):

fsbraun:
@marksweb The diff calculates the differences between the old and new toolbar and then replaces the DOM elements. That led to performance issues (#7069) and was removed in v3 in 3.9 or 3.10. The toolbar is of moderate size, and a full replacement works faster and more reliably.

marksweb (approved):

PR #7793, opened 2024-01-29
docs: Update dev policies

marksweb:
This updates a little on contributing that I think is worth mentioning here.


fsbraun:
@marksweb Do you also want to take a look at https://github.com/django-cms/django-cms/blob/develop-4/docs/contributing/code.rst? There's still reference to aldryn-boilerplate (a bootstrap3 thing). But we could also do this another time.

fsbraun (approved):

PR #7824, opened 2024-02-25
ci: Fix discord notification action inputs

marksweb:
This updates the discord action to match the inputs they list on the readme. It's unclear if this format for `args` is still valid.


Aiden-RC (approved):
looks about right to me but only one way to find out

fsbraun (approved):

PR #7834, opened 2024-03-07
Django 42 upgrade changes

vipulnarang95:
## Description
 Django 4.2 upgrade changes
 Dropped Support for Django 3.1 and below


.github/workflows/test.yml
```diff
@@ -8,12 +8,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [ 3.7, 3.8, 3.9 ]  # latest release minus two
+        python-version: [ 3.8, 3.9, 3.10 ]  # latest release minus two
```
fsbraun, line 11:
```suggestion
        python-version: [ 3.8, 3.9, '3.10' ]  # latest release minus two
```

.github/workflows/test.yml
```diff
@@ -59,12 +57,10 @@ jobs:
...
```
fsbraun, line 60:
```suggestion
        python-version: [ 3.8, 3.9, '3.10' ]  # latest release minus two
```

.github/workflows/test.yml
```diff
@@ -109,12 +105,10 @@ jobs:
...
```
fsbraun, line 108:
```suggestion
        python-version: [ 3.8, 3.9, '3.10' ]  # latest release minus two
```

fsbraun (commented):
yaml will interpret the version `3.10` as float, i.e. `3.1`

.github/workflows/test.yml
vipulnarang95: Done

joshyu:
Hi @fsbraun , please help to review this PR again.

cms/admin/settingsadmin.py
```diff
@@ -33,13 +35,13 @@ def wrapper(*args, **kwargs):
...
                 name='%s_%s_session_store' % info),
-            re_path(r'^cms-toolbar/$',
+            path('cms-toolbar/',
                 wrap(self.get_toolbar),
                 name='%s_%s_get_toolbar' % info),
-            re_path(r'^$',
+            path('',
                 wrap(self.change_view),
                 name='%s_%s_change' % info),
             re_path(r'^(.+)/$',
```
fsbraun, line 47: Can this also be replaced by `re_path`?

fsbraun (approved):
LGTM! Just one place where it might make sense to replace `re_path` with `path`?

cms/admin/settingsadmin.py
joshyu: @fsbraun , Why not use `path` instead of `re_path`? Django official document prefers the use of  `path` also.
Can you tell me the reason why we need to update above `path` with `re_path`?

fsbraun: Sorry, I meant: replace the last `re_path` in line 47 by `path`, too?

vipulnarang95: Updated. Can you review again?

fsbraun: Sure. It's already approved. Just thought you might have missed the last instance of `re_path`.  I'm also fine with keeping the `re_path`.  However, if you replace it, you'd also need to replace the regex... Probably something like `path("<path:id>/")`.

joshyu: @vipulnarang95  please remove the trailing `$` ,that will resolve the failing checks.

joshyu: @fsbraun  please review

joshyu (approved):

PR #7837, opened 2024-03-11
feat: Re-introduce page-level publish permission just as in v4.0

fsbraun:
## Description

In django CMS version 4.1 the `can_publish` permission was removed. With the next release of djangocms-versioning it will support low-level publication permissions if the versioned content model supports it.

This PR restores the `can_publish` permissions and its test as it was available in django CMS 4.0.

Tests have been restored, too.

A (no-op) squashed migration was added to ensure that publish permissions are not lost during migration processes:
* `0035` removes page publish permissions (released in 4.1.0)
* `0036` adds the page publish permissions back
* Squashed migration `0035` replaces `0035` and `0036` with a no-op

## Related resources

* https://github.com/django-cms/djangocms-versioning/pull/390/
* #...

## Checklist


vinitkumar (commented):
@fsbraun I was wondering if it's possible to have test covering this change, just so that if there are regressions in this part of code, we can catch it.

fsbraun:
@vinitkumar The publish permissions are tested. When they were removed, they were removed with their tests... Now, this PR brings both permissions and tests back.

vinitkumar (approved):

PR #7850, opened 2024-03-21
feat: Headless readiness

fsbraun:
## Description

Regular projects distribute their content through serving pages at their corresponding URL as defined in the page tree. This is achieved by adding `path("", include(cms.urls))` to a project's `urls.py`.  Also, the CMS requires `CMS_TEMPLATES` or `CMS_TEMPLATE_DIR` to be set.

This PR allows setting up **headless-ready projects** by:

1. Not requiring `path("", include(cms.urls))` being part of the projects URLs. All editing is done in the admin and the edit, structure and preview endpoints.

   The structure endpoint can be used on read-only objects (such as published pages), too, allowing to view the tree and copy plugins. All modifying actions are disabled for read-only objects. 

2. Not requiring any templates to be defined.  If no templates are available, all interaction runs through the structure endpoint. While the editing experience might be limited, this can be a fast way of setting up a project. 

3. This PR moves the create wizard URL to the placeholder admin (which contains all editing endpoints) (but also keeps the previous endpoint in `cms.urls` for the wizard for compatibility reasons).

4. The data bridge allows the frontend to update the structure board after changes in the plugin tree. This PR changes the HTML response for changes in the plugin tree from inlining the data in a script to a script tag containing JSON objects which are read and passed to the data bridge.

## Related resources

* #7835 
* #7927

## Checklist


vinitkumar:
@fsbraun Great work on this. I would love to see how this works in practise and what we can do with it?

fsbraun:
This PR includes JS and CSS changes to disable UI items in the structure board. Those changes only become visible after running `gulp sass bundle`.

Still, a headless-ready project for testing can be quickly set up using these steps:
* Create a fresh venv: `python3 -m venv .venv`
* Activate it: `source .venv/bin/activate`
* Install PR: `pip install git+https://github.com/fsbraun/django-cms@feat/data-bridge`
* Create project: `djangocms headlesstest --template https://github.com/django-cms/cms-template/tarball/feat/headless`
* Spin up test server: `cd headlesstest && ./manage.py runserver`
* Go to admin in the browser: `http://127.0.0.1/en/admin/`

The template only installs djangocms-text as plugins but works with any plugins.

@vinitkumar @marksweb

docs/how_to/21-headless.rst
```diff
@@ -0,0 +1,114 @@
...
+
+.. code-block:: python
+
+    CMS_PLACEHOLDERS = (
+        ('single', ('content'), _('Single placeholder')),
+        ('two_column', ('left', 'right'), _('Two columns')),
+    )
+
+The ``CMS_PLACEHOLDERS`` setting is a list of tuples. Each tuple represents a
+placeholder configuration. The first element of the tuple is the name of the
```
macolo, line 106: I suggest we link to `CMS_PLACEHOLDER_CONF` (https://docs.django-cms.org/en/latest/reference/configuration.html#cms-placeholder-conf) here, it wasn't intuitive to me at first glance what "placeholder configuration" means.

docs/how_to/21-headless.rst
```diff
@@ -0,0 +1,114 @@
...
+placeholder configuration. The second element is a tuple of placeholder names.
```
macolo, line 107: It's not clear to me why a placeholder can have multiple names.

docs/how_to/21-headless.rst
```diff
@@ -0,0 +1,114 @@
...
+The third element is the verbose description of the placeholder configuration
+which will be shown in the user interface.
```
macolo, line 109: Is this referring to the editor's structure sidebar?

docs/reference/configuration.rst
```diff
@@ -168,6 +164,23 @@ for translation.
...
+CMS_PLACEHOLDERS
+================
+
+default
+    ``(('', ('content',), _("Single placeholder")),)``
+
+A list of placeholders that can be added to a page. The first element of the
+tuple is the name of the placeholder configuration. The second element is a
+tuple of placeholder names. The third element is the verbose description of the
+placeholder configuration which will be shown in the user interface.ed
```
macolo, line 178: Typo at the end of the line?

fsbraun:
@macolo I've tried to improve the clarity of the `CMS_PLACEHOLDERS` setting. It is supposed to replace `CMS_TEMPLATES` in a headless-only config.

cms/admin/pageadmin.py
```diff
@@ -1194,8 +1194,13 @@ def change_template(self, request, object_id):
 
         to_template = request.POST.get("template", None)
```
marksweb, line 1195: What does "to" mean in this context? I'm trying to make sense of the variable naming.

Is it something like the page content is going to go in to this template?

fsbraun: @marksweb Good question. I assume it should signify the value that goes into the `template` of the page content object. This, of course, is the template that the page content will be rendered in. 

A more traditional way of naming would just be `template`.

macolo:
@fsbraun the docs now make a lot of sense to me, easy to understand! 💪  I like the design of this functionality, it seems quite intuitive to me from a developer's perspective. 👍

macolo (approved):

Issue #7849, opened 2024-03-21
[BUG] Unpublished page is shown instead of 404


jrief:
## Description

If a page exists in multiple languages, for instance German and English, and the English version is published but the German version is unpublished, then the German version does not show a "404 – not found page", as one would expect. 

## Steps to reproduce

**djangocms-versioning** is installed and enabled.

The CMS language configuration settings are:

```python
CMS_LANGUAGES = {
    1: [{
        'public': True,
        'code': 'de',
        'hide_untranslated': False,
        'name': "Deutsch",
        'redirect_on_fallback': True,
    }, {
        'public': True,
        'code': 'en',
        'hide_untranslated': True,
        'name': "English",
        'redirect_on_fallback': True,
    }],
    'default': {
        'fallbacks': [],
        'redirect_on_fallback': True,
        'public': True,
        'hide_untranslated': False,
    },
}
```

* Create a page somewhere in the page tree.
* Add content to the German page.
* Publish the German page.
* Create English content for that same page. The URL can be the same as for the German page. Example: `/en/path/to/page` vs. `/de/path/to/page`.
* Publish the English page.
* Switch back to the German page.
* Unpublish the German page.
* The URL `/de/path/to/page` still shows the existing content in German. This is even true for anonymous users.

## Actual & Expected behaviour

One would expect that an unpublished page does not show up and renders a "404 – Not found page"

I also tried to change the setting of `CMS_LANGUAGES['default']['fallbacks'] = ['de', 'en']`. The one would expect to see the English version of a page, because of the fallback value. However, that is not the case.

Btw. Pages which should show up as "404 – Not found page", actually are empty. This means that the template is rendered with empty placeholders.

## Additional information (CMS/Python/Django versions)

* Django-4.1.13
* Django-CMS-4.1 (HEAD on develop-4, commit 21d6a6defc5a6321bd691e43eb57c112508b952f)

## Do you want to help fix this issue?

* [x] Yes, I want to help fix this issue and I will join #workgroup-pr-review on [Slack](https://www.django-cms.org/slack) to confirm with the community that a PR is welcome. 
* [ ] No, I only want to report the issue.

jrief:
In my opinion the culprit is this line of code:
https://github.com/django-cms/django-cms/blob/21d6a6defc5a6321bd691e43eb57c112508b952f/cms/views.py#L228

shouldn't it be
```python
content = page.get_content_obj(language=content_language, fallback=False)
```
instead?

[...]

PR #7853, opened 2024-03-22
fix: `views.details` revealed existence of unpublished language

fsbraun:
## Description
`page.get_languages()` contains a list of languages, for which translations exist. It does not imply those languages should be visible to the public. djangocms-versioning hides languages upon unpublishing.

This PR removes the access to `page.get_languages()` from django CMS' central view in `views.details`. To avoid additional db accesses, the (public) page content objects a page are prefetched and the available languages are determined through the prefetch cache. This might even reduce the number of DB accesses upon accessing a CMS URL by one (not tested).

## Related resources

* #7849 
* #...

## Checklist

### DIFF
--- a/cms/cms_toolbars.py
+++ b/cms/cms_toolbars.py
@@ -429,8 +429,9 @@ class PageToolbar(CMSToolbar):
 
         # if the current page has a parent in the request's current language redirect to it
         if parent_page and language in parent_page.get_languages():
-            with force_language(language):
-                return parent_page.get_absolute_url(language=language)
+            return get_object_preview_url(
+                parent_page.pagecontent_set(manager="admin_manager").latest_content(language=language).first()
+            )
 
         # else redirect to root, do not redirect to Page.objects.get_home() because user could have deleted the last
         # page, if DEBUG == False this could cause a 404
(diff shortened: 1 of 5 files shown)


jrief:
If I now try to open the English version of a page, I get this exception:

```
 File "django-cms/cms/toolbar/toolbar.py", line 510, in _call_toolbar
    getattr(toolbar, func_name)()
  File "djangocms-versioning/djangocms_versioning/cms_toolbars.py", line 340, in populate
    self.page_content = self.get_page_content() if self.page else None
  File "djangocms-versioning/djangocms_versioning/cms_toolbars.py", line 336, in get_page_content
    return get_latest_admin_viewable_content(self.page, language=language, include_unpublished_archived=True)
  File "djangocms-versioning/djangocms_versioning/helpers.py", line 329, in get_latest_admin_viewable_content
    return qs.filter(**extra_grouping_fields).latest_content().first()
AttributeError: 'QuerySet' object has no attribute 'latest_content'
```

btw.:

```
qs.filter(**extra_grouping_fields).first()
```

delivers an object, so presumably the wrong manager is used.

fsbraun:
@jrief  Django's prefetch cache ignores the queryset. That leads to the error you mention.

I've now exchanged the prefetch by a separate call to fill the page content cache. Turns out this is neutral on db hits since it was filled anyways. Can you give it another test?

jrief:
@fsbraun Thanks for the quick fix. This now works as expected.

PR #7868, opened 2024-04-09
fix: #7859: allow special characters in page title

### LINKED ISSUE #7859: [BUG] Special characters in page title
## Description

If a page title contains special characters, such as `&`, then this character is rendered as HTML-entity, for instance `&amp;`.
In browser tabs this does not look right.

## Steps to reproduce

* Create a page template where the page title is used. This usually is done by adding

```html
…
<meta>
  {% page_attribute "page_title" as page_title %}
  <title>{{ page_title}}</title>
  …
</meta>
…
```

* Create or edit a page and use `A & B` as the page title.
* Load the page.

The browser tab shows `A &amp; B`.

## Expected behaviour

The browser tab should show `A & B` as page title.

By patching the templatetag `cms.templatetags.cms_tags.PageAttribute` and [not escaping the value](https://github.com/django-cms/django-cms/blob/fa7b89ceec3bd22e031be8e9cf2328a3f9434b12/cms/templatetags/cms_tags.py#L411), this can be fixed easily. However, I don't know what security implications this fix may have.

I would exempt escaping the string, whenever the name is `page_title`, just as we do it with values of type `datetime`.

Any objections to this?

jrief:
## Description

Instead of escaping the page title, this PR just strips its HTML tags. This prevents XSS-attacks as well but keeps HTML entities. Otherwise a page title containing an `&` will be escaped to `&amp;` which looks ugly in the browser's tabs.

## Related resources

* #7859

## Checklist


cms/tests/test_templatetags.py
```diff
@@ -128,7 +128,7 @@ class FakeRequest:
         template = '{% load cms_tags %}{% page_attribute page_title %}'
         output = self.render_template_obj(template, {}, request)
         self.assertNotEqual(script, output)
-        self.assertEqual(escape(script), output)
+        self.assertEqual(strip_tags(script), output)
```
marksweb, line 131: I feel like we need a second check to be more explicit about the content here. Something to prevent regression and verifies the content renders as we want.

jrief: Isn't `page_attribute page_title` *exclusively* intended to be used for the `<title>…</title>`-tag? Then according to the [specs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title#technical_summary) we should sanitize it so that the permitted content "is not inter-element [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace)".

Then using `escape(…)` is an absolute wrong approach. `strip_tags` is better but still not according to the specs. How could we sanitize the whitespace?

fsbraun: We could rely on Django's auto-escape: If we do not escape or strip tags, Django will escape the result. This should be backwards compatible (we should check) and allow the developer to use the appropriate filter like `strip_tags` or `escape` or `strip_tags`.

fsbraun: I think I have to retract my comment 🙁: It seems that while auto-escape works, still `{% page_attribute page_title as title %}{{ title|striptags }}` does not work. It seems auto-escape is also executed when assigning the tag's result to the `title` context var. In this example, `title` already contains the escaped title.

So, I would suggest to **not** change this and use `{{ request.current_page.get_page_title|striptags }}` as a work-around.

fsbraun (approved):

PR #7880, opened 2024-04-18
fix: CMS widgets need not load if they are read only

### LINKED ISSUE #7879: [BUG] Console error in browser when opening PageContentAdmin
## Description

I'm experiencing this error in the console of the browser:

Screenshot 2024-04-18 at 16 08 30

Note that I bundled the client code using `gulp bundle --debug` to obtain more verbose information.

## Steps to reproduce

* Open the page tree.
* Chose a page which is published, ie. having a green bullet.
* Click on the PageContent editor (button with 3 sliders).

## Expected behaviour

No error in the browser console.

## Actual behaviour

Browser reports:
`window.CMS.Widgets._pageSmartLinkWidgets` is undefined.

Side note:

All the fields of that page's editor are readonly. If I open the editor of a page which is not published, then the editor opens with all the fields, but editable. Then this error does not occur.

Branch `develop-4` current HEAD (92a1086decd082f1645e73bf6c8ab4d501b0e194)

Since I'm not aware of the purpose of this, I probably can't fix this myself.

fsbraun:
## Description

In django CMS 4 form fields that use CMS widgets might be read-only (e.g., if the version is published). In that case, widgets are not loaded into `window.CMS.Widgets`. This PR only activates widgets if they are loaded.

Fixes #7879.

This PR does not change any logic (only prevents a js script error to be thrown), hence no tests.

## Related resources

* #7879 
* #...

## Checklist

### DIFF
--- a/cms/static/cms/js/widgets/forms.pagesmartlinkwidget.js
+++ b/cms/static/cms/js/widgets/forms.pagesmartlinkwidget.js
@@ -95,8 +95,10 @@ require.ensure([], function (require) {
     window.CMS.PageSmartLinkWidget = PageSmartLinkWidget;
 
     $(function () {
-        window.CMS.Widgets._pageSmartLinkWidgets.forEach(function (widget) {
-            new PageSmartLinkWidget(widget);
-        });
+        if (window.CMS.Widgets !== undefined && window.CMS.Widgets._pageSmartLinkWidgets !== undefined) {
+            window.CMS.Widgets._pageSmartLinkWidgets.forEach(function (widget) {
+                new PageSmartLinkWidget(widget);
+            });
+        }
     });
 }, 'admin.widget');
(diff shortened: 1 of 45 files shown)


jrief:
Yes, this PR fixes the problem.

For curiosity, is it intended that `window.CMS.Widgets._pageSelectWidgets` might be undefined?

jrief (changes requested):
Wouldn't it be nicer to write:

```javascript
window.CMS.Widgets?._pageSelectWidgets?.forEach
```

cms/static/cms/js/widgets/forms.pagesmartlinkwidget.js
jrief, line 1:
```
window.CMS.Widgets?._pageSmartLinkWidgets?.forEach
```

fsbraun:
@jrief Good news. 

I hesitate to use optional chaining, since it is "only" supported by newer browsers (~95% according to can I use). django CMS' whole JS code base does not use it and I reckon the two lines are not worth starting to think about compatibility.

marksweb (approved):

PR #7887, opened 2024-04-23
perf: small reduction of DB queries

jrief:
## Description

This PR reduces the number of DB queries by a few.

## Related resources

1. chunk: fallback languages can be cached inside the `Page` object.
2. chunk: Replacing the `.filter(…)` on the queryset against `.all()` actually reduces the number of queries. This is because that queryset in most cases is already consumed. Calling it with `.filter(…)` actually fetches the data a second time. Therefore filtering in Python is cheaper, especially if the number of languages never exceeds a significant number.

Also read https://docs.djangoproject.com/en/5.0/ref/models/querysets/#values

## Checklist

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -198,6 +198,7 @@ class Page(models.Model):
         super().__init__(*args, **kwargs)
         self.urls_cache = {}
         self.page_content_cache = {}
+        self.fallback_languages_cache = {}
 
     def __str__(self):
         try:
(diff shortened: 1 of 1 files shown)


fsbraun:
I like the `all()` optimisation.

How does caching the fallback languages help performance. The fallback languages only depend on the settings, or am I missing something?

vinitkumar:
@jrief Are there any updates on this? Also, would be great to see the number of queries reduction shown by Django Debug Tool or something? In my opinion, it would be help remember the rationale behind the change in future.

fsbraun:
@jrief One change has been merged into `develop-4` (https://github.com/django-cms/django-cms/blob/59d50f2492b574f5ebd083093c846e04792a4049/cms/models/pagemodel.py#L780). The other change (caching fallback languages) imho does not reduce db hits, since fallback languages are a matter of configuration only.

Can we close this?

PR #7888, opened 2024-04-24
ci: Hopefully a fix for discord message

marksweb:
This hopefully fixes the discord message failures;

https://github.com/django-cms/django-cms/actions/workflows/new_contributor_pr.yml


marksweb:
The failure here is because the actions run on the code in the target repo rather than the new code in this PR. Though that's not to say this doesn't still fail - it does remove the comment in the JSON that appears to be causing the current parse error.

fsbraun (approved):

PR #7920, opened 2024-05-14
fix: Only switch to structure endpoint on viewport breakpoint if in edit mode

### LINKED ISSUE #6566: .cms-overflow class on html makes whole page unscrollable after window resize
### Summary

When logged in and not in editing mode, if a make the browser window smaller than 1024px, my html tag gets the `cms-overflow` class, which results in a page that can no more be scrolled at all. A fresh reload solves the problem. To reproduce, make the viewport bigger than 1024px, and smaller again.

### Expected behaviour

The cms leaving scrolling as is.

### More infos

Downgrade to 3.5.2 solves the problem.

### Environment

* Python version: 2.7.14
* Django version: 1.11.latest
* django CMS version: 3.5.3

### LINKED ISSUE #7906: [BUG] Window resize event triggers editing mode
## Description

The Windows resize event triggers a function which put the page in edit mode.

## Steps to reproduce

* As user with editing permission, open any page.
* The toolbar must be active.
* Ensure that the page is in published mode, ie. the canonical URL of the page must be visible.
* Resize the window.

This immediately puts the page into edit mode, ie. the URL changes to something such as: 

`/de/admin/cms/placeholder/object/53/edit/72715/`

## Expected behaviour

Resizing the window should not have an effect in edit mode.

Happens in version 4.1.1 with the latest bundle of `*.js` files.

I was able to drill down a little bit. If I debug the page using the developer tools I can see that the window resize event is registered twice, in `toolbar.js` and `bundle.toolbar.min.js`. If I remove the second event handler, this effect does not happen anymore.

Unfortunately, even rebundling the JS files using `gulp bundle --debug` doesn't show me the handler, where this happens.

fsbraun:
## Description

Fixes #7906 . Fixes #6566 .

* Depending on viewport size the toolbar shows the structure or edit breakpoint. This is done to avoid ~2s latency when loading the edit endpoint on small devices. This must not happen at the preview endpoint.

## Related resources

* #7906 
* #6566 

## Checklist

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -1051,6 +1051,24 @@ class PageContentAdmin(admin.ModelAdmin):
         )
         return can_change_page
 
+    def has_view_permission(self, request, obj=None):
+        """
+        Return true if the current user has permission on the page.
+        Return the string 'All' if the user has all rights.
+        """
+        # Identical to has_change_permission, but will remain untouched by any subclassing
+        # as done, e.g., by djangocms-versioning
+        site = get_site(request)
+
+        if obj:
+            return page_permissions.user_can_change_page(request.user, page=obj.page, site=site)
+        can_view_page = page_permissions.user_can_change_at_least_one_page(
+            user=request.user,
+            site=get_site(request),
+            use_cache=False,
+        )
+        return can_view_page
+
     def has_delete_permission(self, request, obj=None):
         """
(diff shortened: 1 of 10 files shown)


fsbraun:
@jrief Can you give this a test run to verify this fixes your issue?

vinitkumar (approved):

PR #7947, opened 2024-07-03
feat: merge page with node tree

jrief:
## Description

This is the (huge) pull request to merge models `Page` and `TreeNode` into `Page`.

There is some additional cleanup to do, because I haven't yet removed all the legacy code for reference.
All unit tests are working.

This pull request shall go into version 4.2.

@fsbraun Please consider to apply it before merging other pull requests, since this presumably will cause a lot of merging conflicts.

## Related resources

It is based on the discussion #7873 and we talks about this a few times during
the biweekly meetings.

## Checklist


cms/admin/forms.py
```diff
@@ -455,9 +452,7 @@ def get_or_create_root(self):
         Creates the root node used to store all page types
         for the current site if it doesn't exist.
         """
-        root_page = PageType.get_root_page(site=self._site)
-
-        if not root_page:
+        if not (root_page := PageType.get_root_page(site=self._site)):
```
fsbraun, line 455: I would like to opt for better readability here and avoid the walrus operator at least in a nested statement (if not at all).

```suggestion
        root_page = PageType.get_root_page(site=self._site)
        if not root_page:
```

cms/admin/forms.py
```diff
@@ -302,14 +302,12 @@ def clean(self):
             # addressed first.
             return data
 
-        parent_node = data.get("parent_node")
-
-        if parent_node:
-            slug = data["slug"]
-            parent_path = parent_node.item.get_path(self._language)
+        if parent_page := data.get('parent_node'):
```
fsbraun, line 305: Here also, I'd propose to spend one additional line of code for readability:
```suggestion
        parent_page = data.get("parent_node")
        if parent_page:
```

cms/models/query.py
```diff
@@ -4,14 +4,23 @@
 from cms.exceptions import NoHomeFound
 
 
-class PageQuerySet(QuerySet):
+class PageQuerySet(MP_NodeQuerySet):
```
fsbraun, line 7: Can we add `select_related` to this queryset calling `super().select_related` but issuing a warning if `node` or `node__...` is selected (and, of course, removing node from the selection string)?

Similar behaviour for `filter`or `order_by` would be great, e.g.:
```
def order_by(self, *fieldnames):
    modified = [field[6:] if isinstance(field, str) and field.startswith("node__") else field for field in fieldnames]
    if modified != fieldnames:
        warnings.warn(RemovedInDjangoCMS43, "As of django CMS 4.2 the Page model does not have a node property any more. See (link to docs) for more information. This will cause an error in django CMS 4.3 and above.", stacklevel=2)
    return super().order_by(*modified)
```

cms/models/query.py
```diff
@@ -26,6 +35,13 @@ def has_apphooks(self):
...
 
+    def delete_fast(self):
+        # calls django's delete instead of the one from treebeard
+        super(MP_NodeQuerySet, self).delete()
+
+    def root_only(self):
+        return self.filter(depth=1)
+
 
 class PageNodeQuerySet(MP_NodeQuerySet):
```
fsbraun, line 46: I think it's fair to remove `PageNodeQuerySet`.

cms/models/pagemodel.py
```diff
@@ -52,7 +53,7 @@ class TreeNode(MP_Node):
 
     class Meta:
         app_label = 'cms'
-        ordering = ('path',)
+        ordering = ['path']
```
fsbraun, line 56: I might be overlooking something, but can't `TreeNode` be removed. I'm not sure what a custom app would do with it. @marksweb What do you think? Remove or keep for compatibility?

cms/admin/forms.py
```diff
@@ -926,49 +921,42 @@ def __init__(self, *args, **kwargs):
         self._site = kwargs.pop("site", Site.objects.get_current())
         super().__init__(*args, **kwargs)
         self.fields["target"].queryset = Page.objects.filter(
-            node__site=self._site,
+            site=self._site,
             is_page_type=self.page.is_page_type,
         )
 
-    def get_root_nodes(self):
```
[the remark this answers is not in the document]
jrief: The modified pull request now includes:

> A `node` property for Page that returns the page itself. Thereby, custom apps that access page.node.something will not break.

**done**

> Remove `node__` prefixes from `filter`, `order_by`, `select_related` in `PageQuerySet`

**done**

> Rename `page.parent` to `page.page_parent` to be compatible with django CMS 4.1 and earlier

`page.page_parent` now accesses `page.parent` and issues a deprecation warning. By convenience `parent` is the term used in all trees. This was already the case for `TreeNode` and is also used in [django-tree-queries](https://django-tree-queries.readthedocs.io/en/latest/#features-and-limitations).

> Use the RemovedInDjangoCMS43 warning message to mark compatibility code that will need to be removed in the future.

**done**


I also removed all the legacy code including model `TreeNode` itself.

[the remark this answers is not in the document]
jrief: > On the smaller things, I've noticed that you use both single and double quotes (' and "). I'd suggest being consistent and opting for double quotes

Django-CMS is not very consistent on this and Django only since version 4.2.

In my projects I use this rule: If it's intended for humans, I use `"…"`-quotes. If it's an internal string (key, fieldname, attribute, etc.), I use `'…'`-quotes. This idea was propagated by Douglas Crockford many years ago.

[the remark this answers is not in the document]
jrief: > I would like to opt for better readability here and avoid the walrus operator at least in a nested statement (if not at all).

In my opinion less lines of code are easier to read. Therefore Python's decision to add the walrus operator to the language makes a lot of sense. I now removed it in `not(…)`-statements, although I would prefer to keep it.

fsbraun (approved):
Well done! LGTM!

PR #7950, opened 2024-07-08
docs: Update compatibility table

jgadelange:
## Description

According to the release notes since version 3.11.5 Django-CMS is compatible with python 3.12

## Related resources

* https://github.com/django-cms/django-cms/releases/tag/3.11.5

## Checklist


docs/compatibility.include
```diff
@@ -4,6 +4,7 @@ django CMS   Python                                  Django
 \            3.12 3.11 3.10 3.9 3.8 3.7 3.6  5.0 4.2 4.1 4.0 3.2 3.1 3.0 2.2
 ===========  ==== ==== ==== === === === ===  === === === === === === === ===
 4.1.x        ✓    ✓    ✓    ✓   ×   ×   ×    ✓   LTS ✓   ✓   LTS ×   ×   ×
+3.11.5+      ✓    ✓    ✓    ✓   ✓   ✓   ×    ×   LTS ✓   ✓   LTS ×   ×   ×
 3.11.3+      ×    ✓    ✓    ✓   ✓   ✓   ×    ×   LTS ✓   ✓   LTS ×   ×   ×
```
marksweb, line 8: I feel like we shouldn't be specifying the patch release compatibility in here. It's too verbose & makes it more complex.

If someone is on a minor version, it should just be expected that they keep that updated with new patch releases.

@fsbraun what do you think about that?

jgadelange: I agree with you, however knowing which minor version adds the support does have added value. Maybe we could than add a "added in 3.11.5" behind the check mark. This is what [Django does for python version compatibility](https://docs.djangoproject.com/en/4.2/faq/install/#what-python-version-can-i-use-with-django).

For now I followed the same pattern as the three lines below (3.11.0, 3.11.1 and 3.11.3+), I think these lines should all be combined in to one 3.11.x line with maybe some footnotes describing what happened describing the stuff why the multiple lines are needed now.

marksweb: Ok that format/layout for specific version support is good maybe we could do that as something new. 👍

For now, I think people would check the changelog for that level of detail.

docs/compatibility.include
```diff
@@ -4,6 +4,7 @@ django CMS   Python                                  Django
 \            3.12 3.11 3.10 3.9 3.8 3.7 3.6  5.0 4.2 4.1 4.0 3.2 3.1 3.0 2.2
 ===========  ==== ==== ==== === === === ===  === === === === === === === ===
 4.1.x        ✓    ✓    ✓    ✓   ×   ×   ×    ✓   LTS ✓   ✓   LTS ×   ×   ×
+3.11.5+      ✓    ✓    ✓    ✓   ✓   ✓   ×    ×   LTS ✓   ✓   LTS ×   ×   ×
```
fsbraun, line 7:
```suggestion
3.11.6+      ✓    ✓    ✓    ✓   ✓   ✓   ×    ✓   LTS ✓   ✓   LTS ×   ×   ×
```
Since 3.11.6, also django 5.0 is supported. Shall we also add this?

docs/compatibility.include
fsbraun: @marksweb @jgadelange Actually, we might just have one 3.11.6+ line with the current supported versions. That would limit complexity and still contain the most important information. If you needed to know more details, you could always go to the release notes.

docs/compatibility.include
```diff
@@ -4,6 +4,7 @@ django CMS   Python                                  Django
 \            3.12 3.11 3.10 3.9 3.8 3.7 3.6  5.0 4.2 4.1 4.0 3.2 3.1 3.0 2.2
 ===========  ==== ==== ==== === === === ===  === === === === === === === ===
 4.1.x        ✓    ✓    ✓    ✓   ×   ×   ×    ✓   LTS ✓   ✓   LTS ×   ×   ×
+3.11.5+      ✓    ✓    ✓    ✓   ✓   ✓   ×    ✓   LTS ✓   ✓   LTS ×   ×   ×
 3.11.3+      ×    ✓    ✓    ✓   ✓   ✓   ×    ×   LTS ✓   ✓   LTS ×   ×   ×
```
fsbraun, line 8:
```suggestion
```

@marksweb @jgadelange I'd only go for one line here.

docs/compatibility.include
jgadelange: Yeah that's ok!

fsbraun (approved):

PR #7954, opened 2024-07-11
fix: Fail silently when rendering a placeholder on a missing toolbar object

### LINKED ISSUE #7952: [BUG] Placeholder rendering fails loudly without PageContent or Model
## Description

Non-root AppHook pages or Views that aren't tied to a model with placeholders will fail loudly if their parent template has any placeholders that are rendered with `{% placeholder "name" %}`

In CMS3 this works fine when inheriting and it just ignores the placeholders, but in CMS4 it breaks because it seemingly tries to render the placeholders. For example in our base:
```
{% block hero %}
    {% placeholder 'hero' %}
{% endblock hero %}
```

```
class SomeView(DetailView):
    template_name = "app/template.html"

    def get_object(self, queryset=None):
        return request.user    # Something here that doesn't define any placeholders
```
If we leave this block in our template.html it will attempt to render and break with `NoneType should implement get_template` because there's no PageContent or toolbar object.

## Steps to reproduce

1. Create an app with at least 1 non-root view. Make sure there's a `{{ placeholder "name" }}` in the template
2. Add the app to a page
3. Go to the sub-page link

## Expected behaviour

In CMS3 this simply doesn't render the placeholder. The page can load as normal with only the available placeholders

## Actual behaviour

In CMS4 this fails loudly with `NoneType should implement get_template`

Only requires django-cms > 4

## Linked issue
https://github.com/django-cms/django-cms/issues/7712

fsbraun:
## Description
Fixes #7952

If there is no toolbar object available, trying to render a placeholder now raises the `PlaceholderNotFound` exception.

This leads the `{% placeholder %}` template tag to silently fail and return and empty string.

## Related resources

* #7952
* #...

## Checklist

### DIFF
--- a/.github/workflows/new_contributor_pr.yml
+++ b/.github/workflows/new_contributor_pr.yml
@@ -29,6 +29,7 @@ jobs:
     needs: new
     steps:
       - name: Send Discord Webhook
+        continue-on-error: true
         env:
           DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK_URL }}
           DISCORD_EMBEDS: |
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -13,6 +13,7 @@ from django.utils.safestring import mark_safe
 from django.utils.translation import override
 
 from cms.cache.placeholder import get_placeholder_cache, set_placeholder_cache
+from cms.exceptions import PlaceholderNotFound
 from cms.models import PageContent, Placeholder
 from cms.toolbar.utils import (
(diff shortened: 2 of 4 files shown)


.github/workflows/new_contributor_pr.yml
```diff
@@ -29,6 +29,7 @@ jobs:
     needs: new
     steps:
       - name: Send Discord Webhook
+        continue-on-error: true
```
marksweb, line 32: 👍

marksweb (approved):

PR #7956, opened 2024-07-11
fix: Refactor menus app: significant time saving (queries and cpu)

fsbraun:
## Description
Menu creation takes significant time for installations with a large number of pages. This PR reduces time needed for creating the menu and rendering the menu. Measurements were done on a test installation **with ~55,000 pages** (SQLite db).

### Improvement 1: Generating the navigation node list
Creation of menu nodes is cached. So, this is a one-time gain after, say, adding a page.

The efficiency gain comes from identifying the relevant page content object only once per page object when translating the page list into a menu node list.

### Improvement 2: Rendering the menu
The second angle of attack are the menu template tags: They depend on the `cut_levels` implementation, which was very inefficient. Especially, the `extra_inactive` parameter indicating how many inactive children should be present in the navigation had a computational time of O(n) with n being the number of pages/menu entries. The PR makes reduces this O(1).

### Improvement 3: Versioning-induced db hits

An N+1 issue is introduced when generating the menu nodes through djangocms-versioning. Each menu node is checked against the home page to check if it needs to be cut. For this, the home page needs to be not in the navigation. Since, "not in navigation" is a page content property, this check triggers a database hit to find the relevant page content version. This PR only checks once if the home page is in the menu.

### Additional improvement potential

* No partial update of cached node lists.

@jrief Can you help to test this in a realistic environment? 

## Related resources

* #7805
* #7933 (one key argument being here building the menu)

## Checklist


menus/menu_pool.py
```diff
@@ -204,10 +204,36 @@ def _build_nodes(self):
...
-            node.selected = node.is_selected(self.request)
+            if node.is_selected(self.request):
+                node.selected = True
+                self._mark_ancestors(node)
+                self._mark_descendants(node)
+                self._mark_siblings(node, root_nodes)
+                return nodes
         return nodes
 
+    def _mark_ancestors(self, node):
```
marksweb, line 217: It'd be good to have some docstrings on these new functions, despite how simple they may be.

marksweb (approved):

jrief:
@fsbraun 
> @jrief Can you help to test this in a realistic environment?

on our system, this merge request now is running in production and a huge relief for our customers, especially those which are not superusers.

PR #7972, opened 2024-07-29
docs: add documentation for search

filipweidemann:
## Description

Add documentation on how to setup basic search indexes using django-haystack.
Also reference extensions that do most of the work without needing to reinvent it per code base.

It is a docs-only PR, so no changed tests. 

This PR resolves https://github.com/django-cms/django-cms/issues/7571 

## Related resources
* https://github.com/CZ-NIC/djangocms-aldryn-search
* https://github.com/Lfd4/djangocms-haystack

## Checklist

### DIFF
--- a/docs/explanation/searchdocs.rst
+++ b/docs/explanation/searchdocs.rst
@@ -52,12 +52,11 @@ A very simple index could look something like this:
             ).distinct()
 
         def prepare(self, instance: PageContent) -> dict:
-          request = self.get_request_instance(current_language)
           self.prepared_data = super().prepare(instance)
           self.prepared_data["url"] = instance.page.get_absolute_url()
           self.prepared_data["title"] = instance.title
           self.prepared_data["text"] = (
-            self.prepared_data["title"] + instance.page.meta_description or ""
+            self.prepared_data["title"] + (instance.meta_description or "")
           )
           return self.prepared_data
 
--- a/docs/explanation/searchdocs.rst
+++ b/docs/explanation/searchdocs.rst
@@ -63,7 +63,7 @@ A very simple index could look something like this:
 
 The above snippet uses the standard ``text`` field that is recommended by 
 ``django-haystack`` to store all the indexable content. There is also a 
-seperate field for the title because you may want to display it as a heading
+separate field for the title because you may want to display it as a heading
 in your search result, and a field for the URL so you can link to the pages.
 
 The indexed content here is *not* using a template (which is one of the options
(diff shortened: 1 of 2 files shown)


docs/explanation/searchdocs.rst
```diff
@@ -1,2 +1,131 @@
...
+        text = indexes.CharField(document=True, use_template=False)
+        title = indexes.CharField(indexed=False, stored=True)
+        url = indexes.CharField(indexed=False, stored=True)
+
+        def get_model(self):
+            return PageContent
+
+        def index_queryset(self, using=None) -> models.QuerySet:
+            return self.get_model().objects.filter(
+              versions__state="published",
```
fsbraun, line 50: I don't think the `versions_state` filter is necessary (indeed, maybe confusing, since you might not have installed djangocms-versioning). djangocms-versioning or any other versioning package must modify the `objects` manager so that it only returns published objects anyway.

fsbraun (changes requested):
Hi @filipweidemann ! 

This sounds great to me and is an excellent extension for the docs! I only have one comment to it: We should not rely on the details of the djangocms-versioning implementation. 

@marksweb Any thought?

docs/explanation/searchdocs.rst
```diff
@@ -1,2 +1,131 @@
...
+              language=using
+            ).distinct()
```
fsbraun, line 52:
```suggestion
            return self.get_model().objects.filter(language=using)
```

docs/explanation/searchdocs.rst
filipweidemann: Good catch, I didn't know this yet. Your proposed change is committed.

fsbraun: @filipweidemann Could you also test it?

filipweidemann:
The codespell job was failing, I pushed a new commit to correct the spelling error. Should be good now if y'all are happy with it as well :)

fsbraun (approved):

docs/explanation/searchdocs.rst
filipweidemann: Oh I didn't see this, I did test this & merged your PR in our own version earlier today. Works as intended ;)
https://github.com/Lfd4/djangocms-haystack/pull/12#issuecomment-2258246830

fsbraun:
The [djangocms-haystack](https://github.com/Lfd4/djangocms-haystack) package also provides a search index for django-haystack for django CMS >= 4.

PR #7989, opened 2024-09-08
feat: allow plugins to disable their edit modal

fsbraun:
## Description

As discussed in #7982, some plugins' configuration might not be edited. Examples are:
* Column plugins (in many cases)
* Slot-like plugins which do not have their own configuration

This PR allows plugins to mark themselves as **non-editable** by setting
* `edit_disabled` class property of the plugin class, or by setting
* `edit_disabled` instance property on the plugin instance (potentially dynamically)

If either is set, the structure board will not offer the edit button, and double-clicking the plugin on the page or in the structure view will not open the edit modal.

Finally, if the plugin instance has a `add_structureboard_classes` property, those classes will be added to the drag item in the structure board. This can be used to style plugins in the structure board dynamically. Styles would need to be added in a custom `cms_toolbar.py`.

## Example

This PR does **not** include the grey text color. I included this to show which plugins do **not** have edit modals:
* No pencil button
* Double-click does not do anything

image

## Related resources

* #7982 
* #...

## Checklist

## Summary by Sourcery

Allow plugins to disable their edit modal by setting the `edit_disabled` property on the plugin class or instance. Add the ability to style plugins in the structure board dynamically by adding classes to the `add_structureboard_classes` property.

New Features:
- Allow plugins to disable their edit modal.
- Allow plugins to add custom classes to the drag item in the structure board.


macolo:
@fsbraun nice! I thought a bit about this and have two use cases:

1) would dragging a `edit_disabled` plugin outside the controlling parent  be still possible, e.g. can an editor (accidentally?) destroy a parent-child complex designed by a developer so that it becomes unusable, especially as the controlling parent plugin template (probably?) counts on the existence of specific child plugins in a specific order?

2) dragging a `edit_disabled` plugin inside the controlling parent, e.g. changing the order of its child plugins. This might cause a similar issue, the controlling parent's template might expect the controlled children to be in a specific order?

In summary, would it make sense to also disallow direct drag & drop for `edit_disabled` plugins? 

Of course, dragging them via a parent plugin should always be possible, as the whole complex can be dragged to another position in the placeholder, can be cut or copied and pasted elsewhere.

fsbraun:
@macolo Draging is still fully controlled by `allowed_parents` and `allowed_children`. This means I could change the order of columns or move a column to a different row. But I should not be able to drag a column into something else than a row.

Indeed, the parent plugin cannot rely on the existence or order of child plugins. Any management code must be flexible and say recreate a missing child if needed, or might want to merge if there are several children, say, for a single slot. Disabling dragging alone would not solve this issue: There is also pasting or adding plugins.

fsbraun:
@macolo I have now implemented drag-disabling in principle (CSS and JS side). Shall we go for another plugin option that disables drag for child plugins? It would also need to disable add and paste.

cms/static/cms/js/modules/cms.plugins.js
```diff
@@ -419,8 +423,14 @@ var Plugin = new Class({
                 }
                 var name = that.options.plugin_name;
                 var id = that.options.plugin_id;
+                var disabled = $(e.currentTarget).hasClass('.cms-edit-disabled');  // No tooltip for disabled plugins
```
fsbraun, line 426:
```suggestion
                var disabled = $(e.currentTarget).hasClass('cms-edit-disabled');  // No tooltip for disabled plugins
```

Issue #8044, opened 2024-10-28
[BUG] Editing translations of pages using the language tabs inside PageContent creates an inconsistent state 


filipweidemann:
## Description

Consider the following scenario: we got a multi-language site using the languages English (en) and German (de). 

We currently got a page which only exists in the German (de) version. Therefore, we only have a published `PageContent` object for this page in German.

If we change the language switch inside the `PageContent` admin list view (the one on top, just beside the "Main Navigation" label) to `English (en)`, we can see the title of the German Page with a "(de)" behind it. We can then click on the grey circle to create the english version of the page. This all works flawlessly.

However, suppose we have a page which is published in both versions. We now click on the page content settings on the German version and then use the language tab on top of the form to switch the language to English.
This results in an empty form without title, slug or anything that the *already existing english version of this page* provides. If we continue to fill out the form and save it, we can create an inconsistent state inside the DB because the `PageUrl` objects now return 2 results for the english page version. 

If we look at the `get_urls()` method of this page, we can see that the English version now has 2 PageUrl objects:

`<QuerySet [<PageUrl: english-testpage (en)>, <PageUrl: deutsche-testseite-translated (de)>, <PageUrl: adskjfhgdskjfhdsakjh (en)>]>`

If we then select the language with 2 `PageUrl` objects in the `PageContent` list view and try to edit it, we get an error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/django/contrib/admin/options.py", line 688, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/django/utils/decorators.py", line 134, in _wrapper_view
    response = view_func(request, *args, **kwargs)
[...]

PR #8046, opened 2024-10-29
fix: Language tabs didn't show existing content due to caching issue

filipweidemann:
## Description

This PR resolves issue #8044 by not setting the `admin_content_cache` when calling `get_object`. Doing so resulted in partial caches being present for only the currently selected language, meaning that any use of the cache would yield an `EmptyPageContent` object instead of the actual content.

This breaks the language tabs inside the page edit form. Not setting the cache partially now triggers the `set_admin_content_cache` method and resolves this issue.

## Related resources

#8044 

## Checklist

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -740,11 +740,7 @@ class PageContentAdmin(admin.ModelAdmin):
         key is used if no field is provided. Return ``None`` if no match is
         found or the object_id fails validation.
         """
-        obj = super().get_object(request, object_id, from_field)
-
-        if obj:
-            obj.page.admin_content_cache[obj.language] = obj
-        return obj
+        return super().get_object(request, object_id, from_field)
 
     def get_admin_url(self, action, *args):
         url_name = f"{self.opts.app_label}_{self.opts.model_name}_{action}"
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -734,14 +734,6 @@ class PageContentAdmin(admin.ModelAdmin):
         # Block the admin log for change. A signal takes care of this!
         return
 
-    def get_object(self, request, object_id, from_field=None):
-        """
-        Return an instance matching the field and value provided, the primary
-        key is used if no field is provided. Return ``None`` if no match is
-        found or the object_id fails validation.
-        """
-        return super().get_object(request, object_id, from_field)
-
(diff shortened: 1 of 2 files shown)


cms/admin/pageadmin.py
```diff
@@ -740,11 +740,7 @@ def get_object(self, request, object_id, from_field=None):
         key is used if no field is provided. Return ``None`` if no match is
         found or the object_id fails validation.
         """
-        obj = super().get_object(request, object_id, from_field)
-
-        if obj:
-            obj.page.admin_content_cache[obj.language] = obj
-        return obj
+        return super().get_object(request, object_id, from_field)
```
fsbraun, line 743: I think we can delete the whole method; all it does is calling `super()`.

fsbraun (changes requested):
I've added a second change to improve the UX on the language tabs (see here for the reason: https://github.com/django-cms/django-cms/issues/8044#issuecomment-2442531485

Also, I see no performance issue.

cms/admin/pageadmin.py
filipweidemann: Hehe, absolutely. Hadn't had my coffee before removing the lines. Going to change it real quick...

filipweidemann: @fsbraun done :)

fsbraun:
@filipweidemann This looks good to me. Do you have an idea on how we can add a test that ensures this will not break again in the future?

filipweidemann:
> @filipweidemann This looks good to me. Do you have an idea on how we can add a test that ensures this will not break again in the future?

Currently I don't. But I can have another look later. 
But since this is a rather serious bug (in my opinion) I'd say that we should maybe merge this without a test being present *right now*?

I will then create another issue and assign myself to add the missing test cases.

How does that sound? :)

fsbraun:
@filipweidemann I added a class called `AdminCacheDict` that raises an error if you try setting individual values. I think we should be fine with this!

fsbraun (approved):
@filipweidemann Thank you so much for the excellent deep dive!

fsbraun:
Fixes #8044

filipweidemann:
> @filipweidemann Thank you so much for the excellent deep dive!

Sure thing, thanks for the fast merge. I actually just read your idea about the "immutable" AdminCacheDict implementation and wanted to proactively implement this, but then saw that you already did.

Great stuff :)

PR #8049, opened 2024-10-29
chore: Added tests for GetAdminUrlForLanguage template tag

filipweidemann:
## Description
This PR catches up on #8044 and adds two new tests targeting the regressed template tag `GetAdminUrlForLanguage`.
I think this is worth it. Please let me know what you think.

## Related resources
#8044 

## Checklist

### DIFF
--- a/cms/tests/test_templatetags.py
+++ b/cms/tests/test_templatetags.py
@@ -66,7 +66,7 @@ class TemplatetagTests(CMSTestCase):
 
         with force_language('fr'):
             url = GetAdminUrlForLanguage.get_value(None, context={'request': request}, page=page, language="fr")
-            # PageContent for the French languag does not exist, we expect a link to create a new PageContent instance
+            # PageContent for the French language does not exist, we expect a link to create a new PageContent instance
             self.assertEqual(url, f'/fr/admin/cms/pagecontent/add/?cms_page={page.pk}&language=fr')
 
     def test_admin_pagecontent_language_tab_urls(self):
--- a/cms/tests/test_templatetags.py
+++ b/cms/tests/test_templatetags.py
@@ -51,7 +51,7 @@ class TemplatetagTests(CMSTestCase):
 
         # Try to fill the cache with partial data (this should not be possible)
         page.get_content_obj(language='en')  # should not affect admin template tag
-        page.get_admin_content(language='en')
+        page.get_admin_content(language='en')  # Should fill the whole cache
 
         request = RequestFactory().get('/')
         request.current_page = page
--- a/cms/tests/test_templatetags.py
+++ b/cms/tests/test_templatetags.py
@@ -49,24 +49,6 @@ class TemplatetagTests(CMSTestCase):
         english_content = page.pagecontent_set(manager="admin_manager").first()
         german_content = create_page_content("de", "AdminURLTestPage German Content", page)
 
-        request = RequestFactory().get('/')
-        request.current_page = page
-        template = """
-            {% load cms_tags cms_admin %}
-            {% get_admin_url_for_language page_obj 'en' %}
-            {% get_admin_url_for_language page_obj 'de' %}
-            {% get_admin_url_for_language page_obj 'fr' %}
-        """
-        output = self.render_template_obj(template, {'page_obj': page}, request)
-        self.assertIn(f'/en/admin/cms/pagecontent/{english_content.pk}/change/', output)
-        self.assertIn(f'/en/admin/cms/pagecontent/{german_content.pk}/change/', output)
-        self.assertIn(f'/en/admin/cms/pagecontent/add/?cms_page={page.pk}&language=fr', output)
-
-    def test_admin_pagecontent_language_tab_urls_accessed_page(self):
-        page = create_page('AdminURLTestPage English Content', 'nav_playground.html', 'en')
-        english_content = page.pagecontent_set(manager="admin_manager").first()
-        german_content = create_page_content("de", "AdminURLTestPage German Content", page)
-
         # Try to fill the cache with partial data (this should not be possible)
         page.get_content_obj(language='en')
         page.get_admin_content(language='en')
--- a/cms/tests/test_templatetags.py
+++ b/cms/tests/test_templatetags.py
@@ -43,37 +43,33 @@ from cms.utils.placeholder import get_placeholders
 
 
 class TemplatetagTests(CMSTestCase):
-    def test_get_admin_url_for_language(self):
-        # This creates a new page and the corresponding PageContent object for lang 'en'
-        # After that we create a German PageContent object
-        page = create_page("AdminURLTestPage English Content", "nav_playground.html", "en")
-        english_content = page.pagecontent_set(manager="admin_manager").first()
-        german_content = create_page_content("de", "AdminURLTestPage German Content", page)
-
-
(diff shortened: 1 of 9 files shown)


cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,48 @@
...
+        # German PageContent will have pk=2
+        german_content = create_page_content("de", "AdminURLTestPage German Content", page)
+
+
+        request = RequestFactory().get('/')
+        request.current_page = page
+        with force_language('de'):
+            url = GetAdminUrlForLanguage.get_value(None, context={'request': request}, page=page, language="de")
+            # We expect that getting the German URL will return the German page content (change link, not a create link)
+            self.assertEqual(url, '/de/admin/cms/pagecontent/2/change/')
```
fsbraun, line 59:
```suggestion
            self.assertEqual(url, f'/de/admin/cms/pagecontent/{german_content.pk}/change/')
```
It's unsafe to make assumptions on pk's in testing environments.

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,48 @@
...
+
+        with force_language('en'):
+            url = GetAdminUrlForLanguage.get_value(None, context={'request': request}, page=page, language="en")
+            # Same goes for the English URL
+            self.assertEqual(url, '/en/admin/cms/pagecontent/1/change/')
```
fsbraun, line 64:
```suggestion
            self.assertEqual(url, f'/en/admin/cms/pagecontent/{english_content.pk}/change/')
```

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,48 @@
 
 
 class TemplatetagTests(CMSTestCase):
+    def test_get_admin_url_for_language(self):
+        # This creates a new page and the corresponding PageContent object for lang 'en', both with pk=1
+        page = create_page("AdminURLTestPage English Content", "nav_playground.html", "en")
+
+        # German PageContent will have pk=2
```
fsbraun, line 50:
```suggestion
```
It will not necessarily.

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,48 @@
 
 
 class TemplatetagTests(CMSTestCase):
+    def test_get_admin_url_for_language(self):
+        # This creates a new page and the corresponding PageContent object for lang 'en', both with pk=1
+        page = create_page("AdminURLTestPage English Content", "nav_playground.html", "en")
```
fsbraun, line 48:
```suggestion
        page = create_page("AdminURLTestPage English Content", "nav_playground.html", "en")
        
        english_content = page.pagecontent_set(manager="admin_manager").first()
```

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,48 @@
...
+            self.assertEqual(url, '/en/admin/cms/pagecontent/1/change/')
+
+        with force_language('fr'):
+            url = GetAdminUrlForLanguage.get_value(None, context={'request': request}, page=page, language="fr")
+            # PageContent for the French languag does not exist, we expect a link to create a new PageContent instance
+            self.assertEqual(url, '/fr/admin/cms/pagecontent/add/?cms_page=1&language=fr')
+
+    def test_admin_pagecontent_language_tab_urls(self):
+        # Same setup as in the templatetag test above
+        page = create_page('Test', 'nav_playground.html', 'en')
```
fsbraun, line 73:
```suggestion
        page = create_page('Test', 'nav_playground.html', 'en')
        english_coontent = page.pagecontent_set(manager="admin_manager").first()
```

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,48 @@
...
+        request = RequestFactory().get('/')
+        request.current_page = page
+        template = """
+            {% load cms_tags cms_admin %}
+            {% get_admin_url_for_language page_obj 'en' %}
+            {% get_admin_url_for_language page_obj 'de' %}
+            {% get_admin_url_for_language page_obj 'fr' %}
+        """
+        output = self.render_template_obj(template, {'page_obj': page}, request)
+        self.assertIn('/en/admin/cms/pagecontent/1/change/', output)
```
fsbraun, line 85:
```suggestion
        self.assertIn(f'/en/admin/cms/pagecontent/{english_content.pk}/change/', output)
```

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,48 @@
...
+        self.assertIn('/en/admin/cms/pagecontent/2/change/', output)
```
fsbraun, line 86:
```suggestion
        self.assertIn(f'/en/admin/cms/pagecontent/{german_content.pk}/change/', output)
```

fsbraun (commented):
Yes, this is a good idea! 

However, you'll need to avoid assumptions on the primary keys.

filipweidemann:
@fsbraun Yup, that's right. I added your suggestions in one commit & fixed a spelling issue. Should be fine now! Do you have any other suggestions?

fsbraun:
@filipweidemann What exactly is the difference between the two tests? Do we need them both?

I can imagine two tests (based on either of yours).
1. In the test setup call `page.get_content_obj("en")` before calling the template tag.
2. In the test setup call `page.get_admin_content("en")` before calling the template tag.

With a partially filled cache by the two calls, the tests would fail.

filipweidemann:
@fsbraun Yeah wow, I actually forgot the third case, you're right. Was thinking about this before implementing it. I'll add it as well!

About the current ones: I can probably remove the first one and only stick with the "faked" template one. My idea was that we also check if it can be forced to craft the links according to our language but let's be real, it actually is not doing anything here because it uses Django's `reverse` function. So what I am actually testing is Django and yeah, this is kind of unnecessary.

I'll remove the first one and replace it with the one you suggested because this one is actually missing right now!

filipweidemann:
@fsbraun okay so I deleted the first one and added the method calls before rendering the template.

I also verified that this catches the regression by temporarily adding `del self.admin_content_cache["de"]` inside the `set_admin_content_cache`. The test immediately failed.

This is cool. Anything else?

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,28 @@
...
 
 class TemplatetagTests(CMSTestCase):
+    def test_admin_pagecontent_language_tab_urls(self):
+        # Same setup as in the templatetag test above
+        page = create_page('AdminURLTestPage English Content', 'nav_playground.html', 'en')
+        english_content = page.pagecontent_set(manager="admin_manager").first()
+        german_content = create_page_content("de", "AdminURLTestPage German Content", page)
+
+        # Try to fill the cache with partial data (this should not be possible)
+        page.get_content_obj(language='en')
```
fsbraun, line 53:
```suggestion
        page.get_content_obj(language='en')  # should not affect admin template tag
```

cms/tests/test_templatetags.py
```diff
@@ -43,6 +43,28 @@
...
+        page.get_admin_content(language='en')
```
fsbraun, line 54:
```suggestion
        page.get_admin_content(language='en')  # Should fill the whole cache
```

fsbraun (approved):
Nice! Looks good to me!! Thank you so much!

PR #8062, opened 2024-11-05
feat: Add `FrontendEditableAdminMixin` endpoint to plugins

fsbraun:
## Description

django CMS provides editing endpoints and frontend support for general models through `FrontendEditableAdminMixin`. Specifically, it supports editing a subset of fields only.

This PR extends this feature to plugin models which have no model admin registered for themselves, but are managed by the `PlaceholderAdmin` class. Literally, a few lines of code allow us to have partial editing endpoints for plugins. These can be used for inline editing, for example.

### This PR
* Splits the `FrontendEditableAdminMixin` into two classes, a base class (`BaseEditableAdminMixin`) which provides the endpoint, and a specific class (`FrontendEditableAdmin`) that provides the interface to the model admin.
* The `BaseEditableAdminMixin` is added to the `PlaceholderAdmin` along with the appropriate methods to 
  * retrieve the object being edited
  * find its admin instance (which is its plugin class)
* Tweaks some tricks in the existing `FrontendEditableAdminMixin`:
  * Replaces `get_plugin_name` attributes by callables (makes no difference for the template engine, but does for the plugin class)
  * When changing a plugin, the response does not trigger a page reload but just a local update of the plugin through the data bridge
  * Finally, it provides the placeholder endpoints for plugin models instead of the model admin endpoint (which does not exist for placeholders) in all template tags based on `{% render_model %}

### Usage
With this PR, partial edit can be included by replacing `{{ instance.field }}` in a plugin's template by `{% render_model instance "field" "field,other_field" %}`. Upon double-click this will open a modal with two input fields: `field` and `other_fields`.

Especially interesting is the use of `{% render_model instance "field" "field" %}` which will render the field and allow editing just itself. djangocms-text will take this to inline editing if it is active. See this short demo here on an unstyled blog title field and content fields (**not plugins**):

inlnie-fields

Any field type can be edited inline if there is an apps (such as djangocms-text) supporting it for that field type. djangocms-text supports `HTMLField` and `CharField`.

### Docs and tests

... are here now.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -53,7 +53,6 @@ from cms.cache.permissions import clear_permission_cache
 from cms.constants import MODAL_HTML_REDIRECT
 from cms.models import (
     CMSPlugin,
-    EmptyPageContent,
     GlobalPagePermission,
     Page,
     PageContent,
@@ -169,6 +168,7 @@ class PageAdmin(admin.ModelAdmin):
         """Get the admin urls
         """
         info = f"{self.model._meta.app_label}_{self.model._meta.model_name}"
+
         def pat(regex, fn):
             return re_path(regex, self.admin_site.admin_view(fn), name=f'{info}_{fn.__name__}')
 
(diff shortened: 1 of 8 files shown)


vinitkumar (approved):
Looks good to me.

Issue #7948, opened 2024-07-04
[BUG] django-cms and django-modeltranslation 0.19.3+


benzkji:
## Description

Using these two together causes `allow_children` and `child_classes` to be ignored, and `disable_child_plugins` is always set in the structure editor. And maybe more.

## Steps to reproduce

Install django-cms 3.11.X, and django-modeltranslation 0.19.3+. Try the structure editor, plugins can be placed within others - only plugins that have a `child_classes = ["APlugin"]` work correctly. Plugins with a parent are all locked, and cnnot be moved outside the parent.

## Expected behaviour

As before, structure limits respected, no locked plugins.

## Actual behaviour

I tracked the issue to this template: https://github.com/django-cms/django-cms/blob/3.11.6/cms/templates/cms/toolbar/dragitem.html#L3

Using `{{ plugin.get_plugin_class.allow_children|pprint }}` shows `attribute error: XYPlugin has no attribute admin_site`. Not using pprint outputs the model name, i.e. always `True`. YAY. After looking around which admin could be adding some kind of getter or whatever, I found modeltranslation - 0.19.2 and everything is ok. 0.19.3 adds some typing stuff that I don't really understand, and monkeypatches Djangos ModelAdmin - as far as I get it.

Here it is, probably/maybe around the monkeypatch?!: 
https://github.com/deschler/django-modeltranslation/compare/v0.19.2...v0.19.3#diff-70244e2051407abc833c4b5691c54704040cb171edc9a46d71260332fe87ec4bR26

I don't know admin internals that well, but it seems that somewhere an admin_site is required? But where and how, no clue yet.

## Additional information (CMS/Python/Django versions)

Python 3.11, django-cms 3.11.X and django-modeltranslation 0.19.3+ to reproduce.

## Do you want to help fix this issue?

* [x] Yes, I want to help fix this issue and I will join the channel #pr-reviews on [the Discord Server](https://discord-pr-review-channel.django-cms.org) to confirm with the community that a PR is welcome.
* [ ] No, I only want to report the issue.

fsbraun:
@benzkji Sounds like an issue with django-modeltranslation which apparently clouds the class properties that steer the plugin behavior. 

django CMS works fine with [django-parler](https://github.com/django-parler/django-parler) (which offers the additional benefit that adding languages does not create migrations).

Or pin to django-modeltranslation < 0.19.3?

benzkji:
[...]

PR #8071, opened 2024-11-07
fix: Ensure plugin class properties are available to the Django template engine

fsbraun:
## Description

This fix prevents django CMS failing to work if third-party packages inject type hinting into Django (such as django-model-translation). A bug in Django's template engine for Django < 5.2 then hides plugin class properties like `allow_children` from the django CMS frontend.

This PR makes sure they keep being visible to the frontend by actively removing any `__class_get_item__` method that might have been injected into Django's `ModelAdmin`, the parent class of `CMSPlugin`. 

Fixes https://github.com/django-cms/django-cms/issues/7948 

## Related resources

* #7948
* #...

## Checklist


cms/plugin_base.py
```diff
@@ -257,6 +254,15 @@ def _get_render_template(self, context, instance, placeholder):
             raise ValidationError("plugin has no render_template: %s" % self.__class__)
         return template
 
+    if DJANGO_5_1:
```
vinitkumar, line 257: Do you think we should also check for Django 5.2 as well as it will have the same issue and if we account for that here, we won't have to add a fix again when that comes out?

fsbraun: @vinitkumar The bug in the template engine is fixed in django 5.2: https://github.com/django/django/pull/18547, therefore no need to keep it.

vinitkumar:
@fsbraun Then just the conflicts needs to be resolved and then we are good to go! 🚀

PR #8145, opened 2025-02-16
fix: ensures page content translations are created with the same temp…

amandasavluchinske:
## Description

Fixes the issue reported in https://github.com/django-cms/django-cms/issues/8081, in which subsequent created translations did not use the project's default language's template by default, and instead had the "Inherit the template of the nearest ancestor" setting.

## Checklist

* [X] I have opened this pull request against ``develop-4``

### DIFF
--- a/cms/admin/forms.py
+++ b/cms/admin/forms.py
@@ -81,11 +81,7 @@ def get_page_changed_by_filter_choices():
     # This is not site-aware
     # Been like this forever
     # Would be nice for it to filter out by site
-    values = (
-        Page.objects.distinct()
-        .order_by("changed_by")
-        .values_list("changed_by", flat=True)
-    )
+    values = Page.objects.distinct().order_by("changed_by").values_list("changed_by", flat=True)
 
     yield ("", _("All"))
 
@@ -118,9 +114,7 @@ def save_permissions(data, obj):
         for key in ("add", "change", "delete"):
             # add permission `key` for model `model`
             codename = get_permission_codename(key, model._meta)
-            permission = Permission.objects.get(
-                content_type=content_type, codename=codename
-            )
+            permission = Permission.objects.get(content_type=content_type, codename=codename)
             field = f"can_{key}_{name}"
 
             if data.get(field):
@@ -282,9 +276,7 @@ class AddPageForm(BasePageContentForm):
         if root_page:
             # Set the choicefield's choices to the various page_types
             descendants = root_page.get_descendant_pages().filter(is_page_type=True)
-            titles = PageContent.objects.filter(
-                page__in=descendants, language=self._language
-            )
+            titles = PageContent.objects.filter(page__in=descendants, language=self._language)
             choices = [("", "---------")]
             choices.extend((title.page_id, title.title) for title in titles)
             source_field.choices = choices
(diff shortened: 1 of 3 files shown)


amandasavluchinske:
By the way, my pre-commit hooks ran in some files and changed formatting. However, the Sourcery bot is pointing to the correct spots that need to be reviewed.

cms/admin/forms.py
```diff
@@ -389,10 +385,26 @@ def save(self, *args, **kwargs):
...
 
-        translation = self.create_translation(new_page)
+        # If you have djangocms-versioning installed, this will get the latest version of the page content, and ignore drafts.
+        # Get the original page content template if available
+        og_page_content_template = None
+        if new_page:
+            try:
+                main_language = new_page.get_languages()[0]
+                og_page_content = (
+                    PageContent.objects.filter(page_id=new_page.pk, language=main_language).only("template").first()
```
fsbraun, line 398:
```suggestion
                    new_page.get_admin_content(language=main_language)
```

There's an implicit design decision here: 
* You'll get the page's latest (public) content object. The preferred way of doing this would be `new_page.get_content_object(language=main_language)`
* Alternatively, since we're operating in the admin, you could get the latest admin object (see suggestion above). This will also return an admin object if there (with versioning) there is no public page content

fsbraun: Also, what's `og` supposed to stand for?

docs/how_to/04-templates.rst
```diff
@@ -228,3 +228,23 @@ Example: application template
...
 template can dynamically import it.
+
+Template Inheritance in Page Content Translations
+=================================================
+
+Default Behavior (without versioning)
+-------------------------------------
+
+In django CMS (version 4 and above), when creating page content translations, all content is immediately published. When creating a translation of an existing page, the template from the original page content is copied to the new translation.
+This behavior can be seen in the `create_translation` method. The template is explicitly copied from the original page content to ensure consistent layout across languages.
```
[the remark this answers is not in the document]
fsbraun: This will be part of version 4.2 only, so it should read "version 4.2 and above".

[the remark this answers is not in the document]
fsbraun: Can you try breaking the lines at just short of 120 characters?

fsbraun (changes requested):
Hey @amandasavluchinske !

Great work! Thank you. I have made a suggestion which would IMHO better reflect the canoncial way of accessing page content objects within the admin (which is where adding translations happens). Using those page-specific functions ensures that translations are not unnecessarily fetched from the database.

Would you mind taking a look?

cms/admin/forms.py
amandasavluchinske: Nice suggestion! I decided to go with your alternative and use `get_admin_content`. However, there's an important behavior decision to consider, and I’d love your input on it:

By using this function, the translation will always fetch its template from the draft version of the main language page content (if one exists), rather than the published version. If no draft is available, it will fall back to the public version.

IMO that logic makes sense, I just need to adjust the docs to reflect the behavior. What do you think about this?

> Also, what's og supposed to stand for?
og was supposed to stand for original, but I see it's not really self-explanatory even when using the word "original", so I changed it to "main_language" instead :) Thanks for pointing it out!

fsbraun: Using the latest draft is the default behaviour throughout the admin. I second your decision.

vinitkumar:
@fsbraun @amandasavluchinske Sorry, I missed the tag earlier. I just reviewed the code, and it looks good overall—just a small nitpick.

@amandasavluchinske It seems like your editor is running black or another formatter on save. This adds unnecessary noise to the diff, making it harder to see the actual changes. Could you disable it and commit from a clean file? Let me know if that works for you. Thanks!

vinitkumar (approved):
Looks good to me. Let's merge it.

PR #8163, opened 2025-02-26
feat: Adds backwards migration of the `Page`/`TreeNode` model merge

fsbraun:
## Description

This PR adds reverse migration code to allow going back to django CMS 4.1.x after the `Page`/`TreeNode` models have been merged in migrations 0037 to 0039.

I've gone back and forth on a test system several times for testing.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/cms/migrations/0038_alter_page_site.py
+++ b/cms/migrations/0038_alter_page_site.py
@@ -7,8 +7,8 @@ import cms.models.settingmodels
 
 
 def separate_page_treenode(apps, schema_editor):
-    Page = apps.get_model('cms', 'page')
-    TreeNode = apps.get_model('cms', 'treenode')
+    Page = apps.get_model('cms', 'Page')
+    TreeNode = apps.get_model('cms', 'TreeNode')
     db_alias = schema_editor.connection.alias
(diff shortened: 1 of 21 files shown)


cms/migrations/0038_alter_page_site.py
```diff
@@ -6,6 +6,21 @@
 import cms.models.settingmodels
 
 
+def separate_page_treenode(apps, schema_editor):
```
vinitkumar, line 9: I think we should wrap this inside a transaction and try catch as well, so in-case an exception happens we are able to see wha went wrong and with transaction, it only gets committed if all goes fine. What do  you think?

fsbraun: Hold on, I just realize, Django does that automatically:; https://docs.djangoproject.com/en/5.1/topics/migrations/#transactions

vinitkumar: Ah, I didn't knew it did that automatically. I would always do it like this in the migrations to prevent inconsistencies in data if something crash during migration.

Issue #5517, opened 2016-07-08
Refactor PageUser and PageUserGroup


czpython:

gignosko:
Is there an update on this? We're getting the same issue mentioned in 4225 and would love some kind of work around.

mkoistinen:
@czpython @yakky Is there any plans for this issue at this time? Is this field `PageUser.created_by` actually used by anything? (AKAICT, no) Can we deprecate this field starting the next release and plan to remove it in a near-term subsequent release?

creyD:
@gignosko Could you create another PR for this? @mkoistinen As far as I can see it is used for some of the permissions. So I am not quite sure about entire removal, but we could definitely rename it or opt for the One-To-One-Field instead of the PageUser inheriting from the default user.

creyD:
Further discussed in #7225 7225

PR #8167, opened 2025-02-28
fix: deleting users cascaded to deleting PageUser or PageUserGroup

### LINKED ISSUE #4225: Similar field names with PageUser and my custom user model
I have a similar problem as @jrief describes in #3973.
We use django 1.8.2 with a custom user Model and as i tried to install django-cms 3.1.1 i got the following error:

    django.core.exceptions.FieldError: Local field 'created_by' in class 'PageUser' clashes with field of similar name from base class 'Contact'

Unfortunately my custom User model already has a `created_by` Field. And [django does not allow to override any Field attribute from the parent Model](https://docs.djangoproject.com/en/dev/topics/db/models/#field-name-hiding-is-not-permitted).

Since the PageUser model does not add any logic rather than the `created_by` Field, the best solution for me would be to swap the PageUser model like django does with `auth.User`. Are there any Plans to support that functionality? If not, i could invest some time and send a PR.

Or anyone else ran into the same problem and can give a hint?

### LINKED ISSUE #7225: The fields `created_by` of PageUser, PageGroup and `user`/`group` of PagePermission can cause unwanted user/group deletion
## Description

As all of these fields are defined as `on_delete=models.CASCADE`. That means, if I delete User A, that has previously created a `PageUser` B, the user B will be deleted as well. I never was really sure what these are anyway... see also #5517 .

## Steps to reproduce

Create User A. Log in with A, create PageUser B. Delete User A. User B is gone as well.

## Expected behaviour

User B should not be deleted. Instead, one could show (deleted User), and define the fields as `on_delete=models.SET_NULL`. Or remove the fields completly.

## Actual behaviour

User B is gone/deleted.   

Also, discovered that normal new Users that are created trigger the creation of a `PageUser`, if `CMS_PERMISSION` is true. This can cause confusion, and makes it almost impossible to delete useres on such a system.

for discussion since 2016 #5517 

fsbraun:
## Description

This PR ensures that deleting a `User` object does not automatically delete all `PageUser` objects that were created by the 
deleted user. This is achieved by using `models.SET_NULL` on delete and allowing for null objects for the `created_by` field of `PageUser` and `PageUserGroup`.

Fixes #7225, does not fix #4225. 

## Related resources

* #7225 
* #5517 
* #4225 

## Checklist

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where deleting a User would also delete associated PageUser and PageUserGroup objects.


vinitkumar:
@fsbraun The code looks good. Is it good to be merged?

vinitkumar (approved):

PR #8168, opened 2025-03-02
ci: Update publishing to use trusted publishing

marksweb:
This will update pypi releases to use trusted publishing.

Then I'll stop getting emails that we're still using a token!

## Summary by Sourcery

CI:
- Configure PyPI publishing workflows to use trusted publishing.


marksweb:
>     * Consider adding a step to verify the package content before publishing.

How can we add a step to verify the package content before publishing?

fsbraun (approved):

PR #8172, opened 2025-03-06
feat: CONTINUE cms.views - set request language according to request.LANGUAGE_CODE

amandasavluchinske:
## Description

Resumes the work in https://github.com/django-cms/django-cms/pull/7610. Adds tests and removes cache modifications on `placeholder.py`.

## Manual tests checklist

With two languages (example: English and Deutsch), do the following:

- [ ] Check that using the `i18n_patterns` on the `urls.py` file works as expected.
- [ ] Check that default languages work as expected.
- [ ] With the middleware:

```
from django.utils.deprecation import MiddlewareMixin

class LanguageSettingMiddleware(MiddlewareMixin):
    """Middleware to set language based on subdomain prefix."""
    
    def process_request(self, request):            
        # Get the hostname from the request
        hostname = request.get_host().split(':')[0]  # Remove port if present
        
        # Extract language code from the subdomain (first part of the hostname)
        try:
            language_code = hostname.split('.')[0]
            
            if language_code in ['en', 'de']:
                # Set the language code on the request
                request.LANGUAGE_CODE = language_code
        except IndexError:
            pass

```
and the following URLs added to your /etc/hosts file:

- de.example.com
- en.example.com

Check that de.example.com redirects users to the Deustch version, while en.example.com redirects users to the English version.

## Checklist

* [X] I have opened this pull request against ``main``

## Summary by Sourcery

Set request language based on subdomain using middleware, defaulting to German if no valid language is found. Enhance language handling in page cache and menu renderer to use request.LANGUAGE_CODE or default site language. Add tests for language code handling and page cache behavior with language fallbacks.

New Features:
- Introduce middleware to set request language based on subdomain, defaulting to German if no valid language subdomain is found.

Enhancements:
- Improve language handling in page cache by using request.LANGUAGE_CODE if available, and falling back to site's default language if not.
- Enhance menu renderer to correctly handle language code determination using request.LANGUAGE_CODE or default site language.

Tests:
- Add tests for language code handling in menu renderer and page cache, ensuring correct behavior with different languages and subdomains.
- Add tests to verify that the page cache works correctly with language fallbacks and different language versions.


docs/how_to/05-caching.rst
```diff
@@ -55,7 +55,17 @@ Have a look at the following settings to enable/disable various caching behaviou
 - :setting:`CMS_PLACEHOLDER_CACHE`
 - :setting:`CMS_PLUGIN_CACHE`
 
+Language Cache
+=============
```
fsbraun, line 59:
```suggestion
==============
```

cms/tests/test_page.py
```diff
@@ -743,204 +733,275 @@ def test_plugin_loading_queries(self):
...
+
+    def test_page_cache_language_handling(self):
+        """
+        Test that the page cache properly handles different languages
+        """
+        page = self.create_homepage("page", "nav_playground.html", "en")
+        create_page_content("fr", "french home", page)
+
+        # Get the page in English
+        en_url = page.get_absolute_url(language="en")
```
fsbraun, line 894: Are `en_url` and `fr_url` identical? This test should give different cached values even for identical urls (say, root) but different languages.

fsbraun (commented):
Are there any changes to `cms.cache.placeholder`?

amandasavluchinske:
> Are there any changes to cms.cache.placeholder?

@fsbraun Thanks for reviewing! No, just pre-commit changes.

cms/tests/test_page.py
amandasavluchinske: The base URL is identical, but en_url has /en/ at the end and fr_url has /fr/. Is that what we expect?

fsbraun (approved):
Very nice!  Thank you, @amandasavluchinske !

Only one last remark: You custom language middleware can fully replace Django's (and I assume it normally would.) That means the comment on the order of middleware in the docs is not really necessary.

amandasavluchinske:
> Very nice! Thank you, @amandasavluchinske !
> 
> Only one last remark: You custom language middleware can fully replace Django's (and I assume it normally would.) That means the comment on the order of middleware in the docs is not really necessary.

Thank you, @fsbraun ! Adjusted :)

PR #8173, opened 2025-03-06
fix: add a check for parent/child relationship

### LINKED ISSUE #6849: On a CMS plugin defining an incorrect parent_classes list doesn't raise an error or warning and has side effects
## Description

Here's a code snippet that has that problem (the parent class should have been `PersonListPlugin`) - https://gitlab.com/effective-altruism/ea-cms-template/-/blob/fde92411000c59c5abb97b4bad1d63e3805a1ed8/backend/plugins/default/person_list/cms_plugins.py

## Steps to reproduce

Add an incorrect parent_classes value, it still allows to add this plugin to a wrong parent.

Add it as a child, try to drag & drop it or copy-past it.

image

## Additional information

- django cms 3.7.2
- django 2.2.12
- python 3.6

earthcomfy:
## Description

When defining incorrect parent or child classes, raise a warning to avoid side effects. Closes #6849 
I have added a check in the `validate_templates` method, which gets triggered when the cms app is ready in the `setup` method. We need to validate parent/child relationships after all plugins are discovered but before they're used.

## Checklist

## Summary by Sourcery

Validate parent and child plugin classes during setup to prevent incorrect configurations and potential side effects. Refactor test cases for consistency and readability, and add tests to ensure proper validation of plugin class relationships.

Bug Fixes:
- Add validation for parent and child plugin classes during setup to prevent incorrect configurations and potential side effects.

Enhancements:
- Refactor test cases to use consistent string formatting and improve readability.

Tests:
- Add tests to validate the presence of parent and child plugin classes, ensuring that non-existent classes raise an ImproperlyConfigured exception.

### DIFF
--- a/.github/workflows/test_startcmsproject.yml
+++ b/.github/workflows/test_startcmsproject.yml
@@ -13,12 +13,12 @@ jobs:
       fail-fast: false
       matrix:
         django-version: [
-          '4.2', '5.0', '5.1'
+          '4.2', '5.0', '5.1',
         ]
         python-version: ['3.11']
         requirements-file: ['requirements_base.txt']
         os: [
-          ubuntu-20.04,
+          ubuntu-latest,
         ]
 
     steps:
--- a/cms/plugin_pool.py
+++ b/cms/plugin_pool.py
@@ -18,25 +18,22 @@ from cms.utils.helpers import normalize_name
 
 
 class PluginPool:
-
     def __init__(self):
         self.plugins = {}
         self.discovered = False
         self.global_restrictions_cache = {
             None: {},
-            **{key: {} for key in get_cms_setting('PLACEHOLDER_CONF').keys()},
+            **{key: {} for key in get_cms_setting("PLACEHOLDER_CONF").keys()},
         }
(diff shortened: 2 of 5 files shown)


fsbraun:
@earthcomfy Looks nice! I like the tests!

One thought: the test with the django cms template project fails (it's the project template used by the djangocms command) due to djangocms-frontend's use of the `""` empty string plugin as a `parent_classes` to disallow adding to any plugin. I can imagine existing setups where this is done for plugins that are controlled by the developer programatically. 

To not break existing setups I can imagine
* to issue a warning instead of raising an exception
* document how the "non-existing" plugin needs to be named (it seems you have considered `"0"` to be non-existing) - this would still be a breaking change, though
* Move the check to the `./manage.py cms check` command.

My preferences seem to slightly go to the last option. What do you think?

earthcomfy:
Ah, I see. I thought adding the `"0"` would be enough. But yes, the third option seems the most ideal @fsbraun  I'm Going to use that.

fsbraun (approved):
@earthcomfy Elegantly done! Very nice! Thank you.

Issue #4939, opened 2016-01-20
It should be possible to restrict a plugin to be a descendant of another


mkoistinen:
This is different than simple parent-restrictions in that there may be intermediate plugins. Example:

Form Plugin
- Style Plugin
  - Field Plugin

Field plugins should be restricted to be a descendant of Form, even if its not the direct parent.

fsbraun:
I agree: https://github.com/django-cms/djangocms-form-builder/issues/26

fsbraun:
With #8181 merged, django CMS allows to implement this functionality to a reasonable extend.

PR #8181, opened 2025-03-16
fix: child restrictions cached despite children have cache_parent_classes set to False

### LINKED ISSUE #8180: bug: Cannot add form fields indirectly to a form (e.g., below row/col)
## Issue

djangocms-form-builder customizes plugin restrictions such that form fields only can be added as indirect children of a form plugin, e.g.,

* Form
  * Textarea

or

* Form
  * Row
    * Col
      * Textarea    
    * Col
      * Textarea    

To this end it sets the `cache_parent_classes` property of form fields plugins to `False` (since they need to computed dynamically investigating the plugin parents).

django CMS caches the allowed children of the Col plugin in the second example. Hence if a Col plugin appears before outside a form, the cached classes will not include any form fields.

**To Reproduce**

1. Creat a Row with columns
1. Add a "Form" plugin to start a form after the row 
2. Now, you should be able to add form field plugins inside this form plugin as immediate children ("directly")
3. Also, you shold be able to add other plugins, such as rows and columns (by django CMS frontend, e.g.), and form field plugins as their children ("indirectly")

## Solution

django CMS must not cache allowed child plugin classes if one of the children has set `cache_parent_classes` to `False`

fsbraun:
## Description

This PR fixes #8180. 

This seems to be a long-standing flaw in caching plugin restrictions. The fix is necessary for django CMS 5, since plugin restrictions are now cached globally (i.e. not on a per request-basis) and a wrong cache is easily noticeable. It should also  be ported back to django CMS 4.1 and 3.11 imho.

## Related resources

* #8180 
* #4939 

## Checklist

## Summary by Sourcery

Bug Fixes:
- Fixes an issue where child restrictions were being cached even when children had `cache_parent_classes` set to `False`.

## Summary by Sourcery

Fixes an issue where child restrictions were being cached even when children had `cache_parent_classes` set to `False`. This ensures that plugin restrictions are correctly evaluated, especially in environments where caching is more aggressive.

Bug Fixes:
- Fixes an issue where child restrictions were being cached even when children had `cache_parent_classes` set to `False`.
- Ensures that plugin restrictions are correctly evaluated, especially in environments where caching is more aggressive.
- Fixes a long-standing flaw in caching plugin restrictions.


vinitkumar (approved):

PR #8185, opened 2025-03-23
fix: Slug uniqueness not checked when moving page

amandasavluchinske:
## Description

Closes https://github.com/django-cms/django-cms/issues/6840 

Implements slug checking when a page is moved so that there are never two pages at the same level with the same slug.

## Checklist

* [X] I have opened this pull request against ``main``

## Summary by Sourcery

Bug Fixes:
- Ensure slug uniqueness when moving pages to prevent duplicate slugs at the same level.


cms/admin/pageadmin.py
```diff
@@ -561,16 +555,14 @@ def move_page(self, request, page_id, extra_context=None):
         form = self.move_form(request.POST or None, page=page, site=site)
 
         if not form.is_valid():
-            return jsonify_request(HttpResponseBadRequest("error"))
+            return jsonify_request(HttpResponseBadRequest(str(form.errors.get("__all__", _("error")))))
```
amandasavluchinske, line 558: The only line in this file I actually changed.

cms/tests/test_forms.py
```diff
@@ -211,38 +208,150 @@ def get_configs(self):
...
-                        '\\u003B)\\u003B\\u0026lt\\u003B/script\\u0026gt'
-                        '\\u003B' in output)
+        self.assertTrue(
+            "\\u0026lt\\u003Bscript\\u0026gt\\u003Balert("
+            "\\u0026quot\\u003Bbad\\u002Dstuff\\u0026quot"
+            "\\u003B)\\u003B\\u0026lt\\u003B/script\\u0026gt"
+            "\\u003B" in output
+        )
+
+    def test_move_page_form(self):
```
amandasavluchinske, line 242: My changes start in this line.

amandasavluchinske:
Thanks for the suggestions @fsbraun ! The changes have been pushed :)

jrief:
In our private version of django-CMS we actually use these constraints to prevent inconsistencies while moving pages, etc.:

https://github.com/jrief/django-cms/blob/a54ec3d02d80064b64e6d22da45512ad13211f5d/cms/models/pagemodel.py#L1092

cms/models/pagemodel.py
```diff
@@ -1095,19 +1128,24 @@ def get_absolute_url(self, language=None, fallback=True):
...
                 return None
 
-    def get_path_for_base(self, base_path=''):
-        old_base, sep, slug = self.path.rpartition('/')
-        return f'{base_path}/{slug}' if base_path else slug
+    def get_path_for_base(self, base_path=""):
+        old_base, sep, slug = self.path.rpartition("/")
+        return f"{base_path}/{slug}" if base_path else slug
 
+    def save(self, *args, **kwargs):
```
fsbraun, line 1141: Why did you add this? Did we have instances where the `site` property was not set?

fsbraun (commented):
@amandasavluchinske I have not been able to find out what makes the tests fail on mysql. I'm still on it. In the meantime I have two questions.

amandasavluchinske:
> I have not been able to find out what makes the tests fail on mysql.

Thanks @fsbraun , I'll reply to your comments as soon as I can, but in the meantime: I looked the error up, and it seems like it's related to the `condition=` in line 1117 of `pagemodel.py`, which is not supported by MySQL.

fsbraun:
That seems reasonable. But I still try to wrap my head around why we need it.

cms/models/pagemodel.py
amandasavluchinske: I denormalized the site above in line 1102, so I'm saving the site to this model from its page attribute so I can use it as a field in our `UniqueConstraint`.

cms/models/pagemodel.py
```diff
@@ -1078,13 +1099,25 @@ class PageUrl(models.Model):
...
     class Meta:
-        app_label = 'cms'
+        app_label = "cms"
         default_permissions = []
-        unique_together = ('language', 'page')
+        constraints = [
+            UniqueConstraint(
+                fields=["path", "language", "site"],
+                name="unique_together_path_language_site",
+                condition=models.Q(path__isnull=False) if hasattr(models, "Q") else None,
```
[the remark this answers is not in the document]
amandasavluchinske: Good point. I'll need to run some tests later, we may be able to get rid of this.

.github/workflows/test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.9, '3.10', '3.11', '3.12', '3.13.0-rc.2']  # latest release minus two
+        python-version: [3.9, '3.10', '3.11', '3.12', '3.13']  # Starting from earliest supported by Django LTS
```
marksweb, line 15: I'd just drop the comment. I think we're essentially supporting the same python versions as Django does for the versions of that which we support. But it's easier not to have a comment.

cms/models/pagemodel.py
```diff
@@ -1078,13 +1099,24 @@ class PageUrl(models.Model):
...
     objects = PageUrlManager()
 
     class Meta:
-        app_label = 'cms'
+        app_label = "cms"
         default_permissions = []
-        unique_together = ('language', 'page')
+        constraints = [
+            UniqueConstraint(
+                fields=["path", "language", "site"],
```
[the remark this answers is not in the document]
amandasavluchinske: I tried, but the UniqueConstraint only supports direct fields of the model :(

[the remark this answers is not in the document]
fsbraun: OK, then we need to keep this.

fsbraun (approved):
Great work! Thank you, @amandasavluchinske !

PR #8284, opened 2025-07-10
fix: GrouperAdmin Search in content model

mhsiddiqui:
## Related resources

* [Issue](https://github.com/django-cms/django-cms/issues/8277)

## Checklist

## Summary by Sourcery

Enable combined search across grouper and its related content model fields in the Django admin by extending the admin utils and adding comprehensive tests.

New Features:
- Allow GrouperAdmin to search by content model fields alongside grouper model fields
- Support Django admin search modifiers (^startswith and =exact) for both grouper and content fields

Enhancements:
- Override get_search_fields and get_search_results in admin utils to merge search results from the grouper model and its content model
- Refactor test helper createContentInstance to accept custom grouper instances and secret greetings

Tests:
- Add GrouperSearchTestCase with tests covering empty queries, model field searches, content field searches, modifier behaviors, and varied search_fields configurations

### DIFF
--- a/cms/admin/utils.py
+++ b/cms/admin/utils.py
@@ -729,8 +729,7 @@ class GrouperModelAdmin(ChangeListActionsMixin, ModelAdmin):
     def _get_content_search_result(self, request, queryset, search_term):
         """Get search results from content model"""
         content_queryset = self.content_model.admin_manager.all()
-        content_search_fields = self.get_search_fields(request)
-        if content_search_fields:
+        if self.get_search_fields(request):
             content_search_result, __ = super().get_search_results(request, content_queryset, search_term)
         else:
             content_search_result = self.content_model.admin_manager.none()
--- a/cms/admin/utils.py
+++ b/cms/admin/utils.py
@@ -699,128 +699,50 @@ class GrouperModelAdmin(ChangeListActionsMixin, ModelAdmin):
             setattr(form._content_instance, self.grouper_field_name, obj)
             form._content_instance.save()
 
-    def _search_content_model(self, search_term, search_fields):
-        """
-        Search in the related content model with admin_manager.
-        """
-        if not search_fields:
-            return []
-
-        queryset = self.content_model.admin_manager.all()
-        search_results, __ = self._get_search_result(queryset, search_term, search_fields=search_fields)
-
-        return search_results.values_list(f"{self.grouper_field_name}_id", flat=True)
-
-    def _handle_prefix_operators(self, field_name):
-        """Handle special prefix operators (^, =, @)"""
-        if field_name.startswith("^"):
-            return f'{field_name.removeprefix("^")}__istartswith', None
-        elif field_name.startswith("="):
-            return f'{field_name.removeprefix("=")}__iexact', None
-        elif field_name.startswith("@"):
-            return f'{field_name.removeprefix("@")}__search', None
-
-    def construct_search(self, queryset, field_name):
-        """Construct search query from fields in search fields"""
-        prefix_result = self._handle_prefix_operators(field_name)
-        if prefix_result:
-            return prefix_result
-        # Use field_name if it includes a lookup.
-        opts = queryset.model._meta
-        lookup_fields = field_name.split(LOOKUP_SEP)
-        # Go through the fields, following all relations.
-        prev_field = None
-        for i, path_part in enumerate(lookup_fields):
-            if path_part == "pk":
-                path_part = opts.pk.name
-            try:
-                field = opts.get_field(path_part)
-            except FieldDoesNotExist:
-                # Use valid query lookups.
-                if prev_field and prev_field.get_lookup(path_part):
-                    if path_part == "exact" and not isinstance(prev_field, (models.CharField, models.TextField)):
-                        field_name_without_exact = "__".join(lookup_fields[:i])
-                        alias = Cast(
-                            field_name_without_exact,
-                            output_field=models.CharField(),
-                        )
-                        alias_name = "_".join(lookup_fields[:i])
-                        return f"{alias_name}_str", alias
-                    else:
-                        return field_name, None
-            else:
-                prev_field = field
-                if hasattr(field, "path_infos"):
-                    # Update opts to follow the relation.
-                    opts = field.path_infos[-1].to_opts
-        # Otherwise, use the field with icontains.
-        return f"{field_name}__icontains", None
-
-    def _get_search_fields(self, request):
-        search_fields = self.get_search_fields(request)
+    def get_search_fields(self, request):
+        """Return search fields for either grouper model or content model"""
         content_search_fields = []
         grouper_search_fields = []
-        for field_name in search_fields:
+        for field_name in self.search_fields:
             if field_name.startswith(f"{self.content_related_field}__") or field_name.startswith("content__"):
                 content_search_fields.append(re.sub(r"^([@^=]?)[^_]+__", r"\1", field_name))
 
             else:
                 grouper_search_fields.append(field_name)
 
-        return grouper_search_fields, content_search_fields
+        if getattr(self, "_content_fields", False):
+            return content_search_fields
+
+        return grouper_search_fields
 
     def get_search_results(self, request, queryset, search_term):
-        """
-        Get search results from a queryset.
-        """
-        content_results = None
-        grouper_search_fields, content_search_fields = self._get_search_fields(request)
-        if search_term:
-            content_results = self._search_content_model(search_term, content_search_fields)
+        # Reset _content_fields flag
+        self._content_fields = False
+        grouper_search_result, may_have_duplicate_grouper = super().get_search_results(request, queryset, search_term)
 
-        queryset, may_have_duplicates = self._get_search_result(
-            queryset, search_term, search_fields=grouper_search_fields, included_pks=content_results
+        # Set _content_fields for get_search_fields to return content model search fields
+        self._content_fields = True
+        search_result_from_content, may_have_duplicate_content = self._get_content_search_result(
+            request, queryset, search_term
         )
-        return queryset, may_have_duplicates
 
-    def _get_search_result(self, queryset, search_term, search_fields, included_pks=None):
-        """Get search results from a queryset. using search fields"""
-        may_have_duplicates = False
-        if search_fields and search_term:
-            str_aliases = {}
-            orm_lookups = []
-            for field in search_fields:
-                lookup, str_alias = self.construct_search(queryset, str(field))
-                orm_lookups.append(lookup)
-                if str_alias:
-                    str_aliases[lookup] = str_alias
-            queryset = self._resolve_aliases(queryset, str_aliases)
+        return grouper_search_result | search_result_from_content, (
+            may_have_duplicate_grouper & may_have_duplicate_content
+        )
 
-            queryset = queryset.filter(self._build_search_query(search_term, included_pks, orm_lookups))
-            may_have_duplicates |= any(lookup_spawns_duplicates(self.opts, search_spec) for search_spec in orm_lookups)
-        return queryset, may_have_duplicates
-
-    def _resolve_aliases(self, queryset, str_aliases):
-        """Resolve aliases"""
-        if str_aliases:
-            return queryset.alias(**str_aliases)
-        return queryset
-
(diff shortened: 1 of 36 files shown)


fsbraun:
Hey @mhsiddiqui ! Thanks for this interesting PR. It is not an easy task - well done!

It is unfortunate that the Django admin does not expose the `construct_search` function inside `get_search_results`. This leads to a lot of code duplication and will be hard to maintain.

Can I make the following hacky suggestions:

* Let us try reusing the `super().get_search_results()` method. Once it should return the results for the grouper model, then for the content model. The overall result will be the union of both.
* Since `get_search_results` takes the query set as a parameter, we can pass it the content's admin_manager's `all()` qs, if we just could give it the right search fields, too.
* Here comes the hack: We override `get_search_fields` which takes the `request` argument. If that has a certain property, say, `_content_prefix`, it only returns the prefixed search fields (without the prefix for `get_search_result` to work), otherwise it only returns the unprefixed grouper search fields.

I agree that this is a bit hacky patching the `request`. What do you think?

mhsiddiqui:
> Hey @mhsiddiqui ! Thanks for this interesting PR. It is not an easy task - well done!
> 
> It is unfortunate that the Django admin does not expose the `construct_search` function inside `get_search_results`. This leads to a lot of code duplication and will be hard to maintain.
> 
> Can I make the following hacky suggestions:
> 
> * Let us try reusing the `super().get_search_results()` method. Once it should return the results for the grouper model, then for the content model. The overall result will be the union of both.
> * Since `get_search_results` takes the query set as a parameter, we can pass it the content's admin_manager's `all()` qs, if we just could give it the right search fields, too.
> * Here comes the hack: We override `get_search_fields` which takes the `request` argument. If that has a certain property, say, `_content_prefix`, it only returns the prefixed search fields (without the prefix for `get_search_result` to work), otherwise it only returns the unprefixed grouper search fields.
> 
> I agree that this is a bit hacky patching the `request`. What do you think?

I tried approach with union of both queries by

cms/admin/utils.py
```diff
@@ -696,6 +699,51 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+                grouper_search_fields.append(field_name)
+
+        if getattr(self, "_content_fields", False):
+            return content_search_fields
+
+        return grouper_search_fields
+
+    def get_search_results(self, request, queryset, search_term):
+        # Reset _content_fields flag
+        self._content_fields = False
```
fsbraun:
```suggestion
        request._content_fields = False
```
I'm not entirely sure, but to make this more thread-safe, I'd add the flag to the `request` object. The admin object might be called from a different request while the flag is set.

cms/admin/utils.py
```diff
@@ -696,6 +699,51 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+        content_search_fields = self.get_search_fields(request)
+        content_search_result, may_have_duplicate_content = self.content_model.admin_manager.none(), False
+        if content_search_fields:
+            content_search_result, may_have_duplicate_content = super().get_search_results(
+                request, content_queryset, search_term
+            )
+        search_result_from_content = queryset.filter(
+            id__in=content_search_result.values_list(f"{self.grouper_field_name}_id", flat=True)
+        )
+        return search_result_from_content, may_have_duplicate_content
```
fsbraun:
```suggestion
        return search_result_from_content, False
```
The `id__in` filter will never have duplicated content.

cms/admin/utils.py
```diff
@@ -696,6 +699,51 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+        content_search_fields = []
+        grouper_search_fields = []
+        for field_name in self.search_fields:
+            if field_name.startswith(f"{self.content_related_field}__") or field_name.startswith("content__"):
+                content_search_fields.append(re.sub(r"^([@^=]?)[^_]+__", r"\1", field_name))
+
+            else:
+                grouper_search_fields.append(field_name)
+
+        if getattr(self, "_content_fields", False):
```
fsbraun:
```suggestion
        if getattr(request, "_content_fields", False):
```
I'd feel better if the parameter would carry the flag.

cms/admin/utils.py
```diff
@@ -696,6 +699,51 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+
+        return grouper_search_result | search_result_from_content, (
+            may_have_duplicate_grouper & may_have_duplicate_content
+        )
+
+    def _get_content_search_result(self, request, queryset, search_term):
+        """Get search results from content model"""
+        content_queryset = self.content_model.admin_manager.all()
+        content_search_fields = self.get_search_fields(request)
+        content_search_result, may_have_duplicate_content = self.content_model.admin_manager.none(), False
```
fsbraun:
```suggestion
        content_search_result, may_have_duplicate_content = self.content_model.admin_manager.none(), False
```
Can we put this into an else clause?

cms/admin/utils.py
```diff
@@ -696,6 +699,51 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
             setattr(form._content_instance, self.grouper_field_name, obj)
             form._content_instance.save()
 
+    def get_search_fields(self, request):
+        """Return search fields for either grouper model or content model"""
+        content_search_fields = []
+        grouper_search_fields = []
+        for field_name in self.search_fields:
+            if field_name.startswith(f"{self.content_related_field}__") or field_name.startswith("content__"):
```
fsbraun:
```suggestion
            if field_name.startswith(CONTENT_PREFIX):
```
For syntactical consistency, I would like to see the `CONTENT_PREFIX` here. This way all admin configurations behave the same.

cms/admin/utils.py
```diff
@@ -696,6 +699,51 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+                content_search_fields.append(re.sub(r"^([@^=]?)[^_]+__", r"\1", field_name))
```
fsbraun: This can be simplified to just remove `CONTENT_PREFIX`.

fsbraun (changes requested):
Hi @mhsiddiqui ! This is looking good! I have only some minor comments (added). 

Can you check why tests fail on Django < 5.2?

fsbraun:
@mhsiddiqui Once we get this merged, I'd port it back to the 5.0.2 release which the can be pushed out! Thanks!

cms/admin/utils.py
```diff
@@ -696,6 +699,49 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
             setattr(form._content_instance, self.grouper_field_name, obj)
             form._content_instance.save()
 
+    def get_search_fields(self, request):
+        """Return search fields for either grouper model or content model"""
+        content_search_fields = []
+        grouper_search_fields = []
+        for field_name in self.search_fields:
+            if field_name.startswith(CONTENT_PREFIX):
+                content_search_fields.append(field_name.replace(CONTENT_PREFIX, ""))
```
fsbraun:
```suggestion
                content_search_fields.append(field_name[len(CONTENT_PREFIX):]
```
The `CONTENT_PREFIX` could appear multiple times in the search field just accidentally. Let's make sure we only remove it once!

mhsiddiqui:
> @mhsiddiqui Once we get this merged, I'd port it back to the 5.0.2 release which the can be pushed out! Thanks!

@fsbraun all changes recommended by you are done.

cms/admin/utils.py
```diff
@@ -696,6 +696,49 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+        content_queryset = self.content_model.admin_manager.all()
+        content_search_fields = self.get_search_fields(request)
+        if content_search_fields:
+            content_search_result, __ = super().get_search_results(request, content_queryset, search_term)
+        else:
+            content_search_result = self.content_model.admin_manager.none()
+        search_result_from_content = queryset.filter(
+            id__in=content_search_result.values_list(f"{self.grouper_field_name}_id", flat=True)
+        )
+        return search_result_from_content, False
```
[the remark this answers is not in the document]
mhsiddiqui: These changes require change in core Django functionality. We are currently avoiding any such change.

cms/admin/utils.py
```diff
@@ -696,6 +696,49 @@
...
```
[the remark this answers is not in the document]
mhsiddiqui: These changes require change in core Django functionality. We are currently avoiding any such change.

cms/admin/utils.py
```diff
@@ -696,6 +696,49 @@
             setattr(form._content_instance, self.grouper_field_name, obj)
             form._content_instance.save()
 
+    def get_search_fields(self, request):
```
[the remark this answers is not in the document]
mhsiddiqui: These changes require change in core Django functionality. We are currently avoiding any such change.

cms/admin/utils.py
```diff
@@ -696,6 +696,48 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+            may_have_duplicate_grouper & may_have_duplicate_content
+        )
+
+    def _get_content_search_result(self, request, queryset, search_term):
+        """Get search results from content model"""
+        content_queryset = self.content_model.admin_manager.all()
+        if self.get_search_fields(request):
+            content_search_result, __ = super().get_search_results(request, content_queryset, search_term)
+        else:
+            content_search_result = self.content_model.admin_manager.none()
```
fsbraun:
```suggestion
        try:
            # Set flag on request object to get the content search fields. `get_search_results` will call 
            # `get_search_fields` to get the content search fields.
            request._content_fields = True
            if self.get_search_fields(request):
                content_search_result, __ = super().get_search_results(request, content_queryset, search_term)
            else:
                content_serach_results = self.content_model.admin_manager.none()
        finally:
            request._content_fields = False
```
Setting the `_content_fields` flag in a try finally block will ensure that it is reset even if an exception occurs.

~~Also, the admin's `get_search_results` will evaluate the search fields - no need to go through the list twice.~~

cms/admin/utils.py
```diff
@@ -696,6 +696,48 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+
+        return grouper_search_fields
+
+    def get_search_results(self, request, queryset, search_term):
+        # Reset _content_fields flag
+        request._content_fields = False
+        grouper_search_result, may_have_duplicate_grouper = super().get_search_results(request, queryset, search_term)
+
+        # Set _content_fields for get_search_fields to return content model search fields
+        request._content_fields = True
```
fsbraun:
```suggestion
```
same

cms/admin/utils.py
```diff
@@ -696,6 +696,48 @@ def save_model(self, request: HttpRequest, obj: models.Model, form: forms.Form,
...
+                grouper_search_fields.append(field_name)
+
+        if getattr(request, "_content_fields", False):
+            return content_search_fields
+
+        return grouper_search_fields
+
+    def get_search_results(self, request, queryset, search_term):
+        # Reset _content_fields flag
+        request._content_fields = False
```
fsbraun:
```suggestion
```
This can be moved into `_get_content_search_result`

fsbraun (changes requested):
Hi @mhsiddiqui ! This is turning out really nice! The only thing I'd like to ask you to look into is if we can encapsulate the request flag in a try finally block? This would make sure the flag is always removed - even if an exception was raised.

Out of curiosity: What was breaking the tests for Django < 5.2?

cms/admin/utils.py
fsbraun: I probably am wrong about the second point (crossed out now) here: If the search list is empty the `super().get_search_results()` call might return **all** content objects - we only want additional ones.

mhsiddiqui:
> Hi @mhsiddiqui ! This is turning out really nice! The only thing I'd like to ask you to look into is if we can encapsulate the request flag in a try finally block? This would make sure the flag is always removed - even if an exception was raised.
> 
> Out of curiosity: What was breaking the tests for Django < 5.2?

@fsbraun tests were failing due to some unused imports which I actually added for my previous logic but forgot to remove when logic was changed.

fsbraun (approved):
🎉

PR #8418, opened 2025-12-02
fix: clear apphook fields when copying page trees

### LINKED ISSUE #5899: Copy page tree with apphooks throws a server error
### Summary
When trying to __copy a page tree__ that somewhere down the tree has an __apphook__ installed and then pasting the page tree somewhere else, Django-CMS throws a server error. 

It seems that copying apphooks is not implemented and if I try copying a single page w/ apphook I get the appropriate message ``Error: This page cannot be copied because an application is attached to it. See the Page's Advanced settings to manage apphooks.``

I consider this __important__, since the __process for editing an entire page tree__ is to first copy it, make the modifications (while the original tree is still online) and then publish the copy and unpublish the original tree.

### Expected behaviour
I would expect a meaningful error message instead of the 500 server error.

Even better, I would like Django-CMS __copy the full page tree__ w/ apphooks. If the latter is not possible for whatever technical reason, I would like to at least __get the full tree with apphooks removed__ plus a message that they could not be copied.

#### Actual behaviour
Error message ``Error: INTERNAL SERVER ERROR``

#### Environment

* Python version: 3.6
* Django version: 1.8.14
* django CMS version: 3.4.2

#### Thanks
Great work, everyone! Thanks very much!

Vincent-Ngobeh:
## Summary

When copying a page tree containing pages with apphooks, the copy operation would fail with a 500 server error due to unique constraint violations on the application_namespace field.

This change explicitly clears apphook-related fields (`application_urls`, `application_namespace`, `navigation_extenders`) when copying pages, allowing page trees with apphooks to be copied successfully.

## Related resources

* Fixes #5899

## Checklist

* [x] I have opened this pull request against `main`
* [x] I have read the [contribution guidelines](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined [our Discord Server](https://discord-pr-review-channel.django-cms.org) and the channel [#pr-reviews](https://discord.com/channels/800813886689247262/1236299181761630249) to find a "pr review buddy" who is going to review my pull request.

I haven't joined the discord but have read the guidelines.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -38,7 +38,7 @@ coverage==7.7.0
     # via -r requirements.in
 datetime==5.5
     # via -r requirements.in
-django==4.2.26
+django==4.2.27
     # via
     #   -r requirements.in
     #   django-classy-tags


vinitkumar:
@Vincent-Ngobeh Thanks for the contribution. 💯 

Minor nitpick, looks like your editor or pre-commit hook is formatting the code on save, and hence such a big diff for the test file. Ideally, we would like to have just the code and the diff to keep the patch as concise and to the point as possible.

Vincent-Ngobeh:
Thank you very much for the feedback. I will make sure that the formatter is disabled next time.

vinitkumar (approved):
Looks good to me. @fsbraun If this looks good to you, let's get this merged and released 🚢  🚀

cms/models/pagemodel.py
```diff
@@ -469,6 +469,16 @@ def copy(
...
+        # Clear apphook fields to prevent unique constraint violations
+        # and server errors when copying page trees with apphooks.
+        # Fixes: https://github.com/django-cms/django-cms/issues/5899
+        if self.application_urls or self.application_namespace or self.navigation_extenders:
+            new_page.update(
+                application_urls=None,
+                application_namespace=None,
+                navigation_extenders=None,
+            )
+
```
fsbraun:
```suggestion
```

@Vincent-Ngobeh I've read your test, and it confirms that the issue is fixed.

It turns out that this change is not needed for the test to succeed. Maybe someone has fixed the issue (on purpose or otherwise) in the meantime?

I'd love to keep the regression test, but would like to remove these lines. Are you ok with that?

fsbraun (changes requested):
@Vincent-Ngobeh Thanks. Really looks great. I think the changes in `pagemodel.py` are not needed but would like to keep the rest of the PR, esp. the regression test.

cms/models/pagemodel.py
Vincent-Ngobeh: Yes, I am happy with that. Thank you very much.

Vincent-Ngobeh: Hi @fsbraun, I've updated the PR as requested:
- Removed the apphook field clearing code from pagemodel.py
- Kept only the regression test in test_page_admin.py

Let me know if there's anything else needed.

PR #8449, opened 2026-01-18
fix: resolve empty page_title in frontend edit mode for apphooks

### LINKED ISSUE #8427: When editing "outside cms" models, {% page_attribute "page_title" %} does not work
## Description

in an apphook page, I have a model's detail page (/apphookpage/detail/id/random-slug/). When editing, the {% page_attribute "page_title" %} tag does not return anything anymore?

## Steps to reproduce

- prepare a template with {% page_attribute "page_title" %}, check that it outputs the page title
- have your model (poll, project, news... ) configured for frontend editing
- create an apphook page, and urls that have a detail view (apphookpage/detail/id/) for your model
- check that editing works (edit button appears in toolbar)
- click on edit, no page title

## Expected behaviour

output page title

## Actual behaviour

empty string

CMS 4.1, python 3.11, django 4.2

* [X] No, I only want to report the issue.

woolfie1101:
## Description

When editing an external model (non-PageContent) attached via an apphook in frontend editing mode, `request.current_page` was not being correctly resolved in the `render_object_endpoint` view. This caused template tags such as `{% page_attribute "page_title" %}`, which rely on the existence of `current_page`, to return an empty string.

This PR fixes the issue by attempting to resolve `request.current_page` using the `live-url` query parameter provided during frontend editing. It first tries to match a regular CMS page and then falls back to `applications_page_check` to handle apphook-driven URLs.

스크린샷 2026-01-18 오후 5 15 45

## Related resources

* Fixes #8427

## Checklist

## Summary by Sourcery

Ensure frontend editing of apphook-attached models correctly resolves the current CMS page from the live URL so page-dependent template tags work as expected.

Bug Fixes:
- Resolve request.current_page during render_object_endpoint using the live-url query parameter, handling both regular CMS pages and apphook-driven URLs.

Tests:
- Add a regression test verifying that page_attribute-based template rendering works for apphook-attached models in frontend editing mode.


woolfie1101:
I have posted a message in the official Discord `#pr-reviews` channel to find a review buddy, as required by the contribution guidelines.

**Discord Message Link:** https://discord.com/channels/800813886689247262/1236299181761630249/1462350112255905824

fsbraun:
@woolfie1101 Thanks for the PR — great thinking, and really appreciate you jumping in as a new contributor! 🙌

The main concern I have is that the query parameter isn’t guaranteed to be present. In fact, it’s not set by default, so relying on it could lead to inconsistent behavior depending on how or where the view is accessed.

Instead, could you try using the toolbar object’s get_absolute_url (when it exists) and match against that? This should be more robust and aligns better with how the toolbar is intended to be used. It also has the advantage of working consistently with all frontend-editable, apphooked items, regardless of whether additional query params are provided.

If you run into any issues with that approach or want to bounce ideas around, feel free to ask — happy to help iterate on this. And thanks again for the contribution!

woolfie1101:
Hi @fsbraun! Thank you for the insightful feedback.🥹
I've updated the PR to incorporate your suggestions.

[Changes made]
1. Fallback Logic : Now, if the `live-url` query parameter is missing, the view attempts to use the object's `get_absolute_url()` to resolve the page context.
2. Defensive Coding : I've wrapped the `get_absolute_url()` call in a `try-except` block to ensure that any potential `NoReverseMatch` or other errors during URL reversal (common in some complex apphook setups) won't crash the edit view.

[Note on Testing]
I've verified the path resolution logic via the existing automated test. I attempted to write a dedicated test case for the `get_absolute_url` fallback; however, since the test environment re-fetches the object from the database, mocking the method on the instance level was tricky without more invasive changes to the test setup. Nevertheless, the core resolution logic [get_page_from_request] and [applications_page_check] is identical to the one verified by the `live-url` parameter, so the fix is confirmed to be stable.

Let me know if you'd like me to iterate further on the test setup!

cms/views.py
```diff
@@ -320,6 +320,29 @@ def render_object_endpoint(request, content_type_id, object_id, require_editable
...
+    if not getattr(request, "current_page", None):
+        live_url = request.GET.get(get_cms_setting("ENDPOINT_LIVE_URL_QUERYSTRING_PARAM"))
+        if not live_url and hasattr(content_type_obj, "get_absolute_url"):
+            try:
+                live_url = content_type_obj.get_absolute_url()
+            except Exception:
+                live_url = None
+
+        if live_url:
+            from cms.appresolver import applications_page_check
```
fsbraun:
```suggestion
        try:
            object_url = content_type_obj.get_absolute_url()
        except (AttributeError, NoResolverMatch):
            object_url = None

        if object_url:
            from cms.appresolver import applications_page_check
```
I think we can increase maintainability be relying on `get_absolute_url()`, the query string feature does not have to get entangled here.

cms/views.py
```diff
@@ -320,6 +320,29 @@ def render_object_endpoint(request, content_type_id, object_id, require_editable
...
+            # First, check if it's a regular CMS page
+            request.current_page = get_page_from_request(request, use_path=live_url, clean_path=True)
+            if not request.current_page:
+                # If not, check if it's an apphook
+                original_path_info = request.path_info
+                request.path_info = live_url
+                try:
+                    request.current_page = applications_page_check(request)
+                finally:
+                    request.path_info = original_path_info
```
fsbraun:
```suggestion
            # Check if it's an apphook
            original_path_info = request.path_info
            try:
                request.path_info = object_url
                request.current_page = applications_page_check(request)
            finally:
                request.path_info = original_path_info
```
It is for sure not a page at this stage. Calling `applications_page_check()` is sufficient.

cms/tests/test_apphook_frontend_editing.py
```diff
@@ -0,0 +1,73 @@
...
+        
+        original_renderer = cms_extension.toolbar_enabled_models.get(Category)
+        cms_extension.toolbar_enabled_models[Category] = mock_render_category
+        
+        try:
+            # 4. Action: Create a request
+            # We provide live-url to ensure the test passes, 
+            # while the view code now has a fallback to get_absolute_url()
+            live_url = f"/en/apphook-page/category/{category.pk}/"
+            url = f"/admin/cms/placeholder/render-object-edit/{ct.pk}/{category.pk}/?live-url={live_url}"
```
fsbraun:
```suggestion
            url = f"/admin/cms/placeholder/render-object-edit/{ct.pk}/{category.pk}/"
```
Let's see if the test already works! The `Category` model has a `get_absolute_url()` method.

fsbraun (changes requested):
@woolfie1101 Thanks for getting back on this. I looked into it in detail now.

I believe we can save the trouble of getting the query string - lets directly use the `get_absolute_url`. It will simplify maintenance by keeping the features disentangled.

I think the resolution can be simplified and boils down to running the `applications_page_check()` method (which is good since we do not want unnecessary db hits).

When looking at the suggestions be aware that you might also need to import `NoReverseMatch`.

What do you think?

woolfie1101:
Thanks for the additional pointers, @fsbraun! 🥹
I agree that this is a much cleaner and more maintainable approach.
I've now simplified the implementation in this latest iteration...

1. Simplified Resolution : Removed the query string logic entirely. It now relies solely on get_absolute_url() and directly calls applications_page_check() to avoid unnecessary DB hits.
2. Safer Execution : Applied a try-finally block for the path swap and handled NoReverseMatch explicitly to prevent any crashes during resolution.
3. Test Refinement : Cleaned up the test code as well—removing unused variables, organizing imports, and consolidating the setup logic.

The logic feels much more robust now. Ready for your review!

woolfie1101:
@fsbraun Thank you so much for the approval and for the final cleanup of the code! 🙇‍♀️

This is actually my first-ever contribution to an open-source project, so it's incredibly meaningful to me. Your insightful feedback and the final refactoring were extremely helpful and a great learning experience. I also truly appreciate your fast responses throughout the process—it made this first step so much easier.

Thanks again for the great experience and for welcoming me as a contributor! 🥰

fsbraun:
@woolfie1101 I hope we see each other again soon. Are you on discord: https://www.django-cms.org/discord ?

woolfie1101:
@fsbraun Yes, I've already joined the Discord server! I'm looking forward to staying connected and contributing more to the community. See you there! 😊

PR #8254, opened 2025-06-12
fix: Add `default_auto_field` to cms/apps.py

fsbraun:
## Description

Adds the `default_auto_field` property to `cms.apps` and menus.apps`.

Fixes #8253 - no more migrations created when project `DEFAULT_AUTO_FIELD` is not set to `AutoField`.

## Related resources

* #8253 
* #...

## Checklist

* [ ] I have opened this pull request against ``main``
* [ ] I have added or modified the tests when changing logic
* [ ] I have followed [the conventional commits guidelines](https://www.conventionalcommits.org/) to add meaningful information into the changelog
* [ ] I have read the [contribution guidelines ](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined the channel #pr-reviews on our [Discord Server](https://discord-pr-review-channel.django-cms.org) to find a “pr review buddy” who is going to review my pull request.

## Summary by Sourcery

Enhancements:
- Add default_auto_field attribute to CMSConfig and MenusConfig to prevent Django warnings about missing default auto fields

PR #8461, opened 2026-01-26
fix: Revert "Add `default_auto_field` to cms/apps.py (#8254)"

fsbraun:
Reverts django-cms/django-cms#8258

Adding `BigAutoField` to as a `default_auto_field` to the cms and menus app config was an **unintended breaking change**. It needs to be `AutoField`.

* This change **forces downstream creation of migrations** which will need to recreate the pk of all third-party plugins including restrictions  
* No immediate benefits arise:
  - Plugins still limited by `SmallIntergerField` `position`.
  - Pages still limited by django-treebeard
  - Placeholders limited by plugins and pages

## Summary by Sourcery

Revert unintended changes to primary key auto field configuration for CMS and menus apps and drop the associated migrations and upgrade documentation entry.

Bug Fixes:
- Restore the CMS and menus apps to use their previous auto field configuration to avoid forcing unnecessary primary key‑changing migrations in downstream projects.

Documentation:
- Remove the 5.1.0 upgrade note related to the reverted auto field change.


vinitkumar (commented):

vinitkumar (approved):

PR #8471, opened 2026-02-04
fix: Grouper admin kept read-only fields as prepopulated fields

fsbraun:
## Description

Django does not allow for prepopulated fields to contain readonly fields. This PR removes them temprarily from the grouper admin instance. 

## Related resources

* #...
* #...

## Checklist

## Summary by Sourcery

Ensure grouper admin handles prepopulated and read-only fields without violating Django’s admin constraints and align form behavior with the underlying content model.

Bug Fixes:
- Prevent read-only content fields from being used as keys in prepopulated_fields in the grouper admin, complying with Django’s restriction on prepopulated read-only fields.
- Fix boolean detection for list display fields by basing it on the content model’s field type instead of the generated form fields.
- Correct the stored _content_model in the grouper admin form mixin to reference the actual content model class.

Enhancements:
- Generate grouper admin content field accessors from a full content-model ModelForm rather than the current admin form, ensuring all content fields are covered consistently.
- Wrap the admin form class per request to attach _admin and _request metadata cleanly instead of mutating the original form class.
- Cache the original prepopulated_fields configuration on the grouper admin for reuse after read-only filtering.

Tests:
- Add a test verifying that read-only content fields are excluded from prepopulated_fields while remaining in readonly_fields for the grouper admin.


cms/admin/utils.py
```diff
@@ -302,12 +302,16 @@ def __init__(self, model, admin_site):
         if not issubclass(self.form, _GrouperAdminFormMixin):
             self.form = type(
                 "AutoGeneratedGrouperAdminForm",
-                (GrouperAdminFormMixin(self.content_model), self.form),
+                (GrouperAdminFormMixin(self.content_model), copy(self.form)),
```
fsbraun:
```suggestion
                (GrouperAdminFormMixin(self.content_model), self.form),
```

vinitkumar (approved):

PR #8497, opened 2026-02-25
fix: Keep GET params when toggling structure mode

petrklus:
## Description

Fix for the GET parameter retention with https://github.com/django-cms/django-cms/issues/8443 and https://github.com/django-cms/django-cms/issues/8496

## Checklist

## Summary by Sourcery

Preserve relevant GET parameters when generating toolbar edit, preview, and structure URLs so that toggling structure mode retains query state correctly.

Bug Fixes:
- Ensure toolbar object edit, preview, and structure URLs carry through existing GET parameters except language when toggling modes.
- Prevent client-side history manipulation from appending duplicate or stale query strings when switching between edit and structure modes.

Enhancements:
- Mark toolbar URL template variables as safe to allow inclusion of full query strings without escaping issues.

### DIFF
--- a/cms/models/pagemodel.py
+++ b/cms/models/pagemodel.py
@@ -389,7 +389,7 @@ class Page(MP_Node):
         """
         Called from admin interface when page is moved. Should be used on
         all the places which are changing page position. Used like an interface
-        to django-treebeard, but after move is done page_moved signal is fired.
+        to django-treebeard.
         """
         assert isinstance(target_page, Page), f"{target_page} is not an instance of Page."
         inherited_template = self.template == constants.TEMPLATE_INHERITANCE_MAGIC
--- a/cms/signals/__init__.py
+++ b/cms/signals/__init__.py
@@ -1,3 +1,5 @@
+import warnings
+
 from django.conf import settings
 from django.contrib.auth.models import Group, User
 from django.db.models import signals
@@ -32,6 +34,7 @@ from cms.signals.permissions import (
     pre_save_user,
     user_m2m_changed,
 )
+from cms.utils.compat.warnings import RemovedInDjangoCMS60Warning
 from cms.utils.conf import get_cms_setting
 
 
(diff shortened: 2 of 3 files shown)


cms/templates/cms/toolbar/toolbar_javascript.html
```diff
@@ -113,9 +113,9 @@
         "sideframe": { "url": "" },
         "sideframe_enabled": {{ cms_toolbar.sideframe_enabled|yesno:"true,false" }},
         "states": [],
-        "edit": "{{ cms_edit_url }}",
-        "edit_off": "{{ cms_preview_url }}",
-        "structure": "{{ cms_structure_url }}",
+        "edit": "{{ cms_edit_url | escapejs }}",
```
fsbraun, line 116: Why is escapejs needed here? Is this an independent injection prevention?

cms/toolbar/toolbar.py
```diff
@@ -415,19 +415,27 @@ def get_object_pk(self):
             return self.obj.pk
         return ''
 
+    def get_request_extra_params(self, remove_params=None):
```
fsbraun: Do we really need this function? I believe the implementation in `get_object_xxx_url` only updates a language if given?

Can you check what happens if you just do `get_object_xxx_url(self.obj, language=..., params=self.request.GET`?

Would save us repeatedly creating copies of `request.GET`.

cms/templates/cms/toolbar/toolbar_javascript.html
petrklus: We need to either mark safe or use this - to avoid translating & to &amp;. Marking safe would work as well but I think escapejs is more specific and works fine. Wdyt?

cms/toolbar/toolbar.py
petrklus: Just wanted to be explicit here - I've reviewed logic of `get_object_edit_url` and `_get_object_url` and they do seem to prioritise the language param over the language param present in the request.GET, so my addition was redundant. Removed now.

petrklus:
@fsbraun thank you for taking the time to review - I've addressed your comments and now looking at adding unit tests

petrklus:
@fsbraun let me know your thoughts!

PR #8521, opened 2026-03-09
fix: Make sure URL parameters are forwarded for CMS URLs

petrklus:
## Description

This relates to https://github.com/django-cms/django-cms/issues/8443.

There were a few other places where the URL parameters were not forwarded.

To make sure that no URL is missed, the URL factory `updateUrlWithPath` was modified. Care was taken to not override any existing additional URL parameters to avoid any conflicts

## Checklist

## Summary by Sourcery

Ensure CMS URLs preserve and merge existing query parameters when generating updated URLs with cms_path.

Bug Fixes:
- Fix loss of existing URL query parameters when updating CMS URLs via updateUrlWithPath.

Enhancements:
- Prioritize explicitly provided URL parameters over those from the current window location when composing CMS URLs.


petrklus:
@fsbraun I've found a few more places that needed the URL parameter appending, namely the plugin move and others, so I've modified the URL factory.

This was not to only address toolbar issues, but we also had issues with segmentation plugins (we show country-specific content) which was not working properly after the CMS content was updated post plugin move.

fsbraun:
@petrklus Can you describe in which situations this change is needed?

petrklus:
> @petrklus Can you describe in which situations this change is needed?

Of course!

Say you are editing a page, and looking at its content for US by using url?countryCode=US.

If you add a new plugin, move a new plugin, etc. - the re-rendered content does not take into account the countryCode and shows fallback (country detected from the editor IP), which makes editing country-specific content next to impossible.

We have plugins that look at countryCode parameter and show different children based off it, and if the countryCode is lost when adding/moving plugins, this is an issue.

Does the above make sense? If not, please let me know, and I will elaborate further

fsbraun:
Can you track which specific requests lack the query string? I'm trying to pin this to an endpoint and potentially solve this on the backend.

petrklus:
> Can you track which specific requests lack the query string? I'm trying to pin this to an endpoint and potentially solve this on the backend.

I think there will be quite a few - that's why I added the functionality to the path processor. From the initial testing it woudl be:

`django-cms/cms/templates/cms/toolbar/toolbar_javascript.html`, the `request.toolbar` needs URL params to re-render the toolbar correctly
`django-cms/cms/templates/cms/toolbar/plugin.html` - every URL under `urls`. 

These are just the two I've gotten to - I imagine any other operations with the placeholder, such as pasting plugins, would have the same problem.

I've also chosen to update the `updateUrlWithPath` because it needs to be modified anyway - from what I can see, it expects URLs without any query string, for example, the [_loadToolbar](https://github.com/django-cms/django-cms/blob/1a28c25ee2dafe3f632fa19da3af7799b33c8af1/cms/static/cms/js/modules/cms.structureboard.js#L1215) automatically appends a ? a the end of the incoming URL from the HTML, and I imagine there may be others.

The proposed solution always treats params in the URL as secondary to any there may be in the URL.

In a nutshell, anything that causes page content or toolbar re-render needs to be provided with the URL parameters from the source page.

fsbraun:
My thinking goes into updating https://github.com/django-cms/django-cms/blob/1a28c25ee2dafe3f632fa19da3af7799b33c8af1/cms/templates/cms/toolbar/toolbar_javascript.html#L21 and let the toolbar url be generated by the view (like the structure url) including query params.

This is the internal Ajax url to refresh the toolbar. It will need the query params. `_loadToolbar` will need to preserve them and not make the assumption the toolbar url does not carry query params.

Editing changes then can redraw the toolbar for the specific country, e.g. after changing, moving or pasting plugins.

Copy, paste, edit, or add endpoints themselves must not depend on query params. They are admin interactions modifying plugins (independently on how they are rendered).

petrklus:
> My thinking goes into updating
> 
> https://github.com/django-cms/django-cms/blob/1a28c25ee2dafe3f632fa19da3af7799b33c8af1/cms/templates/cms/toolbar/toolbar_javascript.html#L21
> 
> and let the toolbar url be generated by the view (like the structure url) including query params.
> This is the internal Ajax url to refresh the toolbar. It will need the query params. `_loadToolbar` will need to preserve them and not make the assumption the toolbar url does not carry query params.
> 
> Editing changes then can redraw the toolbar for the specific country, e.g. after changing, moving or pasting plugins.
> 
> Copy, paste, edit, or add endpoints themselves must not depend on query params. They are admin interactions modifying plugins (independently on how they are rendered).

I've tried that - I've started by "just" fixing the toolbar - but that was not enough. After the fix, when I, for example, move plugins, the toolbar was regenerated correctly with the correct country, however, the page content was not.

For example, the move_plugin endpoint provides updated plugin content as it returns, and if that plugin content is responding to URL parameters, the move_plugin endpoint needs matching ones to reproduce them.

I think that's why the frontend is a better place for appending the URL parameters in this case.

fsbraun:
Maybe I do not fully grasp the issue:
* The plugins might render differently depending on query params
* Django CMS is not equipped to handle placeholders that have different plugins depending on query params (only by language)
* The plugins shown in the sidebar must always be the same irrespective of the query params. Anything else most likely is going to garble the plugin tree.
* I stand corrected, however, that some actions might need the query params: If they return an Ajax result containing an updated toolbar (such as the toolbar endpoint). The two additional candidates are `copy_plugin` and  `move_plugin`.
* So, the **remaining issue is editing and adding plugins**. They do close the form and render the toolbar without another request response cycle. I have to think about this, since adding arbitrary query params might inject unintended side effects with Django's admin (such as changing fields). But these endpoints did not receive query params in v3 either.

petrklus:
> Maybe I do not fully grasp the issue:
> 
> * The plugins might render differently depending on query params
> * Django CMS is not equipped to handle placeholders that have different plugins depending on query params (only by language)
> * The plugins shown in the sidebar must always be the same irrespective of the query params. Anything else most likely is going to garble the plugin tree.

Correct - the way our segmentation plugins work is that you always have all of the plugins present, see example below:

image

What happens on each render that only some of the plugins are visually present in the rendered page (say if you are in the UK, you only see the content from the GB plugin and not the others). Other plugins just render hidden markup (for the CMS editor compatibility)

> * I stand corrected, however, that some actions might need the query params: If they return an Ajax result containing an updated toolbar (such as the toolbar endpoint). The two additional candidates are `copy_plugin` and  `move_plugin`.

In our example above, the country code is set on the request by the CDN, but in case it is not detected correctly, it needs to be possible to overwrite it in the URL (we have a UI for this). The overwriting is also important for CMS editors as they need to be able to simulate a different country for editing purposes, for which we have built a toolbar country select menu.

> * So, the **remaining issue is editing and adding plugins**. They do close the form and render the toolbar without another request response cycle. I have to think about this, since adding arbitrary query params might inject unintended side effects with Django's admin (such as changing fields). But these endpoints did not receive query params in v3 either.

I thought about another solution to make this more robust - the URL params that really need to be forwarded will likely be a well known set - what if we added a whitelist to the settings? What do you think?

Regarding V3 - I've just tried moving plugins while a `countryCode` override was set and there were no issues - the country selection was not lost. I could see that the entire page was re-rendered via ajax using the  original url including URL parameters.

The whitelisting would work for us fine @fsbraun - there is only a handful of URL parameters we would need to be transferred for each site.

fsbraun:
I think we're moving in on this! Can you try setting `CMS_ALWAYS_REFRESH_CONTENT = True` and see if that works. This would be the immediate solution.

Adding a whitelist for query params to all endpoints (preview, edit, structure, toolbar, move, copy, edit/add) seems a good idea. This will allow you to use the partial refreshes introduced with django CMS 5.0.

petrklus:
> I think we're moving in on this! Can you try setting `CMS_ALWAYS_REFRESH_CONTENT = True` and see if that works. This would be the immediate solution.

Thank you, will try it out!

> Adding a whitelist for query params to all endpoints (preview, edit, structure, toolbar, move, copy, edit/add) seems a good idea. This will allow you to use the partial refreshes introduced with django CMS 5.0.

Would you like me ot add this to the PR? Also, any recommendations on the best spot where to output the whitelisted params to have the frontend pick them up?

petrklus:
@fsbraun I seem to have a broken behaviour with the `CMS_ALWAYS_REFRESH_CONTENT`. When a plugin is updated and the content is refreshed, I get this:

image

Looks like the toolbar is broken and the page title is not set.

We were seeing this occasionaly on V3, but now it seems to be happening all the time with the legacy behaviour in V5. Any ideas how to fix/what may be the cause?

If I subsequently open the structureboard, that fixes the toolbar, but the broken page <title> remains.

Also, there is that odd yellow bar above toolbar that is not present on normal page loads.

I can hack a fix for the toolbar:

```javascript
            // CMS fix for the incomplete toolbar loading
            if (window.matchMedia('(min-width: 740px)').matches) {
                if ($(".cms-toolbar-item-navigation li.cms-toolbar-more:first-child").length == 1) {
                    console.log("Refreshing toolbar again, malformed toolbar detected");
                    // refresh everything
                    setTimeout(function() {
                        CMS.API.StructureBoard._loadStructure();
                    }, 500);
                }
            }
}
```

But the missing page title and the odd yellow bar remains.

And there is quite a bit of broken stuff in the <head> tag:

image

petrklus:
Looks like the issue is in `django-cms/cms/static/cms/js/modules/cms.structureboard.js`, in the `refreshContent` function. I do not think the dd.diff is doing a very good job, have you ever seen this behaviour?

EDIT: @fsbraun , I have also done a litte experiment where I've compared the original page source's <head> with the one generated from the CMS on the content refresh - and I can confirm that they are identical. The initial <head> content and the incoming one from the content refresh are identical, I appreciate that the dd.diff is comapring DOM currently existing at that time, which may have been modified, but something does not seem right.

petrklus:
To add context:

<head> before content refresh:

image


<head> after content refresh:

image

I wonder, rather than comparing to the existing <head>, that may have been modified by other scripts, would it make sense to store a copy of the initial <head> tag, then compare against it? As you can see in the screenshots above, some additional styles added by the consent library were stripped away by the refresh.

petrklus:
Would there be any appetite to provide customisable `refreshContent` function? This honestly seems like quite a task to get right by checking the entire content. In our use-case, the only thing that may concievably change with plugin changes are <style> tags in head, we have done something along these lines in the past:

```javascript
        $(document).find('head > style').detach(); // remove existing styles
        // any nested style would be inside <noscript> and we do not need to worry about that during CMS previews        
        $(newDoc)
            .find('head > style')
            .each(function (index, el) {
                var el = $(this);
                //console.log(el);
                $(document).find('head').append(el.clone());
            });
        //
```

We have hacked the above into the refreshContent function in the past when we were seeing freezes with the dd library trying to do the comparison, and that would still work well and really decrease the potential for errors.


EDIT: Just tried commenting out the `dd.apply(document.head, headDiff);` and just doing our simple `<style>` replacement and that works flawlessly.

petrklus:
Sorry for the barrage of messages @fsbraun , I think that the `dd.apply(document.head, headDiff);` as it is currently use would break a good few things - for example, we have a few widgets that inject <script> and <style> tags into the <head>. This means that the document.head that the dd is doing a diff against is not comparable to the incoming newDoc.head, as the newDoc.head was not modified in the same way.

I cannot think of a good way to resolve this - the only way would be to keep a copy of the original <head> before any scripts ran, and diff against it, but that seems a bit difficult to obtain, perhaps on a page load, fetch a fresh `contentMarkup` and keep its <head> as the initial reference to diff against, but that introduces another content load for any CMS edit mode page load, which seems excessive, but seems like one of the only fully universal ways.

In our case though, I think just parsing out <style> from the incmoing <head> is enough, as in our case this does not cause issues (the widgets we use only inject <script> and <link> tags, so we do not strip their changes).

petrklus:
@fsbraun something like this would work great for us:

```javascript

    refreshContent(contentMarkup) {
        this._requestcontent = null;
        if (!this._loadedStructure) {
            this._requeststructure = null;
        }
        const newDoc = new DOMParser().parseFromString(contentMarkup, 'text/html');

        const structureScrollTop = $('.cms-structure-content').scrollTop();

        const toolbar = $('#cms-top, [data-cms]').detach();
        const newToolbar = $(newDoc).find('.cms-toolbar').clone();

        $(newDoc).find('#cms-top, [data-cms]').remove();

        const messages = this._extractMessages($(newDoc));

        if (messages.length) {
            setTimeout(() =>
                messages.forEach(message => {
                    CMS.API.Messages.open(message);
                })
            );
        }

        this._replaceBodyWithHTML(newDoc.body);
        
        if (typeof window.CMS_CUSTOM_HEAD_REPLACE_FUN === "function") {
            window.CMS_CUSTOM_HEAD_REPLACE_FUN(document.head, newDoc.head);
        } else {
            const headDiff = dd.diff(document.head, nodeToObj(newDoc.head));
            dd.apply(document.head, headDiff);
        }

        toolbar.prependTo(document.body);
        CMS.API.Toolbar._refreshMarkup(newToolbar);

        $('.cms-structure-content').scrollTop(structureScrollTop);
        this._loadedContent = true;
        this._contentChanged();
    }
```

The CMS UI would check for a custom CMS_CUSTOM_HEAD_REPLACE_FUN, which will then handle the <head> update instead of the built-in method.

petrklus:
> > I think we're moving in on this! Can you try setting `CMS_ALWAYS_REFRESH_CONTENT = True` and see if that works. This would be the immediate solution.
> 
> Thank you, will try it out!
> 
> > Adding a whitelist for query params to all endpoints (preview, edit, structure, toolbar, move, copy, edit/add) seems a good idea. This will allow you to use the partial refreshes introduced with django CMS 5.0.
> 
> Would you like me ot add this to the PR? Also, any recommendations on the best spot where to output the whitelisted params to have the frontend pick them up?

Hi @fsbraun, keen to finish up the url parameter appending feature - could you please let me know where would you prefer me to add the whitelist and how to best transfer it to be picked up by the front-end?

fsbraun:
@petrklus Sorry, this sat on my side for a while now. I think we should go with the whitelist, `CMS_REQUIRED_QUERY_PARAMS = [...]`. They need to be included in the CMS.config object to make them available to the JS frontend. What you think? Maybe take that to 5.1?

fsbraun:
@petrklus I have gone through the code once more. I believe only add plugin can be a security issue (since the model admin of django prepopulates fields based on get params). 

Does it make sense to just  drop adding the query params to the add plugin endpoint?

fsbraun:
@petrklus Sorry, this has been dormant for a while. Does the issue persist?

I assume that setting `CMS_ALWAYS_REFRESH_CONTENT = True` helped (the toolbar update has been fixed elsewhere). When  `CMS_ALWAYS_REFRESH_CONTENT = False` you still get the wrong results and this PR would help. While I can see this for copy, paste, or move, I fear that editing or adding a plugin will lose the Query-Params through the edit modal.

Can you give a brief update?

petrklus:
Hi @fsbraun , just to check, is the 5.0.x now forwarding all URL parameters when switching to the CMS editor? Or is there a white-list to be configured anywhere?

fsbraun:
@petrklus I fear not (see my inquiry [above](https://github.com/django-cms/django-cms/pull/8521#issuecomment-4641978945)).

Do you still experience issues with `CMS_ALWAYS_REFRESH_CONTENT = True`?

Please feel free to reopen (stale bot closed it).

petrklus:
> @petrklus I fear not (see my inquiry [above](https://github.com/django-cms/django-cms/pull/8521#issuecomment-4641978945)).
> 
> Do you still experience issues with `CMS_ALWAYS_REFRESH_CONTENT = True`?
> 
> Please feel free to reopen (stale bot closed it).

Thank you for getting back to me - I think I will need to try it again, we are running a fork with a few fixes in, will need to try running a clean version and see if the issues are still there. Will reopen if still an issue!

PR #8528, opened 2026-03-17
fix: `delete_orphaned_plugins` could corrupt the plugin tree order

entuziaz:
## Description

This PR fixes issue #8525 bug where `cms delete-orphaned-plugins` could accidentally break plugin ordering while cleaning up bad rows. 

- Tracing the deletion path showed that `delete-orphaned-plugins` was collecting plugin model instances up front, then deleting them later, while each delete could renumber positions in the same placeholder. The command was still holding old in-memory plugin objects from before that renumbering so later deletions used stale `position` values.

- Work done:
  - instead of storing full plugin objects to delete later, only database IDs are now stored right before deleting each plugin, so a fresh copy is fetched from the database
  - that fresh object has the current position, even if earlier deletions already renumbered the tree

- Tests added:
  - `test_delete_orphaned_plugins_keeps_positions_consecutive` reproduced the bug at the command level by deleting multiple orphaned plugins from one placeholder and asserting the positions should collapse to 1, 2, 3.
  - `test_delete_plugin_uses_stale_position_from_prefetched_instance` isolates the process by showing that after deleting one plugin, a second plugin object still held its old position in memory until `refresh_from_db()`.

- Based on the failed regression test `test_delete_orphaned_plugins_keeps_positions_consecutive`, the `delete_orphaned_plugins.py` was updated to:
  - stop storing plugin instances in the command
  - store only their primary keys
  - re-fetch each plugin with `CMSPlugin.objects.get(pk=...)` immediately before deleting it

That way, every call into `placeholder.delete_plugin(instance)` uses fresh database state instead of stale in-memory positions.

## Related resources

* #...
* #...

## Checklist

## Summary by Sourcery

Add regression coverage and adjust orphaned plugin deletion to avoid corrupting plugin ordering when cleaning up uninstalled or unsaved plugins.

Bug Fixes:
- Ensure the delete-orphaned-plugins management subcommand reloads plugin instances from the database before deletion so placeholder positions remain consistent and consecutive.

Tests:
- Add management command regression test verifying delete-orphaned-plugins keeps remaining plugin positions consecutive after removing bogus plugins.
- Add unit test demonstrating stale in-memory plugin positions after deletions and the need to refresh plugin instances from the database.

### DIFF
--- a/cms/management/commands/subcommands/delete_orphaned_plugins.py
+++ b/cms/management/commands/subcommands/delete_orphaned_plugins.py
@@ -1,4 +1,5 @@
 from cms.management.commands.subcommands.list import plugin_report
+from cms.models.pluginmodel import CMSPlugin
 
 from .base import SubcommandsCommand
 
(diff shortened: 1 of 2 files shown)


fsbraun (approved):
@entuziaz Nice work!

I checked the regression test and it makes sense to me:
* Without the changes they fail
* With the changes they pass.

Now, let's hope that the above idea was the real root cause in my specific case!

Issue #8526, opened 2026-03-13
feat: `./manage.py cms delete-orphaned-plugins` misses plugins of detached placeholders


fsbraun:
**Description**

The management command `manage.py cms delete-orphaned-plugins` does not remove plugins that belong to placeholders which themselves are not attached to any content object.

Since **django CMS 4**, the `Placeholder` model contains a `source` field implemented as a `GenericForeignKey`. Because this relation does not enforce strict database integrity, it is possible for placeholders to remain in the database even though their referenced content object no longer exists. Plugins attached to such placeholders therefore become effectively orphaned.

However, the `delete-orphaned-plugins` command currently does not detect or remove these plugins.

**Important clarification**

This issue **does not concern plugins that belong to unused slots**. Placeholders with valid sources but unused slots are expected to remain intact, and their plugins should not be deleted.

The issue specifically concerns placeholders whose `source` reference no longer resolves to an existing object.

**Steps to reproduce**

1. Create a content object that owns a placeholder.
2. Add one or more plugins to that placeholder.
3. Delete the content object directly (or otherwise remove the referenced object).
4. The placeholder remains with a `source` pointing to a non-existent object.
5. Run:

```bash
python manage.py cms delete-orphaned-plugins
```

**Expected behavior**

Plugins attached to placeholders whose `source` object no longer exists should be considered orphaned and removed.

**Actual behavior**

Plugins remain in the database because the command does not treat plugins attached to such placeholders as orphaned.

**Possible direction**

The command could additionally detect placeholders whose `source` GenericForeignKey cannot be resolved and treat their plugins as orphaned.

granth-agarwal:
Hi @fsbraun, I would like to work on this issue.

### Proposed Solution

The core of the issue is that placeholders remain in the database with a stale `GenericForeignKey` (`content_type` and `object_id`) when their source object is deleted (e.g. bulk deletes, missing signals, or uninstalled apps). This leads to the plugins attached to these placeholders becoming "orphaned" from the perspective of the application, even though the database relationships (`CMSPlugin -> Placeholder`) are still intact.

To resolve this, we can extend the `DeleteOrphanedPluginsCommand` (located in `cms/management/commands/subcommands/delete_orphaned_plugins.py`) to detect and remove these plugins. 

Here is the proposed approach:

**1. Identify Detached Placeholders:**
We can iterate over the distinct `content_type_id`s currently used in the `Placeholder` table. For each `ContentType`:
- Retrieve the associated model class.
[...]

PR #8622, opened 2026-05-09
fix: cascade delete plugins of detached placeholders

hey-granth:
Resolves the issue where the `delete-orphaned-plugins` management command fails to identify plugins attached to detached placeholders.

## Modifications:

* Modified `cms/management/commands/subcommands/delete_orphaned_plugins.py` to evaluate `Placeholder` instances for stale `GenericForeignKey` references.
* Implemented logic iterating through distinct `content_type_id` values to flag placeholders where the source model class is deleted or the specific `object_id` is absent from the source table.
* Executed database deletion on identified detached placeholders, triggering native cascading deletes for all associated `CMSPlugin` records.
* Updated command standard output to report the quantity of removed detached placeholders and plugins.
* Reorganized the import hierarchy in `cms/management/commands/subcommands/delete_orphaned_plugins.py` and `cms/tests/test_management.py` to standardize module structure and resolve `I001` linter violations.
* Added automated unit tests in `cms/tests/test_management.py` simulating content object deletion and asserting successful database removal of stale placeholders and associated plugins via the management command.

## Related resources

* #8526

## Checklist

* [x] I have opened this pull request against `main`
* [x] I have followed [[the conventional commits guidelines](https://www.conventionalcommits.org/)](https://www.conventionalcommits.org/) to add meaningful information into the changelog
* [x] I have read the [[contribution guidelines](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst)](https://github.com/django-cms/django-cms/blob/develop/CONTRIBUTING.rst) and I have joined [[our Discord Server](https://discord-pr-review-channel.django-cms.org/)](https://discord-pr-review-channel.django-cms.org) and the channel [[#pr-reviews](https://discord.com/channels/800813886689247262/1236299181761630249)](https://discord.com/channels/800813886689247262/1236299181761630249) to find a “pr review buddy” who is going to review my pull request.

## Summary by Sourcery

Extend the delete-orphaned-plugins management command to remove detached placeholders and their plugins.

Bug Fixes:
- Ensure the delete-orphaned-plugins command detects placeholders whose content objects or models no longer exist and cleans up their plugins.

Enhancements:
- Include detached placeholder detection in the delete-orphaned-plugins command output and reporting.

Tests:
- Add management command tests covering detached placeholders and verifying associated plugins are removed and output is updated.


fsbraun (commented):
Thanks for tackling this, @hey-granth — detached placeholders leaking orphaned plugins is a real pain, and the overall approach (one pass per distinct `content_type_id`, `in_bulk` for the content types, and an `Exists`/`OuterRef` existence check) reads cleanly. 🙌 I have one important thing to flag before merge, plus a couple of small cleanups.

## 🔴 Important: existence check should use `_base_manager`, not `_default_manager`

```python
object_exists=Exists(
    model_class._default_manager.filter(pk=OuterRef('object_id'))
)
```

The default manager isn't a reliable "does this row exist in the table?" check in django-cms 4. With **djangocms-versioning** installed (very common in production), `PageContent.objects` only returns *published* content — drafts, unpublished and archived rows physically exist in the table but are hidden by the default manager. The same goes for other versioned content models such as djangocms-alias' `AliasContent`.

That means placeholders attached to non-published content would be classified as "detached" and deleted, cascading to all of their plugins — i.e. silently destroying draft/unpublished content. The current test passes only because it runs without versioning, so the issue stays hidden in CI.

The fix is to use `_base_manager`, which bypasses manager-level filtering and queries the raw table:

```python
object_exists=Exists(
    model_class._base_manager.filter(pk=OuterRef('object_id'))
)
```

This is already the established pattern in the codebase — see `cms/views.py` (`PageContent._base_manager...`), which uses it for exactly this reason. It'd be great to also add a regression test using a model whose default manager filters out an existing row (simulating versioning) and assert its placeholder is **not** deleted, so this can't silently regress later.

## 🟡 A few unused imports in the test

The PR mentions resolving `I001`, but the test adds three imports that aren't used and will trip `F401`:

```python
from cms.toolbar.toolbar import CMSToolbar      # unused
from cms.utils import get_current_site          # unused
from cms.utils.conf import get_cms_setting       # unused
```

(`User` and `ContentType` are used and fine.)

## 🟡 Avoid importing the User model directly

```python
from django.contrib.auth.models import User
```

django-cms supports swappable user models, so hardcoding `auth.User` can break under a custom `AUTH_USER_MODEL`. `get_user_model()` is the safe choice (or use any simple model like `Site` for the fixture). Tiny style nit too: `get_for_model(User) # ...` wants two spaces before the `#` (E261).

## 🟢 Minor, non-blocking thoughts

- The `model_class() is None` branch will also flag placeholders whose app is merely *not installed in the current settings* (vs. truly removed). Given the destructive cascade, a one-line note in the command help/docs that it must run with the full production app set would be worth it.
- Wrapping the deletions in `transaction.atomic()` would make this destructive command a bit safer against partial failures.
- The refactor of the existing report assertions into a `for plugin in report` loop drops the link/text instance-count checks — minor, but you could keep those.

Really nice work overall — once the `_base_manager` change (plus a covering test) is in and the import cleanup is done, this should be good to go. Happy to re-review whenever you're ready! 😊

cms/management/commands/subcommands/delete_orphaned_plugins.py
```diff
@@ -7,9 +12,44 @@
 class DeleteOrphanedPluginsCommand(SubcommandsCommand):
     help_string = ('Delete plugins from the CMSPlugins table that should have instances '
                    'but don\'t, and ones for which a corresponding plugin model can no '
-                   'longer be found')
+                   'longer be found. MUST be executed within a complete production application environment.')
```
fsbraun:
```suggestion
                   'longer be found. ATTENTION: Also removes plugins of apps currently not present in INSTALLED_APPS. '
                   'Ensure nothing is temporarily commented out.')
```
I understand that my original comment was not very telling. I hope this is clearer.

cms/tests/test_management.py
```diff
@@ -740,6 +754,36 @@ def test_djangocms_command_returns_version(self):
...
+                proxy = True
+                app_label = 'auth'
+
+            class MyManager(models.Manager):
+                def get_queryset(self):
+                    return super().get_queryset().none()
+
+            objects = MyManager()
+
+        content_type = ContentType.objects.get_for_model(FilteredUser)
```
fsbraun:
```suggestion
        content_type = ContentType.objects.get_for_model(FilteredUser, for_concrete_model=False)
        self.assertEqual(content_type.model_class(), FilteredUser)
```
I believe `for_concrete_model` is necessary here, to not get the (concrete) `User` model back.

fsbraun (changes requested):
Hi @hey-granth ! 

Thanks for the update! Looks great now. Last thing: To make the test for the `_base_manager` work, we need to ensure the test actually works on the restricted model.

Then this is ready to go!

granth-agarwal:
Thanks for the review and the helpful suggestions. I'll make these changes and update the PR shortly.

granth-agarwal:
hey @fsbraun, I've applied both suggestions and pushed the updates. Thanks again for the review.

fsbraun (approved):
Great! Thank you, @hey-granth ! LGTM!

granth-agarwal:
hii @fsbraun!
The `codecov/patch` check failed to hit the target. This is due to the `model_class is None` branch lacking explicit test coverage. Given the approval, do you want to merge this as-is, or should I write an additional test to cover the uninstalled app edge case?

fsbraun:
@hey-granth This is ready for check-in!

PR #8629, opened 2026-05-18
fix: Compatibility with formtools 2.6+ caching

### LINKED ISSUE #8628: django-formtools 2.6+ breaks the page creation wizard (get_form_list caching)
### Description

django-formtools 2.6 / 2.6.1 breaks the django-cms wizard (`WizardCreateView`), and therefore the test suite.

`cms.tests.test_wizards` fails with:

```
TypeError: BaseForm.__init__() got an unexpected keyword argument 'wizard_request'
```

### Root cause

formtools 2.6 (PR [jazzband/django-formtools#291](https://github.com/jazzband/django-formtools/pull/291)) changed `WizardView.get_form_list()` to **cache** its result. Previously it was *"always generated on the fly"*.

`WizardCreateView` declares a placeholder `Form` for step `'1'` and swaps in the real, dynamically-built step-2 form at runtime by mutating `self.form_list` inside `get_form()`. formtools' `get_form()` resolves the form class via `self.get_form_list()[step]`. With 2.6's caching, `get_form_list()` returns a **stale** cached list still containing the placeholder `Form`, which is then instantiated with the `wizard_request` / `wizard_site` / ... kwargs meant for the real form — raising the `TypeError`.

The cache key in 2.6 only reflects `condition_dict`, not `self.form_list`, so the runtime swap is never picked up. Reported upstream as [jazzband/django-formtools#295](https://github.com/jazzband/django-formtools/issues/295).

### Affected versions

- formtools 2.6 and 2.6.1 break the wizard.
- formtools 2.5.1 works.

`django-cms` does not pin `django-formtools`, so existing installations pick up 2.6.x automatically.

### Workaround for existing installations

Pin formtools back to the last working release:

```
django-formtools==2.5.1
```

### Proposed fix

Rather than mutating the shared `self.form_list` inside `get_form()`, perform the step-2 form substitution in an overridden `get_form_list()`, which returns a fresh/copied list and is therefore always honoured regardless of formtools' caching. This works against both formtools 2.5.1 and 2.6.x, so no version pin is required.

This is the change applied in `cms/wizards/views.py`:

- Add a `get_form_list()` override that injects the dynamically-built step-2 form into the list returned by `super().get_form_list()`.
- Guard it with an instance flag against infinite recursion (building the step-2 form needs step 0's cleaned data, which re-resolves the form list).
- Drop the `self.form_list[step] = ...` mutation from `get_form()`.

Verified: `cms.tests.test_wizards` and `cms.tests.test_cms_config_wizards` pass on **both** formtools 2.5.1 and 2.6.1 with this change.

A PR with this fix will follow.

## Related

* https://github.com/jazzband/django-formtools/issues/295

fsbraun:
## Summary by Sourcery

Bug Fixes:
- Ensure dynamically generated second-step forms are correctly substituted when formtools caches `get_form_list` results, preventing issues with form caching in newer formtools versions.

## Related resources

* fixes #8628
* https://github.com/jazzband/django-formtools/issues/295

## Checklist

### DIFF
--- a/cms/wizards/views.py
+++ b/cms/wizards/views.py
@@ -60,6 +60,30 @@ class WizardCreateView(SessionWizardView):
             context['wizard_entry'] = self.get_selected_entry()
         return context
 
+    def get_form_list(self):
+        """
+        Resolve the wizard's form list.
+
+        Step ``'1'`` is declared with a placeholder ``Form`` because the real
+        form depends on the entry chosen in step ``'0'``. Once step 0 has been
+        completed, the real form is built and substituted here. Doing the
+        substitution in ``get_form_list`` -- instead of mutating the shared
+        ``self.form_list`` -- keeps the swap visible to django-formtools 2.6+,
+        which caches ``get_form_list`` results.
+        """
+        form_list = super().get_form_list()
+        # ``_building_step_2`` guards against infinite recursion: building the
+        # step-2 form needs step 0's cleaned data, which itself resolves the
+        # form list again.
+        if '1' in form_list and not getattr(self, '_building_step_2', False):
+            self._building_step_2 = True
+            try:
+                if self.get_cleaned_data_for_step('0') is not None:
+                    form_list['1'] = self.get_step_2_form()
+            finally:
+                self._building_step_2 = False
+        return form_list
+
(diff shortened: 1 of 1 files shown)


vinitkumar (approved):

PR #8636, opened 2026-05-27
fix: GrouperModelAdmin shadowed prepopulated_fields class attribute

### LINKED ISSUE #8634: BUG: GrouperModelAdmin snapshots prepopulated_fields at __init__, shadowing later class-level changes
## Summary

In `cms.admin.utils.GrouperModelAdmin`, `prepopulated_fields` is captured into an instance attribute at construction time and then overwritten as a side effect of `get_readonly_fields`. The instance attribute permanently shadows the class attribute, so any code that mutates `cls.prepopulated_fields` after the admin is instantiated (test setup, dynamic configuration, subclass `__init_subclass__`, etc.) is silently ignored.

This was introduced in #8471 (commit fd4aac9) — the bug fix itself is correct, but the implementation strategy creates a downstream-visible regression.

## The pattern today

`cms/admin/utils.py`, in `GrouperModelAdmin.__init__`:

```python
# Store prepopulated fields to be used in get_readonly_fields
self._prepopulated_fields = self.prepopulated_fields
```

and later in `GrouperModelAdmin.get_readonly_fields`:

```python
# Ensure no read-only fields are in prepopulated_fields
self.prepopulated_fields = {
    key: value for key, value in self._prepopulated_fields.items() if key not in fields
}
return fields
```

Because `ModelAdmin` instances are process-wide singletons (created once during `admin.site.register` / autodiscover), `self._prepopulated_fields` is whatever the class attribute happened to be at import time. The first call to `get_readonly_fields` (which Django invokes from `get_form` on every change/add view) then assigns `self.prepopulated_fields = {…}` — an *instance* attribute that masks the class attribute from that point on.

## Reproducer

```python
modeladmin = admin.site._registry[Poll]                # PollAdmin(GrouperModelAdmin)
modeladmin.__class__.prepopulated_fields = {"name": ("content__text",)}

# Anything that triggers get_form/get_readonly_fields (a request, a test client GET):
modeladmin.get_readonly_fields(request, obj)

modeladmin.prepopulated_fields            # {}  ← instance attribute, derived from {} captured at __init__
modeladmin.__class__.prepopulated_fields  # {"name": ("content__text",)}  ← ignored
```

## Downstream impact

This breaks `djangocms-versioning`'s `ExtendedGrouperVersionAdminMixin.get_prepopulated_fields`, which calls `super().get_prepopulated_fields()` and then filters the result against the rendered form's fields (to drop entries whose *dependencies* are readonly, which the django-cms filter does not check). Two regression tests start failing against django-cms `main`:

- `tests.test_admin.DefaultGrouperAdminTestCase.test_prepopulated_fields_excluded_when_readonly`
- `tests.test_admin.DefaultGrouperAdminTestCase.test_prepopulated_fields_add_change_add_sequence`

Both tests patch `modeladmin.__class__.prepopulated_fields` (the documented `ModelAdmin` extension point) and expect `get_prepopulated_fields` to see it. With the snapshot pattern, they don't — `super().get_prepopulated_fields()` returns the stale instance attribute (`{}`).

More generally, the pattern silently breaks any caller that assumes `ModelAdmin.prepopulated_fields` behaves like a normal class attribute, which is the standard Django contract.

## Proposed fix

Drop the snapshot and the side-effecting assignment in `get_readonly_fields`. Do the read-only filtering inside `get_prepopulated_fields()` (Django's documented hook for exactly this) and read the class attribute dynamically:

```python
def get_prepopulated_fields(self, request, obj=None):
    prepopulated = super().get_prepopulated_fields(request, obj)
    if not prepopulated:
        return prepopulated
    readonly = set(self.get_readonly_fields(request, obj))
    return {k: v for k, v in prepopulated.items() if k not in readonly}
```

Then remove `self._prepopulated_fields = self.prepopulated_fields` from `__init__` and the `self.prepopulated_fields = {…}` assignment from `get_readonly_fields`.

This:
- Keeps the original fix (#8471) intact: read-only keys are still stripped before `AdminForm` renders.
- Removes the instance-attribute shadowing of a documented Django class attribute.
- Uses Django's intended override point (`get_prepopulated_fields`) instead of side-effecting `get_readonly_fields`.
- Makes the existing checks (`GrouperModelAdminChecks._check_prepopulated_fields_*`) the single source of truth on the class attribute.

Happy to put up a PR if this direction looks right.

## Environment

- django-cms `main` (5.1.0a1, commit including fd4aac9)
- Django 6.2 (`main`)
- Surfaced via djangocms-versioning's test suite

fsbraun:
## Summary

Closes #8634.

`GrouperModelAdmin` snapshotted `prepopulated_fields` into `self._prepopulated_fields` at `__init__` time and then overwrote `self.prepopulated_fields` from inside `get_readonly_fields`. The instance attribute permanently shadowed the class attribute, so anything that mutated `cls.prepopulated_fields` after the admin was registered (test setup, dynamic configuration, downstream subclasses) was silently ignored.

This PR:

- Drops `self._prepopulated_fields = self.prepopulated_fields` from `GrouperModelAdmin.__init__`.
- Removes the side-effecting assignment to `self.prepopulated_fields` inside `get_readonly_fields`.
- Adds a `get_prepopulated_fields` override that reads `self.prepopulated_fields` dynamically and filters out keys that are currently read-only — the same outcome as #8471, but via the documented Django hook and without shadowing the class attribute.

The new `get_prepopulated_fields` is what `_changeform_view` calls right before constructing `AdminForm`, so read-only keys still get stripped before they could raise `KeyError`. Downstream subclasses can declare `prepopulated_fields` the normal way and `super().get_prepopulated_fields()` returns the live class-level value.

## Why

The snapshot pattern is observable in downstream code. Concretely, djangocms-versioning's `ExtendedGrouperVersionAdminMixin.get_prepopulated_fields` calls `super().get_prepopulated_fields()` and filters the result further (it also checks *dependencies*, not just keys). With the old pattern, the super call returned the empty snapshot captured at import time, regardless of what callers had set on the class. This regressed two djangocms-versioning tests against `main`:

- `tests.test_admin.DefaultGrouperAdminTestCase.test_prepopulated_fields_excluded_when_readonly`
- `tests.test_admin.DefaultGrouperAdminTestCase.test_prepopulated_fields_add_change_add_sequence`

Both pass again with this change.

## Test plan

- [x] `python manage.py test cms.tests.test_grouper_admin` — 67 tests pass.
- [x] `python manage.py test cms.tests.test_grouper_admin cms.tests.test_admin` — 111 tests pass.
- [x] djangocms-versioning regression tests (`DefaultGrouperAdminTestCase`) pass against this branch + Django `main`.
- [x] Updated `test_prepopulated_fields_exclude_readonly_fields` to assert on the new hook output (and that the class attribute is no longer mutated as a side effect).
- [x] Added `test_prepopulated_fields_respects_class_attribute_after_init` to lock in the dynamic behavior.

## Summary by Sourcery

Fix GrouperModelAdmin prepopulated field handling to respect dynamic class-level configuration and avoid mutating instance attributes in readonly field logic.

Bug Fixes:
- Ensure GrouperModelAdmin no longer snapshots and shadows prepopulated_fields so late-bound class-level changes are honored.
- Prevent KeyError by dynamically filtering prepopulated_fields to exclude fields that are currently read-only.

Tests:
- Update existing prepopulated_fields readonly test to assert on get_prepopulated_fields output without side effects on the class attribute.
- Add a regression test to verify prepopulated_fields respects class attribute changes made after admin instantiation.

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -19,6 +19,7 @@ jobs:
       with:
         node-version: ${{ env.NODE_VERSION }}
     - run: npm install
+    - run: rm -rf ~/.cache/puppeteer/chrome
     - run: npx puppeteer browsers install chrome
     - run: npm install -g gulp@latest
     - run: gulp unitTest
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -19,6 +19,7 @@ jobs:
       with:
         node-version: ${{ env.NODE_VERSION }}
     - run: npm install
+    - run: npx puppeteer browsers install chrome
     - run: npm install -g gulp@latest
     - run: gulp unitTest
     - run: gulp lint
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -19,11 +19,17 @@ jobs:
       with:
         node-version: ${{ env.NODE_VERSION }}
     - run: npm install
-    - name: Install Chrome for Testing and export its path
+    - name: Resolve Chrome binary and export its path
       run: |
-        rm -rf ~/.cache/puppeteer/chrome
-        CHROME_BIN="$(npx puppeteer browsers install chrome --format '{{path}}' | tail -n1)"
-        echo "Installed Chrome at: $CHROME_BIN"
+        # Prefer the Chrome that ships preinstalled on the runner image.
+        CHROME_BIN="$(command -v google-chrome || command -v google-chrome-stable || command -v chromium-browser || true)"
+        # Fall back to downloading Chrome for Testing via puppeteer.
+        if [ -z "$CHROME_BIN" ]; then
+          rm -rf ~/.cache/puppeteer/chrome
+          npx puppeteer browsers install chrome
+          CHROME_BIN="$(node -e "console.log(require('puppeteer').executablePath())")"
+        fi
+        echo "Using Chrome at: $CHROME_BIN"
         test -x "$CHROME_BIN"
         echo "CHROME_BIN=$CHROME_BIN" >> "$GITHUB_ENV"
     - run: npm install -g gulp@latest
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -19,8 +19,13 @@ jobs:
       with:
         node-version: ${{ env.NODE_VERSION }}
     - run: npm install
-    - run: rm -rf ~/.cache/puppeteer/chrome
-    - run: npx puppeteer browsers install chrome
+    - name: Install Chrome for Testing and export its path
+      run: |
+        rm -rf ~/.cache/puppeteer/chrome
+        CHROME_BIN="$(npx puppeteer browsers install chrome --format '{{path}}' | tail -n1)"
+        echo "Installed Chrome at: $CHROME_BIN"
+        test -x "$CHROME_BIN"
+        echo "CHROME_BIN=$CHROME_BIN" >> "$GITHUB_ENV"
     - run: npm install -g gulp@latest
     - run: gulp unitTest
     - run: gulp lint
--- a/cms/admin/utils.py
+++ b/cms/admin/utils.py
@@ -304,9 +304,6 @@ class GrouperModelAdmin(ChangeListActionsMixin, ModelAdmin):
                 dict(),
             )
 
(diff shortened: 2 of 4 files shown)


vinitkumar (approved):

Issue #8649, opened 2026-06-05
BUG: CMS_PLACEHOLDER_CONF plugin restrictions ignored in edit/structure mode (regression in 5.0.7)


rm-wholestack:
## Description

  `CMS_PLACEHOLDER_CONF` plugin restrictions are completely ignored in edit/structure mode
  when rendering CMS page placeholders. All registered plugins (including third-party ones
  like `TextPlugin`) are offered in the "Add plugin" menu regardless of the `plugins` list
  configured for the placeholder.

  ## Steps to reproduce

  1. Configure `CMS_PLACEHOLDER_CONF` with a restricted plugin list for a named placeholder:
     ```python
     CMS_PLACEHOLDER_CONF = {
         'pages/default.html header': {
             'plugins': ('HeaderVideoPublisher', 'HeaderImagePublisher'),
             'name': 'Header',
         },
     }
  2. Enter edit/structure mode on a CMS page using that template.
  3. Click "Add plugin" on the header placeholder. 
  4. Observe the plugin menu.

  Expected behaviour

  Only the plugins listed in CMS_PLACEHOLDER_CONF['plugins'] for that placeholder
  should appear — in this case only HeaderVideoPublisher and HeaderImagePublisher.

  Actual behaviour

  All registered plugins are shown, including third-party plugins (e.g. TextPlugin)
  that are not listed in the plugins restriction. The CMS_PLACEHOLDER_CONF restriction
  is effectively bypassed entirely.

  Root cause (identified via code diff)

  The regression was introduced in cms/plugin_rendering.py between 5.0.6 and 5.0.7.

  In the render_page_slot method, the call to render_placeholder was refactored to
  pass language=language but accidentally dropped page=current_page:

  5.0.6 (correct):
  content = self.render_placeholder(
      placeholder,
      context=context,
      page=current_page,   # ← page passed correctly
      editable=editable,
      use_cache=True,
      nodelist=None,
  )

  5.0.7 (broken):
  content = self.render_placeholder(
      placeholder,
      context=context,
      language=language,   # ← language added, but page dropped!
      editable=editable,   
      use_cache=True,
      nodelist=None,
  )   

  This causes the following chain of failures:

  1. render_placeholder(page=None) → get_editable_placeholder_context(page=None)
  2. → get_placeholder_toolbar_js(placeholder, page=None)
  3. → plugin_pool.get_all_plugins(placeholder.slot, page=None)
  4. → template = None (because page is None)
[...]

PR #8651, opened 2026-06-05
fix: template-specific CMS_PLACEHOLDER_CONF keys ignored when renderi…

rm-wholestack:
Description:
  When rendering a page placeholder, `render_placeholder()` was called with
  `page=placeholder.source` instead of `page=current_page`. This meant
  `get_template()` could not resolve the page template, so template-specific
  `CMS_PLACEHOLDER_CONF` keys (e.g. `'mytemplate.html myslot'`) were silently
  ignored and all plugins shown without restriction.

  Fix: pass `page=current_page` — already resolved as `page or self.current_page`
  on line 413.

  Related resources:
  * #8649

  Checklist — tick these three:
  - [x] I have opened this pull request against main
  - [x] I have added or modified the tests when changing logic
  - [x] I have followed the conventional commits guidelines to add meaningful information into the changelog

## Summary by Sourcery

Ensure template-specific placeholder configuration is respected when rendering page placeholders.

Bug Fixes:
- Fix placeholder rendering to use the current page so template-specific CMS_PLACEHOLDER_CONF keys are applied correctly.

Tests:
- Add regression test covering template-specific CMS_PLACEHOLDER_CONF behavior for placeholder extra context.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,7 @@ Features:
 
 Bug Fixes:
 ----------
+* Template-specific ``CMS_PLACEHOLDER_CONF`` keys were ignored when rendering page placeholders (#8649) -- Ralph Mueller
 * get_permissions failed for missing global permissions (#8468) (65d84a929) -- Fabian Braun
 * Create page wizard not available on empty install (#8463) -- Fabian Braun
 * Make Placeholder.add_plugin() set plugin.instance to self (#8442) (#8447) -- Jasper Bok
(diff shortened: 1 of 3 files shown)


fsbraun:
@rm-wholstack Thanks for the quick PR! It should (only) go against the `release/5.0.x` branch. The page argument was dropped accidentally, when back porting a bugfix PR from `main`.

The page argument should be populated by `current_page` in django CMS 5.0.

Also, no need to update the changelog. It is created automatically during the release process.

I appreciate your adding a test!

Can you reopen this PR against `release/5.0.x`? I will likely port forward the test.

rm-wholestack:
Reopened against `release/5.0.x` as requested — see #8652.

PR #8652, opened 2026-06-06
fix: template-specific CMS_PLACEHOLDER_CONF keys ignored when rendering page placeholders

### LINKED ISSUE #8649: BUG: CMS_PLACEHOLDER_CONF plugin restrictions ignored in edit/structure mode (regression in 5.0.7)

rm-wholestack:
Fixes #8649

Pass `page=current_page` to `render_placeholder()` inside `render_page_slot()` so that `get_placeholder_conf` can match template-scoped keys like `'mytemplate.html slot'`. The argument was accidentally dropped during a backport, causing all template-specific placeholder conf to be silently ignored.

Includes two regression tests:
- `test_placeholder_extra_context_template_specific` — verifies template-specific keys are respected
- `test_placeholder_extra_context_template_specific_takes_precedence` — verifies template-specific key wins over generic slot key

Reopens #8651 against the correct base branch.

## Summary by Sourcery

Ensure template-specific placeholder configuration is honored when rendering page placeholders and guard against regression with tests.

Bug Fixes:
- Restore passing the current page to placeholder rendering so template-scoped CMS_PLACEHOLDER_CONF keys are correctly applied.

Tests:
- Add regression tests verifying template-specific CMS_PLACEHOLDER_CONF entries are respected and take precedence over generic slot keys.

### DIFF
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -447,6 +447,7 @@ class ContentRenderer(BaseRenderer):
                 placeholder,
                 context=context,
                 language=language,
+                page=current_page,
                 editable=editable,
                 use_cache=True,
                 nodelist=None,
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -447,7 +447,7 @@ class ContentRenderer(BaseRenderer):
                 placeholder,
                 context=context,
                 language=language,
-                # page=current_page,
+                page=current_page,
                 editable=editable,
                 use_cache=True,
                 nodelist=None,
--- a/cms/plugin_rendering.py
+++ b/cms/plugin_rendering.py
@@ -447,7 +447,7 @@ class ContentRenderer(BaseRenderer):
                 placeholder,
                 context=context,
                 language=language,
-                page=current_page,
+                # page=current_page,
                 editable=editable,
                 use_cache=True,
                 nodelist=None,
--- a/cms/tests/test_rendering.py
+++ b/cms/tests/test_rendering.py
@@ -371,6 +371,31 @@ class RenderingTestCase(CMSTestCase):
             r = self.render(self.test_page4, template=t)
         self.assertEqual(r, self.test_data4['extra'])
 
+    def test_placeholder_extra_context_template_specific(self):
+        """Template-specific CMS_PLACEHOLDER_CONF keys must be respected (regression #8649)."""
+        t = '{% load cms_tags %}{% placeholder "extra_context" %}'
+        template_specific_conf = {
+            'extra_context.html extra_context': self.test_data4['placeholderconf']['extra_context'],
+        }
+        r = self.render(self.test_page4, template=t)
+        self.assertEqual(r, self.test_data4['no_extra'])
+        cache.clear()
+        with override_placeholder_conf(CMS_PLACEHOLDER_CONF=template_specific_conf):
+            r = self.render(self.test_page4, template=t)
+        self.assertEqual(r, self.test_data4['extra'])
+
+    def test_placeholder_extra_context_template_specific_takes_precedence(self):
+        """Template-specific CMS_PLACEHOLDER_CONF key takes precedence over generic slot key."""
+        t = '{% load cms_tags %}{% placeholder "extra_context" %}'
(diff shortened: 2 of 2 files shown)


cms/plugin_rendering.py
```diff
@@ -447,6 +447,7 @@ def render_page_placeholder(
                 placeholder,
                 context=context,
                 language=language,
+                page=current_page,
```
fsbraun:
```suggestion
                # page=current_page,
```

cms/plugin_rendering.py
```diff
@@ -447,6 +447,7 @@ def render_page_placeholder(
                 placeholder,
                 context=context,
                 language=language,
+                # page=current_page,
```
fsbraun:
```suggestion
                page=current_page,
```

fsbraun (approved):
@rm-wholestack Thank you! Looks great! Tests pass only with the fix. Good to go!

PR #8666, opened 2026-06-10
fix: Minor fixes of long-standing bugs

fsbraun:
## Summary by Sourcery

Fix multiple long-standing permission, placeholder configuration, admin view, and utility bugs across the CMS and extend tests to guard against regressions.

Bug Fixes:
- Ensure page delete permission checks correctly evaluate plugin delete permissions for all placeholders on a page.
- Fix permission cache invalidation so user and group signal handlers work with custom `AUTH_USER_MODEL` and use correct dispatch identifiers.
- Correct placeholder configuration inheritance to resolve through chained inherit keys without mutating base configs or leaking unrelated settings.
- Adjust page admin title edit and translation delete views to raise 404 before permission checks when the target object does not exist.
- Update plugin media path resolution to rely on the placeholder’s page relationship and avoid legacy fallback logic.
- Fix page permission placeholder lookup to use the `PageContent` content type and the correct loop variable when checking delete-plugin permissions.
- Correct utility check messaging methods to call the proper wrapper APIs and fix a typo in the skip path.
- Fix template tag editable context generation so the `attribute_name` is derived safely from `edit_fields` only when it is a non-string iterable.
- Correct the configuration decorator to return the wrapper function instead of the original function so the decoration is effective.

Tests:
- Add regression tests verifying placeholder permission checks when deleting pages with plugins.
- Add regression tests ensuring permission cache invalidation is triggered on user save/delete and group membership changes, including with custom user models.

## Related resources

* #...
* #...

## Checklist


vinitkumar (approved):
LGTM

PR #8667, opened 2026-06-10
fix: Corrupted plugin positions are now automatically healed

### LINKED ISSUE #8665: [BUG] IntegrityError (duplicate position) in `Placeholder._recalculate_plugin_positions` when plugin positions contain gaps (PostgreSQL)
## Description

`Placeholder._recalculate_plugin_positions()` compacts the per
`(placeholder, language)` plugin positions back to `1..n` with a single
statement (postgresql/sqlite branch):

```sql
UPDATE cms_cmsplugin
SET position = subquery.new_pos
FROM (
  SELECT id, ROW_NUMBER() OVER (ORDER BY position, id) AS new_pos
  FROM cms_cmsplugin WHERE placeholder_id=%s AND language=%s
) subquery
WHERE cms_cmsplugin.id = subquery.id
```

PostgreSQL enforces the `(placeholder_id, language, position)` unique constraint **per row, while the statement is executing**, and the row processing order is whatever the query plan yields (typically heap order
when the planner hashes the filtered subquery and drives the update from a seq scan of `cms_cmsplugin`). The method's docstring acknowledges it requires callers to have shifted plugins out far enough beforehand — and when positions are dense that contract holds. But as soon as the position sequence contains a *gap*, a row can be assigned a position that another, not-yet-processed row still holds, and the statement fails:

```
IntegrityError: duplicate key value violates unique constraint
"cms_cmsplugin_placeholder_id_language_position_bad60ebe_uniq"
DETAIL:  Key (placeholder_id, language, "position")=(2822, en, 37) already exists.
```

Whether it fails depends on the chosen plan and the physical row order, so in production it is intermittent and page-dependent, which makes it very confusing to diagnose.

Gaps occur in real installations:

1. Third-party code deleting plugins with plain `plugin.delete()` instead of `placeholder.delete_plugin()` — e.g. djangocms-text's `AbstractText.clean_plugins()` does this on every text save that removes a child plugin (link/image deleted in the editor, or a cancelled plugin dialog via `revert_on_cancel`). Nothing on `CMSPlugin` prevents or heals this. (We are reporting that separately against djangocms-text.)
2. django CMS's own edit flow is not atomic: `PlaceholderAdmin.edit_plugin` / `add_plugin` (unlike `move_plugin` / `copy_plugins`) carry no `transaction.atomic`, and `Placeholder.add_plugin()` runs shift → `instance.save()` → compaction as separate autocommit statements. Any failure after the shift leaves the shift committed, i.e. a large gap.
3. `PluginAddValidationForm` does not clamp the client-supplied `plugin_position` to the end of the tree (it only validates it against the parent's position), so a stale value from the frontend creates a trailing gap.

These two defects amplify each other into a vicious circle: once a gap exists, editing any plugin that is not last in the placeholder funnels through `CMSPluginBase.save_model` → `Placeholder.add_plugin` → `_recalculate_plugin_positions`, which now randomly raises the IntegrityError — and because the view is not atomic, the already-committed shift from that failed edit widens the gap further. Affected pages degrade until editors cannot save at all, and the data needs manual repair. (Note the unique constraint means duplicates can never exist *at rest*; gaps are the only persistent corruption, the duplicates are transient, mid-statement self-collisions.)

## Steps to reproduce

ORM level (no third-party plugins needed; `Placeholder.add_plugin` does not validate the plugin type):

```python
from cms.models import CMSPlugin, Placeholder

ph = Placeholder.objects.create(slot="content")
for pos in range(1, 7):
    CMSPlugin.objects.create(placeholder=ph, plugin_type="TextPlugin", language="en", position=pos)

# 1. A gap appears (third-party plugin.delete(), or a previous request that
#    died between the shift and the compaction in Placeholder.add_plugin):
CMSPlugin.objects.get(placeholder=ph, position=2).delete()

# 2. Edit any plugin that is not last in the placeholder — the admin funnels
#    plugin edits (change=True) through Placeholder.add_plugin:
plugin = CMSPlugin.objects.get(placeholder=ph, position=4)
ph.add_plugin(plugin)   # intermittently raises IntegrityError on PostgreSQL
```

Whether step 2 raises depends on the query plan, so here is a deterministic distillation that reproduces the mechanism in seconds on PostgreSQL (15.7 used below). It recreates the production shape — a large table, the compaction filtered to one small group — which makes the planner hash the subquery and drive the UPDATE from a seq scan in heap order:

```sql
CREATE TABLE demo (
  id serial PRIMARY KEY,
  placeholder_id int NOT NULL,
  language varchar(15) NOT NULL,
  position smallint NOT NULL,
  UNIQUE (placeholder_id, language, position)
);

-- background data: 300 dense placeholders
INSERT INTO demo (placeholder_id, language, position)
SELECT ph, 'en', pos FROM generate_series(1, 300) ph, generate_series(1, 20) pos;

-- placeholder 9999: gap at position 2 (positions 1, 3..40); higher positions
-- live in newer heap tuples, as naturally happens when plugins were inserted
-- mid-tree over time (every insert rewrites the shifted rows)
INSERT INTO demo (placeholder_id, language, position)
SELECT 9999, 'en', pos FROM generate_series(40, 3, -1) pos;
INSERT INTO demo (placeholder_id, language, position) VALUES (9999, 'en', 1);
ANALYZE demo;

-- the exact statement from Placeholder._recalculate_plugin_positions:
UPDATE demo
SET position = subquery.new_pos
FROM (
  SELECT id, ROW_NUMBER() OVER (ORDER BY position, id) AS new_pos
  FROM demo WHERE placeholder_id=9999 AND language='en'
) subquery
WHERE demo.id = subquery.id;
```

Observed output (PostgreSQL 15.7):

```
ERROR:  duplicate key value violates unique constraint "demo_placeholder_id_language_position_key"
DETAIL:  Key (placeholder_id, language, "position")=(9999, en, 39) already exists.
```

with this plan (Seq Scan on the outer side = heap-order row processing):

```
Update on demo
  ->  Hash Join
        Hash Cond: (demo.id = subquery.id)
        ->  Seq Scan on demo
        ->  Hash
              ->  Subquery Scan on subquery
                    ->  WindowAgg
                          ->  Sort (position, id)
                                ->  Bitmap Heap Scan on demo demo_1 (placeholder_id = 9999 ...)
```

The same statement *succeeds* when the planner happens to drive the update from the position-sorted subquery instead — that plan/order dependence is exactly why production failures look random.

## Expected behaviour

Re-saving a plugin compacts the placeholder's positions back to a dense `1..n` without errors, regardless of pre-existing gaps — especially since gaps can be produced by django CMS's own non-atomic edit flow. A failed compaction must not leave the placeholder more corrupted than before.

## Actual behaviour

Editing a plugin in an affected placeholder intermittently 500s with the IntegrityError above, and each failure widens the gap (the preceding shift UPDATE from `Placeholder.add_plugin` stays committed because `PlaceholderAdmin.edit_plugin` is not atomic). Affected pages eventually become un-editable until positions are repaired by hand.

Production traceback (django CMS 5.0.6, plugin edited inside a TextPlugin):

```
  File "cms/admin/placeholderadmin.py", line 666, in edit_plugin
    response = plugin_instance.change_view(request, str(plugin_id))
  ...
  File "django/contrib/admin/options.py", line 1895, in _changeform_view
    self.save_model(request, new_object, form, change)
  File "djangocms_text/cms_plugins.py", line 734, in save_model
    super().save_model(request, obj, form, change)
  File "cms/plugin_base.py", line 591, in save_model
    self.saved_object = pl.add_plugin(obj)
  File "cms/models/placeholdermodel.py", line 495, in add_plugin
    self._recalculate_plugin_positions(instance.language)
  File "cms/models/placeholdermodel.py", line 729, in _recalculate_plugin_positions
    cursor.execute(sql, [self.pk, language])
  ...
IntegrityError: duplicate key value violates unique constraint
"cms_cmsplugin_placeholder_id_language_position_bad60ebe_uniq"
DETAIL:  Key (placeholder_id, language, "position")=(2822, en, 37) already exists.
```

## Screenshots

Not applicable — tracebacks and SQL output above.

* django CMS 5.0.6 — the statement is unchanged on `main` and
  `release/5.0.x` as of 2026-06-10
* Django 5.2.15, Python 3.14, PostgreSQL 15.7
* djangocms-text 0.9.8 (the gap source in our installation),
  djangocms-versioning 2.5.1
* SQLite shares the same single-statement branch and also enforces
  uniqueness per row during UPDATE; the MySQL/Oracle branch likely has the
  same hazard but was not tested.

**Suggested fix** (running in our production as a monkeypatch; happy to turn it into a PR): replace the single-statement compaction with a two-phase, order-independent one — park every row of the group at the *negative* of its final position, then flip the sign, inside one transaction:

```python
park_sql = (
    "UPDATE {0} "
    "SET position = -subquery.new_pos "
    "FROM ("
    "  SELECT id, ROW_NUMBER() OVER (ORDER BY position, id) AS new_pos "
    "  FROM {0} WHERE placeholder_id=%s AND language=%s "
    ") subquery "
    "WHERE {0}.id=subquery.id"
).format(table)
flip_sql = "UPDATE {0} SET position = -position WHERE placeholder_id=%s AND language=%s AND position < 0".format(table)

with transaction.atomic(using=db_alias):
    cursor.execute(park_sql, [self.pk, language])
    cursor.execute(flip_sql, [self.pk, language])
```

Phase 1 cannot collide because all live rows in the group hold positive positions while every target is negative; phase 2 cannot collide because all rows in the group are negative while every target is positive; targets are distinct in both phases. It is therefore safe under *any* pre-existing state, needs no schema change (`position` is a signed `SmallIntegerField` with no check constraint, and intermediate values are bounded by the plugin count), and turns every successful edit into a self-heal of existing gaps.

Alternatives/complements: make the unique constraint `DEFERRABLE INITIALLY DEFERRED` on PostgreSQL (schema migration), and wrap `PlaceholderAdmin.edit_plugin` / `add_plugin` in `transaction.atomic` like `move_plugin` already is, so a failed save can no longer commit a partial shift.

fsbraun:
## Summary by Sourcery

#8665 introduced the idea of self-healing plugin position recalculations. This PR  ensures plugin position recalculation is **collision-free and self-healing** when adding, moving, copying, or deleting plugins, **even in the presence of corrupted or gapped position data**.

Main driver here is, that not only the django CMS core needs to be hardened, but also custom plugins might cause inconsistencies (see example in #8665). The core can correct them.

Together with optimizing the move processes, these are cost-neutral for all operations except plugin move across placeholders (since it "heals" two placeholders there), which now is 9 transactions as opposed to 8 before.

Bug Fixes:
- Prevent `IntegrityError` and duplicate position issues during plugin position recalculation across supported databases when gaps or corrupted positions exist.
- Correct plugin move logic within and across placeholders so that shifts use a single safe offset and final compaction heals pre-existing gaps.
- Fix `add_plugin` behavior to correctly park and squash plugin positions based on a known safe upper-bound offset, avoiding collisions.

Enhancements:
- Refine `_recalculate_plugin_positions` to use a two-phase park-and-unpark strategy with an optional caller-provided upper bound, improving robustness and avoiding extra queries when possible.
- Restrict recalculation SQL to known-supported database backends with clearer error signaling for others.
- Wrap add, edit, and copy plugin admin operations in atomic transactions to ensure plugin tree consistency on failure.

Tests:
- Add regression tests covering recalculation with gaps and reversed ordering, move operations that must heal pre-existing gaps, and cross-placeholder moves that compact both source and target placeholders.

## Related resources

* fixes #8665 
* #...

## Checklist


vinitkumar (approved):
This is some smart code. Good use of @transaction/atomic to prevent corruptions.

PR #8675, opened 2026-06-16
feat: Allow wildcards (*, ?, []) for plugin restrictions, and "auto" `child_classes` detection

fsbraun:
## Summary by Sourcery

Extend plugin parent/child restriction handling to support glob-style patterns and automatic child detection, and simplify frontend droppability checks to rely solely on child restrictions.

New Features:
- Allow glob-style patterns in plugin `child_classes` and `parent_classes` to match registered plugin names, such as `"Theme*"` (anything starting with "Theme")
- Introduce a special `child_classes="auto"` mode that automatically permits plugins which declare the current plugin as a parent.

Enhancements:
- Clarify and enforce the semantics of empty `child_classes` and `parent_classes` lists as disallowing all children/parents.
- Stop sending parent restriction data to the frontend and derive droppability exclusively from child (bounds) restrictions, with server-side APIs retaining `parent_classes` for backwards compatibility.
- Cache and reuse expanded plugin restriction patterns for efficient evaluation.

Documentation:
- Update plugin restriction and configuration documentation to describe wildcard support, "auto" child detection, and new restriction semantics.
- Add upgrade documentation for 5.1.0 explaining the change in parent restriction handling and new wildcard features.

Tests:
- Add backend and frontend tests covering glob expansion, "auto" child detection, empty restriction semantics, and the removal of frontend parent restriction handling.

## Related resources

* #...
* #...

## Checklist


cms/models/pluginmodel.py
```diff
@@ -279,11 +280,21 @@ def get_plugin_info(self, children=None, parents=None):
             'plugin_language': self.language or '',
             'plugin_parent': self.parent_id or '',
             'plugin_restriction': children or [],
-            'plugin_parent_restriction': parents or [],
             'is_slot': plugin_class.is_slot,
             'disable_child_plugins': plugin_class.disable_child_plugins,
             'urls': self.get_action_urls(),
         }
+        if parents is not None:
```
jrief, line 290: I actually would use `hasattr(parents, '__iter__')` here to explicitly test for a list or tuple. In the future we might use magic values for `parent_classes`. In addition to that, line 296 can be simplified.

cms/plugin_base.py
```diff
@@ -777,7 +786,10 @@ def get_child_classes(
         # Get all child plugin candidates
         installed_plugins = cls.get_child_plugin_candidates(slot, page)
 
-        if child_classes:
+        if child_classes is not None:
```
jrief: I actually would use `hasattr(child_classes, '__iter__')` here to explicitly test for a list or tuple. In the future we might use magic values for `child_classes`. In addition to that, line 794 is safer, because it expects an iterable.

jrief (changes requested):

cms/models/pluginmodel.py
fsbraun: This is a check if a `parent` argument has been provided to this method. The method ignores it and issues a warning.

cms/plugin_base.py
fsbraun: I think your suggestion #8678 should go here.

jrief (commented):
I just tested the `child_classes = "auto"` feature in djangocms-cascade on a dozen of plugins and it works as expected.

However, now the attribute `allow_children` is redundant. If it's set to `True` and a plugin does not have children, the “+” sign now is muted as expected (bugfix). But then, setting this to `False` does not really make sense unless someone wants the explicitly prevent third party plugins to become children of its own plugin.

fsbraun:
> However, now the attribute `allow_children` is redundant. If it's set to `True` and a plugin does not have children, the “+” sign now is muted as expected (bugfix). But then, setting this to `False` does not really make sense unless someone wants the explicitly prevent third party plugins to become children of its own plugin.

Yes, we need to keep it for exactly that reason: Some people might set it to avoid having child plugins.

jrief (approved):

PR #8692, opened 2026-06-23
fix: Missing authorization in `render_object_structure` disclosed non-PageContent placeholder structure

fsbraun:
## Summary by Sourcery

Enforce object-level authorization for the render_object_structure endpoint to protect placeholder structures of non-PageContent objects.

Bug Fixes:
- Prevent staff users without change permissions from accessing placeholder/plugin structures of non-PageContent objects via the render_object_structure endpoint.

Enhancements:
- Introduce a reusable helper to check whether a user may change placeholders attached to a given object, including support for custom placeholder change permission hooks.

Tests:
- Add tests ensuring the structure endpoint denies unauthorized staff, and allows properly permissioned staff and superusers, for non-PageContent objects using placeholder relation fields.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/cms/cms_toolbars.py
+++ b/cms/cms_toolbars.py
@@ -26,7 +26,9 @@ from cms.utils.i18n import get_language_dict, get_language_tuple
 from cms.utils.page_permissions import (
     user_can_change_page,
     user_can_delete_page,
+    user_can_view_page,
 )
+from cms.utils.permissions import user_can_view_placeholder_source
 from cms.utils.urlutils import add_url_parameters, admin_reverse
 from menus.utils import DefaultLanguageChanger
 
@@ -157,6 +159,17 @@ class PlaceholderToolbar(CMSToolbar):
             return False
         return any(ph for ph in self.placeholders if ph.has_change_permission(self.request.user))
 
+    def _has_page_view_perm(self):
+        if self.page and user_can_view_page(self.request.user, self.page):
+            return True
+        return False
+
+    def _has_placeholder_view_perm(self):
+        obj = self.toolbar.obj
+        if obj is None:
+            return False
+        return user_can_view_placeholder_source(self.request.user, obj)
+
     def _can_add_button(self):
         if self._has_page_change_perm():
             return True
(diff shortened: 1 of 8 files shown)


marksweb:
@fsbraun That Sourcery point got me curious and there's two potential things that could also be addressed.

1. Medium risk: `render_object_structure()` still only checks `user_can_view_page()` for `PageContent` objects, so a staff user who can view a page but cannot change it can still fetch the structure endpoint directly. That is weaker than the toolbar's own gate, which only offers structure mode when the user can change the page or one of its placeholders (`cms/cms_toolbars.py:167-175`). The new guard closes the non-page-object leak, but the page-backed path still has a direct-URL bypass at `cms/views.py:304-316`.

2. Low risk: The new `_can_change_placeholder_object()` helper in `cms/views.py:330-343` duplicates the permission logic already on `Placeholder.has_change_permission()` in `cms/models/placeholdermodel.py:94-111`. That gives you two places to keep in sync if placeholder authorization changes later, which is the drift the sourcery comment was warning about.

fsbraun:
@marksweb Thanks, both good catches — they made me rethink the model. Here's the result (pushed in a13f5605d):
1. On point 1 (page branch only checks `user_can_view_page`): you're right that it's weaker than the toolbar's change-gate — but I think the fix is to make the toolbar consistent with the endpoint, not the other way round. Structure view is a read affordance: for  **headless setups** a reviewer who can view a page should be able to pull up its structure board to inspect content, even without change rights. So the endpoint requiring only view permission is intentional.
   - So I loosened _can_add_structure_mode to gate on view permission (`_has_page_view_perm` / `_has_placeholder_view_perm`) instead of change — the `cms.use_structure` requirement stays.
   - The board renders read-only when the user can't change it: `placeholder_is_immutable` also requires `has_change_permission`, so view-only users get the `cms-drag-disabled` / "Placeholder not editable" board (mutations were already gated at the plugin endpoints; this just stops showing edit affordances that would fail).

   Net: view-only/headless users can see structure but not edit it; restricted pages are still denied by `user_can_view_page`.
  
2. On point 2 (duplicated permission logic): dissolved it. The non-PageContent branch now  calls a shared `user_can_view_placeholder_source()` util (single source of truth, honors a `has_placeholder_view_permission` hook, falls back to model/object view-or-change), and the read-only board logic reuses Placeholder.has_change_permission. No more parallel copy to keep in sync.

To seal this up, I added tests that confirm the can view but cannot change situation (structure board is rendered read-only).

PR #8700, opened 2026-06-27
fix: GrouperAdmin only showed current content versions

fsbraun:
## Summary by Sourcery

Improve grouper and page content admin behavior around outdated content versions and language switching.

Bug Fixes:
- Prevent language switching on outdated content from silently navigating to the latest version instead of the currently viewed version by hiding language selectors for non-latest content.

Enhancements:
- Allow selecting and displaying a specific content object (including non-latest versions) in the grouper change view via a dedicated GET parameter and a content change redirect URL.
- Expose a stable admin change URL for content models that redirects to the unified grouper change form when the content model is not separately registered.
- Add a hook to detect whether a content object is the latest version for its grouper and current grouping fields to control language selector visibility.

Tests:
- Add regression tests for grouper admin content selection, content change redirect behavior, and language selector visibility for latest vs non-latest content.
- Add tests for page content change views to assert language tabs are shown only for the latest content.

This PR fixes two issues:

* Unexpected user experience when switching languages on an outdated page version: Going back to original language does not return to the viewed content but to the latest content
* GrouperAdmin did not allow for viewing specific outdated content objects.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -954,7 +954,13 @@ class PageContentAdmin(PageDeleteMessageMixin, admin.ModelAdmin):
             "language_tabs": get_language_tuple(site.pk),
             "filled_languages": self.get_filled_languages(request, obj.page),
         }
-        context["show_language_tabs"] = len(context["language_tabs"])
+        # Only offer the language selector for the latest content. Switching the language always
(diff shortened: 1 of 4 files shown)


vinitkumar (approved):

PR #8708, opened 2026-06-29
chore: Update test and compatibility setup for django >= 5.2

### LINKED ISSUE #8702: Windows test failure: test_filters_date uses unsupported strftime("%-d")
### Summary

Running the django CMS test suite on Windows fails in `cms.tests.test_toolbar.EditModelTemplateTagTest.test_filters_date`.

The failure happens because the test uses `strftime("%-d")` when building the expected formatted date string. The `%-d` directive is supported on POSIX platforms, but it is not supported by the Windows C runtime, so the test raises `ValueError: Invalid format string`.

This appears to be a test portability issue, not an application behavior bug.

### Environment

* OS: Windows
* Python: 3.13.13
* Django: 5.2.15
* django CMS: 5.1.0a1
* Branch: `main`
* Commit tested: `db9a163`

### Reproduction

```powershell
python -m venv .venv
.\.venv\Scripts\python.exe -m pip install -r test_requirements\django-5.2.txt
.\.venv\Scripts\python.exe manage.py test cms.tests.test_toolbar.EditModelTemplateTagTest.test_filters_date -v 2
```

### Expected behavior

The test should pass on Windows.

### Actual behavior

The test fails with:

```text
ValueError: Invalid format string
```

The failure occurs while constructing the expected assertion value.

### Root cause

`cms/tests/test_toolbar.py` builds the expected formatted date using:

```python
ex1.date_field.strftime("%b. %-d, %Y")
```

The `%-d` directive is platform-specific and is not supported on Windows. Because of that, the test fails before it can complete the assertion.

### Suggested fix

Use Django’s date formatting utilities instead of platform-specific `strftime`, for example:

```python
from django.utils.formats import date_format
```

and then format the expected value with Django’s configured date formatting behavior.

This would make the test portable across platforms while keeping the assertion aligned with Django’s own date formatting.

### Notes

I could not find an existing open issue or PR covering this specific Windows failure.

If this approach looks reasonable, I'd be happy to prepare a focused PR to make the test portable across platforms while keeping the existing behavior unchanged.

fsbraun:
## Summary by Sourcery

Raise the minimum supported Django version to 5.2 and remove legacy compatibility paths and workarounds, updating tests and CI to match.

New Features:
- Gracefully skip adding the sample2-root navigation node in StaticMenu4 when its URL cannot be reversed, avoiding noisy menu errors.

Enhancements:
- Simplify toolbar lazy value handling by standardizing on the modern Promise API.
- Clean up compatibility utilities by retaining version markers only for backward import compatibility and aligning sitemap protocol handling with current defaults.

Build:
- Raise the core dependency requirement to Django >= 5.2 and refresh package classifiers to reflect supported Django versions.

CI:
- Adjust GitHub Actions test matrix to drop Django 4.2 runs and align supported Python/Django combinations with Django 5.2+.
- Update release-test and migration-test workflows to test only against Django 5.2 and newer.

Tests:
- Update admin and toolbar tests to assert the current error markup, input field attributes, and standardized URL protocol.
- Remove tests and plugin hooks that targeted historical Django template type-hinting bugs fixed in Django 5.2.

## Related resources

* fixes #8702 
* #...

## Checklist

### DIFF
--- a/cms/utils/compat/__init__.py
+++ b/cms/utils/compat/__init__.py
@@ -6,7 +6,8 @@ from packaging.version import Version
 DJANGO_VERSION = get_version()
 PYTHON_VERSION = python_version()
 
-# These means "less than or equal to DJANGO_FOO_BAR"
+# These mean "less than or equal to DJANGO_FOO_BAR"
+# Kept for backward compatibility, none of them used currently
 DJANGO_2_2 = Version(DJANGO_VERSION) < Version('3.0')
 DJANGO_3_0 = Version(DJANGO_VERSION) < Version('3.1')
 DJANGO_3_1 = Version(DJANGO_VERSION) < Version('3.2')
@@ -15,3 +16,5 @@ DJANGO_3 = Version(DJANGO_VERSION) < Version('4.0')
 DJANGO_4_1 = Version(DJANGO_VERSION) < Version('4.2')
 DJANGO_4_2 = Version(DJANGO_VERSION) < Version('4.3')
 DJANGO_5_1 = Version(DJANGO_VERSION) < Version('5.2.dev')  # To allow testing against django's main branch
+
+# Currently used version markers:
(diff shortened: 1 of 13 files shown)


cms/utils/compat/__init__.py
```diff
@@ -15,3 +16,5 @@
 DJANGO_4_1 = Version(DJANGO_VERSION) < Version('4.2')
 DJANGO_4_2 = Version(DJANGO_VERSION) < Version('4.3')
 DJANGO_5_1 = Version(DJANGO_VERSION) < Version('5.2.dev')  # To allow testing against django's main branch
```
fsbraun:
```suggestion
DJANGO_5_1 = Version(DJANGO_VERSION) < Version('5.2')  # To allow testing against django's main branch
```

vinitkumar (approved):

PR #8727, opened 2026-07-16
fix: Avoid N+1 queries when copying pages or plugins between languages

vinitkumar:
## Summary

   Fix N+1 query patterns when copying plugins between page languages.

   ## Problem

   Both copy paths queried related data inside the placeholder loop:

   - `cms/api.py::copy_plugins_to_language`
     - Queried target plugin existence once per placeholder.
     - Queried source plugins once per placeholder.

   - `cms/admin/pageadmin.py::copy_language`
     - Queried target placeholders once per source placeholder.
     - Queried source plugins once per source placeholder.

   For a page with `N` placeholders, these queries grew linearly with `N`, in addition to
 the work required to copy the plugins themselves.

   ## Evidence

   A regression test using two populated placeholders previously observed 38 queries
 during the copy operation.

   The new API regression test verifies that source plugins are loaded with one batched
 query using:

   ```sql
   WHERE language = 'en'
   AND placeholder_id IN (...)
 ```

 instead of one query per placeholder.

 Changes

 - Load all source placeholders before iterating.
 - Load all target placeholders into a slot map.
 - Load all source plugins in one query.
 - Group plugins by placeholder ID in memory.
 - Batch target plugin existence checks using an aggregate query.
 - Preserve missing-target-placeholder behavior in the admin path.

 Validation

 Passed:

 ```text
   27 Django tests
   ruff check
   git diff --check
 ```

 Relevant coverage includes:

 - API plugin language-copy query regression test.
 - Admin plugin language-copy tests.
 - Missing target placeholder behavior.
 - Full management command test module.

 The copy lang management command was not changed in this PR.
## Checklist

## Summary by Sourcery

Optimize plugin language-copy operations to avoid N+1 queries when copying placeholders between page languages.

New Features:
- Add an API regression test to ensure plugin language copy uses a single batched query for source plugins instead of per-placeholder queries.

Bug Fixes:
- Prevent per-placeholder database queries when copying plugins between languages in the public API path.
- Prevent per-placeholder database queries when copying plugins between languages in the admin copy-language action while preserving missing-placeholder warnings.

Enhancements:
- Refactor plugin copy-to-language logic to prefetch source placeholders, group plugins by placeholder, and batch target plugin existence checks.
- Refactor admin page language copy to prefetch source and target placeholders and reuse grouped plugins for more efficient copying.

Tests:
- Add query-count regression coverage for copy_plugins_to_language to guard against reintroducing N+1 query patterns.
- Extend admin language-copy tests to cover missing target placeholder behavior under the new query strategy.

### DIFF
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -84,7 +84,7 @@ from cms.utils.i18n import (
     get_site_language_from_request,
 )
 from cms.utils.permissions import clear_permission_lru_caches
-from cms.utils.plugins import copy_plugins_to_placeholder
+from cms.utils.plugins import copy_plugins_to_placeholder, downcast_plugins
 from cms.utils.urlutils import admin_reverse, static_with_version
 
 require_POST = method_decorator(require_POST)
@@ -1240,10 +1240,12 @@ class PageContentAdmin(PageDeleteMessageMixin, admin.ModelAdmin):
             placeholder.slot: placeholder
             for placeholder in target_page_content.get_placeholders()
         }
-        source_plugins = CMSPlugin.objects.filter(
-            placeholder_id__in=[placeholder.pk for placeholder in source_placeholders],
-            language=source_page_content.language,
-        ).order_by("position")
+        source_plugins = downcast_plugins(
+            CMSPlugin.objects.filter(
+                placeholder_id__in=[placeholder.pk for placeholder in source_placeholders],
+                language=source_page_content.language,
+            ).order_by("position")
+        )
         plugins_by_placeholder = defaultdict(list)
         for plugin in source_plugins:
             plugins_by_placeholder[plugin.placeholder_id].append(plugin)
@@ -1257,7 +1259,12 @@ class PageContentAdmin(PageDeleteMessageMixin, admin.ModelAdmin):
 
             if not target.has_add_plugins_permission(request.user, plugins):
                 return HttpResponseForbidden(_("You do not have permission to copy these plugins."))
-            copy_plugins_to_placeholder(plugins, target, language=target_language)
+            copy_plugins_to_placeholder(
+                plugins,
+                target,
+                language=target_language,
+                plugins_are_downcast=True,
+            )
         return HttpResponse("ok")
 
     def delete_view(self, request, object_id, extra_context=None):
--- a/cms/admin/pageadmin.py
+++ b/cms/admin/pageadmin.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 import re
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 
 import django
 from django.conf import settings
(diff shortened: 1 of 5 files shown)


fsbraun:
@vinitkumar Great catch! Looking into the code base I see a similar issue in `cms.api.copy_plugins_to_language`. It seems these are the only two places where multiple placeholder are copied.

`Placeholder.get_plugins_list()` does not return downcast plugins—it only materializes the base `CMSPlugin` queryset. The downcasting currently happens inside `copy_plugins_to_placeholder()` via `get_bound_plugins()`.

Even with the proposed changes, the base plugins are fetched in one query, but `copy_plugins_to_placeholder()` is still called separately for every placeholder. Each call can therefore query the concrete plugin tables again, leaving an N+1 pattern per placeholder/plugin type. (The new regression test only looks at queries against `cms_cmsplugin`, so queries against concrete tables such as the text plugin table are not detected.)

I propose both `cms.api.copy_plugins_to_language` and `cms.admin.pageadmin.PageAdmin.copy_language` should also batch the plugin downcasting across all source placeholders: **Could we fetch and downcast the plugins across all source placeholders using either `assign_plugins()` or `downcast_plugins()`, and then group the resulting concrete instances by placeholder?**

`copy_plugins_to_placeholder()` could receive an option indicating that its input is already downcast, allowing it to skip `get_bound_plugins()`. By default it would keep the current behavior and downcast itself. If the two methods above already provide downcast plugins, `get_bound_plugins()` would not be called avoiding another N+1 query.

It would also be useful for the regression test to verify the concrete plugin-table queries or the total query count, so that both the initial fetch and the downcasting remain batched.

What do you think?

vinitkumar:
> Great catch! Looking into the code base I see a similar issue in `cms.api.copy_plugins_to_language`. It seems these are the only two places where multiple placeholder are copied.
> 
> `Placeholder.get_plugins_list()` does not return downcast plugins—it only materializes the base `CMSPlugin` queryset. The downcasting currently happens inside `copy_plugins_to_placeholder()` via `get_bound_plugins()`.
> 
> Even with the proposed changes, the base plugins are fetched in one query, but `copy_plugins_to_placeholder()` is still called separately for every placeholder. Each call can therefore query the concrete plugin tables again, leaving an N+1 pattern per placeholder/plugin type. (The new regression test only looks at queries against `cms_cmsplugin`, so queries against concrete tables such as the text plugin table are not detected.)
> 
> I propose both `cms.api.copy_plugins_to_language` and `cms.admin.pageadmin.PageAdmin.copy_language` should also batch the plugin downcasting across all source placeholders: **Could we fetch and downcast the plugins across all source placeholders using either `assign_plugins()` or `downcast_plugins()`, and then group the resulting concrete instances by placeholder?**
> 
> `copy_plugins_to_placeholder()` could receive an option indicating that its input is already downcast, allowing it to skip `get_bound_plugins()`. By default it would keep the current behavior and downcast itself. If the two methods above already provide downcast plugins, `get_bound_plugins()` would not be called avoiding another N+1 query.
> 
> It would also be useful for the regression test to verify the concrete plugin-table queries or the total query count, so that both the initial fetch and the downcasting remain batched.
> 
> What do you think?

Let me check this out. Thanks for the detailed review.

vinitkumar:
@fsbraun Pushed a new commit last night. Please have a look and see if it does as per your suggestion.

fsbraun (approved):
@vinitkumar Looks great! Let's get this merged!

PR #8744, opened 2026-07-28
feat: Offer replacement for djangotree-beard dependency

fsbraun:
## Summary by Sourcery

django-treebeard deprecated its API on node objects in release 6.0. django-cms only uses a small subset of its features, node methods extensively, however.  The api change will remove many methods from the `Page` model which is a subclass of django-treebeards `MP_Node`. 

Introduce a configurable, dependency-free materialized-path tree backend (`MP_Node`) for the Page model and make django-treebeard optional based on a runtime backend selector.

Enhancements:
- Make Page, PageManager, and PageQuerySet use backend-selector helpers so treebeard is only imported when the treebeard backend is active.
- Add a MaterializedPath-based tree backend with a treebeard-compatible API (django-treebeard < 6) and identical field layout, allowing backend swaps without schema migrations.
- Update Page queryset deletion logic to keep numchild caches correct when using the new backend and to bypass tree-fixup logic when needed.
- Relax setup-time treebeard checks so the dependency is only required when the treebeard backend is configured.

Tests:
- Add extensive prototype, backend, and coverage tests to validate the new materialized-path backend, its parity with treebeard, and the swappable backend configuration.

## Related resources

* #...
* #...

## Checklist

## Summary by Sourcery

Introduce a configurable, dependency-free materialized-path tree backend for the Page model and make the tree backend swappable at runtime, keeping schema compatibility with django-treebeard.

New Features:
- Add a MaterializedPath-based tree backend and mixin with a treebeard-compatible API and field layout for Page and other tree models.
- Make the Page model, manager, and queryset select their tree backend and base classes at import time so django-treebeard becomes an optional dependency.
- Support backend selection via CMS_TREE_BACKEND, allowing deployments to switch between treebeard and the new mptree backend without migrations.

Bug Fixes:
- Ensure bulk Page queryset deletions correctly maintain parent numchild caches under the new tree backend and avoid underflow on unsigned columns.

Enhancements:
- Adjust PageQuerySet and PageManager to avoid importing treebeard when the mptree backend is active and to provide a backend-agnostic delete_fast path.
- Relax setup-time dependency checks so django-treebeard is only required when the treebeard backend is configured.
- Optimize tree operations to use set-based, database-friendly updates while preserving existing tree semantics and ordering.

CI:
- Run the django-main CI jobs against the mptree backend via CMS_TREE_BACKEND to validate the new backend in the primary test matrix.

Tests:
- Add extensive tests for the new materialized-path backend, including field parity with treebeard, Page tree behavior, queryset delete semantics, driver-level operations, invariants, and performance benchmarks.


fsbraun:
The tests **contain a benchmark versus django-treebeard** when run locally. On Postgres:
* Move: Roughly on par
* Fix-tree: 30% performance improvement (x1,5)

On mysql:
* Move: On par
* Fix-tree: 80% performance improvement (x5)

On sqlite:
* Move: on par
* fix-tree: 90% performance improvement (x10)

cms/utils/mptree.py
```diff
@@ -0,0 +1,735 @@
...
+
+    ``treebeard`` is imported lazily (only by the selectors below, only when this
+    returns ``"treebeard"``), so the ``mptree`` backend never imports it.
+    """
+    import os
+
+    from django.conf import settings
+
+    return getattr(settings, "CMS_TREE_BACKEND", None) or os.environ.get(
+        "CMS_TREE_BACKEND", "mp-tree"
```
PeterW-LWL: Shouldn't the default be `treebeard`?
```suggestion
        "CMS_TREE_BACKEND", "treebeard"
```

Otherwise the doc above is wrong.

In any case `mp-tree` is wrong.

fsbraun: @PeterW-LWL That's a good question, and honestly, I do not have a strong opinion:
1. `"treebeard"` is the safer choice since it is production-proven **if pinned below version 6**. With current versions it'll start throwing warnings, with an anticipated version 7 it will break
2. `"mptree"` is the sustainable choice since it will **keep the API of the Page object** 

For a quick (patch) release  the first option is better.

What are your thoughts?

pyproject.toml
```diff
@@ -44,7 +44,7 @@ dependencies = [
   "Django >= 5.2",
   "django-classy-tags >=0.7.2",
   "django-formtools >=2.1",
-  "django-treebeard >=4.3",
+#  "django-treebeard >=4.3",
```
fsbraun:
```suggestion
  "django-treebeard >=4.3",
```
undo temp  change

cms/utils/mptree.py
PeterW-LWL: > For a quick (patch) release the first option is better.

I'm fully supporting this.

---

In which version is this change expected to land?
- 5.0 
- 5.1 
- 5.2 
- 6.0 

I think for version 5.0 keeping "treebeard" as the default as this is the LTS version and after that "mptree".

Edit: after tinking about this a little more 5.1 might also keep "treebeard" as the default because it is already released

cms/utils/mptree.py
```diff
@@ -0,0 +1,735 @@
...
+            type(self)._base_manager.filter(pk=parent_id).update(
+                numchild=models.F("numchild") - 1
+            )
+        return result
+
+
+def get_tree_backend() -> str:
+    """
+    The active page-tree backend: ``"treebeard"`` (default) or ``"mptree"``,
+    from the ``CMS_TREE_BACKEND`` setting. An env var of the same name overrides
```
PeterW-LWL, line 732:
```suggestion
    The active page-tree backend: ``"treebeard"`` or ``"mptree"`` (default),
    from the ``CMS_TREE_BACKEND`` setting. An env var of the same name overrides
```

cms/utils/mptree.py
fsbraun: It's either 5.1 or whatever comes next. The default to treebeard in 5.0 and 5.1 and mptree later is a good choice.

cms/utils/mptree.py
```diff
@@ -0,0 +1,735 @@
...
+
+    ``treebeard`` is imported lazily (only by the selectors below, only when this
+    returns ``"treebeard"``), so the ``mptree`` backend never imports it.
+    """
+    import os
+
+    from django.conf import settings
+
+    return getattr(settings, "CMS_TREE_BACKEND", None) or os.environ.get(
+        "CMS_TREE_BACKEND", "mptree"
```
fsbraun:
```suggestion
        "CMS_TREE_BACKEND", "treebeard"
```

vinitkumar (commented):
This looks really good. Let's get it merged and tested.

vinitkumar (approved):

PR #8748, opened 2026-07-30
fix: Squashed migration with custom user model

pierreben:
## Description

In the squashed migration `0001_initial_squashed_0022_auto_20180620_1551` (introduced in #8484),  the `PageUser` model hardcodes:

- parent link field: `user_ptr`
- bases: `("auth.user",)`

This causes an issue with a custom user model defined in `settings.AUTH_USER_MODEL` : 

```
django.core.exceptions.FieldError: Auto-generated field 'user_ptr' in class 'PageUser' for parent_link to base class 'User' clashes with declared field of the same name.
```
The following commit dynamically computes these two values, as was already done in https://github.com/django-cms/django-cms/blob/main/cms/migrations/0002_auto_20140816_1918.py

## Summary by Sourcery

Bug Fixes:
- Resolve migration failures when using a custom user model by avoiding hardcoded references to auth.User in the PageUser model definition.

### DIFF
--- a/cms/migrations/0001_initial_squashed_0022_auto_20180620_1551.py
+++ b/cms/migrations/0001_initial_squashed_0022_auto_20180620_1551.py
@@ -4,14 +4,11 @@ import django.contrib.auth.models
 import django.db.models.deletion
 import django.utils.timezone
 from django.conf import settings
-from django.contrib.auth import get_user_model
 from django.db import migrations, models
 
 # Handle custom AUTH_USER_MODEL as in 0002_auto_20140816_1918.py
-User = get_user_model()
-
-user_model_label = '{}.{}'.format(User._meta.app_label, User._meta.model_name)
-user_ptr_name = '%s_ptr' % User._meta.object_name.lower()
+user_model_label = settings.AUTH_USER_MODEL.lower()
+user_ptr_name = f"{settings.AUTH_USER_MODEL.rsplit('.', 1)[1].lower()}_ptr"
 
 
 class Migration(migrations.Migration):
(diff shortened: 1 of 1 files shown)


fsbraun (approved):

cms/migrations/0001_initial_squashed_0022_auto_20180620_1551.py
```diff
@@ -4,8 +4,15 @@
...
 import django.utils.timezone
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.db import migrations, models
 
+# Handle custom AUTH_USER_MODEL as in 0002_auto_20140816_1918.py
+User = get_user_model()
+
+user_model_label = '{}.{}'.format(User._meta.app_label, User._meta.model_name)
+user_ptr_name = '%s_ptr' % User._meta.object_name.lower()
```
fsbraun:
```suggestion
from django.db import migrations, models

# Handle custom AUTH_USER_MODEL as in 0002_auto_20140816_1918.py
user_model_label = settings.AUTH_USER_MODEL.lower()
user_ptr_name = f"{settings.AUTH_USER_MODEL.rsplit('.', 1)[1].lower()}_ptr"
```

We only need the name, migrations will resolve to the model's state at migration time. `get_user_model` does provide the current user model.

pierreben: I'm okay with this.
Do you think this could be included in a patch soon?
Django CMS 5.1 doesn’t work at all on projects with a custom AUTH_USER_MODEL without this patch.

fsbraun: @pierreben I'm planning 5.1.1 for the next days.

pierreben: Thanks a lot !

PR #8784, opened 2026-08-18
fix: Add display limit to plugin deletion context

fsbraun:
## Summary

- forward the ModelAdmin deletion display limit from the custom plugin deletion view
- add the same context value to placeholder clearing confirmations
- retain compatibility with Django versions that do not define the setting
- cover both confirmation endpoints with regression tests

Django 6.1 admin deletion templates require delete_confirmation_max_display when rendering deletion trees. The custom django CMS confirmation contexts omitted it, causing VariableDoesNotExist while deleting plugins.

## Testing

- 31 tests in cms.tests.test_placeholder_admin pass on Django 6.1
- ruff passes for the changed files

## Summary by Sourcery

Provide the deletion display limit in custom plugin deletion confirmation views for compatibility with newer Django admin templates.

Bug Fixes:
- Pass Django’s deletion display limit to plugin deletion and placeholder-clearing confirmation contexts to prevent deletion-template rendering errors.

Enhancements:
- Preserve compatibility with Django versions that do not define the deletion display limit setting.

Tests:
- Add regression coverage for the deletion confirmation context in both custom admin endpoints.

### DIFF
--- a/cms/admin/placeholderadmin.py
+++ b/cms/admin/placeholderadmin.py
@@ -1223,6 +1223,7 @@ class PlaceholderAdmin(BaseEditableAdminMixin, admin.ModelAdmin):
             "protected": protected,
             "opts": opts,
             "app_label": opts.app_label,
+            "delete_confirmation_max_display": getattr(self, "delete_confirmation_max_display", None),
         }
         request.current_app = self.admin_site.name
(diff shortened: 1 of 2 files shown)


vinitkumar (approved):

PR #8822, opened 2026-09-01
fix: AddPageForm did not check all required CMS permissions

fsbraun:
## Summary by Sourcery

Enforce all required CMS permissions when creating pages through AddPageForm.

Bug Fixes:
- Validate CMS add-page permissions against the parent page submitted in the form, preventing unauthorized page creation through manipulated requests.

Tests:
- Add coverage for permitted and denied page creation, forged parent values, root-page attempts, and translation creation.

## Related resources

* #...
* #...

## Checklist

### DIFF
--- a/cms/admin/forms.py
+++ b/cms/admin/forms.py
@@ -52,7 +52,12 @@ from cms.utils.compat.forms import UserChangeForm
 from cms.utils.conf import get_cms_setting
 from cms.utils.i18n import get_language_list, get_site_language_from_request
 from cms.utils.page import get_clean_username
-from cms.utils.page_permissions import user_can_change_page, user_can_view_page
+from cms.utils.page_permissions import (
+    user_can_add_page,
+    user_can_add_subpage,
+    user_can_change_page,
+    user_can_view_page,
+)
 from cms.utils.permissions import (
     get_current_user,
     get_model_permission_codename,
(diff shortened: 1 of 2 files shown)


fatihsugar:
Thanks for the quick fix and for the extra root-page variant - that generalization (falling back to `user_can_add_page()` when no parent is submitted) is the right call and closes a boundary my original report didn't cover.

The change reads correctly to me: re-validating the POST-bound `parent_page` in `clean_parent_page()` mirrors `CreateCMSPageForm.clean_parent_page()`, so the admin form and the wizard now enforce the same invariant. The four cases (allowed-parent add, forbidden-parent add, forbidden-parent-in-body via divergent query string, and empty-parent root creation) plus the translation case cover the surface I'd want tested.

One small, optional maintainability note rather than a defect:

- The early `if self.cleaned_data.get("cms_page"): return parent_page` guard is what keeps the translation path working (and `test_add_translation_ignores_the_parent` green). Its comment explains *why* it skips, but the part that isn't visible at the call site is that it depends on `cms_page` being cleaned before `parent_page` - i.e. on field declaration order - so `cleaned_data` already holds `cms_page` when `clean_parent_page()` runs. A short "relies on cms_page being declared/cleaned before parent_page" note would protect that test from silently breaking if the fields are ever reordered.

- Adjacent surface, not this bug: the page-type subclass's `clean_parent_page()` calls `super().clean_parent_page()`, so it inherits the check - good. (The separate `source`/page-type copy path is unrelated to this PR; just noting I looked.)

Nothing blocking from my side - LGTM. Thanks again.

vinitkumar (approved):

PR #8831, opened 2026-09-03
fix: Slug uniqueness ignored site

### LINKED ISSUE #8830: [BUG] Slug-uniqueness-check maybe not respecting the site
## Description

When moving pages  in the tree of a site the slug uniqueness will be checked because changing the depth of a page (means changing the level in the tree) could produce 2 pages with the same slug in the same level.
This has been fixed with [6840](https://github.com/django-cms/django-cms/issues/6840).
Now when moving a page a corresponding error message appears when 2 pages with the same slug in the same language on the same level could exist when the move would be done.

BUT

When you have multiple sites and the editors build the pages on the first site before the developer copies the tree to one or more other sites (with [copy site](https://docs.django-cms.org/en/stable/reference/cli.html#cms-copy-site)) you have 2 or more identical trees.

When you now move a root page (or folder) in site 1 and lang en and change the position (not the level) the slug-uniqueness-validator will be called with a parent of none:

```python
    def _validate_slug_uniqueness(self, parent, language, slug):
        target_siblings = Page.objects.filter(parent=parent).exclude(pk=self.page.pk)
        if target_siblings.filter(urls__slug=slug, urls__language=language).exists():
            raise ValidationError(
                _(
                    "You cannot have two pages with the same slug at the same page level. "
                    "Please alter the slug of one of the pages before trying to move again."
                )
            )
```
So now the target_siblings queryset would contain all pages with a parent of None but also from all sites.

Right?

## Steps to reproduce

* Build a page tree on one site and language
* Play around and move some pages on the root level without changing the level
* [Copy the tree](https://docs.django-cms.org/en/stable/reference/cli.html#cms-copy-site) to another site
* Log in to admin of site 2 and move a root page to another position on the same level
-> Error message should appear but should not.

## Expected behaviour
When on a single tree all slugs are valid when moving root pages it should be on a copied tree too.

Django 5.2.17
Django-CMS 5.0.* and 5.1.2

fsbraun:
## Summary by Sourcery

Ensure page moves validate slug uniqueness within the correct site and page type while refining toolbar link styling.

Bug Fixes:
- Restrict page slug uniqueness checks to sibling pages on the same site and of the same page type, preventing false collisions across sites.

Enhancements:
- Prevent toolbar menu links from displaying underlined text.

Tests:
- Add coverage for moving root pages between sites with identical slugs.

## Related resources

* fixes #8830
* #...

## Checklist

### DIFF
--- a/cms/admin/forms.py
+++ b/cms/admin/forms.py
@@ -1017,7 +1017,11 @@ class MovePageForm(PageTreeForm):
         return target_page.parent
 
     def _validate_slug_uniqueness(self, parent, language, slug):
-        target_siblings = Page.objects.filter(parent=parent).exclude(pk=self.page.pk)
+        target_siblings = Page.objects.filter(
+            parent=parent,
+            site=self._site,
+            is_page_type=self.page.is_page_type,
+        ).exclude(pk=self.page.pk)
         if target_siblings.filter(urls__slug=slug, urls__language=language).exists():
             raise ValidationError(
                 _(
--- a/cms/static/cms/sass/components/_toolbar.scss
+++ b/cms/static/cms/sass/components/_toolbar.scss
@@ -123,6 +123,7 @@
         padding: $toolbar-menu-item-padding;
         line-height: $toolbar-height;
         height: $toolbar-height;
+        text-decoration: none;
         cursor: default;
     }
     li ul {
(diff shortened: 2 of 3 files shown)


vinitkumar (approved):

PR #8863, opened 2026-09-13
fix: CurrentUserMiddleware hang in async mode

### LINKED ISSUE #8859: [BUG] CurrentUserMiddleware can hang ASGI requests when restoring a lazy user
## Description

A synthetic local reproduction using django CMS 5.1.2, Django 5.2.17, and asgiref 3.9.2 shows that `CurrentUserMiddleware` can leave an ASGI request hanging when the lazy authentication user is evaluated during async context restoration.

This report contains only a local synthetic reproducer and public dependency behavior.

## Steps to reproduce

In a local django CMS project configured with SQLite, save the following as `reproduce.py`. It needs no database tables or real users. The synthetic user loader performs only `SELECT 1`, which exercises Django's real async-safety guard.

Run it in the project's environment with a process deadline:

```sh
DJANGO_SETTINGS_MODULE=myproject.settings timeout --kill-after=2s 15s python reproduce.py
```

Use local test settings: the query uses the configured default database.

```python
import asyncio
from unittest.mock import patch

import django

django.setup()

from django.db import connection
from django.http import HttpResponse
from django.test import AsyncClient, override_settings
from django.urls import path


def load_user(request):
    with connection.cursor() as cursor:
        cursor.execute("SELECT 1")
    return get_user_model()(pk=1)


async def health(request):
    return HttpResponse("ok")


async def main():
    response = await AsyncClient().get("/health/")
    print(response.status_code, response.content, flush=True)


with override_settings(
    ROOT_URLCONF=type("ProbeUrls", (), {"urlpatterns": [path("health/", health)]}),
    MIDDLEWARE=[
        "django.contrib.sessions.middleware.SessionMiddleware",
        "django.contrib.auth.middleware.AuthenticationMiddleware",
        "cms.middleware.user.CurrentUserMiddleware",
    ],
    SESSION_ENGINE="django.contrib.sessions.backends.signed_cookies",
), patch("django.contrib.auth.middleware.get_user", load_user):
    asyncio.run(main())
```

## Expected behaviour

The synthetic request completes with HTTP 200 and body `ok`.

## Actual behaviour

The local example raises `django.core.exceptions.SynchronousOnlyOperation` and does not return a response before the process deadline.

The relevant public-library sequence is:

1. Django's authentication middleware creates a lazy `request.user`.
2. CMS stores that unresolved object in its current-user `ContextVar`.
3. asgiref's `_restore_context` compares context values on the event-loop thread.
4. The comparison evaluates the lazy user, invoking the synthetic database query in an async context.
5. In asgiref 3.9.2, restoration in `AsyncToSync.main_wrap` occurs before the exception handler that completes its result future. The synchronous caller therefore keeps waiting.

## Screenshots

Not applicable.

The submitted reproduction was verified locally with:

- django CMS 5.1.2
- Django 5.2.17
- asgiref 3.9.2
- Python 3.12.14
- Linux and SQLite

The current upstream `main` middleware also passes the user unchanged to `set_current_user`; a full `main` checkout has not been tested.

### Local workaround

A synchronous subclass can resolve the user before CMS stores it:

```python
from cms.middleware.user import CurrentUserMiddleware as CMSCurrentUserMiddleware


class CurrentUserMiddleware(CMSCurrentUserMiddleware):
    def __call__(self, request):
        _ = request.user.is_authenticated
        return super().__call__(request)
```

Use this subclass in place of the stock CMS middleware entry, after Django's authentication middleware. The upstream implementation still handles resetting its context token.

Local synthetic ASGI checks with this workaround completed successfully for concurrent authenticated/anonymous requests and subsequent health requests on Python 3.11 and 3.12. The same ASGI checks hang with the stock middleware. Context cleanup on both success and exception was also checked.

- [ ] Yes, I want to help fix this issue and I will join the channel #pr-reviews on the Discord Server to confirm with the community that a PR is welcome.
- [x] No, I only want to report the issue.

The synthetic reproduction and locally verified workaround are supplied above.

fsbraun:
## Summary by Sourcery

Prevent CurrentUserMiddleware from hanging in async mode when handling lazy authenticated users.

Bug Fixes:
- Prevent ASGI requests from hanging when CurrentUserMiddleware processes an unevaluated lazy user.
- Ensure current-user context state is restored correctly and does not leak between requests.

Enhancements:
- Expose current-user context access through safe getter, setter, and reset helpers while preserving lazy-user evaluation behavior.

Tests:
- Add regression coverage for async middleware requests, context management, lazy-user comparisons, and request isolation.

## Related resources

* fixes #8859 
* #...

## Checklist


vinitkumar (approved):

